PR #134, opened 2016-11-04
Progressive Results Part 2: Result Completion Event

benrr101:
(aka Electric Boogaloo)

The main change in this pull request is to add a new event that will be fired upon completion of a resultset but before the completion of a batch. This event will only fire if a resultset is available and generated.

Changes:
* ConnectionService - Slight changes to enable mocking, cleanup 
* Batch - Moving summary generation into ResultSet class, adding generation of ordinals for resultset and locking of result set list (which needs further refinement, but would be outside scope of this change)
* Adding new event and associated parameters for completion of a resultset. Params return the resultset summary
* Adding logic for assigning the event a handler in the query execution service
* Adding unit tests for testing the new event /making sure the existing tests work
* Refactoring some private properties into member variables


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecuteResultSetCompleteNotification.cs
```diff
@@ -0,0 +1,22 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.SqlTools.ServiceLayer.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
+{
+    public class QueryExecuteResultSetCompleteParams
+    {
+        public ResultSetSummary ResultSetSummary { get; set; }
```
benrr101: Done and done!

PR #151, opened 2016-11-16
Fix Code Coverage

benrr101:
This is another large code review. I want to make a few more changes, but since these changes will stand on their own, I'll hold back on making this change set any larger than it already is.

Changes in this request:
* To address Microsoft/vscode-mssql#326, instead of doing `taskkill` on the service layer when WaitForExit is executed, we now make an educated guess at which service layer was spawned when the test starts and do a Process.Kill on it when we shut down the test.
* All the perf tests have been moved into a new project. This was done to keep them easily separated from code coverage test runs. At the same time the perf tests were separated into separate classes for logical categorization. This process will likely be repeated on the stress tests. The tests can still easily be ran from Visual Studio Test Explorer
* To address Microsoft/vscode-mssql#349, a new SelfCleaningFile class was created to allow for easy cleanup of temporary files generated for integration tests via using blocks.
* Due to some of the refactoring done while moving the perf tests to a new project, the TestBase class had to be switched to more of a helper class style. As such, all tests that use inherit from TestBase now create a TestBase object on start via a using block. This also simplifies the cleanup at the end of the test.


llali:
> using System;

you should still add copy right even for tests

---
Refers to: test/Microsoft.SqlTools.ServiceLayer.PerfTests/Tests/Common.cs:1 in 29fa528. [](commit_id = 29fa52860498ba7cdef0dfc6a5ed358d712d401d, deletion_comment = False)

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Utility/ConnectionTestUtils.cs
```diff
@@ -22,7 +22,7 @@ public class ConnectionTestUtils
         public static Setting Setting = InitSetting();
 
         private static readonly Lazy<ConnectParams> azureTestServerConnection =
-            new Lazy<ConnectParams>(() => GetConnectionFromVsCodeSettings("***REMOVED***"));
```
llali: you probably don't want to change this

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Utility/TestRunner.cs
```diff
@@ -0,0 +1,73 @@
...
+            if (methodInfo == null)
+            {
+                Console.WriteLine("Invalid method name");
+            }
+            else
+            {
+                using (var typeInstance = (IDisposable)Activator.CreateInstance(type))
+                {
+                    PropertyInfo testNameProperty = type.GetProperty("TestName");
+                    testNameProperty?.SetValue(typeInstance, testName);
```
llali: these two lines should be removed. probably copied from my changes

test/Microsoft.SqlTools.ServiceLayer.PerfTests/Tests/IntellisenseTests.cs
```diff
@@ -0,0 +1,211 @@
...
+                    TextDocument = new VersionedTextDocumentIdentifier
+                    {
+                        Version = 2,
+                        Uri = queryFile.FilePath
+                    }
+                };
+
+                TestTimer timer = new TestTimer();
+                await testBase.RequestChangeTextDocumentNotification(changeParams);
+                await Common.ExecuteWithTimeout(timer, 60000, async () =>
```
llali: >ExecuteWithTimeout [](start = 29, length = 18)

thanks for refactoring these tests :)

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Utility/SelfCleaningFile.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.TestDriver.Utility
+{
+    public class SelfCleaningFile : IDisposable
+    {
+        private bool disposed;
+
+        public SelfCleaningFile()
+        {
+            FilePath = Path.GetTempFileName();
```
llali: >GetTempFileName [](start = 28, length = 15)

Add copyright

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Utility/SelfCleaningFile.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.IO;
+
+namespace Microsoft.SqlTools.ServiceLayer.TestDriver.Utility
+{
+    public class SelfCleaningFile : IDisposable
+    {
+        private bool disposed;
+
+        public SelfCleaningFile()
```
llali: SelfCleaningTempFile?

llali:
:clock1:

benrr101:
The latest changes should take care of the requested changes

PR #152, opened 2016-11-17
*duplicate* Peek definition support for tables, views and stored procedures

sharonravindran:
Initial code review for peek definition - tables, view and stored procedures

Does not include temp file cleanup and scripting options.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Hosting/ServiceHost.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Hosting/ServiceHost.cs
@@ -150,7 +150,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Hosting
                     Capabilities = new ServerCapabilities
                     {
                         TextDocumentSync = TextDocumentSyncKind.Incremental,
-                        DefinitionProvider = false,
+                        DefinitionProvider = true,
                         ReferencesProvider = false,
                         DocumentHighlightProvider = false,
                         HoverProvider = true,
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -4,6 +4,7 @@
 //
 
 using System;
+using System.IO;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -41,6 +42,8 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         internal const int BindingTimeout = 500;
 
+        internal const int PeekDefinitionTimeout = 30000;
+
         internal const int OnConnectionWaitTimeout = 300000;
 
         private static ConnectionService connectionService = null;
@@ -53,10 +56,10 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         private ConnectedBindingQueue bindingQueue = new ConnectedBindingQueue();
 
-        private ParseOptions defaultParseOptions =  new ParseOptions(
+        private ParseOptions defaultParseOptions = new ParseOptions(
             batchSeparator: LanguageService.DefaultBatchSeperator,
-            isQuotedIdentifierSet: true,  
-            compatibilityLevel: DatabaseCompatibilityLevel.Current, 
+            isQuotedIdentifierSet: true,
+            compatibilityLevel: DatabaseCompatibilityLevel.Current,
             transactSqlVersion: TransactSqlVersion.Current);
 
         /// <summary>
@@ -99,13 +102,13 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         private static readonly Lazy<LanguageService> instance = new Lazy<LanguageService>(() => new LanguageService());
 
-        private Lazy<Dictionary<string, ScriptParseInfo>> scriptParseInfoMap 
+        private Lazy<Dictionary<string, ScriptParseInfo>> scriptParseInfoMap
             = new Lazy<Dictionary<string, ScriptParseInfo>>(() => new Dictionary<string, ScriptParseInfo>());
 
         /// <summary>
         /// Gets a mapping dictionary for SQL file URIs to ScriptParseInfo objects
         /// </summary>
-        internal Dictionary<string, ScriptParseInfo> ScriptParseInfoMap 
+        internal Dictionary<string, ScriptParseInfo> ScriptParseInfoMap
         {
             get
             {
@@ -160,7 +163,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             {
                 if (LanguageService.workspaceServiceInstance == null)
                 {
-                    LanguageService.workspaceServiceInstance =  WorkspaceService<SqlToolsSettings>.Instance;
+                    LanguageService.workspaceServiceInstance = WorkspaceService<SqlToolsSettings>.Instance;
                 }
                 return LanguageService.workspaceServiceInstance;
             }
@@ -198,7 +201,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             // Register the requests that this service will handle
 
             // turn off until needed (10/28/2016)
-            // serviceHost.SetRequestHandler(DefinitionRequest.Type, HandleDefinitionRequest);
+            serviceHost.SetRequestHandler(DefinitionRequest.Type, HandleDefinitionRequest);
             // serviceHost.SetRequestHandler(ReferencesRequest.Type, HandleReferencesRequest);
             // serviceHost.SetRequestHandler(SignatureHelpRequest.Type, HandleSignatureHelpRequest);
             // serviceHost.SetRequestHandler(DocumentHighlightRequest.Type, HandleDocumentHighlightRequest);
@@ -253,22 +256,22 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                 await Task.FromResult(true);
             }
             else
-            {                
+            {
                 // get the current list of completion items and return to client 
                 var scriptFile = LanguageService.WorkspaceServiceInstance.Workspace.GetFile(
                     textDocumentPosition.TextDocument.Uri);
 
                 ConnectionInfo connInfo;
                 LanguageService.ConnectionServiceInstance.TryFindConnection(
-                    scriptFile.ClientFilePath, 
+                    scriptFile.ClientFilePath,
                     out connInfo);
 
                 var completionItems = Instance.GetCompletionItems(
                     textDocumentPosition, scriptFile, connInfo);
-               
-                   await requestContext.SendResult(completionItems);
-                }  
+
+                await requestContext.SendResult(completionItems);
             }
+        }
 
         /// <summary>
         /// Handle the resolve completion request event to provide additional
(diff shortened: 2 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/Hosting/ServiceHost.cs
```diff
@@ -150,7 +150,7 @@ public void RegisterInitializeTask(InitializeCallback callback)
                     Capabilities = new ServerCapabilities
                     {
                         TextDocumentSync = TextDocumentSyncKind.Incremental,
-                        DefinitionProvider = false,
+                        DefinitionProvider = true,
```
benrr101: Any time we can add a new capability to the list it's 👍 💯

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -293,15 +295,23 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
 
-// turn off this code until needed (10/28/2016)
-#if false
-        private static async Task HandleDefinitionRequest(
-            TextDocumentPosition textDocumentPosition,
-            RequestContext<Location[]> requestContext)
+        private async Task HandleDefinitionRequest(TextDocumentPosition textDocumentPosition, RequestContext<Location[]> requestContext)
         {
-            await Task.FromResult(true);
+            //TODO: check for settings
```
benrr101: When will we be checking for settings?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -293,15 +295,23 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+
+            Location[] locations = this.GetDefinition(textDocumentPosition);
+            if (locations != null)
+            {
+                await requestContext.SendResult(locations);
+            }
+            else
+            {
+                await Task.FromResult(true);
```
benrr101: I believe this else branch isn't necessary. As long as there's at least one await in the async method, there's no need to add a separate `Task.FromResult(true)`

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -802,21 +812,127 @@ internal Hover GetHoverItem(TextDocumentPosition textDocumentPosition, ScriptFil
...
+                                            return null;
+                                    }
+                                }
+                            }
+                            return null;
+
+                        },
+                        timeoutOperation: (bindingContext) =>
+                        {
+                            return null;
```
benrr101: What happens in the UI if the operation times out? Do users get some kind of notice that the peek definition timed out?

I'm mostly asking because I've seen some issues with the omnisharp where going go to definition doesn't work sometimes, no notices, no nothing and it's a really frustrating experience.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+            if (this.connectionInfo.SqlConnection != null)
+            {
+                Server server = new Server(this.connectionInfo.SqlConnection.DataSource);
+                Database database = server.Databases[this.connectionInfo.SqlConnection.Database];
+                Table table = database.Tables[tableName];
+                string tempFileName = tempPath + tableName + ".sql";
+
+                if (table != null)
+                {
+                    using (StreamWriter scriptFile = new StreamWriter(File.Open(tempFileName, FileMode.Create, FileAccess.ReadWrite)))
```
benrr101: I forget if this is the case, but it may be worthwhile to double check that disposing the streamwriter also disposes the filestream

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+                    {
+                        StringCollection scripts = table.Script();
+                        foreach (string script in scripts)
+                        {
+                            if (script.Contains("CREATE TABLE"))
+                            {
+                                scriptFile.WriteLine(script);
```
benrr101: Is it possible to have more than one script that contains "CREATE TABLE"?

Could we possibly simplify this code with some LINQ, like `scripts.FirstOrDefault(s => s.Contains("CREATE TABLE")`?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <param name="schemaName">Schema name </param>
+        /// <returns>Location object representing URI and range of the script file</returns>
+        internal Location[] GetViewDefinition(string viewName, string schemaName)
+        {
+            if (this.connectionInfo.SqlConnection != null)
+            {
+                Server server = new Server(this.connectionInfo.SqlConnection.DataSource);
+                Database database = server.Databases[this.connectionInfo.SqlConnection.Database];
+                View view = (schemaName != null) ? database.Views[viewName, schemaName] : database.Views[viewName];
+                string tempFileName = (schemaName != null) ? tempPath + schemaName + "." + viewName + ".sql"
```
benrr101: Rather than concatenate the strings with `+` let's either use `string.Format` or 🌟[string interpolation](https://msdn.microsoft.com/en-us/library/dn961160.aspx)🌟

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+                                                    : tempPath + viewName + ".sql";
+
+                if (view != null)
+                {
+                    using (StreamWriter scriptFile = new StreamWriter(File.Open(tempFileName, FileMode.Create, FileAccess.ReadWrite)))
+                    {
+                        StringCollection scripts = view.Script();
+                        foreach (string script in scripts)
```
benrr101: Same comment as above regarding simplifying using LINQ and what happens if there are more than one "CREATE VIEW"

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            if (this.connectionInfo.SqlConnection != null)
+            {
+                Server server = new Server(this.connectionInfo.SqlConnection.DataSource);
+                Database database = server.Databases[this.connectionInfo.SqlConnection.Database];
+                View view = (schemaName != null) ? database.Views[viewName, schemaName] : database.Views[viewName];
+                string tempFileName = (schemaName != null) ? tempPath + schemaName + "." + viewName + ".sql"
+                                                    : tempPath + viewName + ".sql";
+
+                if (view != null)
```
benrr101: It might be a good idea to reverse this condition to eliminate a level of nesting

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <returns>Location object representing URI and range of the script file</returns>
+        internal Location[] GetStoredProcedureDefinition(string storedProcedureName, string schemaName)
+        {
+            if (this.connectionInfo.SqlConnection != null)
+            {
+                Server server = new Server(this.connectionInfo.SqlConnection.DataSource);
+                Database database = server.Databases[this.connectionInfo.SqlConnection.Database];
+                StoredProcedure storedProcedure = (schemaName != null) ? database.StoredProcedures[storedProcedureName, schemaName] :
+                                                    database.StoredProcedures[storedProcedureName];
+                string tempFileName = (schemaName != null) ? tempPath + schemaName + "." + storedProcedureName + ".sql"
```
benrr101: Same comment as above with regards to string interpolation/formatting

benrr101:
Additionally, can we get some unit tests for these methods in the PeekDefinition class?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: This is an initial PR, I'm working on adding the TODOs and the scripting options.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: Right now, that's what we do too. No peek window is displayed and the handler returns a null. The language service protocol expects a valid URI or a null.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+                                                    : tempPath + viewName + ".sql";
+
+                if (view != null)
+                {
+                    using (StreamWriter scriptFile = new StreamWriter(File.Open(tempFileName, FileMode.Create, FileAccess.ReadWrite)))
+                    {
+                        StringCollection scripts = view.Script();
+                        foreach (string script in scripts)
+                        {
+                            if (script.Contains("CREATE VIEW"))
```
llali: >script [](start = 32, length = 6)

shouldn't it ignore case?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: There can be only one CREATE I think. I'm looking into scripting options that would let us remove this check.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        /// <param name="storedProcedureName">Stored Procedure name</param>
+        /// <param name="schemaName">Schema Name</param>
+        /// <returns>Location object representing URI and range of the script file</returns>
+        internal Location[] GetStoredProcedureDefinition(string storedProcedureName, string schemaName)
+        {
+            if (this.connectionInfo.SqlConnection != null)
+            {
+                Server server = new Server(this.connectionInfo.SqlConnection.DataSource);
+                Database database = server.Databases[this.connectionInfo.SqlConnection.Database];
```
llali: these two lines are in other methods too. can they share the code?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Will do.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Yes, I'll extract those lines into a method.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: I checked and it does dispose the filestream. 👍

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: It seems to be in caps always but I'm double checking. I'm also looking into Scripting Options that will let us eliminate this check.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: Removed

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: There aren't any settings to check yet

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
benrr101: I'd caution that, at least from my perspective, not doing anything at the UI isn't very intuitive. It suggests to me that the extension is broken or incapable of doing what it should. But, that would be a discussion for the extension

sharonravindran:
Apologies. I messed up my commits and created a new branch with all the code review changes. 
I've created a new pull request( with a newer branch) here:  #160

kburtram:
@sharonravindran if this is a duplicate PR should we just close it out and work from the other one, or is there a reason to keep this active?

PR #153, opened 2016-11-18
Implemented function signature help, and added tests

MitchellSternke:
https://github.com/Microsoft/vscode-mssql/issues/351


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
```diff
@@ -680,5 +680,79 @@ private static string GetCompletionItemInsertName(Declaration autoCompleteItem)
                 return null;
             }
         }
+
+        /// <summary>
+        /// Converts a SQL Parser List of MethodHelpText objects into a VS Code SignatureHelp object
+        /// </summary>
+        internal static SignatureHelp ConvertMethodHelpTextListToSignatureHelp(List<Babel.MethodHelpText> methods, Babel.MethodNameAndParamLocations locations, int line, int column)
```
kevcunnane: >methods [](start = 106, length = 7)

Please do argument validation on these (e.g. Validate.IsNotNull()). Internal is more or less equivalent to public, in that we shouldn't trust the caller does the right thing.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
```diff
@@ -680,5 +680,79 @@ private static string GetCompletionItemInsertName(Declaration autoCompleteItem)
...
+                };
+            }).ToArray();
+
+            if (help.Signatures.Length == 0)
+            {
+                return null;
+            }
+
+            // Find the matching method signature at the cursor's location
+            int signatureIndex = methods.FindIndex(method => method.Name.Contains(locations.Name));
```
kevcunnane: >locations [](start = 82, length = 9)

Should this be earlier? Otherwise don't we convert the methods to signatures unnecessarily? Probably not a huge perf concern, it's just that ideally we validate early exit conditions first, then do business logic. Not a blocker.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -309,20 +309,29 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+#endif
 
-        private static async Task HandleDocumentHighlightRequest(
+        private static async Task HandleSignatureHelpRequest(
             TextDocumentPosition textDocumentPosition,
-            RequestContext<DocumentHighlight[]> requestContext)
+            RequestContext<SignatureHelp> requestContext)
         {
-            await Task.FromResult(true);
+            var scriptFile = WorkspaceService<SqlToolsSettings>.Instance.Workspace.GetFile(
```
kevcunnane: Best practice is not to use var in source if the right-hand side isn't clearly typed. Here, I'm wondering is this a File, FileInfo, or some different type of object. Could you use full names for this and the "var help"?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -309,20 +309,29 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                textDocumentPosition.TextDocument.Uri);
+
+            var help = LanguageService.Instance.GetSignatureHelp(textDocumentPosition, scriptFile);
+            if (help != null)
+            {
+                await requestContext.SendResult(help);
+            }
+            
+            await requestContext.SendResult(new SignatureHelp());
```
kevcunnane: >await requestContext.SendResult(new SignatureHelp()); [](start = 12, length = 53)

Isn't this double-sending? Should this be in an else clause?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -691,6 +700,78 @@ internal Hover GetHoverItem(TextDocumentPosition textDocumentPosition, ScriptFil
         }
 
         /// <summary>
+        /// Get function signature help for the current position
+        /// </summary>
+        /// <param name="textDocumentPosition"></param>
+        /// <param name="scriptFile"></param>
```
kevcunnane: Nitpick: remove params if they don't have additional documentation.

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Tests/LanguageServiceTests.cs
```diff
@@ -279,5 +279,137 @@ public class LanguageServiceTests : TestBase
...
+
+        [Fact]
+        public async Task FunctionSignatureCompletionReturnsCorrectFunction()
+        {
+            string sqlText = "EXEC sys.fn_isrolemember ";
+            string ownerUri = System.IO.Path.GetTempFileName();
+
+            try
+            {
+                System.IO.File.WriteAllText(ownerUri, sqlText);
```
kevcunnane: >System.IO. [](start = 16, length = 10)

nitpick: use System.IO and remove the fully-qualified name unless there's a conflict?

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Tests/LanguageServiceTests.cs
```diff
@@ -279,5 +279,137 @@ public class LanguageServiceTests : TestBase
...
+            }
+            finally
+            {
+                System.IO.File.Delete(ownerUri);
+                WaitForExit();
+            }
+        }
+
+        [Fact]
+        public async Task FunctionSignatureCompletionReturnsCorrectParametersAtEachPosition()
```
kevcunnane: We should have some negative test, e..g for when the location.Name doesn't match. Could you add this one?

kevcunnane:
:clock1:

kevcunnane:
Overall looks good, would just like a little more tests and some fixes before merging. Thanks!

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -309,20 +309,29 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
             TextDocumentPosition textDocumentPosition,
-            RequestContext<SignatureHelp> requestContext)
+            RequestContext<DocumentHighlight[]> requestContext)
         {
             await Task.FromResult(true);
         }
+#endif
 
-        private static async Task HandleDocumentHighlightRequest(
+        private static async Task HandleSignatureHelpRequest(
```
kevcunnane: >HandleSignatureHelpRequest [](start = 34, length = 26)

I think you missed adding a check to only process if suggestions are enabled:
if (!WorkspaceService<SqlToolsSettings>.Instance.CurrentSettings.IsSuggestionsEnabled)
            {
                await Task.FromResult(true);
            }

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
MitchellSternke: I agree, this makes sense. I'm not sure why the protocol dictates returning all signatures (perhaps to list multiple signatures if there are overloads? (not sure if this applies to SQL)). OmniSharp appears to follow this pattern:  https://github.com/OmniSharp/omnisharp-roslyn/blob/cdd99aef194660835e61c58f19fb51ab3dabc52e/src/OmniSharp.Roslyn.CSharp/Services/Signatures/SignatureHelpService.cs#L24

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
MitchellSternke: Oops, for sure!

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
MitchellSternke: My bad! Yes, you're correct. I must've interpreted the await as a return while I was writing this since they have the same highlighting color...

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Tests/LanguageServiceTests.cs
MitchellSternke: Will do

MitchellSternke:
I'll merge this in at the end of today if there's no more feedback. I think I've addressed everything so far.

kevcunnane:
Looks good to me, :shipit:

---
In reply to: [263722019](https://github.com/Microsoft/sqltoolsservice/pull/153#issuecomment-263722019) [](ancestors = 263722019)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -691,6 +710,76 @@ internal Hover GetHoverItem(TextDocumentPosition textDocumentPosition, ScriptFil
...
+
+            if (scriptParseInfo != null && scriptParseInfo.ParseResult != null)
+            {
+                if (Monitor.TryEnter(scriptParseInfo.BuildingMetadataLock))
+                {
+                    try
+                    {
+                        QueueItem queueItem = this.BindingQueue.QueueBindingOperation(
+                            key: scriptParseInfo.ConnectionKey,
+                            bindingTimeout: LanguageService.HoverTimeout,
```
kburtram: I would use BindingTimeout here instead of HoverTimeout...though they both are 500ms at this point, they *could* diverge in the future.

MitchellSternke: Good catch - I originally copied this code from the Hover handler and must have missed that 😃

PR #154, opened 2016-11-18
Feature/more tests

MitchellSternke:
A few more connection related tests for code coverage


benrr101:
:shipit: 
Though, I'd like to see some comments of the "given, if, then" format. But no worries otherwise.

MitchellSternke:
@benrr101 Added comments. This is actually really nice for readability, I think I'll adopt this format in the future 😃

PR #155, opened 2016-11-21
Perf Test fixing bugs and Update command

llali:
- Fixed some bugs that was caused by refactoring the code. 
- Ran the tests again to make sure they all are running again. 
- Added a command to update the dlls to an external location to run the tests automatically after each build 
- Using a test database in the server used for running the tests instead of a fixed db. Create the database if doesn't  exist before running each test.


test/Microsoft.SqlTools.ServiceLayer.PerfTests/Microsoft.SqlTools.ServiceLayer.PerfTests.xproj
```diff
@@ -7,7 +7,7 @@
   <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
   <PropertyGroup Label="Globals">
     <ProjectGuid>7e5968ab-83d7-4738-85a2-416a50f13d2f</ProjectGuid>
-    <RootNamespace>Microsoft.SqlTools.ServiceLayer.PerfTests</RootNamespace>
+    <RootNamespace>Microsoft.SqlTools.ServiceLayer.PerfTests.Tests</RootNamespace>
```
benrr101: The root namespace should be without the .Tests, otherwise anything in the tests folder should technically be "Microsoft.SqlTools.ServiceLayer.PerfTests.Tests.Tests". It's not a huge issue, but without it, we'll get warnings from ReSharper. If there's a functional reason for the change, then that's fine.

test/Microsoft.SqlTools.ServiceLayer.PerfTests/Tests/Common.cs
```diff
@@ -12,7 +12,7 @@
 using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
 using Xunit;
 
-namespace Microsoft.SqlTools.ServiceLayer.PerfTests.Tests
+namespace Microsoft.SqlTools.ServiceLayer.PerfTests
```
benrr101: This combo of changes to remove the .Tests from the namespaces and add .Tests to the default namespace doesn't seem right.

test/Microsoft.SqlTools.ServiceLayer.PerfTests/Tests/IntellisenseTests.cs
```diff
@@ -202,7 +202,7 @@ public class IntellisenseTests
         private static async Task ValidateCompletionResponse(TestHelper testHelper, string ownerUri, string query, [CallerMemberName] string testName="")
         {
             TestTimer timer = new TestTimer();
-            await Common.ExecuteWithTimeout(timer, 60000, async () =>
+            await Common.ExecuteWithTimeout(timer, 500000, async () =>
```
benrr101: Keep in mind this will keep the test running, even if it fails, for ~8min. I'm guessing that's the expected behavior, but it seems a little excessive for something we'd expect to complete in under a minute

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Utility/TestRunner.cs
```diff
@@ -23,9 +23,8 @@ public class TestRunner
                     bool containsTestName = testName.Contains(".");
                     var className = containsTestName ? testName.Substring(0, testName.LastIndexOf('.')) : testName;
                     var methodName = containsTestName ? testName.Substring(testName.LastIndexOf('.') + 1) : null;
-
-
-                    var type = Type.GetType(testNamespace + className);
+                    Assembly assembly = Assembly.GetEntryAssembly();
```
benrr101: I think rather than getting the entry assembly, we should get the assembly of the current type, ie TestRunner.GetType().Assembly (or similar). The issue I foresee happening is that the entry assembly won't be the what we expect when we run it from Test Explorer

benrr101:
Apologies for any bugs that were introduced with the refactor. Take a look at the comments, tho. I think there's some issues we might want to address

test/Microsoft.SqlTools.ServiceLayer.PerfTests/Tests/Common.cs
llali: I fixed the namespace in the project

---
In reply to: [88957713](https://github.com/Microsoft/sqltoolsservice/pull/155#discussion_r88957713) [](ancestors = 88957713)

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Utility/TestRunner.cs
llali: the current type (Testrunner) doesn't have those test methods because the methods are in different assembly (PerfTest)

---
In reply to: [88958836](https://github.com/Microsoft/sqltoolsservice/pull/155#discussion_r88958836) [](ancestors = 88958836)

llali:
no worries 

---
In reply to: [262027794](https://github.com/Microsoft/sqltoolsservice/pull/155#issuecomment-262027794) [](ancestors = 262027794)

PR #157, opened 2016-11-21
Fixed issue where no-op queries would never complete

MitchellSternke:
Fixes https://github.com/Microsoft/vscode-mssql/issues/341

Also added unit test

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecuteRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecuteRequest.cs
@@ -32,6 +32,12 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
         /// Connection error messages. Optional, can be set to null to indicate no errors
         /// </summary>
         public string Messages { get; set; }
+
+        /// <summary>
+        /// Optional, can be set to true to indicate that the messages in the result should not be treated as an error,
+        /// but rather as an informational message.
+        /// </summary>
+        public bool? HasInfoMessages { get; set; }
     }
 
     public class QueryExecuteRequest
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecuteRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecuteRequest.cs
@@ -29,15 +29,9 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
     public class QueryExecuteResult
     {
         /// <summary>
-        /// Connection error messages. Optional, can be set to null to indicate no errors
+        /// Informational messages from the query runner. Optional, can be set to null.
         /// </summary>
         public string Messages { get; set; }
-
-        /// <summary>
-        /// Optional, can be set to true to indicate that the messages in the result should not be treated as an error,
-        /// but rather as an informational message.
-        /// </summary>
-        public bool? HasInfoMessages { get; set; }
     }
 
     public class QueryExecuteRequest
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
@@ -424,6 +424,25 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                     OwnerUri = executeParams.OwnerUri,
                     BatchSummaries = q.BatchSummaries
                 };
+
+                // If there were no batch results in the query, send back a message that the commands were completed successfully
+                if (eventParams.BatchSummaries.Length == 0)
+                {
+                    eventParams.BatchSummaries = new BatchSummary[] { new BatchSummary()
+                    {
+                        Messages = new ResultMessage[] { new ResultMessage(SR.QueryServiceCompletedSuccessfully) }
+                    }};
+
+                    var batchParams = new QueryExecuteBatchCompleteParams();
+                    batchParams.BatchSummary = eventParams.BatchSummaries[0];
+                    batchParams.OwnerUri = executeParams.OwnerUri;
+
+                    // HACK: this is here to prevent the QueryExecuteCompleteEvent from being sent before the QueryExecuteResult is processed
+                    System.Threading.Thread.Sleep(100);
+
+                    await requestContext.SendEvent(QueryExecuteBatchCompleteEvent.Type, batchParams);
+                }
+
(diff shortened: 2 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -424,6 +424,25 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    {
+                        Messages = new ResultMessage[] { new ResultMessage(SR.QueryServiceCompletedSuccessfully) }
+                    }};
+
+                    var batchParams = new QueryExecuteBatchCompleteParams();
+                    batchParams.BatchSummary = eventParams.BatchSummaries[0];
+                    batchParams.OwnerUri = executeParams.OwnerUri;
+
+                    // HACK: this is here to prevent the QueryExecuteCompleteEvent from being sent before the QueryExecuteResult is processed
+                    System.Threading.Thread.Sleep(100);
```
kevcunnane: Following up offline chat: I'd suggest using a semaphore here. The other event will lift it, this will block on it. That way if already called we don't wait. If it hasn't gone through yet, this will cleanly wait on that.

---
In reply to: [89245031](https://github.com/Microsoft/sqltoolsservice/pull/157#discussion_r89245031) [](ancestors = 89245031)

MitchellSternke:
I ended up changing my approach to this after discovering that using a semaphore in the service layer to ensure message ordering didn't solve the problem. The race condition would still happen in the extension between the two requests/events.

I solved this by refactoring it to only send the initial QueryExecuteResult with the informational message as part of the response. I had to add another parameter to QueryExecuteResult to flag whether the messages were informational or an error since the extension would always interpret messages in the result as errors.

Corresponding extension changes can be found in this pull request: https://github.com/Microsoft/vscode-mssql/pull/422

Please review again. Hopefully this is the last round 😃

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Tests/QueryExecutionTests.cs
```diff
@@ -323,5 +324,38 @@ public class QueryExecutionTests
                 await testHelper.Disconnect(queryTempFile.FilePath);
             }
         }
+
+        [Fact]
+        public async Task NoOpQueryReturnsMessage()
```
kevcunnane: Shouldn't we also have a test that covers the error message case? I don't see one of these in this test file

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecuteRequest.cs
```diff
@@ -32,6 +32,12 @@ public class QueryExecuteResult
         /// Connection error messages. Optional, can be set to null to indicate no errors
         /// </summary>
         public string Messages { get; set; }
```
kevcunnane: In an ideal world we'd use one of 2 patterns:
1. Each message has a MessageType (Error, Warn, Info). 
2. We have 2 streams, 1 for error messages, the other for InfoMessages.

This would be less restrictive than the current approach. It's not vital and your solution requires minimal code change on the mssql extension side, but it's something to think about and see if it's worth it

kevcunnane:
I'm signing off as this is shippable as-is. I think we could improve the contract slightly and should ensure we have a test for the error case, but otherwise this looks fine. You can always check this in as-is and follow up on those in a separate PR or won't-fix the suggestions

kevcunnane:
:shipit:

benrr101:
I'm confused... Wasn't the original version of this just fixing a case where the query completed event wasn't being fired? I'm not sure the benefit of the "hasInfoMessage" changes. Can you clarify what was wrong and how this fixes it?

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Tests/QueryExecutionTests.cs
MitchellSternke: Yeah, that would make sense. After looking further into the code it appears that the message field is actually unused (aside from this use case for the bug) which I'm guessing is an artifact of changing query execution to be asynchronous and moving the messages to a different event. I had originally thought that a test should have already existed, but looks like we'll need one for the error message case too.

benrr101: We changed a while back to not send a result back if there was an error. Instead we would send an error back on the request context if there was an error. This did leave the message field behind as not necessary.

MitchellSternke:
@benrr101 Yes - my original solution to the bug was to simply add a query completed event with a message. However, there was a race condition in the extension where the completed event would sometimes finish processing before the initial ACK from the query request could be handled, so the UI would still get "stuck" thinking that the query was still running since the ACK got processed after the completion message. (i.e. the QueryExecuteResult would get processed out-of-order after the QueryExecuteComplete event).

So my initial change to fix this race condition was to introduce a semaphore to ensure that the messages would get sent to the client in the correct order, but this didn't work since the extension could still process the requests in the wrong order due to how promises work.

I then tried just sending the message down with the initial QueryExecuteResult, which eliminated the race condition, but would show up as an error due to how the messages field was interpreted by the UI. So, I added the "hasInfoMessages" flag to allow the message type to be sent to allow to the UI to handle both errors and info messages without breaking any code that already depended on the existing messages field.

Hope this clarifies everything!

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecuteRequest.cs
MitchellSternke: I'll quickly switch the parameter out to have a message type instead - I like that idea (and it's an easy change)

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Tests/QueryExecutionTests.cs
MitchellSternke: I'll address this in a separate PR after I get this checked in

benrr101:
Ok, that makes more sense now, thanks! However, I think we can simplify this further. The messages that are passed back with the "ack" from starting the query are no longer error messages. Any error messages that occur are sent back as an Error result, basically leaving the messages field unused. In which case, you could omit the "hasInfoMessage" since all messages would be informational anyways.

However, I'm not sure what scenario would result in a query generating zero batches. Even if the only line in the query is a comment, it is still parsed into a batch and sent to the server. What was your test query for this scenario?

MitchellSternke:
That's a good point, I'll just change the documentation to reflect that the messages field is purely informational since we're not using it for anything else.

I used "GO" and "-- comment" which both result in the batch array having zero entries.

PR #158, opened 2016-11-22
Fixed Turkish I problem with autocomplete keyword suggestions

MitchellSternke:
Fixes https://github.com/Microsoft/vscode-mssql/issues/362

I debugged into this and found we weren't using the invariant culture when upper/lower casing the keywords list. Tested on Windows Server 2012 with the Turkish locale with and without the fix applied.


benrr101:
:shipit:
Good find! Also good learning for the rest of us what the difference between ToUpper and ToUpperInvariant is

MitchellSternke:
@benrr101 Yeah, everyone should definitely be aware of this. Never use ToUpper/ToLower directly in C# code unless you know what you are doing.

http://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/

kevcunnane:
:shipit:

PR #160, opened 2016-11-22
Peek definition support for tables, views and stored procedures

sharonravindran:
Add support for peek/go to definition
Add unit tests

Closes Microsoft/vscode-mssql#253 and Microsoft/vscode-mssql#268

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -704,10 +704,6 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                             return peekDefinition.GetScript(declarationItems, tokenText, schemaName);
                             
 
-                        },
-                        timeoutOperation: (bindingContext) =>
-                        {
-                            return null;
                         });
 
                     // wait for the queue item
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -297,15 +297,18 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         internal static async Task HandleDefinitionRequest(TextDocumentPosition textDocumentPosition, RequestContext<Location[]> requestContext)
         {
-            // Retrieve document and connection
-            ConnectionInfo connInfo;
-            var scriptFile = LanguageService.WorkspaceServiceInstance.Workspace.GetFile(textDocumentPosition.TextDocument.Uri);
-            LanguageService.ConnectionServiceInstance.TryFindConnection(scriptFile.ClientFilePath, out connInfo);
-
-            Location[] locations = LanguageService.Instance.GetDefinition(textDocumentPosition, scriptFile, connInfo);
-            if (locations != null)
+            if (WorkspaceService<SqlToolsSettings>.Instance.CurrentSettings.IsIntellisenseEnabled)
             {
-                await requestContext.SendResult(locations);
+                // Retrieve document and connection
+                ConnectionInfo connInfo;
+                var scriptFile = LanguageService.WorkspaceServiceInstance.Workspace.GetFile(textDocumentPosition.TextDocument.Uri);
+                LanguageService.ConnectionServiceInstance.TryFindConnection(scriptFile.ClientFilePath, out connInfo);
+
+                Location[] locations = LanguageService.Instance.GetDefinition(textDocumentPosition, scriptFile, connInfo);
+                if (locations != null)
+                {
+                    await requestContext.SendResult(locations);
+                }
             }
         }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -33,7 +33,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
     /// </summary>
     public sealed class LanguageService
     {
-        private const int ONE_SECOND = 1000;
+        private const int OneSecond = 1000;
 
         internal const string DefaultBatchSeperator = "GO";
 
@@ -43,9 +43,9 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         internal const int BindingTimeout = 500;
 
-        internal const int OnConnectionWaitTimeout = 300 * ONE_SECOND;
+        internal const int OnConnectionWaitTimeout = 300 * OneSecond;
 
-        internal const int PeekDefinitionTimeout = 10 * ONE_SECOND;
+        internal const int PeekDefinitionTimeout = 10 * OneSecond;
 
         private static ConnectionService connectionService = null;
 
@@ -299,7 +299,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         internal static async Task HandleDefinitionRequest(TextDocumentPosition textDocumentPosition, RequestContext<Location[]> requestContext)
         {
-            if (WorkspaceService<SqlToolsSettings>.Instance.CurrentSettings.IsIntellisenseEnabled)
+            if (WorkspaceService<SqlToolsSettings>.Instance.CurrentSettings.IsIntelliSenseEnabled)
             {
                 // Retrieve document and connection
                 ConnectionInfo connInfo;
@@ -667,7 +667,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
             if (RequiresReparse(scriptParseInfo, scriptFile))
             {
-                ParseAndBind(scriptFile, connInfo);
+                scriptParseInfo.ParseResult = ParseAndBind(scriptFile, connInfo);
             }
 
             // Get token from selected text
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -677,7 +677,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             }
             // Strip "[" and "]"(if present) from the token text to enable matching with the suggestions.
             // The suggestion title does not contain any sql punctuation
-            string tokenText = selectedToken.Text.Replace("]",String.Empty).Replace("[",String.Empty);
+            string tokenText = TextUtilities.RemoveSquareBracketSyntax(selectedToken.Text);
 
             if (scriptParseInfo.IsConnected && Monitor.TryEnter(scriptParseInfo.BuildingMetadataLock))
             {
@@ -739,7 +739,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             {
                 return null;
             }
-            return GetToken(scriptParseInfo, position.Line + 1, startColumn - 1).Text.Replace("]",String.Empty).Replace("[",String.Empty);
+            return TextUtilities.RemoveSquareBracketSyntax(GetToken(scriptParseInfo, position.Line + 1, startColumn - 1).Text);
         }
 
 
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -653,16 +653,6 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         /// <returns> Location with the URI of the script file</returns>
         internal Location[] GetDefinition(TextDocumentPosition textDocumentPosition, ScriptFile scriptFile, ConnectionInfo connInfo)
         {
-            Location[] locations = new[] { 
-                    new Location {
-                        Uri = new Uri("C:\\test.txt").AbsoluteUri,
-                        Range = new Range {
-                            Start = new Position { Line = 0, Character = 1},
-                            End = new Position { Line = 1, Character = 1}
-                        }
-                    }
-            };
-
             // Parse sql
             ScriptParseInfo scriptParseInfo = GetScriptParseInfo(textDocumentPosition.TextDocument.Uri);
             if (RequiresReparse(scriptParseInfo, scriptFile))
@@ -673,7 +663,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             // Return if parse failed
             if (scriptParseInfo == null)
             {
-                return locations;
+                return null;
             }
 
             // Get token from selected text
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -34,7 +34,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
     public sealed class LanguageService
     {
         private const int ONE_SECOND = 1000;
-        
+
         internal const string DefaultBatchSeperator = "GO";
 
         internal const int DiagnosticParseDelay = 750;
@@ -660,15 +660,14 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         {
             // Parse sql
             ScriptParseInfo scriptParseInfo = GetScriptParseInfo(textDocumentPosition.TextDocument.Uri);
-            if (scriptParseInfo != null && RequiresReparse(scriptParseInfo, scriptFile))
-            {
-                ParseAndBind(scriptFile, connInfo);
-            }
-
-            // Return if parse failed
             if (scriptParseInfo == null)
             {
                 return null;
+            }
+
+            if (RequiresReparse(scriptParseInfo, scriptFile))
+            {
+                ParseAndBind(scriptFile, connInfo);
             }
 
             // Get token from selected text
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -735,6 +735,10 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             string sql = scriptFile.GetLine(position.Line + 1);
             int startColumn = TextUtilities.PositionOfPrevDelimeter(sql, 0, position.Character);
             // if no schema name, returns null
+            if( sql[startColumn - 1] != '.')
+            {
+                return null;
+            }
             return GetToken(scriptParseInfo, position.Line + 1, startColumn - 1).Text.Replace("]",String.Empty).Replace("[",String.Empty);
         }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -33,6 +33,8 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
     /// </summary>
     public sealed class LanguageService
     {
+        private const int ONE_SECOND = 1000;
+        
         internal const string DefaultBatchSeperator = "GO";
 
         internal const int DiagnosticParseDelay = 750;
@@ -41,9 +43,9 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         internal const int BindingTimeout = 500;
 
-        internal const int OnConnectionWaitTimeout = 300000;
+        internal const int OnConnectionWaitTimeout = 300 * ONE_SECOND;
 
-        internal const int PeekDefinitionTimeout = 10000;
+        internal const int PeekDefinitionTimeout = 10 * ONE_SECOND;
 
         private static ConnectionService connectionService = null;
 
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -709,7 +709,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                             // Match token with the suggestions(declaration items) returned
                             DeclarationType type = 0;
                             string schemaName;
-                            PeekDefinition pd = new PeekDefinition(connInfo);
+                            PeekDefinition peekDefinition = new PeekDefinition(connInfo);
                             foreach (Declaration declarationItem in declarationItems)
                             {
                                 if (declarationItem.Title.Equals(tokenText))
@@ -719,13 +719,13 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                                     switch (type)
                                     {
                                         case DeclarationType.Table:
-                                            return pd.GetTableDefinition(tokenText);
+                                            return peekDefinition.GetTableDefinition(tokenText);
                                         case DeclarationType.View:
                                             schemaName = this.GetSchemaName(scriptParseInfo, textDocumentPosition.Position, scriptFile);
-                                            return pd.GetViewDefinition(tokenText, schemaName);
+                                            return peekDefinition.GetViewDefinition(tokenText, schemaName);
                                         case DeclarationType.StoredProcedure:
                                             schemaName = this.GetSchemaName(scriptParseInfo, textDocumentPosition.Position, scriptFile);
-                                            return pd.GetStoredProcedureDefinition(tokenText, schemaName);
+                                            return peekDefinition.GetStoredProcedureDefinition(tokenText, schemaName);
                                         default:
                                             return null;
                                     }
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -43,6 +43,8 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
         internal const int OnConnectionWaitTimeout = 300000;
 
+        internal const int PeekDefinitionTimeout = 10000;
+
         private static ConnectionService connectionService = null;
 
         private static WorkspaceService<SqlToolsSettings> workspaceServiceInstance;
@@ -198,7 +200,6 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             // Register the requests that this service will handle
 
             // turn off until needed (10/28/2016)
-            // serviceHost.SetRequestHandler(DefinitionRequest.Type, HandleDefinitionRequest);
             // serviceHost.SetRequestHandler(ReferencesRequest.Type, HandleReferencesRequest);
             // serviceHost.SetRequestHandler(SignatureHelpRequest.Type, HandleSignatureHelpRequest);
             // serviceHost.SetRequestHandler(DocumentHighlightRequest.Type, HandleDocumentHighlightRequest);
@@ -206,6 +207,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             serviceHost.SetRequestHandler(CompletionResolveRequest.Type, HandleCompletionResolveRequest);
             serviceHost.SetRequestHandler(HoverRequest.Type, HandleHoverRequest);
             serviceHost.SetRequestHandler(CompletionRequest.Type, HandleCompletionRequest);
+            serviceHost.SetRequestHandler(DefinitionRequest.Type, HandleDefinitionRequest);
 
             // Register a no-op shutdown task for validation of the shutdown logic
             serviceHost.RegisterShutdownTask(async (shutdownParams, shutdownRequestContext) =>
@@ -293,15 +295,22 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             }
         }
 
-// turn off this code until needed (10/28/2016)
-#if false
-        private static async Task HandleDefinitionRequest(
-            TextDocumentPosition textDocumentPosition,
-            RequestContext<Location[]> requestContext)
+        internal static async Task HandleDefinitionRequest(TextDocumentPosition textDocumentPosition, RequestContext<Location[]> requestContext)
         {
-            await Task.FromResult(true);
+            // Retrieve document and connection
+            ConnectionInfo connInfo;
+            var scriptFile = LanguageService.WorkspaceServiceInstance.Workspace.GetFile(textDocumentPosition.TextDocument.Uri);
+            LanguageService.ConnectionServiceInstance.TryFindConnection(scriptFile.ClientFilePath, out connInfo);
+
+            Location[] locations = LanguageService.Instance.GetDefinition(textDocumentPosition, scriptFile, connInfo);
+            if (locations != null)
+            {
+                await requestContext.SendResult(locations);
+            }
         }
 
+// turn off this code until needed (10/28/2016)
+#if false
         private static async Task HandleReferencesRequest(
             ReferencesParams referencesParams,
             RequestContext<Location[]> requestContext)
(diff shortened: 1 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +645,131 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+        /// </summary>
+        /// <param name="textDocumentPosition"></param>
+        /// <param name="scriptFile"></param>
+        /// <param name="connInfo"></param>
+        /// <returns> Location with the URI of the script file</returns>
+        internal Location[] GetDefinition(TextDocumentPosition textDocumentPosition, ScriptFile scriptFile, ConnectionInfo connInfo)
+        {
+            Location[] locations = new[] { 
+                    new Location {
+                        Uri = new Uri("C:\\test.txt").AbsoluteUri,
```
MitchellSternke: I'm guessing we shouldn't be using a test URI?

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -0,0 +1,228 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+#define LIVE_CONNECTION_TESTS
```
MitchellSternke: We define this in one of the build configurations at compile-time, so this probably shouldn't be defined here.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: Good catch! I meant to remove this.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
sharonravindran: I'll remove this then.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -43,6 +43,8 @@ public sealed class LanguageService
 
         internal const int OnConnectionWaitTimeout = 300000;
 
+        internal const int PeekDefinitionTimeout = 10000;
```
kevcunnane: >10000 [](start = 51, length = 5)

Minor: Is this 10 seconds? One thing we do in other projects is to define a constant for 1 second and use 10 * Seconds to do this kind of definition. It's handy as it makes clear the unit of measurement. Alternatively a comment also keep this clear.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -293,15 +295,22 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
             }
         }
 
-// turn off this code until needed (10/28/2016)
-#if false
-        private static async Task HandleDefinitionRequest(
-            TextDocumentPosition textDocumentPosition,
-            RequestContext<Location[]> requestContext)
+        internal static async Task HandleDefinitionRequest(TextDocumentPosition textDocumentPosition, RequestContext<Location[]> requestContext)
```
kevcunnane: Shouldn't we skip if intellisense is disabled just like for other request types like CompletionResolveRequest?

kevcunnane: Or else if IsQuickInfoEnabled is false? There should be some config flag that skips processing

---
In reply to: [89247609](https://github.com/Microsoft/sqltoolsservice/pull/160#discussion_r89247609) [](ancestors = 89247609)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +645,121 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+        {
+            // Parse sql
+            ScriptParseInfo scriptParseInfo = GetScriptParseInfo(textDocumentPosition.TextDocument.Uri);
+            if (RequiresReparse(scriptParseInfo, scriptFile))
+            {
+                ParseAndBind(scriptFile, connInfo);
+            }
+
+            // Return if parse failed
+            if (scriptParseInfo == null)
```
kevcunnane: This needs to be checked immediately after GetScriptParseInfo - that's the method that returns null. Otherwise the RequiresReparse check could cause a null ref as it accesses a null info object

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +645,121 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+            {
+                return null;
+            }
+
+            // Get token from selected text
+            Token token = GetToken(scriptParseInfo, textDocumentPosition.Position.Line + 1, textDocumentPosition.Position.Character);
```
kevcunnane: >token [](start = 18, length = 5)

This could use a more descriptive variable name - e.g. What is this token? Is it the schema name?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +645,121 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+            if (token == null)
+            {
+                return null;
+            }
+            // Strip "[" and "]"(if present) from the token text to enable matching with the suggestions.
+            // The suggestion title does not contain any sql punctuation
+            string tokenText = token.Text.Replace("]",String.Empty).Replace("[",String.Empty);
+
+
+            if (scriptParseInfo.IsConnected && Monitor.TryEnter(scriptParseInfo.BuildingMetadataLock))
```
kevcunnane: minor: remove unnecessary whitespace

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +645,121 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+                try
+                {
+                    // Queue the task with the binding queue    
+                    QueueItem queueItem = this.BindingQueue.QueueBindingOperation(
+                        key: scriptParseInfo.ConnectionKey,
+                        bindingTimeout: LanguageService.PeekDefinitionTimeout,
+                        bindOperation: (bindingContext, cancelToken) =>
+                        {
+                            // Get suggestions for the token
+                            IEnumerable<Declaration> declarationItems;
```
kevcunnane: >IEnumerable<Declaration> declarationItems; [](start = 28, length = 42)

Please join declaration and assignment here. E.g. IEnumerable<Declaration> declarationItems = Resolver.FindCompletions(...

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,171 @@
...
+using System;
+using System.IO;
+using System.Collections.Specialized;
+using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlTools.ServiceLayer.Connection;
+using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
+{
+    public class PeekDefinition
```
kevcunnane: Please document all public classes and methods

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +645,121 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+                                            return peekDefinition.GetTableDefinition(tokenText);
+                                        case DeclarationType.View:
+                                            schemaName = this.GetSchemaName(scriptParseInfo, textDocumentPosition.Position, scriptFile);
+                                            return peekDefinition.GetViewDefinition(tokenText, schemaName);
+                                        case DeclarationType.StoredProcedure:
+                                            schemaName = this.GetSchemaName(scriptParseInfo, textDocumentPosition.Position, scriptFile);
+                                            return peekDefinition.GetStoredProcedureDefinition(tokenText, schemaName);
+                                        default:
+                                            return null;
+                                    }
```
kevcunnane: This logic should be inside the PeekDefinition class. This service shouldn't know about these implementation details. Given you return null if not found, cleanest thing is move this info the peekDefinition class as-is for now and just call into it.
Separately, we're going to run into scale issues here - there are lots of types in SQL, so we shouldn't rely on switch statement as we'll have 100 entries before long. Options to replace:
- Use a Dictionary<Type,Func<string,string, Location[]> mapping type to a method that supports getting a definition.
- plenty of other patterns, but I'd guess this will work well enough for now. In the long term it might be nice to use dependency injection and have classes that register with supported types so the main logic class never needs to know about this, but it's overkill for now.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,171 @@
...
+            connectionInfo = connInfo;
+            tempPath = Path.GetTempPath();
+        }
+
+        /// <summary>
+        /// Script a table using SMO and write to a file.
+        /// </summary>
+        /// <param name="tableName">Table name</param>
+        /// <returns>Location object representing URI and range of the script file</returns>
+        internal Location[] GetTableDefinition(string tableName)
```
kevcunnane: Can't tables have schema names too?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,171 @@
...
+                                scriptFile.WriteLine(script);
+                            }  
+                        }
+
+                        
+                    }
+                    return GetLocationFromFile(tempFileName, 0);
+                }
+            }
+            return null;
```
kevcunnane: The code here is very duplicative for Table, View and Stored Procedure. We're going to have a lot more checks so we can simplify this dramatically. What's different is:
- SMO query to retries the object (table, view, proc)
- script.IndexOf check. The variable we check for is slightly different (basically string.format("CREATE {0}", typeName) should be used)

Based on this, please consider 1 common method that has the main logic, and takes as parameters 1) a getter for the SMO object and 2) The name to query for when trying to find the Create script. in scripts.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -0,0 +1,228 @@
...
+
+
+        /// <summary>
+        /// Tests the definition event handler. When called with no active connection, no definition is sent
+        /// </summary>
+        [Fact]
+        public void DefinitionsHandlerWithNoConnectionTest()
+        {
+
+            InitializeTestObjects();
```
kevcunnane: Minor: minimize unnecessary whitespace

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -0,0 +1,228 @@
...
+            // request the completion list
+            Task handleCompletion = LanguageService.HandleDefinitionRequest(textDocument, requestContext.Object);
+            handleCompletion.Wait(TaskTimeout);
+
+            // verify that send result was not called
+            requestContext.Verify(m => m.SendResult(It.IsAny<Location[]>()), Times.Never());
+
+        }
+
+#if LIVE_CONNECTION_TESTS
```
kevcunnane: When do these get run? Do your tests basically require real SMO as it's too hard to mock it out otherwise?

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// <summary>
+        /// Test get definition for a table object with active connection
+        /// </summary>
+        [Fact]
+        public void GetTableDefinitionTest()
+        {
+            ScriptFile scriptFile;
+            ConnectionInfo connInfo = TestObjects.InitLiveConnectionInfo(out scriptFile);
```
kevcunnane: What are the script contents? It's not clear what the source of this test is (e.g. source script, backing database) so without this how do I know that "test_table" is present?

kevcunnane:
:clock1:

test/Microsoft.SqlTools.ServiceLayer.TestDriver/Tests/LanguageServiceTests.cs
```diff
@@ -210,6 +210,49 @@ public class LanguageServiceTests
...
+                };
+
+                await testHelper.RequestOpenDocumentNotification(openParams);
+                  
+                Thread.Sleep(500);
+
+                bool connected = await testHelper.Connect(queryTempFile.FilePath, ConnectionTestUtils.LocalhostConnection);
+                Assert.True(connected, "Connection is successful");
+
+                Thread.Sleep(10000);
```
sharonravindran: We wait 10 secs for the language service binding. I'll try and optimize this.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: Modified code.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
sharonravindran: Yes, these need real SMO to test the scripting and the declarations. These get run when you build and run tests with the integration option. 
Yeah we should be making these required, there are a bunch of tests that are marked live connection.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: Yes, we should add the check here. There are no options for definition per se but I'll add the check for is intellisense enabled

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +648,102 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+
+            // Get token from selected text
+            Token selectedToken = GetToken(scriptParseInfo, textDocumentPosition.Position.Line + 1, textDocumentPosition.Position.Character);
+            if (selectedToken == null)
+            {
+                return null;
+            }
+            // Strip "[" and "]"(if present) from the token text to enable matching with the suggestions.
+            // The suggestion title does not contain any sql punctuation
+            string tokenText = selectedToken.Text.Replace("]",String.Empty).Replace("[",String.Empty);
```
kevcunnane: We need to handle escaped [ ] characters. It's possible to add a table called [Strange\[Name] for example. Please put this in a Utils file since we'll probably need to handle cases like this in other places.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +650,101 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+        /// <param name="position"></param>
+        /// <param name="scriptFile"></param>
+        /// <returns> schema nama</returns>
+        private string GetSchemaName(ScriptParseInfo scriptParseInfo, Position position, ScriptFile scriptFile)
+        {
+            // Get schema name 
+            string sql = scriptFile.GetLine(position.Line + 1);
+            int startColumn = TextUtilities.PositionOfPrevDelimeter(sql, 0, position.Character);
+            // if no schema name, returns null
+            if( sql[startColumn - 1] != '.')
```
kevcunnane: >sql[startColumn - 1] != '.') [](start = 16, length = 28)

We should get the last non-whitespace token instead of assuming the previous character works. The problem is that you can do "SELECT dbo. T1" with a space between the . and the table name. This is  valid and we have to cover these cases. Please also include a test or this scenario :)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// Peek Definition/ Go to definition implementation
+    /// Script sql objects and write create scripts to file
+    /// </summary>
+    internal class PeekDefinition
+    {
+        private ConnectionInfo connectionInfo;
+        private string tempPath;
+        
+        // Dictionary that holds the script getter for each type
+        private Dictionary<DeclarationType, Func<string, string, StringCollection>> sqlScriptGetters =
```
kevcunnane: >Func<string, string, StringCollection>> [](start = 44, length = 39)

Minor: we should probably declare this as a delegate (see [https://msdn.microsoft.com/en-us/library/ms173172.aspx](https://msdn.microsoft.com/en-us/library/ms173172.aspx)) to simplify this and shorten the definition. That would look like 
private delegate StringCollection GetScripts(string schemaName, string objectName). 
Also note that in this pattern, we put schemaName first, then objectName (such as table) as this is the standard we use in other codebases for looking up objects.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// <summary>
+        /// Add getters for each sql object supported by peek definition
+        /// </summary>
+        private void Initialize()
+        {
+            //Add script getters for each sql object
+
+            //Add tables to supported types
+            sqlScriptGetters.Add(DeclarationType.Table, GetTableScripts);
+            sqlObjectTypes.Add(DeclarationType.Table, "Table");
```
kevcunnane: When doing the codegen work you'll likely want to make a method that adds to both of these dictionaries in 1 operation. This will make it easier since there's less code and we guarantee that the value is in both dictionaries (which is assumed later on, but right now requires everyone to do the right thing when declaring a new handler)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,210 @@
...
+                    : database.Tables[tableName].Script();
+        }
+
+        /// <summary>
+        /// Script a view using SMO
+        /// </summary>
+        /// <param name="viewName">View name</param>
+        /// <param name="schemaName">Schema name </param>
+        /// <returns>String collection of scripts</returns>
+        internal StringCollection GetViewScripts(string viewName, string schemaName)
```
kevcunnane: >internal [](start = 8, length = 8)

Can't these be private now? You use the generic method in tests to get these so they no longer need internal accessibility.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.SqlTools.ServiceLayer.Connection;
+using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
+{
+    /// <summary>
+    /// Peek Definition/ Go to definition implementation
+    /// Script sql objects and write create scripts to file
+    /// </summary>
+    internal class PeekDefinition
```
kevcunnane: Is there a reason this isn't still public?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,210 @@
...
+
+                if (scripts != null)
+                {
+                    int lineNumber = 0;
+                    using (StreamWriter scriptFile = new StreamWriter(File.Open(tempFileName, FileMode.Create, FileAccess.ReadWrite)))
+                    {
+                        
+                        foreach (string script in scripts)
+                        {
+                            if (script.IndexOf(String.Format("CREATE {0}", objectType), StringComparison.OrdinalIgnoreCase) >= 0)
```
kevcunnane: >String [](start = 47, length = 6)

please use "string" instead of "String" as that's the code guideline we follow. It's nice if we keep this consistent

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,210 @@
...
+                            {
+                                scriptFile.WriteLine(script);
+                                lineNumber = GetStartOfCreate(script, String.Format("CREATE {0}", objectType));
```
kevcunnane: Again, please use "string" here

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -0,0 +1,236 @@
...
+            Location[] locations = peekDefinition.GetSqlObjectDefinition(peekDefinition.GetStoredProcedureScripts, objectName, schemaName, objectType);
+            Assert.NotNull(locations);
+            Cleanup(locations);
+        }
+
+        /// <summary>
+        /// Test get definition for a stored procedure object with active connection and no schema
+        /// </summary>
+        [Fact]
+        public void GetStoredProcedureDefinitionWithoutSchemaTest()
```
kevcunnane: Where is the test covering how we handle an unsupported object type? It is important to test negative scenarios and verify we don't crash but instead return empty results cleanly. WIth the new GetSqlObjectDefinition this should be possible? Or is this covered in a different set of tests?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: It's only expected to be called by the LanguageService so I thought I'll make it internal.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: I pass these getters to the generic function so they are still internal.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Modified

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Modified

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Modified

src/Microsoft.SqlTools.ServiceLayer/SqlContext/SqlToolsSettings.cs
```diff
@@ -90,6 +90,17 @@ public bool IsQuickInfoEnabled
                     && this.SqlTools.IntelliSense.EnableQuickInfo.Value;
             }
         }
+
+        /// <summary>
+        /// Gets a flag determining if IntelliSense is enabled
+        /// </summary>
+        public bool IsIntellisenseEnabled
```
kburtram: should be IsIntelliSenseEnabled

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// <param name="schemaName">Schema name or null</param>
+        /// <param name="objectType">Type of SQL object</param>
+        /// <returns>Location object representing URI and range of the script file</returns>
+        internal Location[] GetSqlObjectDefinition(
+                ScriptGetter sqlScriptGetter, 
+                string objectName, 
+                string schemaName, 
+                string objectType) 
+        {
+            if (this.connectionInfo.SqlConnection != null)
```
kburtram: why do we null check this here

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                if (scripts != null)
+                {
+                    int lineNumber = 0;
+                    using (StreamWriter scriptFile = new StreamWriter(File.Open(tempFileName, FileMode.Create, FileAccess.ReadWrite)))
+                    {
+                        
+                        foreach (string script in scripts)
+                        {
+                            if (script.IndexOf(string.Format("CREATE {0}", objectType), StringComparison.OrdinalIgnoreCase) >= 0)
```
kburtram: I'd assign string.Format("CREATE {0}", objectType) to a variable to avoid calling same function twice

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// <summary>
+        /// Script a table using SMO
+        /// </summary>
+        /// <param name="tableName">Table name</param>
+        /// <param name="schemaName">Schema name</param>
+        /// <returns>String collection of scripts</returns>
+        internal StringCollection GetTableScripts(string tableName, string schemaName)
+        {
+            return (schemaName != null) ? database.Tables[tableName, schemaName]?.Script()
+                    : database.Tables[tableName].Script();
```
kburtram: does need to be ?. like the other Tables access

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// </summary>
+        /// <param name="declarationItems"></param>
+        /// <param name="tokenText"></param>
+        /// <param name="schemaName"></param>
+        /// <returns>Location object of the script file</returns>
+        internal Location[] GetScript(IEnumerable<Declaration> declarationItems, string tokenText, string schemaName)
+        {
+            foreach (Declaration declarationItem in declarationItems)
+            {
+                if (declarationItem.Title.Equals(tokenText))
```
kburtram: Title can't be null?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+        internal delegate StringCollection ScriptGetter(string objectName, string schemaName);
+        
+        // Dictionary that holds the script getter for each type
+        private Dictionary<DeclarationType, ScriptGetter> sqlScriptGetters =
+            new Dictionary<DeclarationType, ScriptGetter>();
+
+        //Dictionary that holds the object name (as appears on the TSQL create statement)
+        private Dictionary<DeclarationType, string> sqlObjectTypes = new Dictionary<DeclarationType, string>();
+
+        private Database database
```
kburtram: property names should be Capitalized..so **private Database Database** would be a better definition here.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+            new Dictionary<DeclarationType, ScriptGetter>();
+
+        //Dictionary that holds the object name (as appears on the TSQL create statement)
+        private Dictionary<DeclarationType, string> sqlObjectTypes = new Dictionary<DeclarationType, string>();
+
+        private Database database 
+        {
+            get
+            {
+                Server server = new Server(this.connectionInfo.SqlConnection.DataSource);
```
kburtram: should we null check this.connectionInfo.SqlConnection here?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+                return server.Databases[this.connectionInfo.SqlConnection.Database];
+            }
+        }
+        
+        internal PeekDefinition(ConnectionInfo connInfo)
+        {
+            connectionInfo = connInfo;
```
kburtram: let's use a "this." prefix when access instance fields and properties.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -0,0 +1,218 @@
...
+            DirectoryInfo tempScriptDirectory = Directory.CreateDirectory( Path.GetTempPath()+ "mssql_definition");
```
kburtram: nit: spacing..should be **Directory.CreateDirectory(Path.GetTempPath() + "mssql_definition");**

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
kevcunnane: OK sounds good

---
In reply to: [90150351](https://github.com/Microsoft/sqltoolsservice/pull/160#discussion_r90150351) [](ancestors = 90150351)

erickangMSFT:
Please try Go To Definition and Peek Definition with the attached database.

1. Run 1 setup.sql to create ProductCatalog database
2. Run setup-temporal.sql to populate data
3. Using temporal.sql connected to ProductCatalog database, try the features.

It seems IntelliSense suggestion list and Go To Definition and Peek Definition do not function as expected.
The ProductCatalog database is a sample demo from https://github.com/Microsoft/sql-server-sample 
[4 temporal.zip](https://github.com/Microsoft/sqltoolsservice/files/623550/4.temporal.zip)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Yes. I should have move the check up when I refactored the code. Modifying this.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -636,6 +650,110 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
...
+            string tokenText = TextUtilities.RemoveSquareBracketSyntax(selectedToken.Text);
+
+            if (scriptParseInfo.IsConnected && Monitor.TryEnter(scriptParseInfo.BuildingMetadataLock))
+            {
+                try
+                {
+                    // Queue the task with the binding queue    
+                    QueueItem queueItem = this.BindingQueue.QueueBindingOperation(
+                        key: scriptParseInfo.ConnectionKey,
+                        bindingTimeout: LanguageService.PeekDefinitionTimeout,
```
sharonravindran: Nothing. The users sees the peek window on return. I've set it to ten seconds to give it enough time to find resolutions(worst case).

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: I think if IntelliSense is disabled in general then PeekDefinition should be disabled too but the menu option that doesn't do anything is not good. I'll ask Eric what he wants.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: This check is here since the database get doesn't check for null. I'll refactor the checks.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
sharonravindran: I've added the test for the unsupported scenario

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
kburtram: this will block Intellisense during this time which might be not the best UX.  We can start with this and see how it works.

PR #162, opened 2016-11-28
Remove enabling MARS

kburtram:
The change to enable MARS merged down from master, so we need to remove it again.


llali:
:shipit:

PR #164, opened 2016-12-02
Intellisense for when the text starts with bracket

llali:
- Added '[' as completion trigger character 
- Moved the code for creating completion item out of AutoCompleteHelper and added tests for the publish functions

- Added brackets to the completion item label when the given token text has bracket

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/SqlCompletionItemTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/SqlCompletionItemTests.cs
@@ -1,0 +1,179 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlServer.Management.SqlParser.Intellisense;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices.Contracts;
+using Xunit;
+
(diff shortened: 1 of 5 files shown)


test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/SqlCompletionItemTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            SqlCompletionItem item = new SqlCompletionItem(declarationTitle, declarationType, tokenText);
+            CompletionItem completionItem = item.CreateCompletionItem(0, 1, 2);
+
+            Assert.Equal(completionItem.Label, expected);
+            Assert.Equal(completionItem.InsertText, expected);
+            Assert.Equal(completionItem.Detail, expected);
+        }
+
+        [Fact]
+        public void LabelShouldIncludeBracketGivenTokenWithBrackets()
```
kevcunnane: I think we should have a test covering escaped brackets. E.g. what happens in the following case? Will auto-complete work correctly? So long as this passes I'm happy to sign off

```

-- Create the table in the specified schema
CREATE TABLE [Bracket\[]
(
    C1 INT NOT NULL PRIMARY KEY, -- primary key column
    C2 [NVARCHAR](50) NOT NULL,
    C3 [NVARCHAR](50) NOT NULL
   -- specify more columns here
);
GO
select C1 from dbo.[Bracket\[] 
```

wpostma: Wow that's awesome. Left Square Bracket as the name of a table.  Just before I quit somewhere, I'm going to introduce that as a table name.

PR #166, opened 2016-12-03
Reenable second test driver execution

kburtram:
I am not sure the previous code coverage fix was correct.  This change adds ~2.5% coverage when I run code coverage locally on my machine.  The -mergeoutput argument should prevent one test run from overwriting the previous test runs.


kburtram:
Closing since a fix was applied in separate PR.

PR #167, opened 2016-12-05
adding a new event for when definition is requested

llali:
- Calculating number of milliseconds that will take to create auto complete list and send back to client, that amount will be aggregated in each bucket and then the result will be sent to client when the connection is disconnected.  These aggregated numbers will be created for different file size groups, There are 4 file size groups are defined now. {100, 1000, 2000 and 5000} the number defines the number of characters in the file. for example for file size group 2000 (the files will less than 1000 characters and more than 100) the telemetry event will be sent that includes the number of milliseconds needed to create auto complete list. 

  - Refactored Language server and moved the code to create auto complete to CompletionService and added unit tests

 - Created a wrapper class for SQL parser to be able to create mock in the tests

- added a general telemetry notification for sending telemetry request to client

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -358,21 +358,17 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                 return false;
             }
 
-            foreach (var item in info.IntellisenseMetrics.Quantile)
+            // Send a telemetry notification for intellisense performance metrics
+            ServiceHost.SendEvent(TelemetryNotification.Type, new TelemetryParams()
             {
-                // Send a telemetry notification for intellisense performance metrics
-                ServiceHost.SendEvent(TelemetryNotification.Type, new TelemetryParams()
-                {
-                    Properties = new Dictionary<string, string>
+                Properties = new Dictionary<string, string>
                     {
-                        { "IsAzure", info.IsAzure ? "1" : "0" },
-                        { "FileSize", item.Key }
+                        { "IsAzure", info.IsAzure ? "1" : "0" }
                     },
-                    EventName = TelemetryEvenNames.IntellisensePersentile,
-                    Measures = item.Value.Quantile
-                });
-            }
-            
+                EventName = TelemetryEvenNames.IntellisensePersentile,
+                Measures = info.IntellisenseMetrics.Quantile
+            });
+
 
             // Close the connection            
             info.SqlConnection.Close();
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -361,12 +361,15 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             // Send a telemetry notification for intellisense performance metrics
             ServiceHost.SendEvent(TelemetryNotification.Type, new TelemetryParams()
             {
-                Properties = new Dictionary<string, string>
+                Params = new TelemetryProperties
+                {
+                    Properties = new Dictionary<string, string>
                     {
                         { "IsAzure", info.IsAzure ? "1" : "0" }
                     },
-                EventName = TelemetryEvenNames.IntellisensePersentile,
-                Measures = info.IntellisenseMetrics.Quantile
+                    EventName = TelemetryEventNames.IntellisenseQuantile,
+                    Measures = info.IntellisenseMetrics.Quantile
+                }
             });
 
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -15,6 +15,7 @@ using System.Threading.Tasks;
 using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection;
 using Microsoft.SqlTools.ServiceLayer.Hosting.Protocol;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices.Contracts;
 using Microsoft.SqlTools.ServiceLayer.SqlContext;
 using Microsoft.SqlTools.ServiceLayer.Utility;
 using Microsoft.SqlTools.ServiceLayer.Workspace;
@@ -290,6 +291,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                     AzureVersion = serverInfo.AzureVersion,
                     OsVersion = serverInfo.OsVersion
                 };
+                connectionInfo.IsAzure = serverInfo.IsCloud;
             }
             catch(Exception ex)
             {
@@ -355,6 +357,22 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             {
                 return false;
             }
+
+            foreach (var item in info.IntellisenseMetrics.Quantile)
+            {
+                // Send a telemetry notification for intellisense performance metrics
+                ServiceHost.SendEvent(TelemetryNotification.Type, new TelemetryParams()
+                {
+                    Properties = new Dictionary<string, string>
+                    {
+                        { "IsAzure", info.IsAzure ? "1" : "0" },
+                        { "FileSize", item.Key }
+                    },
+                    EventName = TelemetryEvenNames.IntellisensePersentile,
+                    Measures = item.Value.Quantile
+                });
+            }
+            
 
             // Close the connection            
             info.SqlConnection.Close();
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/AutoCompletionResult.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/AutoCompletionResult.cs
@@ -1,0 +1,68 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.SqlServer.Management.SqlParser.Intellisense;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion
+{
+    /// <summary>
+    /// Includes the objects created by auto completion service
+    /// </summary>
+    public class AutoCompletionResult
+    {
+        /// <summary>
+        /// Creates new instance
+        /// </summary>
+        public AutoCompletionResult()
+        {
+            Stopwatch = new Stopwatch();
+            StartProcesssing();
+        }
+
+        /// <summary>
+        /// Starts processing the result
+        /// </summary>
+        public void StartProcesssing()
+        {
+            Stopwatch.Start();
+        }
+
+        private Stopwatch Stopwatch { get; set; }
+
+        /// <summary>
+        /// Completes the results to calculate the duration
+        /// </summary>
+        public void CompleteResult(IEnumerable<Declaration> suggestions, CompletionItem[] completionItems)
+        {
+            Stopwatch.Stop();
+            Suggestions = suggestions;
+            CompletionItems = completionItems;
+        }
+
+        /// <summary>
+        /// The number of milliseconds to process the result
+        /// </summary>
+        public double Duration
+        {
+            get
+            {
+                return Stopwatch.ElapsedMilliseconds;
+            }
+        }
+        
+        /// <summary>
+        /// Suggestion list
+        /// </summary>
+        public IEnumerable<Declaration> Suggestions { get; private set; }
+
+        /// <summary>
+        /// Completion list
+        /// </summary>
+        public CompletionItem[] CompletionItems { get; private set; }
+    }
+}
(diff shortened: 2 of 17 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Definition.cs
```diff
@@ -14,5 +14,22 @@ public static readonly
...
+    /// Parameters sent back with a definition event
+    /// </summary>
+    public class DefinitionSentParams
+    {
+    }
+
+    /// <summary>
+    /// Event sent when the language service sent the definition
+    /// </summary>
+    public class DefinitionSentNotification
```
kevcunnane: I was thinking more of a generic SendTelemetryNotification that asks for telemetry to be published? This would make it much cheaper to add new telemetry from the Service side since we can add them there and not update the VSCode-MSSQL side at all. What do you think?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -356,6 +358,22 @@ public bool Disconnect(DisconnectParams disconnectParams)
...
+            {
+                // Send a telemetry notification for intellisense performance metrics
+                ServiceHost.SendEvent(TelemetryNotification.Type, new TelemetryParams()
+                {
+                    Properties = new Dictionary<string, string>
+                    {
+                        { "IsAzure", info.IsAzure ? "1" : "0" },
+                        { "FileSize", item.Key }
+                    },
+                    EventName = TelemetryEvenNames.IntellisensePersentile,
```
kburtram: >Persentile [](start = 63, length = 10)

typo ('s' to 'c')..IntellisensePercentile

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/InteractionMetrics.cs
```diff
@@ -0,0 +1,97 @@
...
+        public InteractionMetrics(int[] metrics)
+        {
+            Validate.IsNotNull("metrics", metrics);
+            if(metrics.Length == 0)
+            {
+                throw new ArgumentOutOfRangeException("metrics");
+            }
+
+            Counters = new ConcurrentDictionary<string, T>();
+            Metrics = IsSorted(metrics) ? metrics : metrics.OrderBy( i => i).ToArray();
```
kburtram: >metrics.OrderBy( i => i).ToArray( [](start = 52, length = 33)

If you do an in-place sort it looks like you could avoid an unnecessary additional array allocation..e.g., [https://msdn.microsoft.com/en-us/library/kwx6zbd4(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/kwx6zbd4(v=vs.110).aspx)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/InteractionMetrics.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Update metric value given new number
+        /// </summary>
+        public void UpdateMetrics(double duration, T newValue, Func<string, T, T> updateValueFactory)
+        {
+           Task.Factory.StartNew(() =>
```
kburtram: >Task.Factory.StartNew [](start = 11, length = 21)

any reason we run this as a task...could it take a long time to run?  I don't see any locking, is it thread safe?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/InteractionMetrics.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        /// <summary>
+        /// Returns the quantile
+        /// </summary>
+        public Dictionary<string, T> Quantile
+        {
+            get
+            {
+                return Counters.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
kburtram: >Counters.ToDictionary [](start = 23, length = 21)

any reason not to return Counters directly here?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/InteractionMetrics.cs
llali: Counters is ConcurrentDictionary so it's thread safe and debugging was easier this way because without it the completion list would time out most of the times  

---
In reply to: [91594323](https://github.com/Microsoft/sqltoolsservice/pull/167#discussion_r91594323) [](ancestors = 91594323)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/InteractionMetrics.cs
llali: I didn't want to return the Concurrent Dictionary

---
In reply to: [91594488](https://github.com/Microsoft/sqltoolsservice/pull/167#discussion_r91594488) [](ancestors = 91594488)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -871,12 +875,19 @@ internal SignatureHelp GetSignatureHelp(TextDocumentPosition textDocumentPositio
...
                                    endColumn + 1,
                                    bindingContext.MetadataDisplayInfoProvider);
-                                
-                                // convert from the parser format to the VS Code wire format
-                                return AutoCompleteHelper.ConvertMethodHelpTextListToSignatureHelp(methods, 
-                                    methodLocations,
-                                    startLine + 1,
-                                    endColumn + 1);
+
+                                if (methodLocations != null)
```
kburtram: >if (methodLocations != null) [](start = 32, length = 28)

you may push the null check into the helper method..this method is already pretty long so additional conditionals and returns adds to that complexity a little.

llali: there's a null check in the method and it throws exception when it's null so when I debugging, I was getting so many exceptions because for some reason this is null many times 

---
In reply to: [91594909](https://github.com/Microsoft/sqltoolsservice/pull/167#discussion_r91594909) [](ancestors = 91594909)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/AutoCompletionResult.cs
```diff
@@ -0,0 +1,68 @@
...
+        public AutoCompletionResult()
+        {
+            Stopwatch = new Stopwatch();
+            StartProcesssing();
+        }
+
+        /// <summary>
+        /// Starts processing the result
+        /// </summary>
+        public void StartProcesssing()
```
kburtram: >public [](start = 8, length = 6)

I couldn't see this method being used outside the class (may have missed it though)..if not used outside you could make private or just move Stopwatch.Start into constructor directly.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
```diff
@@ -0,0 +1,172 @@
...
+        private QueueItem AddToQueue(
+            ConnectionInfo connInfo,
+            ScriptParseInfo scriptParseInfo,
+            ScriptDocumentInfo scriptDocumentInfo,
+            bool useLowerCaseSuggestions)
+        {
+            // queue the completion task with the binding queue    
+            QueueItem queueItem = this.BindingQueue.QueueBindingOperation(
+                key: scriptParseInfo.ConnectionKey,
+                bindingTimeout: 60000,
```
kburtram: >60000 [](start = 32, length = 5)

the timeout is going to be different based on which type of operation is being queued.  this should be a parameter somewhere.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
```diff
@@ -0,0 +1,172 @@
...
+                    // wait for the queue item
+                    queueItem.ItemProcessed.WaitOne();
+                    var completionResult = queueItem.GetResultAsT<AutoCompletionResult>();
+                    if (completionResult != null && completionResult.CompletionItems != null)
+                    {
+                        result = completionResult;
+                    }
+                    else if (!ShouldShowCompletionList(scriptDocumentInfo.Token))
+                    {
+                        result.CompleteResult(scriptDocumentInfo.ScriptParseInfo.CurrentSuggestions, AutoCompleteHelper.EmptyCompletionList);
```
kburtram: >scriptDocumentInfo.ScriptParseInfo.CurrentSuggestions, [](start = 46, length = 54)

what are we doing here?  If I recall correctly CurrentSuggestions was a cache to assist in the resolve completion callback which doesn't have the script context available.  What purpose it it serving here?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
```diff
@@ -0,0 +1,166 @@
...
+                    case (int)Tokens.LEX_MULTILINE_COMMENT:
+                    case (int)Tokens.LEX_END_OF_LINE_COMMENT:
+                        result = false;
+                        break;
+                }
+            }
+            return result;
+        }
+
+        private ConnectedBindingQueue BindingQueue { get; set; }
```
kburtram: >private ConnectedBindingQueue BindingQueue  [](start = 8, length = 43)

can we move this toward the top of class with other properties?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        private AutoCompletionResult CreateDefaultCompletionItems(ScriptParseInfo scriptParseInfo, ScriptDocumentInfo scriptDocumentInfo, bool useLowerCaseSuggestions)
+        {
+            AutoCompletionResult result = new AutoCompletionResult();
+            CompletionItem[] completionList = AutoCompleteHelper.GetDefaultCompletionItems(scriptDocumentInfo, useLowerCaseSuggestions);
```
kburtram: >AutoCompleteHelper [](start = 46, length = 18)

I think the refactor starts to mix-up responsibilities between the AutoCompleteHelper and the CompletionService..i.e., not really sure what should go in which class.  It's probably fine for this check-in, but we should rationalize why we have to similar classes and when to use which one, or should we get rid of the helper and move that functionality into here.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/TelemetryNotification.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /// <summary>
+    /// Event sent when the language service needs to add a telemetry event
+    /// </summary>
+    public class TelemetryNotification
+    {
+        public static readonly
+            EventType<TelemetryParams> Type =
+            EventType<TelemetryParams>.Create("textDocument/telemetry");
```
kburtram: >textDocument/telemetry [](start = 47, length = 22)

did we consider using the language service protocol "telemetry/event" event type?  [https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#telemetry_event](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#telemetry_event)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/TelemetryNotification.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /// <summary>
+    /// List of telemetry events
+    /// </summary>
+    public static class TelemetryEvenNames
+    {
+        /// <summary>
+        /// telemetry even name for auto complete response time
```
kburtram: >even [](start = 22, length = 4)

typo..event (same in comment below and in class name.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/TelemetryNotification.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        public const string IntellisensePersentile = "IntellisenseQuantile";
```
kburtram: >s [](start = 43, length = 1)

's' should be 'c'

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
```diff
@@ -0,0 +1,166 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Generic;
```
kburtram: just a general comment is that there's quite a bit of refactoring and probably some functional changes as well, this code is pretty hard to test all the various cases (timeouts, errors, etc.).  if we want to take for preview #2 release let's get this in and spend some focused testing time to make sure nothing regressed.

kburtram:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
llali: actually I don't need it I'll remove it 

---
In reply to: [91597181](https://github.com/Microsoft/sqltoolsservice/pull/167#discussion_r91597181) [](ancestors = 91597181)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -356,6 +358,18 @@ public bool Disconnect(DisconnectParams disconnectParams)
...
             }
 
+            // Send a telemetry notification for intellisense performance metrics
+            ServiceHost.SendEvent(TelemetryNotification.Type, new TelemetryParams()
+            {
+                Properties = new Dictionary<string, string>
+                    {
+                        { "IsAzure", info.IsAzure ? "1" : "0" }
+                    },
+                EventName = TelemetryEvenNames.IntellisensePersentile,
```
kevcunnane: Fix typos :-)

PR #169, opened 2016-12-06
Batch Start Notification

benrr101:
This change is part of the progressive results code. It will submit a notification from the service layer to indicate when execution of a batch has completed. This notification will contain the selection for batch, execution start time, and its ID. This will enable the extension to produce a header for the batch before the batch completes, in order to make it more clear to the user that execution is going on.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -184,7 +184,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
         {
             get
             {
-                // Basic summary with basic information
+                // Batch summary with information available at start
                 BatchSummary summary = new BatchSummary
                 {
                     HasError = HasError,
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -75,7 +75,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
             this.outputFileFactory = outputFileFactory;
         }
 
-        #region Properties
+        #region Events
 
         /// <summary>
         /// Asynchronous handler for when batches are completed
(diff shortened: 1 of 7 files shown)


benrr101:
Ping! Can someone take a look a this?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -175,17 +184,25 @@ public BatchSummary Summary
         {
             get
             {
-                return new BatchSummary
+                // Basic summary with basic information
```
kevcunnane: Minor: that's a lot of "basics" in 1 comment. Just "basic summary with key information"?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -100,9 +100,14 @@ public Query(string queryText, ConnectionInfo connection, QueryExecutionSettings
         public event Batch.BatchAsyncEventHandler BatchCompleted;
 
         /// <summary>
+        /// Event to be called when a batch starts execution;
```
kevcunnane: Minor: "." not ";" at end of comment

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Execution/ServiceIntegrationTests.cs
```diff
@@ -49,14 +51,23 @@ public class ServiceIntegrationTests
...
-            VerifyQueryExecuteCallCount(requestContext, Times.Once(), Times.Once(), Times.Once(), Times.Never(), Times.Never());
+            VerifyQueryExecuteCallCount(requestContext, Times.Once(), Times.Once(), Times.Once(), Times.Once(), Times.Never(), Times.Never());
             Assert.Null(result.Messages);
 
             Assert.Equal(1, completeParams.BatchSummaries.Length);
             Assert.Empty(completeParams.BatchSummaries[0].ResultSetSummaries);
             Assert.NotEmpty(completeParams.BatchSummaries[0].Messages);
 
+            // ... Batch start summary should not contain result sets, messages, but should contain owner URI
+            Assert.NotNull(batchStartParams);
```
kevcunnane: Minor: you're doing a lot of the same checks in multiple tests (for batchStart, always verify no messages, summaries, but verify URI is set as expected). Could we have some common VerifyBatchStart methods that you call instead? The tests are pretty long and again if you need a new standard property test in the future, you'd update less places.

benrr101: I agree. It mostly has to do with the weirdness of having a callback that sets a variable. I'll see if I can improve upon it in the future, but for now it's a bit of a large change to make.

PR #173, opened 2016-12-09
Adding Milliseconds to DateTime fields

benrr101:
This is a slightly larger change than anticipated due the difference between `DATETIME`, `DATETIME2`, and `DateTime`. The `DATETIME` type always uses 3 decimal points of a second, while the `DATETIME2` type has 7 (although since `DATETIME2(7)` is default in SSMS suggesting that it is a variable precision type). Regardless of the db type, the engine returns `DateTime` C# type. The db types are only made visible via the column info, via the numeric precision and numeric scale. My findings were as such:
`DATETIME `: Precision = 23, Scale = 3
`DATETIME2`: Precision = 255, Scale = 7

The scale corresponds neatly with the number of second decimal points to show. The buffer file writer was modified to store both the scale and the number of ticks. Then the buffer file reader was modified to read in the precision and the number of ticks and generate the ToString version of the DateTime to add "f" as many times as there is scale, which corresponds to milliseconds.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
@@ -260,11 +260,21 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
         /// <returns>A DateTime</returns>
         public FileStreamReadResult ReadDateTime(long offset)
         {
-            return ReadCellHelper(offset, length =>
-            {
-                long ticks = BitConverter.ToInt64(buffer, 0);
-                return new DateTime(ticks);
-            });
+            int precision = 0;
+
+            return ReadCellHelper(offset,
+                length =>
+                {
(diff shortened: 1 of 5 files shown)


benrr101:
Anyone who wanted to see what it looks like now
millisecondview

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -260,11 +260,21 @@ public FileStreamReadResult ReadDecimal(long offset)
...
+            return ReadCellHelper(offset,
+                length =>
+                {
+                    precision = BitConverter.ToInt32(buffer, 0);
+                    long ticks = BitConverter.ToInt64(buffer, 4);
+                    return new DateTime(ticks);
+                }, null,
+                time =>
+                {
+                    // Output the number milliseconds equivalent to the precision
```
kevcunnane: For the dumb among us, mention that new string('f', precision) output ffff for precision = 4

kevcunnane:
:shipit:

benrr101:
For proof that the DateTime2(0) works
vscodedatetime2 0

PR #174, opened 2016-12-09
Feature/delete peek scripts

sharonravindran:
Deletes the temp folder containing peek definition scripts during shutdown. Closes Microsoft/vscode-mssql/380


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1293,5 +1295,14 @@ private bool IsPreviewWindow(ScriptFile scriptFile)
                 return false;
             }
         }
+
+        internal void DeletePeekDefinitionScripts()
+        {
+            // Delete temp folder created to store peek definition scripts
+            if (FileUtils.SafeDirectoryExists(FileUtils.PeekDefinitionTempFolder))
```
kevcunnane: As mentioned offline, let's ensure that each workspace has its own temp folder by creating a sub-folder for this the 1st time we need to save a file. Otherwise if you have 2 workspaces open and close one, the definitions files are deleted for the other workspace

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
```diff
@@ -8,6 +8,46 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
...
+                try
+                {
+                    // use tempDirectory name created previously
+                    DirectoryInfo tempScriptDirectory = Directory.CreateDirectory(FileUtils.PeekDefinitionTempFolder);
+                    tempPath = tempScriptDirectory.FullName;
+                }
+                catch (Exception)
+                {
+                    // swallow exception and use temp folder to store scripts
+                    tempPath = Path.GetTempPath();
```
benrr101: If this condition happens, then when the service shuts down, the user's entire temp directory will attempt to be wiped out. That's a big deal!

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
```diff
@@ -8,6 +8,46 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
...
+                    // create new temp folder
+                    string tempFolder = string.Format("{0}_{1}", FileUtils.PeekDefinitionTempFolder, DateTime.Now.ToString("yyyyMMddHHmmssffff"));
+                    DirectoryInfo tempScriptDirectory = Directory.CreateDirectory(tempFolder);
+                    FileUtils.PeekDefinitionTempFolder = tempPath = tempScriptDirectory.FullName;
+                    PeekDefinitionTempFolderCreated = true;
+                }
+                catch (Exception)
+                {
+                    // swallow exception and use temp folder to store scripts
+                    tempPath = Path.GetTempPath();
```
benrr101: Same issue here, if this condition happens, then when the service shuts down, the user's entire temp directory will attempt to be wiped out. That's a big deal!

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
```diff
@@ -8,6 +8,46 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
...
+                }
+            }
+            else
+            {
+                try
+                {
+                    // use tempDirectory name created previously
+                    DirectoryInfo tempScriptDirectory = Directory.CreateDirectory(FileUtils.PeekDefinitionTempFolder);
```
benrr101: If the path already exists, we want to just get the info for the path instead of create the folder anew, right? We can do that with `new DirectoryInfo(FileUtils.PeekDefinitionTempFolder)`

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
```diff
@@ -8,6 +8,46 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
 {
     internal static class FileUtils
     {
+        internal static string PeekDefinitionTempFolder = Path.GetTempPath() + "mssql_definition";
```
benrr101: Since all this code is prefixed with PeekDefinition, I think it would work better to put it in the language service instead of the globally available FileUtils class.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -184,6 +185,33 @@ public void GetSchemaFromDatabaseQualifiedNameWithInvalidNameTest()
...
+        /// </summary>
+        [Fact]
+        public void DeletePeekDefinitionScriptsWhenFolderDoesNotExistTest()
+        {
+            var languageService = LanguageService.Instance;
+            PeekDefinition peekDefinition = new PeekDefinition(null);
+            FileUtils.SafeDirectoryDelete(FileUtils.PeekDefinitionTempFolder, true);
+            Assert.False(Directory.Exists(FileUtils.PeekDefinitionTempFolder));
+            // Expected not to throw any exception
+            languageService.DeletePeekDefinitionScripts();
```
benrr101: just a suggestion, for future unit tests on this feature, we should try to create mock that will keep us from writing to the filesystem. Cleaning up after filesystem writes can be a huge pain for unit tests, if we can avoid it, let's do so.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -184,6 +185,33 @@ public void GetSchemaFromDatabaseQualifiedNameWithInvalidNameTest()
             string actualSchemaName = peekDefinition.GetSchemaFromDatabaseQualifiedName(validDatabaseQualifiedName, objectName);
             Assert.Equal(actualSchemaName, expectedSchemaName);
         }
+
+        /// <summary>
+        /// Test Deletion of peek definition scripts for a valid temp folder that exists
+        /// </summary>
+        [Fact]
```
benrr101: I'd really like to see more unit tests that check the various branches of code that were added here. Especially the branches for using TempPath and what happens on service shutdown.

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
sharonravindran: No, shutdown only attempts to delete FileUtils.PeekDefinitionTempFolder which is never set to the entire temp directory.

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
sharonravindran: CreateDirectory does the same thing. It doesn't create if the directory already exists

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
```diff
@@ -8,6 +8,46 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
...
+        internal static string GetPeekDefinitionTempFolder()
+        {
+            string tempPath;
+            if (!PeekDefinitionTempFolderCreated)
+            {
+                
+                try
+                {
+                    // create new temp folder
+                    string tempFolder = string.Format("{0}_{1}", FileUtils.PeekDefinitionTempFolder, DateTime.Now.ToString("yyyyMMddHHmmssffff"));
```
sharonravindran: Temp folder.  It will create a subfolder under the temp folder

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
sharonravindran: No, shutdown only attempts to delete FileUtils.PeekDefinitionTempFolder which is never set to the entire temp directory.  When this condition happens we don't set  FileUtils.PeekDefinitionTempFolder  to take the value of the temp folder

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1293,5 +1295,14 @@ private bool IsPreviewWindow(ScriptFile scriptFile)
...
+            {
+                FileUtils.SafeDirectoryDelete(FileUtils.PeekDefinitionTempFolder, true);
```
sharonravindran: True. This just seemed cleaner

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -184,6 +185,29 @@ public void GetSchemaFromDatabaseQualifiedNameWithInvalidNameTest()
...
+            Assert.False(Directory.Exists(FileUtils.PeekDefinitionTempFolder));
+        }
+
+        /// <summary>
+        /// Test Deletion of peek definition scripts for a temp folder that does not exist
+        /// </summary>
+        [Fact]
+        public void DeletePeekDefinitionScriptsWhenFolderDoesNotExistTest()
+        {
+            var languageService = LanguageService.Instance;
```
sharonravindran: Added check

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
sharonravindran: We create a folder in the temp directory. We don't create a folder in the workspace

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
benrr101: In that case it's a minor change, but it makes it clearer that we're not creating the directory here, just getting information on it.

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
benrr101: Ok, that's good then!

PR #175, opened 2016-12-09
Fixed null ref exception in GetSignatureHelp

coquagli:
- Fixed a few places where ScriptParseInfo was being used before it was verified to be non-null
- Added 1 basic test for GetSignatureHelp scenario


benrr101:
Looks good to me, :shipit:

PR #176, opened 2016-12-10
Reuse query connection

coquagli:
Fixes [483](https://github.com/Microsoft/vscode-mssql/issues/483) in vscode-mssql.

- Replace the `DbConnection `with `Dictionary<ConnectionType, DbConnection>` in `ConnectionInfo` so that any number of simultaneous connections can be supported
- Change `ConnectionService` to support conneciton, disconneciton, and cancellation of any number of simultanious connecitons. 
- Add `ConnectionType.cs`, and add a `ConnectionType` variable to numerous parameter classes
- Change query execution so the first query for a URI will open a new `DbConnection` dedicated to query execution. All subsequent queries on that URI use the same `DbConnection`
- Change existing tests to reflect connection logic changes
- Add new tests for transactions, temp tables, and multiple `DbConnection` logic

I continued to use a `DbConnection` for the connection. I noticed that some other classes are using `SqlConnection` instead of `DbConnection`. Notably `ConnectedBindingQueue` (which is reused in `PeekDefinition`), `DbConnectionWrapper`, and `ReliableSqlConnection`.


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -52,6 +52,12 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
         public DbConnection SqlConnection { get; set; }
 
         /// <summary>
+        /// The connection to the SQL database that queries will be run against.
+        /// A single instance is created and re-used after the first query is run 
+        /// </summary>
+        public DbConnection QueryConnection { get; set; }
```
kburtram: >QueryConnection [](start = 28, length = 15)

don't we need one connection per SQL file?  The connectioninfo object is shared across multiple editor windows.

kburtram:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
coquagli: Yes, good catch.

benrr101: Karl's comment needs us to rethink the solution a bit. Apologies for leading you down the wrong path on this one.

coquagli: I haven't fully investigated everything, but it looks like one ConnectionInfo is created per SQL file, so this actually (surprisingly) creates one QueryConnection per SQL file. 

It looks like the same connection is reused when the same SQL file is opened in 2 separate tabs. However, when disconnecting then reconnecting, a crash occurs. I'll look into this more tomorrow.

coquagli:
I just did some testing:
- Running `BEGIN TRANSACTION` in one query then `ROLLBACK TRANSACTION` in another query in the same file succeeds
- Running `BEGIN TRANSACTION` in one query then `ROLLBACK TRANSACTION` in another query in the same file opened in another window succeeds
- Running `BEGIN TRANSACTION` in one query then `ROLLBACK TRANSACTION` in another query in a different file fails

I think this is the behavior we're looking for?

coquagli:
I am going to add some more tests for connecting and disconnecting scenarios.

coquagli:
Pushing some unfinished local changes to the server in order to not lose them over the holidays in case of a machine failure.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -358,6 +403,37 @@ public bool Disconnect(DisconnectParams disconnectParams)
                 return false;
             }
 
+            // If we are disconnecting all existing connections for the provided URI
+            if (disconnectParams.Type == ConnectionType.AllConnections)
+            {
+                foreach (KeyValuePair<ConnectionType, DbConnection> entry in info.ConnectionTypeToConnectionMap)
+                {
+                    entry.Value.Close();
```
benrr101: Should this be wrapped in a try/catch to handle the case where closing the connection fails for some reason?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -45,11 +48,23 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
+
+        /// <summary>
+        /// A map containing all connections to the database that are associated with OwnerUri.
+        /// </summary>
+        public readonly Dictionary<ConnectionType, DbConnection> ConnectionTypeToConnectionMap = new Dictionary<ConnectionType, DbConnection>();
+
+        /// <summary>
+        /// A map containing all CancellationTokenSource objects that are associated with a ConnectionType. 
+        /// </summary>
+        public readonly ConcurrentDictionary<ConnectionType, CancellationTokenSource> ConnectionTypeToCancellationTokenSourceMap = new ConcurrentDictionary<ConnectionType, CancellationTokenSource>();
```
coquagli: Yes, good idea.

kburtram:
Please merge dev and get code coverage and other checks passing.

coquagli:
I'm not sure what the deal with this is. If I run the test on my local machine, it passes. If I log into the test server with the Jenkins account it also passes.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -382,11 +539,8 @@ public bool Disconnect(DisconnectParams disconnectParams)
                 }
             }
 
-            // Close the connection            
-            info.SqlConnection.Close();
-
-            // Remove URI mapping
-            ownerToConnectionMap.Remove(disconnectParams.OwnerUri);
+            // TODO do these activities depend on the type of connection?
```
coquagli: Wondering what others think of this. Should `InvokeOnConnectionActivities` be run after every call to `Connect()`? Likewise for `InvokeOnDisconnectionActivities` and `Disconnect()`.

benrr101: I'm thinking only on the default connection. We run the risk of duplicating actions (that may or may not be idempotent) if we run this for every connection. Not only that but extra actions may slow down the service.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -160,29 +196,49 @@ public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connec
                 };
             }
 
-            // Resolve if it is an existing connection
-            // Disconnect active connection if the URI is already connected
+            // If there is no ConnectionInfo in the map, create a new ConnectionInfo, 
+            // but wait until later when we are connected to add it to the map.
             ConnectionInfo connectionInfo;
-            if (ownerToConnectionMap.TryGetValue(connectionParams.OwnerUri, out connectionInfo) )
+            if (!ownerToConnectionMap.TryGetValue(connectionParams.OwnerUri, out connectionInfo))
```
kevcunnane: If there is a connection in progress we need to handle this. We should probably return a message stating a connection is in progress and to cancel or wait.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -137,9 +145,37 @@ public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connec
...
+
+            // return null upon success
+            return null;
+        }
+
+        /// <summary>
+        /// Open a connection with the specified ConnectParams
         /// </summary>
-        /// <param name="connectionParams"></param>
         public async Task<ConnectionCompleteParams> Connect(ConnectParams connectionParams)
```
kevcunnane: An overall comment on the Connect & Disconnect methods: they were already large and have now hit the "too large" level. Most of the functionality can be split into discrete methods and the main method flow will then be a sequence of steps that you call, instead of the complex logic right now. Please work on this and see if it reveals any common code you can combine :)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -192,11 +248,11 @@ public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connec
                     {
                         // If the URI is currently connecting from a different request, cancel it before we try to connect
                         CancellationTokenSource currentSource;
```
kevcunnane: Looks like I was wrong in my above comment - current behavior is to cancel current connection if we get a new one in? I guess we can preserve this logic...

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -345,17 +413,106 @@ public bool Disconnect(DisconnectParams disconnectParams)
...
+                        connectionsToRemove.Add(entry.Key);
+                    }
+                    catch (System.Data.Common.DbException e)
+                    {
+                        closeFailed = true;
+                    }
+                }
+
+                // If there we no errors calling Close(), simply clear the map and continue
+                if (!closeFailed)
```
kevcunnane: If an exception happened, it more likely means something is wrong with the connection, not that it'll still be usable. I'd recommend removing all connections either way - it's the stated intent and there's not much we can do if close errors. The connection's still broken and the best thing to do is clear up

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -345,17 +413,106 @@ public bool Disconnect(DisconnectParams disconnectParams)
...
+            }
+            // If we are disconnecting only a single type of connection
+            else
             {
-                return false;
+                // Get non-nullable ConnectionType
+                ConnectionType connectionType = (ConnectionType)disconnectParams.Type;
+
+                // Cancel if we are in the middle of connecting
+                if (CancelConnect(new CancelConnectParams() { OwnerUri = disconnectParams.OwnerUri, Type = connectionType }))
```
kevcunnane: Separating parts of this out will show that there's a lot of duplication in the "disconnect all" vs "disconnect one" cases. Please simplify this when you're refactoring.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kevcunnane: 👍 yep only for default connection

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -51,7 +51,13 @@ public static ConnectionService Instance
            
         private readonly Dictionary<string, ConnectionInfo> ownerToConnectionMap = new Dictionary<string, ConnectionInfo>();
 
-        private readonly ConcurrentDictionary<string, CancellationTokenSource> ownerToCancellationTokenSourceMap = new ConcurrentDictionary<string, CancellationTokenSource>();
+        /// <summary>
+        /// A map containing all CancellationTokenSource objects that are associated with a given URI/ConnectionType pair. 
+        /// Entries in this map correspond to DbConnection instances that are in the process of connecting. 
+        /// Tuple is used here becasue of it's predictable hash code behavior. 
+        /// </summary>
+        private readonly ConcurrentDictionary<Tuple<string, ConnectionType>, CancellationTokenSource> CancelTupleToCancellationTokenSourceMap =
```
coquagli: Per the second point, `CancelConnectParams` already has the type and URI. We could change it so it overrides the `Equals()` and `HashCode()` methods so it could be used in the `ConcurrentDictionary`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
coquagli: I agree, as per our discussion earlier. Where can I put the constant string for the message you described?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
coquagli: Good catch.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
```diff
@@ -0,0 +1,103 @@
...
+                // e.g. loop and do it over again
+                service.Disconnect(new DisconnectParams() { OwnerUri = connectionInfo.OwnerUri });
+
+                // We should be left with an empty connection map
+                Assert.Equal(0, service.OwnerToConnectionMap.Count);
+            }
+        }
+
+        [Fact]
+        public void DatabaseChangesAffectAllConnections()
```
coquagli: I'm not sure about either one. I will check.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -45,11 +46,12 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
         public ConnectionDetails ConnectionDetails { get; private set; }
-        
+
         /// <summary>
-        /// The connection to the SQL database that commands will be run against.
+        /// A map containing all connections to the database that are associated with 
+        /// this ConnectionInfo's OwnerUri.
         /// </summary>
-        public DbConnection SqlConnection { get; set; }
+        public readonly Dictionary<ConnectionType, DbConnection> ConnectionTypeToConnectionMap = new Dictionary<ConnectionType, DbConnection>();
```
coquagli: I didn't think about the case for 3rd party services. I'll make that change.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -355,7 +357,8 @@ private void OnInfoMessage(object sender, SqlInfoMessageEventArgs args)
                 // Did the database context change (error code 5701)?
                 if (error.Number == DatabaseContextChangeErrorNumber)
                 {
-                    ConnectionService.Instance.ChangeConnectionDatabaseContext(editorConnection.OwnerUri, conn.Database);
+                    newDatabase = conn.Database;
+                    //ConnectionService.Instance.ChangeConnectionDatabaseContext(editorConnection.OwnerUri, conn.Database);
```
coquagli: I honestly don't remember my reasoning for this. I'll investigate and revert if needed.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kevcunnane: We should avoid overloading anything in the Contracts folders - these are serializable objects and in the future we might auto-gen the contracts. I'd recommend subclassing that and implementing the equals and hashcode in the subclass instead. That would separate the data interchange objects from stuff that can be used internally and have more complex methods / additional data defined. What do you think?

coquagli: That sounds good to me.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
coquagli: Unless there is somewhere other than `OnInfoMessage()` in `Query.cs` that notifies the LanguageService of a database change, I'm not even sure if the Intellisense parser is rebuilt upon database change.

src/Microsoft.SqlTools.ServiceLayer/Connection/CancelTokenKey.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public override bool Equals(object obj)
+        {
+            if (!(obj is CancelTokenKey))
+            {
+                return false;
+            }
+
+            CancelTokenKey other = obj as CancelTokenKey;
+            return String.Equals(other.OwnerUri, OwnerUri) && String.Equals(other.Type, Type);
```
kburtram: >String [](start = 19, length = 6)

we should use "string" instead of "String" when access methods in this code...i.e., prefer calling methods against the primitive type instead of the corresponding BCL class.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -60,5 +65,84 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
+        /// <returns></returns>
+        public List<string> GetAllConnectionTypes()
+        {
+            return ConnectionTypeToConnectionMap.Keys.ToList();
+        }
+
+        /// <summary>
+        /// Gets the count of DbConnectioninstances held by this ConnectionInfo 
+        /// </summary>
+        public int CountConnections()
```
kburtram: >CountConnections [](start = 19, length = 16)

maybe GetConnectionsCount aligns better with the naming used for the other methods

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -60,5 +65,84 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
+        /// false otherwise</returns>
+        public bool RemoveConnection(string connectionType)
+        {
+            // Remove a single DbConnection
+            if (connectionType != null)
+            {
+                ConnectionTypeToConnectionMap.Remove(connectionType);
+            }
+            // Remove all DbConnections 
+            else
```
kburtram: not sure have connectionType == null is the best way for the user to specify remove all connections.  May connectionType == null should throw an exception and we have a different overload for Remove All

coquagli: Good call. I would agree.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -249,58 +366,63 @@ public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connec
...
-            connectionInfo.ConnectionDetails.DatabaseName = connectionInfo.SqlConnection.Database;
-            response.ConnectionSummary = new ConnectionSummary
+        /// <summary>
+        /// Gets the existing connection with the given URI and connection type string. If none exists, 
+        /// creates a new connection. This cannot be used to create a default connection or to create a 
+        /// connection if a default connection does not exist.
+        /// </summary>
+        public async Task<DbConnection> GetOrOpenConnection(string ownerUri, string connectionType)
+        {
+            if (String.IsNullOrEmpty(ownerUri) || String.IsNullOrEmpty(connectionType))
```
kburtram: >String [](start = 16, length = 6)

let's use 'string' instead of 'String' as noted elsewhere.

coquagli:
Waiting on #214 before I merge with `dev`

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/DisconnectParams.cs
```diff
@@ -15,5 +15,11 @@ public class DisconnectParams
         /// or a virtual file representing an object in a database.         
         /// </summary>
         public string OwnerUri { get; set; }
+
+        /// <summary>
+        /// The type of connection we are disconnecting. If null, we will disconnect all
+        /// connections. This is null by default.  
+        /// </summary>
+        public string Type { get; set; } = null;
```
kburtram: >= null [](start = 41, length = 6)

I think the value would be null by default if you don't set it, so this seems unneeded in this case.

kburtram:
Please make sure the coveralls code coverage number doesn't decrease from this PR.

src/Microsoft.SqlTools.ServiceLayer/Connection/CancelTokenKey.cs
benrr101: (This is one thing ReSharper will catch for you 😄)

src/Microsoft.SqlTools.ServiceLayer/Connection/CancelTokenKey.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading.Tasks;
+using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Connection
+{
+    /// <summary>
+    /// Used to uniquely identify a CancellationTokenSource associated with both
+    /// a string URI and a string connection type.
+    /// </summary>
+    public class CancelTokenKey : CancelConnectParams
```
benrr101: If we're implementing Equals and GetHashToken, should we also have this class implement IEquatable<CancelTokenKey>?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -39,17 +41,20 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
-        
+
         /// <summary>
-        /// The connection to the SQL database that commands will be run against.
+        /// A map containing all connections to the database that are associated with 
+        /// this ConnectionInfo's OwnerUri.
+        /// This is internal for testing access only
         /// </summary>
-        public DbConnection SqlConnection { get; set; }
+        internal readonly Dictionary<string, DbConnection> ConnectionTypeToConnectionMap =
```
benrr101: Any particular reason we went with a string for the key instead of having an enum?

Additionally, any chance we might need to make this a ConcurrentDictionary?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
benrr101: This should definitely be turned into a getter-only property (but keep the same name)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -60,5 +65,84 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
+        public List<DbConnection> GetAllConnections()
+        {
+            return ConnectionTypeToConnectionMap.Values.ToList();
+        }
+
+        /// <summary>
+        /// Get a list of all connection type strings held by this ConnectionInfo
+        /// </summary>
+        /// <returns></returns>
+        public List<string> GetAllConnectionTypes()
```
benrr101: Personally, I think any method with a Get* name that just returns a value and doesn't take any parameters should be a getter-only property.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -60,5 +65,84 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
+
+            return ConnectionTypeToConnectionMap.TryGetValue(connectionType, out connection);
+        }
+
+        /// <summary>
+        /// Get a List of all DbConnection instances held by this ConnectionInfo
+        /// </summary>
+        public List<DbConnection> GetAllConnections()
+        {
+            return ConnectionTypeToConnectionMap.Values.ToList();
```
benrr101: We don't need/want the list this generates to be editable, since any changes wouldn't be reflected in the dictionary. My personal preference would be to change these Get* methods to getter-only properties that return IEnumerable<whatever> (or ICollection<>) and just return Map.Values/Map.Keys without newing up a list.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -60,5 +65,84 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
+        }
+
+        /// <summary>
+        /// Adds a DbConnection to this object and associates it with the given 
+        /// connection type string. If a connection already exists with an identical 
+        /// connection type string, it is overwritten. Ignores calls where connectionType = null
+        /// </summary>
+        public void AddConnection(string connectionType, DbConnection connection)
+        {
+            if (connectionType != null)
```
benrr101: These checks for null suggest to me that we should definitely be using an Enum here.
Alternatively, we should consider using the Utils.Validate* methods (or whatever they're called) to check for null/whitespace and throw on those values.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -160,29 +168,137 @@ public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connec
                 };
             }
 
-            // Resolve if it is an existing connection
-            // Disconnect active connection if the URI is already connected
+            // return null upon success
+            return null;
```
benrr101: This code should eventually take a page from the QueryExecutionService and send Error on failure, and a Result on success. This would simplify API quite a bit, but would be a fairly substantial change. I say it's outside the scope for this change, but we should backlog it.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
benrr101: I have a suspicion @coquagli is a former Java developer 😝

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -249,58 +366,63 @@ public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connec
...
+        {
+            if (String.IsNullOrEmpty(ownerUri) || String.IsNullOrEmpty(connectionType))
             {
-                ServerName = connectionInfo.ConnectionDetails.ServerName,
-                DatabaseName = connectionInfo.ConnectionDetails.DatabaseName,
-                UserName = connectionInfo.ConnectionDetails.UserName,
-            };
+                return null;
+            }
+            if (String.Equals(connectionType, ConnectionType.Default))
```
benrr101: Generally unless you need case insensitive or other special case comparisons, you can just do `str1 == str2`

Although this code would also be neater if we were using enums

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -314,9 +436,15 @@ public bool CancelConnect(CancelConnectParams cancelParams)
                 return false;
             }
 
+            CancelTokenKey cancelKey = new CancelTokenKey()
```
benrr101: Parens are unnecessary when doing object initializer syntax

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -346,58 +474,118 @@ public bool Disconnect(DisconnectParams disconnectParams)
...
+            }
+
+            // Remove the disconnected connections from the map
+            if (info.RemoveConnection(disconnectParams.Type))
+            {
+                ownerToConnectionMap.Remove(disconnectParams.OwnerUri);
+            }
+
+            // Handle Telemetry disconnect events if we are disconnecting the default connection
+            if (disconnectParams.Type == null || String.Equals(disconnectParams.Type, ConnectionType.Default))
```
benrr101: Same comment as before with doing direct string ==

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -741,10 +937,14 @@ public void ChangeConnectionDatabaseContext(string ownerUri, string newDatabaseN
             {
                 try
                 {
-                    if (info.SqlConnection.State == ConnectionState.Open)
+                    foreach (DbConnection connection in info.GetAllConnections())
                     {
-                        info.SqlConnection.ChangeDatabase(newDatabaseName);
+                        if (connection.State == ConnectionState.Open)
```
benrr101: I'm not totally sure about this, but can a connection go into a non-open state by being idle? If so, would we still need to change database anyways?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -768,13 +968,65 @@ public void ChangeConnectionDatabaseContext(string ownerUri, string newDatabaseN
...
             }
         }
+
+        /// <summary>
+        /// Invokes the fianl on-disconnect activities if the provided DisconnectParams represents the default
+        /// connection or is null - representing that all connections are being disconnected.
+        /// </summary>
+        private void InvokeOnDisconnectionActivities(ConnectionInfo connectionInfo)
+        {
+            foreach (var activity in this.onDisconnectActivities)
```
benrr101: Another thing we should backlog: moving these activities to events. This works fine, but it'll be cleaner to do this via events.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionType.cs
```diff
@@ -0,0 +1,19 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.Connection
```
benrr101: I didn't see the original comment, why was doing this as an enum not beneficial?

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectParams.cs
```diff
@@ -22,5 +22,10 @@ public class ConnectParams
         /// connection properties to the same database.
         /// </summary>
         public ConnectionDetails Connection { get; set; }
+
+        /// <summary>
+        /// The type of this connection. By default, this is set to ConnectionType.Default.
+        /// </summary>
+        public string Type { get; set; } = ConnectionType.Default;
```
benrr101: Oooooooh, this is a new syntax to me 👍 💯

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -275,68 +276,48 @@ public Task<ResultSetSubset> GetSubset(int batchIndex, int resultSetIndex, int s
                 return;
             }
 
-            // Open up a connection for querying the database
-            string connectionString = ConnectionService.BuildConnectionString(editorConnection.ConnectionDetails);
-            // TODO: Don't create a new connection every time, see TFS #834978
-            using (DbConnection conn = editorConnection.Factory.CreateSqlConnection(connectionString))
+            // Locate and setup the connection
```
benrr101: I think the changes to the query execution service and the connection service could have happened in separate pull requests and made the big changes easier to deal with. Buuuuut, here we are. File that away for the future 😃

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
benrr101: Hmmm... that seems like logic that should be moved into the connection service

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
coquagli: Got me...

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
coquagli: Agreed.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
coquagli: It was originally an enum and Kevin wanted it changed:

"We can't control the connection types. Instead use a string mapping, and use some constants file with Default and Query as the existing options. That way any 3rd party service can register their own connection type for a URI without changing the core code. You could potentially accept null and map to Default too"

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
coquagli: That is a good question. I moved this line, so it is marked as an addition by git, but this is the current behavior of the service.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
coquagli: I think the issue comes from the fact that this PR used to be for 10 lines of code, but slowly grew to 1000.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
coquagli: Clarification: `OnInfoMessage()` calls the database change logic in `ChangeConnectionDatabaseContext()` in the connection service. Nether function rebuilds the intellisense parser.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
benrr101: Ah, well, I guess that makes sense. It's too bad we don't have partial enums in C#. At least I'd hope we specify the strings as constants in the services that will be calling the connection service.

In that case, I'd say we don't really need a whole class for two constants. But I'll leave that up to you.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
benrr101: Ok, if it's not new, then I'd say it's fine to leave it be. But we should remember this for the future just in case.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
coquagli: I don't particularly have an opinion on where the constants should be located.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectParams.cs
coquagli: Fancy, huh?

coquagli:
Anyone have any idea for the source of the coveralls drop? `82a9659` had 86%, but `e69d212` (one hour later, minimal changes) had 71%.

PR #177, opened 2016-12-12
Add error handling for Azure Exceptions

sharonravindran:
Added error pipeline and message for Azure Connection Exceptions


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Definition.cs
```diff
@@ -14,5 +14,37 @@ public static readonly
...
+        /// <summary>
+        /// Error message 
+        /// </summary>
+        public string message { get; set; }
+    }
+
+    /// <summary>
+    /// Result object for Definition
+    /// </summary>
+    public class DefinitionResult
```
kevcunnane: I'd move this out of the Contracts/Definition.cs file - probably create a new file at the LanguageServices folder level. The reasoning is that Contracts files map to data and events sent over the JSON-RPC layer. As this class is used internally and won't be sent on that channel, we shouldn't have it here.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Definition.cs
```diff
@@ -14,5 +14,37 @@ public static readonly
...
+    {
+        /// <summary>
+        /// True, if definition error occured
+        /// </summary>
+        public bool Error;
```
kevcunnane: This should be HasError. Boolean properties should follow standard "IsXyz", "HasXyz", "DidXyz" patterns.

kevcunnane: You could also use "IsErrorResult" if you prefer.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -54,6 +56,8 @@ private Database Database
                     catch(Exception ex)
                     {
                         Logger.Write(LogLevel.Error, "Exception at PeekDefinition Database.get() : " + ex.Message);
+                        this.error = true;
+                        this.errorMessage = SR.PeekDefinitionAzureError;
```
kevcunnane: This is a dangerous assumption. Many different errors could have caused a general exception. We should only give this kind of message if we're sure.

I'd recommend 2 improvements:
- Always include the actual error message back to the user. The default error should be something like string.Format("An unexpected error occurred during Peek Definition execution: '{0}', ex.Message);
- Catch the most specific exception you can, e.g. ConnectionFailureException. Then _only_ if the target is Azure should you return this message. And you should still return the actual error message too unless you're 100% sure of the cause.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -163,17 +167,23 @@ internal Location[] GetScript(IEnumerable<Declaration> declarationItems, string
                         // This workaround ensures that a schema name is present by attempting 
                         // to get the schema name from the declaration item 
                         // If all fails, the default schema name is assumed to be "dbo"
-                        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && string.IsNullOrEmpty(schemaName))
+                        if (connectionInfo.ConnectionDetails.AuthenticationType.Equals("SqlLogin") && string.IsNullOrEmpty(schemaName))
```
kevcunnane: SqlLogin should be a constant (in a Constants.cs file somewhere) given it's used around the codebase. Can you fix?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Modified

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -723,27 +731,57 @@ internal Location[] GetDefinition(TextDocumentPosition textDocumentPosition, Scr
...
+                                Locations = null
+                            };
                         });
 
                     // wait for the queue item
                     queueItem.ItemProcessed.WaitOne();
-                    return queueItem.GetResultAsT<Location[]>();
+                    return queueItem.GetResultAsT<DefinitionResult>();
+                }
+                catch (Exception ex)
```
sharonravindran: This is to catch any exceptions in the Intellisense call. PeekDefinition handles its exceptions separately

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -48,6 +48,8 @@ public sealed class LanguageService
 
         internal const int PeekDefinitionTimeout = 10 * OneSecond;
 
+        internal static ServiceHost ServiceHost= ServiceHost.Instance;
```
kevcunnane: You don't use this at all, as far as I can tell. Shouldn't it be removed?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -7,12 +7,13 @@
...
 using System.Collections.Specialized;
 using System.Data.SqlClient;
-using System.Runtime.InteropServices;
 using Microsoft.SqlServer.Management.Smo;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlServer.Management.SqlParser.Intellisense;
 using Microsoft.SqlTools.ServiceLayer.Utility;
 using Microsoft.SqlTools.ServiceLayer.Connection;
+using Microsoft.SqlTools.ServiceLayer.Hosting.Protocol;
+
```
kevcunnane: Please remove empty line

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -62,20 +66,29 @@ private Database Database
...
-                            Server server = new Server(peekConnection);                        
-                            this.database = new Database(server, peekConnection.DatabaseName);                        
+                            Server server = new Server(peekConnection);
+                            this.database = new Database(server, peekConnection.DatabaseName);
+                        }
+                        catch (ConnectionFailureException cfe)
+                        {
+                            Logger.Write(LogLevel.Error, "Exception at PeekDefinition Database.get() : " + cfe.Message);
+                            this.error = true;
+                            this.errorMessage = connectionInfo.IsAzure ? SR.PeekDefinitionAzureError(cfe.Message) : SR.PeekDefinitionError(cfe.Message); ;
```
kevcunnane: You never validate connectionInfo is not null. Please check here since it's possible (in the tests there are multiple places you don't define it, in which case this would throw null ref exception)

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -231,23 +243,27 @@ public void GetTableDefinitionInvalidObjectTest()
...
         }
 
         /// <summary>
-        /// Test GetDefinition with an unsupported type(function)
+        /// Test GetDefinition with an unsupported type(function). Expect a error result.
         /// </summary>
         [Fact]
-        public void GetUnsupportedDefinitionForFullScript()
+        public void GetUnsupportedDefinitionErrorTest()
         {
```
kevcunnane: Minor: remove whitespace lines at top & bottom of method

kevcunnane: Minor: please add a `// given I am querying an unsupported object (Schema)` to match given/when/then pattern and make clear that it's the dbo schema you are getting the definition for

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -257,27 +273,112 @@ public void GetUnsupportedDefinitionForFullScript()
...
+            // When I call the language service
+            var result = languageService.GetDefinition(textDocument, scriptFile, connInfo);
+
+            // Then I expect null locations and an error to be reported
+            Assert.NotNull(result);
+            Assert.True(result.IsErrorResult);
+
+        }
+
+
```
kevcunnane: Minor: remove unnecessary whitespace (2 empty lines instead of 1)

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -257,27 +273,112 @@ public void GetUnsupportedDefinitionForFullScript()
...
+            string objectName = "from";
+
+            List<Declaration> declarations = new List<Declaration>();
+            DefinitionResult result = peekDefinition.GetScript(declarations, objectName, null);
+
+            Assert.NotNull(result);
+            Assert.True(result.IsErrorResult);
+            Assert.Equal(SR.PeekDefinitionNoResultsError, result.Message);
+        }
+
```
kevcunnane: Minor: remove unnecessary whitespace (2 empty lines instead of 1)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
sharonravindran: Yes, it should be removed. I added this while fixing the tests.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -203,9 +225,9 @@ internal string GetSchemaFromDatabaseQualifiedName(string fullObjectName, string
             string[] tokens = fullObjectName.Split('.');
             for (int i = tokens.Length - 1; i > 0; i--)
             {
-                if(tokens[i].Equals(objectName))
+                if (tokens[i].Equals(objectName))
                 {
-                    return tokens[i-1];
+                    return tokens[i - 1];
```
llali: does the name always has schema? shouldn't you check for when tokens has only one item in it?

sharonravindran: If it has only one item then there isn't a schema name. This was a workaround for the SMO DefaultSchema bug

llali: ok so my problem is that you're getting tokens[i - 1] without checking if the tokens has more than item 

---
In reply to: [92512474](https://github.com/Microsoft/sqltoolsservice/pull/177#discussion_r92512474) [](ancestors = 92512474)

PR #179, opened 2016-12-12
Peek definition for Azure DB objects

kburtram:
Various updates to make Peek Definition work with Azure DB connections.  This change also requires a SMO fix to properly read custom attributes for Types in .Net Core builds (that change is out for review against the SMO repo).

This is related to this issue https://github.com/Microsoft/vscode-mssql/issues/492.


kevcunnane:
:shipit:

PR #180, opened 2016-12-12
fixed the broken perf test

llali:
the Intellisense perf tests were broken since the change for adding brackets to auto complete list was merged. Verified the test are passing now


test/Microsoft.SqlTools.ServiceLayer.PerfTests/Tests/IntellisenseTests.cs
```diff
@@ -257,7 +257,7 @@ public class IntellisenseTests
                 {
                     string query = Scripts.SelectQuery;
                     CompletionItem[] completions = await testHelper.RequestCompletion(ownerUri, query, 0, query.Length + 1);
-                    return completions != null && completions.Any(x => x.Label == databaseName);
+                    return completions != null && (completions.Any(x => x.Label==databaseName || completions.Any(y => y.Label == "[" + databaseName + "]")));
```
benrr101: Just wondering, should these comparisons be done case insensitive?

Plus we could be cool and use the string interpolation feature to shrink the code down to $"[{databaseName}]"

PR #181, opened 2016-12-12
Move Save As to ResultSet

benrr101:
No doubt, this will be a massive code review. It's an overhaul of the Save As mechanism to utilize the file reader/writer classes to better align with the patterns laid out by the rest of the query execution. Why make this change? This change makes our code base more uniform and adherent to the patterns/paradigms we've set up. This change also helps with the encapsulation of the classes to "separate the concerns" of each component of the save as function. 

* Replumbing the save as execution to pass the call down the query stack as QueryExecutionService->Query->Batch->ResultSet
    * Each layer performs it's own parameter checking
        * QueryExecutionService checks if the query exists
        * Query checks if the batch exists
        * Batch checks if the result set exists
        * ResultSet checks if the row counts are valid and if the result set has been executed
    * Success/Failure delegates are passed down the chain as well
* Determination of whether a save request is a "selection" moved to the SaveResultsRequest class to eliminate duplication of code and creation of utility classes
* Making the IFileStream* classes more generic
    * Removing the requirements of max characters to store from the GetWriter method, and moving it into the constructor for the temporary buffer writer - the values have been moved to the settings and given defaults
    * Removing the individual type writers from IFileStreamWriter
    * Removing the individual type writers from IFIleStreamReader
* Adding a new overload for WriteRow to IFileStreamWriter that will write out data, given a row's worth of data and the list of columns
* Creating a new IFileStreamFactory that creates a reader/writer pair for reading from the temporary files and writing to CSV files
* Creating a new IFileStreamFactory that creates a reader/writer pair for reading from the temporary files and writing to JSON files
* Dramatically simplified the CSV encoding functionality
* Removed duplicated logic for saving in different types and condensed down to a single chain that only differs based on what type of factory is provided
* Removing the logic for managing the list of save as tasks, since the ResultSet now performs the actual saving work, there's no real need to expose the internals of the ResultSet
* Adding new strings to the sr.strings file for save as error messages
* Completely rewriting the unit tests for the save as mechanism. Very fine grained unit tests now that should cover majority of cases (aside from race conditions)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -494,6 +465,43 @@ public void InitializeService(ServiceHost serviceHost)
...
+            }
+
+            //Setup the callback for completion of the save task
+            ResultSet.SaveAsAsyncEventHandler successHandler = async parameters =>
+            {
+                await requestContext.SendResult(new SaveResultRequestResult());
+            };
+            ResultSet.SaveAsFailureAsyncEventHandler errorHandler = async (parameters, reason) =>
+            {
+                // TODO: Put this into the strings file
```
kburtram: >// TODO: Put this into the strings file [](start = 16, length = 39)

is this comment still valid?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -282,6 +279,89 @@ public Task<ResultSetSubset> GetSubset(int startRow, int rowCount)
...
+                    {
+                        await failureHandler(saveParams, e.Message);
+                    }
+                }
+            });
+
+            // If saving the task fails, return a failure
+            if (!SaveTasks.TryAdd(saveParams.FilePath, saveAsTask))
+            {
+                throw new InvalidOperationException(SR.QueryServiceSaveAsMscStartingError);
```
kburtram: >Msc [](start = 73, length = 3)

It's not clear to me what 'Msc' means in this context.  Maybe better to use the whole word(s) this is abbreviating

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -301,10 +381,10 @@ protected virtual void Dispose(bool disposing)
 
             IsBeingDisposed = true;
             // Check if saveTasks are running for this ResultSet
-            if (!saveTasks.IsEmpty)
+            if (!SaveTasks.IsEmpty)
             {
                 // Wait for tasks to finish before disposing ResultSet
-                Task.WhenAll(saveTasks.Values.ToArray()).ContinueWith((antecedent) =>
+                Task.WhenAll(SaveTasks.Values.ToArray()).ContinueWith((antecedent) =>
```
kburtram: >Task.WhenAll(SaveTasks.Values.ToArray()) [](start = 16, length = 40)

is this going to prevent the service host from exiting if a file save is in progress?  if so, is that the intended and reviewed design...saving a file can take several minutes, for example.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsWriterBase.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.IO;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
+{
+    /// <summary>
+    /// Abstract class for implementing writers that save results to file. Stores some basic info
+    /// that all save as writer would need.
+    /// </summary>
+    public abstract class SaveAsStreamWriter : IFileStreamWriter
```
kburtram: >SaveAsStreamWriter [](start = 26, length = 18)

should this implement IDisposable?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -143,12 +143,64 @@ public IList<DbCellValue> ReadRow(long fileOffset, IEnumerable<DbColumnWrapper>
...
+        #region Private Helpers
+
+        /// <summary>
+        /// Creates a new buffer that is of the specified length if the buffer is not already
+        /// at least as long as specified.
+        /// </summary>
+        /// <param name="newBufferLength">The minimum buffer size</param>
+        private void AssureBufferLength(int newBufferLength)
+        {
+            if (buffer.Length < newBufferLength)
```
kburtram: >< [](start = 30, length = 1)

<=

kburtram:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
benrr101: Nope, just forgot to remove it. Good catch!

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
benrr101: Yes, I believe this was the reviewed design. We can change this to something else in the future or come up with something different for shutdown of the service versus disposal of the object.

benrr101:
After digging through the Travis error messages, it looks like there is a race condition in the way that events are sent for query start vs all other events. This is not related to this code review, so I will submit it as is. However, I'll make a note that we need to fix this race condition.

PR #182, opened 2016-12-12
Catch exceptions raised sending telemetry events

kburtram:
Catch exceptions raised sending on the "OnConnection" telemetry event.  Sending telemetry during unit tests caused an unhandled exception and is causing code coverage runs to fail.


llali:
:shipit:

PR #185, opened 2016-12-13
Fix broken peek definition integration tests

kburtram:
These integration tests broke with my previous peek definition check-in...I forgot that several tests don't build by default with the LIVE_CONNECTION conditional.


kburtram:
These changes are included in the larger code coverage PR that is pending review.

PR #191, opened 2016-12-15
reserved words with brackets in autocomplete

abist:
This fixes [@473](https://github.com/Microsoft/vscode-mssql/issues/473).


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
```diff
@@ -343,6 +343,14 @@ internal static CompletionItem[] EmptyCompletionList
         }
 
         /// <summary>
+        /// Retrieves the reserved word list 
+        /// </summary>
+        public static string[] GetReservedWordList()
+        {
+            return DefaultCompletionText;
```
kburtram: >DefaultCompletionText [](start = 19, length = 21)

I think we can just make DefaultCompletionText internal instead of adding this method.

kburtram: If you don't want to add a separate assessor, please  add it as a property similar to EmptyCompletionList above.

---
In reply to: [92691614](https://github.com/Microsoft/sqltoolsservice/pull/191#discussion_r92691614) [](ancestors = 92691614)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
```diff
@@ -45,7 +46,7 @@ private void Init()
         {
             InsertText = GetCompletionItemInsertName();
             Label = DeclarationTitle;
-            if (StartsWithBracket(TokenText))
+            if (StartsWithBracket(TokenText) || IsReservedWord(InsertText))
```
kburtram: >IsReservedWord [](start = 48, length = 14)

have you verified we don't add '[' when autocompleted keywords outside the column or table completion context...e.g., we don't escape SELECT if it's typed by its self.

kburtram:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
abist: Yes, I did and it works as intended.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
abist: Changed it

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
```diff
@@ -202,5 +203,12 @@ private string WithBracket(string text)
                 return text;
             }
         }
+
+        private bool IsReservedWord(string text)
```
abist: Done.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
```diff
@@ -45,7 +46,7 @@ private void Init()
         {
             InsertText = GetCompletionItemInsertName();
             Label = DeclarationTitle;
-            if (StartsWithBracket(TokenText))
+            if (StartsWithBracket(TokenText) || AutoCompleteHelper.IsReservedWord(InsertText))
             {
                 Label = WithBracket(Label);
                 InsertText = WithBracket(InsertText);
```
llali: I think for your scenario you should only add the bracket to insert text not the label. because now user see everything in the list without bracket except for objects with reserved word name

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
```diff
@@ -45,7 +46,7 @@ private void Init()
         {
             InsertText = GetCompletionItemInsertName();
             Label = DeclarationTitle;
-            if (StartsWithBracket(TokenText))
+            if (StartsWithBracket(TokenText) || AutoCompleteHelper.IsReservedWord(InsertText))
             {
```
llali: also I don't see any test tp verify new scenarios. can you please add tests?

llali:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
abist: sure, on it

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
abist: I was trying to be consistent with other words that start with brackets. Is it better to remove it from the label?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
```diff
@@ -343,6 +343,27 @@ internal static CompletionItem[] EmptyCompletionList
         }
 
         /// <summary>
+        /// Retrieves the reserved word list 
+        /// </summary>
+        internal static string[] GetReservedWordList
```
kburtram: > internal static string[] GetReservedWordList [](start = 7, length = 45)

this property doesn't appear to be used anywhere now since IsReserveredWord was moved into this class...if it's no longer needed can we delete it?

abist: agreed, removed it

penwon:
Is it possible to disable escaping for temp table names? For example in query like: select * from #MyTable. #MyTable transformed to [#MyTable].

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/SqlCompletionItemTests.cs
```diff
@@ -13,6 +13,304 @@ namespace Microsoft.SqlTools.ServiceLayer.Test.LanguageServer
 {
     public class SqlCompletionItemTests
     {
+        private static readonly string[] ReservedWords = new string[]
+        {
+            "all",
```
kburtram: I thought we already had this list in the language server code.  Perhaps it makes more sense to have the test access the list from there rather than duplicate it into the test code.

abist:
@penwon Thanks for the suggestion. Opened a new issue for this here - [#692](https://github.com/Microsoft/vscode-mssql/issues/692)

PR #193, opened 2016-12-15
Add positive test for GetSignatureHelp

coquagli:
Completes [496](https://github.com/Microsoft/vscode-mssql/issues/496) in vscode-mssql. 

Also moved `GetTestDocPosition()` into `TestObjects`.


test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/LanguageServiceTests.cs
```diff
@@ -290,6 +280,44 @@ public void AutoCompleteFindCompletions()
...
+            };
+
+            // If we have a valid ScriptParseInfo and the SQL has already been parsed
+            var service = LanguageService.Instance;
+            ScriptParseInfo parseInfo = service.GetScriptParseInfo(scriptFile.ClientFilePath);
+            await service.UpdateLanguageServiceOnConnection(connInfo);
+
+            // We should get back a non-null SignatureHelp 
+            SignatureHelp signatureHelp = service.GetSignatureHelp(textDocument, scriptFile);
+            Assert.NotNull(signatureHelp);
```
kburtram: Are there any useful properties on the signatureHelp object we could also validate to make the test more complete?

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/LanguageServiceTests.cs
```diff
@@ -290,6 +280,44 @@ public void AutoCompleteFindCompletions()
...
+            scriptFile.Contents = queryWithFunction;
+            TextDocumentPosition textDocument = new TextDocumentPosition
+            {
+                TextDocument = new TextDocumentIdentifier
+                {
+                    Uri = scriptFile.ClientFilePath
+                },
+                Position = new Position
+                {
+                    Line = 1,
```
kburtram: Could you verify if Position.Line should be 0-based?  I think the SqlParser methods are 1-based and most every thing else is 0-based if I recall correctly?

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/LanguageServiceTests.cs
coquagli: It looks like `FunctionSignatureCompletionReturnsCorrectParametersAtEachPosition()` in `LanguageServiceTests.cs` takes care of the other properties already. The changes here are more for handling a specific scenario where GetSignatureHelp was unexpectedly returning `null`.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/LanguageServiceTests.cs
coquagli: You are correct that they are 0-based. I updated this in the test.

PR #195, opened 2016-12-16
Add fix for SendEvent causing test to fail 

coquagli:


test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/LanguageServiceTests.cs
```diff
@@ -308,10 +312,21 @@ public void AutoCompleteFindCompletions()
...
+            // (this will perform the same operations as UpdateLanguageServiceOnConnection, but without
+            // the ServiceHost.Instance.SendEvent, which will throw in a test environment) 
+            var bindingQueue = LanguageService.Instance.BindingQueue;
+            if (Monitor.TryEnter(scriptInfo.BuildingMetadataLock, LanguageService.OnConnectionWaitTimeout))
+            {
+                scriptInfo.ConnectionKey = bindingQueue.AddConnectionContext(info);
+                scriptInfo.IsConnected = true;
+                Monitor.Exit(scriptInfo.BuildingMetadataLock);
+            }
+            AutoCompleteHelper.PrepopulateCommonMetadata(info, scriptInfo, bindingQueue);
```
kburtram: this call is for perf optimization and probably isn't needed in test context.

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/LanguageServiceTests.cs
```diff
@@ -2,12 +2,16 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
+//#define LIVE_CONNECTION_TESTS
```
kburtram: we probably don't need this if it's commented out

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/LanguageServiceTests.cs
```diff
@@ -308,10 +312,21 @@ public void AutoCompleteFindCompletions()
...
+            // If we have a valid ScriptParseInfo
             var service = LanguageService.Instance;
-            ScriptParseInfo parseInfo = service.GetScriptParseInfo(scriptFile.ClientFilePath);
-            await service.UpdateLanguageServiceOnConnection(connInfo);
+            ScriptParseInfo scriptInfo = service.GetScriptParseInfo(scriptFile.ClientFilePath);
+
+            // And we parse the SQL 
+            // (this will perform the same operations as UpdateLanguageServiceOnConnection, but without
+            // the ServiceHost.Instance.SendEvent, which will throw in a test environment) 
+            var bindingQueue = LanguageService.Instance.BindingQueue;
```
kburtram: ideally we don't duplicate product code into the tests and figure out how to prevent the product code from failing, for example, may we should add a "send telemetry event" helper method" that try\catch and ignores errors...this is a pretty simple snippet so it's probably fine in this case, but better to avoid if possible.

coquagli:
@kburtram: Are the changes I just added what you had in mind?

src/Microsoft.SqlTools.ServiceLayer/Utility/TelemetryHelper.cs
```diff
@@ -0,0 +1,42 @@
...
+            EventType<TParams> eventType,
+            TParams eventParams)
+        {
+            try
+            {
+                return ServiceHost.Instance.SendEvent(eventType, eventParams);
+            }
+            catch (Exception ex)
+            {
+                Logger.Write(LogLevel.Verbose, "Exeception in SendEvent " + ex.ToString());
```
kburtram: typo..."Exception"

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -617,7 +617,7 @@ public ParseResult ParseAndBind(ScriptFile scriptFile, ConnectionInfo connInfo)
                 AutoCompleteHelper.PrepopulateCommonMetadata(info, scriptInfo, this.BindingQueue);
 
                 // Send a notification to signal that autocomplete is ready
-                ServiceHost.Instance.SendEvent(IntelliSenseReadyNotification.Type, new IntelliSenseReadyParams() {OwnerUri = info.OwnerUri});
+                TelemetryHelper.SendEventAndIgnoreErrors(IntelliSenseReadyNotification.Type, new IntelliSenseReadyParams() {OwnerUri = info.OwnerUri});
```
kburtram: oops, I thought this fix was about telemetry..this looks like a product notification..sorry for the confusion.

In this scenario, I think we should continue to log through ServiceHost but have some type of "test flag" to prevent throwing as you were suggesting.

coquagli:
@kburtram done. The diff looks a little crazy because of an indentation change, but I really only changed 10 lines or so.

src/Microsoft.SqlTools.ServiceLayer/Hosting/Protocol/ProtocolEndpoint.cs
```diff
@@ -194,37 +203,52 @@ public void WaitForExit()
...
+                if (SendEventIgnoreExceptions)
+                {
+                    Logger.Write(LogLevel.Verbose, "Exception in SendEvent " + ex.ToString());
+                }
+                else
+                {
+                    throw;
+                }
             }
+            return null;
```
coquagli: I'm unsure if returning null is the correct thing to do here.

src/Microsoft.SqlTools.ServiceLayer/Hosting/Protocol/ProtocolEndpoint.cs
```diff
@@ -29,6 +30,14 @@ public class ProtocolEndpoint : IMessageSender, IProtocolEndpoint
             new Dictionary<string, TaskCompletionSource<Message>>();
 
         /// <summary>
+        /// When true, SendEvent will ingnore exceptions and write them
```
kburtram: typo 'ignore'

src/Microsoft.SqlTools.ServiceLayer/Hosting/Protocol/ProtocolEndpoint.cs
```diff
@@ -29,6 +30,14 @@ public class ProtocolEndpoint : IMessageSender, IProtocolEndpoint
...
+        /// to the log instead. Intended to be used for test scenarios
+        /// where SendEvent throws exceptions unrelated to what is
+        /// being tested.
+        /// </summary>
+        public static bool SendEventIgnoreExceptions = false;
```
kburtram: test-only variables should be internal.

src/Microsoft.SqlTools.ServiceLayer/Hosting/Protocol/ProtocolEndpoint.cs
kburtram: I think Task.FromResult(false) may be better to avoid nullrefs in other test scenarios

PR #196, opened 2016-12-17
Date & Time Type Fixes

benrr101:
This basically is a replacement to the fix for Adding Milliseconds to DateTime fields. I didn't take into consideration that `DATE` columns would report as DateTime type. Date columns have a numeric scale set to 255, leading to the formatting string for the date time to include 255 millisecond places, which is invalid.

This change also reverses the change to store DateTime precision in the buffer file. Instead, the column metadata is now used when deserializing the rows from the db. `DATETIME` and `DATETIME2` columns are differentiated by their numeric scale while `DATE` columns are differentiated by their datatype name field.

More unit tests were added. Additionally, this fixes an unreported bug that `DATE` columns were being displayed with times, which is incorrect.


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -257,37 +274,54 @@ public FileStreamReadResult ReadDecimal(long offset)
...
-                length =>
+            return ReadCellHelper(offset, length =>
+            {
+                long ticks = BitConverter.ToInt64(buffer, 0);
+                return new DateTime(ticks);
+            }, null, dt =>
+            {
+                // Switch based on the type of column
+                string formatString;
+                if (col.DataTypeName.Equals("DATE", StringComparison.OrdinalIgnoreCase))
```
kburtram: >col. [](start = 20, length = 4)

is it possible this could be null in any scenario?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -147,8 +154,9 @@ public IList<DbCellValue> ReadRow(long fileOffset, IEnumerable<DbColumnWrapper>
         /// Reads a short from the file at the offset provided
         /// </summary>
         /// <param name="fileOffset">Offset into the file to read the short from</param>
+        /// <param name="col">Not used</param>
         /// <returns>A short</returns>
-        public FileStreamReadResult ReadInt16(long fileOffset)
+        public FileStreamReadResult ReadInt16(long fileOffset, DbColumnWrapper col)
```
kburtram: >, DbColumnWrapper col [](start = 61, length = 21)

why are we adding an unused parameter to so many functions?

kburtram:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
benrr101: This is mostly to get all the signatures to be the same, that way we can utilize the "method group" style of assigning delegates. I'm not thrilled with the idea either, so I'll go back and tweak it.

PR #198, opened 2016-12-17
Fix/iForgotHowPolymorphismWorks

benrr101:
This change is a refactor of the DbColumnWrapper class. This fixes a huge oversight/gotcha in the DbColumnWrapper where treating the DbColumnWrapper as a DbColumn (ie, polymorphism) would result in almost all the fields of the column being null. This is a highly unexpected behavior. This change fixes that.


kburtram:
:shipit:

PR #204, opened 2016-12-22
Fix for query execution event race condition

benrr101:
This is a small change to fix a race condition in the query execution service. I discovered a day ago that it is possible for a batch start event to occur before the query execute result is sent. This would result in the extension not registering an event listener until after an event was missed. The fix is to move the code that sends the query execute result before the code that launches the query.


coquagli:
Do you think this is related to the behavior I was seeing with #176?

benrr101:
I'm thinking it will. Of course, you're more than welcome to pull down the changes and try out your scenario that generates the bug.

PR #207, opened 2017-01-03
Turning off a broken test until a fix can be merged

kburtram:


kburtram:
This is blocking test runs so I'm automerging.  Please add any comments to the commit for follow-up.

coquagli:
I'm not sure what the deal with this is. If I run the test on my local machine, it passes. If I log into the test server with the Jenkins account it also passes.

PR #208, opened 2017-01-03
Feature: Progressive Messages

benrr101:
This change is a reworking of the way that messages are sent to clients from the service layer. It is also a reworking of the protocol to ensure that all formulations of query send back events to the client in a deterministic ordering. To support the first change:
* Added a new event that will be sent when a message is generated
* Messages now indicate which Batch (if any) generated them
* Messages now indicate if they were error level
* Removed message storage in Batch objects and BatchSummary objects
* Batch objects no longer have error state


kburtram:
Please make sure the AppVeyor is clean prior to merging.

benrr101:
Of course, working on that right now

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -84,11 +83,22 @@ internal Batch(string batchText, SelectionData selection, int ordinalId, IFileSt
...
+
+        /// <summary>
         /// Event that will be called when the batch has completed execution
         /// </summary>
         public event BatchAsyncEventHandler BatchCompletion;
 
         /// <summary>
+        /// Event that will be called when a message has been emitted
+        /// </summary>
+        public event BatchAsyncMessageHandler BatchMessage;
```
llali: I think the better name would be BatchMessageSent, BatchMessage doesn't really say why happened to the message.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -430,30 +429,30 @@ internal void StatementCompletedHandler(object sender, StatementCompletedEventAr
         /// </summary>
         /// <param name="sender">Object that fired the event</param>
         /// <param name="args">Arguments from the event</param>
-        private void StoreDbMessage(object sender, SqlInfoMessageEventArgs args)
+        private void ServerMessageHandler(object sender, SqlInfoMessageEventArgs args)
         {
-            resultMessages.Add(new ResultMessage(args.Message));
+            SendMessage(args.Message, false).Wait();
```
llali: sendMessage returns null if the message is null, should you check for null here before running .Wait()?

benrr101: I don't think that's the case. SendMessage is a async method, so it always returns a Task.

benrr101:
@kburtram Coveralls shows unknown changes... Is that good or no good?

llali:
:shipit:

benrr101:
Code coverage is approximately the same as it was before (according to a manually ran Jenkins build). I'll go ahead and check this in.

PR #211, opened 2017-01-07
fixed the file paths in codecoverage.bat

llali:
The test projects were building the dlls in  %DOTNETCONFIG%\netcoreapp1.0 folder but the codecoverage.bat had hard coded the path to debug\netcoreapp1.0. the change should fix the broken code coverage report


coquagli:
Looks like the most recent commit that changed the tests is causing merge conflicts with this. After fixing those, I say :shipit:

llali:
actually I'm going to close this branch. the change didn't fix the problem with code coverage 

---
In reply to: [272010418](https://github.com/Microsoft/sqltoolsservice/pull/211#issuecomment-272010418) [](ancestors = 272010418)

PR #213, opened 2017-01-10
Feature execution plan settings and request implementation 

Raymondd:
Implements https://github.com/Microsoft/vscode-mssql/issues/575
Only implemented for the XML estimated and actual showplans because SSMS treats Per-Yukon showplans (all non-XML based ones) as standard output tables. The XML can either be displayed by the client or transformed (as SSMS does) into a standardized output of our choice. 

**HandleExecutionPlanRequest Request**
- adds `HandleExecutionPlanRequest` to handler (takes in query URI, batch ID, and result ID)
- adds `GetExecutionPlan `to query which calls batch which calls result set `GetExecutionPlan`

**Query Options added**
- adds `ExecutionPlanOptions` parameter to the `QueryExecuteRequest `via the `QueryExecuteParams `class
- adds `ExecutionPlanOptions` to `QueryExecutionSettings `in order to propagate it from `QueryExecuteParams`
- adds `processSpecialAction `to batch and result set classes to determine which action occurred
- adds `SpecialAction `class to handle execution plan (purposely verbose to handle future special actions)
- adds `SpecialAction `to returned with the batch and result summary to notify client


Raymondd:
Currently adding tests while this is being reviewed since it is a lot of code.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -60,5 +60,16 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
...
+
+        /// <summary>
+        /// Returns true if the sql connection is to a DW instance
+        /// </summary>
+        public bool IsSqlDW { get; set; }
+
+        /// <summary>
+        /// Returns the major version number of the db we are connected to 
+        /// </summary>
+        public int majorVersion { get; set; }
```
llali: Property name should start with capital letter

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -473,6 +498,20 @@ private void ServerMessageHandler(object sender, SqlInfoMessageEventArgs args)
             }
         }
 
+        /// <summary>
+        /// Aggregates all result sets in the batch into a single special action 
+        /// </summary>
+        private SpecialAction processResultSetSpecialActions()
```
llali: Method name should start with capital letter

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -473,6 +498,20 @@ private void ServerMessageHandler(object sender, SqlInfoMessageEventArgs args)
...
+        {
+            SpecialAction batchSpecialAction = new SpecialAction();
```
llali: Create an instance from this in the class to use, no need to create it in the method

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -473,6 +498,20 @@ private void ServerMessageHandler(object sender, SqlInfoMessageEventArgs args)
...
+            foreach (ResultSet resultSet in resultSets) 
+            {
+                batchSpecialAction.combineSpecialAction(resultSet.Summary.SpecialAction);
```
llali: C should be upper case

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -374,6 +472,14 @@ private void OnInfoMessage(object sender, SqlInfoMessageEventArgs args)
             }
         }
 
+        /// <summary>
+        /// Function to add a new batch to a Batch set
+        /// </summary>
+        private void addBatch(string query, List<Batch> batchSet, IFileStreamFactory outputFactory)
```
llali: >addBatch [](start = 21, length = 8)

put space after each comma

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -89,7 +122,32 @@ public Query(string queryText, ConnectionInfo connection, QueryExecutionSettings
...
+            AfterBatches = new List<Batch>();
+
+            // Checking settings for execution plan options 
+            if (querySettings.ExecutionPlanOptions.IncludeEstimatedExecutionPlanXml) 
+            {
+                // Check support level 
+                if (this.doesSupportExecutionPlan(connection))
+                {
+                    // Enable set showplan xml
+                    addBatch(string.Format(s_SetShowPlanXml, s_On), BeforeBatches, streamOutputFactory);
```
llali: >s_On [](start = 61, length = 4)

these are duplicate code, move the login to a method or even another class to handle options

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -403,6 +509,14 @@ protected virtual void Dispose(bool disposing)
             disposed = true;
         }
 
+        /// <summary>
+        /// Does this connection support XML Execution plans
+        /// </summary>
+        private bool doesSupportExecutionPlan(ConnectionInfo connectionInfo) {
```
llali: method name should start with upper case letter

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,122 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
+{
+    /// <summary>
+    /// Class that represents a Special Action which occured by user request during the query 
+    /// </summary>
+    public class SpecialAction {
+        bool _None;
+        bool _ExpectActualYukonXmlShowPlan;
```
llali: private fields starts with lowercase letter

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
Raymondd: are you referring to the addBatch function?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
Raymondd: i think this was referring to the next line?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,132 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
+{
+    /// <summary>
+    /// Class that represents a Special Action which occured by user request during the query 
+    /// </summary>
+    public class SpecialAction {
+        
+        #region Private Class variables 
+        bool none;
```
kburtram: >bool [](start = 8, length = 4)

add the access level explicitly to these fields. i.e., 'private' or 'internal'

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,132 @@
...
+                if (value)
+                {
+                    expectActualYukonXmlShowPlan = false;
+                    expectEstimatedYukonXmlShowPlan = false;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Actual XML execution plan may be returned
```
kburtram: >may be  [](start = 38, length = 7)

why does this say "may be"...should it say "will be" or is it really nondeterministic whether plan will be returned?   perhaps rewording would clarify.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// <summary>
+        /// a type of XML execution plan may be returned  
+        /// </summary>
+        public bool ExpectYukonXMLShowPlan 
+        {
+            get { return ExpectEstimatedYukonXmlShowPlan || ExpectActualYukonXmlShowPlan; }
+            set 
+            { 
+                ExpectEstimatedYukonXmlShowPlan = value;
+                ExpectActualYukonXmlShowPlan = value;
```
kburtram: so when this is set we expect both an actual and estimated plan?  if that's not the case I'd make this a read-only field.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+        
+        #endregion
+
+        #region Private Helper Functions 
+        /// <summary>
+        /// Check to see if all properties are false, other than none 
+        /// </summary>
+        private bool areAllFalse()
```
kburtram: >areAllFalse [](start = 21, length = 11)

methods should always start with UpperCase letter..same below.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            if (!ExpectActualYukonXmlShowPlan && !ExpectEstimatedYukonXmlShowPlan)
```
kburtram: >!ExpectActualYukonXmlShowPlan && !ExpectEstimatedYukonXmlShowPlan [](start = 16, length = 65)

method could be rewritten as "return !ExpectActualYukonXmlShowPlan && !ExpectEstimatedYukonXmlShowPlan;"

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -26,6 +29,14 @@ public class QueryExecutionSettings
         /// </summary>
         private const int DefaultMaxXmlCharsToStore = 2097152; // 2 MB - QE default
 
+        /// <summary>
+        /// Default selection of returning an actual XML showplan with all batches
+        /// </summary>
+        private ExecutionPlanOptions DefaultExecutionPlanOptions = new ExecutionPlanOptions() {
```
kburtram: >{ [](start = 94, length = 1)

'{' should be on line by itself

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -26,6 +29,14 @@ public class QueryExecutionSettings
         /// </summary>
         private const int DefaultMaxXmlCharsToStore = 2097152; // 2 MB - QE default
 
+        /// <summary>
+        /// Default selection of returning an actual XML showplan with all batches
+        /// </summary>
+        private ExecutionPlanOptions DefaultExecutionPlanOptions = new ExecutionPlanOptions() { 
+            IncludeActualExecutionPlanXml = false,
+            IncludeEstimatedExecutionPlanXml = false
+        }; // do not return any execution plan by default
```
kburtram: >// do not return any execution plan by default [](start = 11, length = 46)

could you add this comment into the summary block instead of down here

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -26,6 +29,14 @@ public class QueryExecutionSettings
         /// </summary>
         private const int DefaultMaxXmlCharsToStore = 2097152; // 2 MB - QE default
 
+        /// <summary>
+        /// Default selection of returning an actual XML showplan with all batches
+        /// </summary>
+        private ExecutionPlanOptions DefaultExecutionPlanOptions = new ExecutionPlanOptions() { 
+            IncludeActualExecutionPlanXml = false,
```
kburtram: >false [](start = 44, length = 5)

I think false is the default value for a bool so you probably don't need the object initialization block in this case IIRC

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/ExecutionPlanTests.cs
```diff
@@ -0,0 +1,251 @@
...
+using System.Threading.Tasks;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Test.Utility;
+using Microsoft.SqlTools.ServiceLayer.SqlContext;
+using Xunit;
+
+namespace Microsoft.SqlTools.ServiceLayer.Test.QueryExecution
+{
+    public class ExecutionPlanTests
```
kburtram: >ExecutionPlanTests [](start = 17, length = 18)

please make sure that this change does not decrease the code coverage and add additional test methods if necessary

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
Raymondd: by 'may be' I meant that another request can be sent to the service to fetch the entire plan itself. I will reword to 'Contains an XML execution plan result set'.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
llali: yes, adding a batch before and after based on an option

---
In reply to: [95651178](https://github.com/Microsoft/sqltoolsservice/pull/213#discussion_r95651178) [](ancestors = 95651178)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -316,16 +401,29 @@ public Task<ResultSetSubset> GetSubset(int batchIndex, int resultSetIndex, int s
...
+                    // Execute beforeBatches synchronously, before the user defined batches 
+                    foreach (Batch b in BeforeBatches)
+                    {
+                        await b.Execute(conn, cancellationSource.Token);
+                    }
+
                     // We need these to execute synchronously, otherwise the user will be very unhappy
                     foreach (Batch b in Batches)
                     {
+                        // Attach extension callbacks to the these batches because the user explicity ran them
```
benrr101: I'm not sure what you mean by this comment?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -374,6 +472,14 @@ private void OnInfoMessage(object sender, SqlInfoMessageEventArgs args)
             }
         }
 
+        /// <summary>
+        /// Function to add a new batch to a Batch set
+        /// </summary>
+        private void addBatch(string query, List<Batch> batchSet, IFileStreamFactory outputFactory)
+        {
+            batchSet.Add(new Batch(query, new SelectionData(0, 0, 0, 0), batchSet.Count, outputFactory));
```
benrr101: Ideally, I think if we can send null for the selectiondata we might be better off. I'm thinking how this will play out for our ability to execute arbitrary queries.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -208,6 +209,56 @@ public void InitializeService(ServiceHost serviceHost)
...
+        public async Task HandleExecutionPlanRequest(QueryExecutionPlanParams planParams,
+            RequestContext<QueryExecutionPlanResult> requestContext)
+        {
+            try
+            {
+                // Attempt to load the query
+                Query query;
+                if (!ActiveQueries.TryGetValue(planParams.OwnerUri, out query))
+                {
+                    await requestContext.SendResult(new QueryExecutionPlanResult
```
benrr101: We should definitely be doing this via SendError instead of sending a message back via the result.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -208,6 +209,56 @@ public void InitializeService(ServiceHost serviceHost)
...
+                // Return the error as a result
+                await requestContext.SendResult(new QueryExecutionPlanResult
+                {
+                    Message = ioe.Message
+                });
+            }
+            catch (ArgumentOutOfRangeException aoore)
+            {
+                // Return the error as a result
+                await requestContext.SendResult(new QueryExecutionPlanResult
```
benrr101: Same here, use SendError

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -208,6 +209,56 @@ public void InitializeService(ServiceHost serviceHost)
...
+                {
+                    Message = aoore.Message
+                });
+            }
+            catch (Exception e)
```
benrr101: In general, I think we can have the other exception types just be caught using the general Exception catch and return all of them via SendError

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
Raymondd: When it is set, it basically lets us know that there is an XML execution plan that will be returned, but we are not sure which one. It is not possible to know which kind of plan is returned unless I couple the query execution options to the result set summaries (i.e. Save a copy of the QueryExecutionSettings in the resultSet object to later refer to when when setting up the special action). I am not sure exactly what to do in this case.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,125 @@
+// 
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
+{
+    /// <summary>
+    /// Class that represents a Special Action which occured by user request during the query 
+    /// </summary>
+    public class SpecialAction {
```
benrr101: I'm having a really hard time wrapping my head around this class. If these are just boolean flags that we'd set as part of execution, why can't they just be part of the resultset object itself? If there's some kind of action that needs to be taken, I think it should be setup more like a SpecialAction list and have the SpecialAction be an interface that has a method for the action to take.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,125 @@
...
+        public SpecialAction()
+        {
+            None = true;
+        }
+
+        #region Public Functions
+        /// <summary>
+        /// No Special action performed 
+        /// </summary>
+        public bool None {
```
benrr101: This code particularly has me confused. So if you set None to true, you get both fields set to false, but none set to true? I'm ultimately not sure what you're trying to do with this, but I think there's a better way to do it.

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -26,6 +29,16 @@ public class QueryExecutionSettings
         /// </summary>
         private const int DefaultMaxXmlCharsToStore = 2097152; // 2 MB - QE default
 
+        /// <summary>
+        /// Default selection of returning an actual XML showplan with all batches
+        /// Do not return any execution plan by default
+        /// </summary>
+        private ExecutionPlanOptions DefaultExecutionPlanOptions = new ExecutionPlanOptions()
```
benrr101: One thing I'm wondering is: is anyone actively using these settings the way they're written?

benrr101: (I'm asking more of an existential question than directly asking about these specific settings)

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Common.cs
```diff
@@ -183,7 +212,7 @@ public static ISqlConnectionFactory CreateMockFactory(Dictionary<string, string>
         {
             var mockFactory = new Mock<ISqlConnectionFactory>();
             mockFactory.Setup(factory => factory.CreateSqlConnection(It.IsAny<string>()))
-                .Returns(() => CreateTestConnection(data, throwOnRead));
+                .Returns<string>((s) => CreateTestConnection(data, throwOnRead));
```
benrr101: Oh whoops, I don't think we need this change anymore.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/ExecutionPlanTests.cs
```diff
@@ -0,0 +1,251 @@
...
+        public void ExecutionPlanInvalid()
+        {
+            // Setup:
+            // ... I have a batch that has been executed
+            Batch b = Common.GetBasicExecutedBatch();
+
+            // If:
+            // ... I have a result set and I ask for an execution plan that doesn't exist
+            ResultSet planResultSet = b.ResultSets.First();
+            ExecutionPlan plan = planResultSet.GetExecutionPlan().Result;
```
benrr101: Shouldn't it throw?

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/ExecutionPlanTests.cs
```diff
@@ -0,0 +1,251 @@
...
+        [InlineData(2)]   // Invalid result set, too high
+        public void BatchExecutionPlanInvalidParamsTest(int resultSetIndex)
+        {
+            // If I have an executed batch which has an execution plan 
+            Batch b = Common.GetExecutedBatchWithExecutionPlan();
+
+            // ... And I ask for an execution plan with an invalid result set index
+            // Then: 
+            // ... It should throw an exception
+            Assert.ThrowsAsync<ArgumentOutOfRangeException>(() => b.GetExecutionPlan(resultSetIndex)).Wait();
```
benrr101: One cool thing we get from XUnit is the ability to have async test methods. Although it's not a big deal, since the code isn't very different, we can avoid using the .Wait and instead using the await keyword.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/ExecutionPlanTests.cs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        #endregion
+
+        #region Query Class Tests
+
+        [Theory]
+        [InlineData(-1)]  // Invalid batch, too low
+        [InlineData(2)]   // Invalid batch, too high
+        public void QueryExecutionPlanInvalidParamsTest(int batchIndex)
```
benrr101: It'd be my preference to try and isolate these sort of tests to the class they test. In this case, the batch class. So instead of creating a whole Query object, I'd just create the Batch and test from there. But, this works, too.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/ExecutionPlanTests.cs
```diff
@@ -0,0 +1,251 @@
...
+                IncludeEstimatedExecutionPlanXml = true
+            };
+            var executeRequest = RequestContextMocks.Create<QueryExecuteResult>(null);
+            await queryService.HandleExecuteRequest(executeParams, executeRequest.Object);
+            await queryService.ActiveQueries[Common.OwnerUri].ExecutionTask;
+            queryService.ActiveQueries[Common.OwnerUri].Batches[0].ResultSets[0].hasBeenRead = false;
+
+            // ... And I then ask for a valid execution plan from it 
+            var executionPlanParams = new QueryExecutionPlanParams { OwnerUri = Common.OwnerUri, ResultSetIndex = 0, BatchIndex = 0 };
+            var executionPlanRequest = new EventFlowValidator<QueryExecutionPlanResult>()
```
benrr101: Thank you for using my EventFlowValidator! 👍 💯

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryExecutionPlanRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    }
+
+    /// <summary>
+    /// Parameters for the query execution plan request
+    /// </summary>
+    public class QueryExecutionPlanResult
+    {
+        /// <summary>
+        /// Executionb plan request error messages. Optional, can be set to null to indicate no errors
```
Raymondd: is there an example of how to implement that in another request handler i can look at?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
Raymondd: I'll rephrase. I was just trying to say that these have completion handlers because they aren't injected batches like the beforeBatches and afterBatches.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/ExecutionPlanTests.cs
Raymondd: previously we were returning null, but yes i think i should have it throw instead.

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
Raymondd: Hmm I'm not sure. I just assumed it was fallback I needed to put in place.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/ExecutionPlanTests.cs
Raymondd: I went with the way the other request handlers formatted their tests. Maybe at some point we can regroup some of the more similar ones like you mentioned above, but I think that might be out of scope for this PR.

benrr101: Agreed, no real need to redo all that at this point.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -0,0 +1,89 @@
...
+        public SpecialAction()
+        {
+            flags = ActionFlags.None;
+        }
+
+        #region Public Functions
+        /// <summary>
+        /// No Special action performed 
+        /// </summary>
+        public bool None
```
benrr101: So, I think you don't need setters for these properties. Additionally, you can avoid storing booleans for none, etc by just returning `flags == ActionFlags.None`. For the others, you can just do `flags & ActionFlags.Yukon > 0`

If you're adding the setters because you need to be able to deserialize a SpecialActions class from json, you should explicitly check the value provided. If it's true, `|=` the value to the flags, if it's false, `&= ~(value)` to the flags.

PR #217, opened 2017-01-21
query/executeString 

benrr101:
This is a small API addition that allows us to execute queries directly as strings. This will make it easier to execute queries outside the confines of a workspace like VS Code.

* Refactor execution requests and events are now named less redundantly and moved into a separate namespace for organization. This is the bulk of the changes.
    * QueryExecuteBatchNotification -> ExecuteRequests/BatchEvents
    * QueryExecuteMessageNotification -> ExecuteRequests/MessageEvent
    * QueryExecuteCompleteNotification -> ExecuteRequests/QueryCompleteEvent
    * QueryExecuteResultSetCompleteNotification -> ExecuteRequests/ResultSetEvents
    * QueryExecuteSubsetRequest -> SubsetRequest.cs
* Creating an inheritance pattern where
    * `ExecuteRequestParamsBase` has execution options and ID for a query execute request
    * `ExecuteDocumentSelectionParams` inherits from `ExecuteRequestParamsBase` and provides a document selection
    * `ExecuteStringParams`  inherits from `ExecuteRequestParamsBase` and provides the query text
* Adding a helper method to get SQL text based on request type
* Through the AWESOME POWER OF POLYMORPHISM, we are able to create a request for executing straight SQL basically for free.
* **Breaking change:** query/execute => query/executeDocumentSelection to make it more obvious what is expected.
* Adding unit tests for the code that gets SQL text


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -539,6 +511,50 @@ private static void ExecuteAndCompleteQuery(QueryExecuteParams executeParams, Re
...
+            }
+
+            // If it is an ExecuteStringParams, return the text as is
+            if (requestType == typeof(ExecuteStringParams))
+            {
+                return ((ExecuteStringParams) request).Query;
+            }
+
+            // Note, this shouldn't be possible due to inheritance rules
+            throw new InvalidCastException("Invalid request type");
```
coquagli: Would this be the correct exception to throw?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -539,6 +511,50 @@ private static void ExecuteAndCompleteQuery(QueryExecuteParams executeParams, Re
             }
         }
 
+        // Internal for testing purposes
+        internal string GetSqlText(ExecuteRequestParamsBase request)
```
coquagli: Seems like this should be an abstract function that the subclasses override.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Common.cs
```diff
@@ -96,6 +97,11 @@ public static Query GetBasicExecutedQuery()
         public static Query GetBasicExecutedQuery(QueryExecutionSettings querySettings)
         {
             ConnectionInfo ci = CreateTestConnectionInfo(new[] {StandardTestData}, false);
+
+            // Query won't be able to request a new query DbConnection unless the ConnectionService has a 
+            // ConnectionInfo with the same URI as the query, so we will manually set it
+            ConnectionService.Instance.OwnerToConnectionMap[ci.OwnerUri] = ci;
```
coquagli: This is unfortunately true and may have been a bad decision on my part. The intent was to fail early when no ConnectionInfo is present, because ConnectionInfo contains the details needed to create a connection string in a real non-test scenario. This works for now, but it could be changed to prevent the need to do this in the future.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
benrr101: I agree, it *should*, but I weighed this against separation of concerns. If I put this in ExecuteRequestParamsBase, the ExecuteDocumentSelectionParams class would need access to the WorkspaceService. This isn't ideal, since really only the *Service classes should interact with the other services. On the other hand, if we pass in the WorkspaceService to the GetSqlText method on the ExecuteDocumentSelectionParams, we'd also have to pass it into the ExecuteStringParams implementation, which isn't useful.

In this case, my opinion is that we should do this kinda annoying cast in order to preserve the contracts from not needing to know about the services that utilize them.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Common.cs
benrr101: I agree. I tried to mess with the services a bit to try and prevent us from needing to do this, but I realized it would be a much larger and annoying change than I hoped.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
coquagli: For the second concern, isn't `WorkspaceService` a singleton? You would call `WorkspaceService<SqlToolsSettings>.Instance` to get the instance. No need to pass it as a parameter.

benrr101:
Ping on this!

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -539,6 +511,50 @@ private static void ExecuteAndCompleteQuery(QueryExecuteParams executeParams, Re
...
+        {
+            Type requestType = request.GetType();
+
+            // If it is a document selection, we'll retrieve the text from the document
+            if (requestType == typeof(ExecuteDocumentSelectionParams))
```
llali: FxCop will give you warning for these two lines (typeof and casting to a type) instead you should do documentRequest =  request as ExecuteDocumentSelectionParams and then check if documentRequest  is null

benrr101: Agreed, I think I wrote it like this assuming there was the ability to do a switch on types. I'll modify this as per your suggestion.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteRequestParamsBase.cs
```diff
@@ -0,0 +1,20 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests
+{
+    public abstract class ExecuteRequestParamsBase
```
kevcunnane: Minor: please document the class definition

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -539,6 +511,50 @@ private static void ExecuteAndCompleteQuery(QueryExecuteParams executeParams, Re
...
 
+        // Internal for testing purposes
+        internal string GetSqlText(ExecuteRequestParamsBase request)
+        {
+            // If it is a document selection, we'll retrieve the text from the document
+            ExecuteDocumentSelectionParams docRequest = request as ExecuteDocumentSelectionParams;
+            if (docRequest != null)
+            {
+                // Get the document from the parameters
+                ExecuteDocumentSelectionParams documentRequest = (ExecuteDocumentSelectionParams) request;
```
kevcunnane: You should use `docRequest` since you already cast it and validated it's not null - this 2nd declaration is redundant

kevcunnane: Also to be nitpicky - this is a bunch of independent logic in an If clause. For clarity, why not move everything in the if clause into its own method and just call `return GetSqlTextFromDocRequest`? That'll make it easier to read this method since we can see the overall logic there and ignore this detailed lookup logic.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Common.cs
kevcunnane: Let's consider this for future fixing - it points to some non-optimal decisions. Also, I still don't like that we're directly accessing a dictionary here. That's just not a good design (we should protect it with explicit method access, not let you query it from anywhere in the app). It's outside the scope of this review but would be good to fix up later.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteStringRequest.cs
```diff
@@ -0,0 +1,27 @@
...
+//
+
+using Microsoft.SqlTools.ServiceLayer.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests
+{
+    /// <summary>
+    /// Parameters for executing a query directly
+    /// </summary>
+    public class ExecuteStringParams : ExecuteRequestParamsBase
```
kevcunnane: Where is the unit test/tests for the ExecuteStringRequest? I don't see this in the PR. Any new API should have unit tests covering it.

benrr101: Not sure why, but GitHub isn't expanding the ServiceIntegrationTests file. I added new unit tests to test the "getsqltext" method, which is the only place where the ExecuteStringRequest and ExecuteDocumentRequest code paths differ. Then I added a note:
```
﻿// NOTE: In order to limit test duplication, we're running the ExecuteDocumentSelection
// version of execute query. The code paths are almost identical.
```
I *can* go an duplicate all the tests for the ExecuteStringRequest, but I'd hate to add a ton of unnecessary tests. Do you want me to go ahead and do that?

That being said, I don't have unit tests for the params classes themselves, but since they're just storage classes that we haven't unit tested before, it seems unnecessary.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
benrr101: I can see where you're coming from, but I think the logic is small enough that it doesn't need to be broken out into a separate set of methods. If we ever get more types to check against, I'd be more apt to consider it, but for now I think it's fine as-is.

Definitely agree on the redundant declarations. Coffee hadn't kicked in this morning. Will address right away.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteRequestParamsBase.cs
benrr101: Sure, will do.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteStringRequest.cs
kevcunnane: Ah - no that's fine. it's a small change so as long as we've covered it there it's fine. I just couldn't spot with the GitHub PR review tool

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
kevcunnane: Yep, not a deal breaker, just a personal preference having seen the result (the next person adds another clause, then it just gets hard to track).

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
benrr101: True, but, we don't always want to use that instance. Using the WorkspaceService.Instance is the default case, but sometimes we want to override the default WorkspaceService -- unit testing for instance. Additionally, we wouldn't use the WorkspaceService if we have a ExecuteStringParams. If we ever introduce a different execute type (executefile, for instance) we wouldn't need the workspace service there either.

I suppose we could have the WorkspaceService be a static property of the ExecuteDocumentSelection and allow it to be reset for testing purposes. What say you to that idea?

coquagli: I think that would be good solution.

benrr101: After looking at this some more, I think the changes required are going to be too massive for this code review (large changes to connection service). Since I want to get moving on the Edit Data stuff, I'm going to decline to implement these changes for now and stick with the slightly less than ideal implementation that's currently in review. I have most of it written up (see https://github.com/Microsoft/sqltoolsservice/tree/fix/queryConnectionIsolation), but I want to make sure that I get the unit tests right -- which will likely take a fair amount of time.

llali:
:shipit:

PR #220, opened 2017-01-25
Fix to get the current database context for peek definition 

sharonravindran:
Added code to get the database name from the query connection object (if one exists)

Fix for Microsoft/vscode-mssql#644


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -73,7 +74,22 @@ private Database Database
                         {
                             // Reuse existing connection 
                             Server server = new Server(this.serverConnection);
-                            this.database = new Database(server, this.serverConnection.DatabaseName);
+                            // The default database name is the database name of the server connection
+                            string dbName = this.serverConnection.DatabaseName;
```
kevcunnane: Can we have a unit test to verify this scenario?

sharonravindran: I've added tests for this scenario

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -73,7 +74,22 @@ private Database Database
                         {
                             // Reuse existing connection 
                             Server server = new Server(this.serverConnection);
-                            this.database = new Database(server, this.serverConnection.DatabaseName);
+                            // The default database name is the database name of the server connection
+                            string dbName = this.serverConnection.DatabaseName;                          
+                            if (this.connectionInfo != null)
+                            {
+                                // If there is a query DbConnection, use that connection to get the database name
+                                // This is preferred since it has the most current database name (in case of database switching)
```
coquagli: I think this logic should be moved to the ConnectionService when #645 is fixed. It is a bug if different DbConnection instances for the same URI have different `Database` fields.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -61,7 +62,7 @@ internal PeekDefinition(ServerConnection serverConnection, ConnectionInfo connIn
             Initialize();
         }
 
-        private Database Database
+        internal Database Database
```
coquagli: Since the check for a new `Database` field only occurs if `this.database == null`, Is it possible for the following scenario to occur?
- `Database` getter is accessed, setting `this.database` to not null and `Database` field to `master`
- `use tempdb` is called, changing the `Database` field for all open connections to `tempdb`
- `Database` getter is accessed again, returning the non-null `this.database` with `Database = master` instead of `Database = tempdb`.

sharonravindran: No because the peek definition object persists for only one peek definition call.

sharonravindran: In this scenario, a new peek definition call would be made after the use tempdb execution and that will have a new peek definition object.

sharonravindran:
Ping.

llali:
Please fix the problem with the code coverage. Merging from dev might help with

---
In reply to: [276252015](https://github.com/Microsoft/sqltoolsservice/pull/220#issuecomment-276252015) [](ancestors = 276252015)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
coquagli: Looking at this again, this seems to be more of a flaw in the ServerConnection passed into PeekDefinition's constructor than it is for PeekDefinition. It seems like the Database name for the BindingQueue's ServerConnection will never be updated. Updating the database name for the binding queue [in the connection service's database changed callback](https://github.com/Microsoft/sqltoolsservice/blob/dev/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs#L942) might be a better solution than checking for the database name of a Query connection.

benrr101:
I've restarted your Travis CI build on OSX. It has been very flaky lately.

sharonravindran:
Ping.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
coquagli: It might be better for a backlog task, but managing the DatabaseName for the SqlConnection opened in ConnectedBindingQueue might be a cleaner solution.

PR #221, opened 2017-01-27
FIx: SqlDw connection setup change

Raymondd:
Fixes https://github.com/Microsoft/vscode-mssql/issues/563

- Added a sqldw check before database specific properties are sent to server. 
- Added caching for sqldw upon first connection to database

I am aware that the caching side of things is a bit odd, but I decided to keep it consistent with DacFX dev updates.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
@@ -137,15 +137,15 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
             }
         }
 
-        public static void TryGetIsSqlDw(IDbConnection connection, out bool isSqlDw)
+        public static bool TryGetIsSqlDw(IDbConnection connection, out bool isSqlDw)
         {
             Validate.IsNotNull(nameof(connection), connection);
 
             SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connection.ConnectionString);
-            TryGetIsSqlDw(builder.DataSource, out isSqlDw);
+            return TryGetIsSqlDw(builder.DataSource, out isSqlDw);
         }
 
-        public static void TryGetIsSqlDw(string dataSource, out bool isSqlDw)
+        public static bool TryGetIsSqlDw(string dataSource, out bool isSqlDw)
         {
             Validate.IsNotNullOrWhitespaceString(nameof(dataSource), dataSource);
             CachedInfo info;
@@ -154,12 +154,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
             if(hasFound)
             {
                 isSqlDw = info.IsSqlDw;
+                return true;
             }
-            else
-            {
-                throw new Exception(Resources.ServerInfoCacheMiss);
-            }
-
+            
+            isSqlDw = false;
+            return false;
         }
 
         private static string SafeGetDataSourceFromConnection(IDbConnection connection)
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
@@ -124,7 +124,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
             AddOrUpdateIsSqlDw(builder.DataSource, isSqlDw);
         }
 
-        public static void AddOrUpdateIsSqlDw(string  dataSource, bool isSqlDw)
+        public static void AddOrUpdateIsSqlDw(string dataSource, bool isSqlDw)
         {
             Validate.IsNotNullOrWhitespaceString(nameof(dataSource), dataSource);
             CachedInfo info;
@@ -175,8 +175,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
         
         public static void TryGetIsSqlDw(string dataSource, out bool isSqlDw)
         {
-            Validate.IsNotNullOrWhitespaceString(nameof(dataSource), dataSource);
-
             Validate.IsNotNullOrWhitespaceString(nameof(dataSource), dataSource);
             CachedInfo info;
             bool hasFound = _cache.TryGetValue(dataSource, out info);
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
@@ -17,6 +17,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
     /// </summary>
     internal static class CachedServerInfo
     {
+        public enum CacheVariable {
+            IsSqlDw,
+            IsAzure
+        }
+
         private struct CachedInfo
         {
             public bool IsAzure;
@@ -28,7 +33,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
         private static object _cacheLock;
         private const int _maxCacheSize = 1024;
         private const int _deleteBatchSize = 512;
-
         private const int MinimalQueryTimeoutSecondsForAzure = 300;
 
         static CachedServerInfo()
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -26,6 +27,8 @@ internal static class ReliableConnectionHelper
         private const string ApplicationIntent = "ApplicationIntent";
         private const string MultiSubnetFailover = "MultiSubnetFailover";
         private const string DacFxApplicationName = "DacFx";
+
+        private const int SqlDwEngineEditionId = (int)DatabaseEngineEdition.SqlDataWarehouse;
```
kburtram: >(int) [](start = 49, length = 5)

any reason to store as an int rather than as than as the enum type?

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -460,6 +463,56 @@ private static bool IsCloudEngineId(int engineEditionId)
...
+            bool isSqlDw = false;
+            try
+            {
+                isSqlDw = executeCommand(SqlConnectionHelperScripts.EngineEdition);
+            }
+            catch (SqlException)
+            {
+                // The default query contains a WITH (NOLOCK).  This doesn't work for Azure DW, so when things don't work out, 
+                // we'll fall back to a version without NOLOCK and try again.
+                isSqlDw = executeCommand(SqlConnectionHelperScripts.EngineEditionWithLock);
```
kburtram: > isSqlDw = executeCommand(SqlConnectionHelperScripts.EngineEditionWithLock); [](start = 15, length = 76)

should this be inside a try/catch too?  or is someone else catch exceptions thrown from this call?

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
```diff
@@ -115,6 +116,80 @@ public static void AddOrUpdateIsAzure(string  dataSource, bool isAzure)
...
 
+        public static void AddOrUpdateIsSqlDw(IDbConnection connection, bool isSqlDw)
+        {
+            Validate.IsNotNull(nameof(connection), connection);
+
+            SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connection.ConnectionString);
+            AddOrUpdateIsSqlDw(builder.DataSource, isSqlDw);
+        }
+
+        public static void AddOrUpdateIsSqlDw(string dataSource, bool isSqlDw)
```
kburtram: >AddOrUpdateIsSqlDw [](start = 27, length = 18)

this looks basically the same as AddOrUpdateIsAzure..is it possible to factor common logic into a shared method?

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
Raymondd: The engine version it is compared to is an integer, so I can either cast it here or later on when I compare it. I decided to do it here because the DacFx code had an explicit '= 6' instead of a reference the enum. So if we ever port over any other code that uses this variable, things will be consistent.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
Raymondd: Checkout the function `OpenConnection()` on line 87 of this same file. It calls `open()` which propagates to all these executed commands. Any thrown exceptions are caught and bubbled up as errors to the client

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
Raymondd: I added an enum to categorize each variable and combined all the logic. It is a departure from the way DacFX was doing it, but I do think this design is more sustainable. (especially if we decide to cache more connection information).

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/CachedServerInfo.cs
```diff
@@ -108,13 +123,45 @@ public static void AddOrUpdateIsAzure(string  dataSource, bool isAzure)
...
 
+        public static void TryGetIsSqlDw(IDbConnection connection, out bool isSqlDw)
+        {
+            Validate.IsNotNull(nameof(connection), connection);
+
+            SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connection.ConnectionString);
+            TryGetIsSqlDw(builder.DataSource, out isSqlDw);
+        }
+
+        public static void TryGetIsSqlDw(string dataSource, out bool isSqlDw)
```
kevcunnane: TryGet pattern: return bool indicating if found or not. Do not throw an exception. In references to the code, use `if (!TryGetIsSqlDw(...)` and in the if clause read the value and update the cache if it wasn't already set.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableSqlConnection.cs
```diff
@@ -96,6 +97,30 @@ protected override void Dispose(bool disposing)
...
+        /// <param name="conn">A connection object.</param>
+        private bool IsSqlDwConnection(IDbConnection conn)
+        {
+            //Set the connection only if it has not been set earlier.
+            //This is assuming that it is highly unlikely for a connection to change between instances.
+            //Hence any subsequent calls to this method will just return the cached value and not 
+            //verify again if this is a SQL DW database connection or not.
+            try
+            {
+                CachedServerInfo.TryGetIsSqlDw(conn, out _isSqlDwDatabase);
```
kevcunnane: Per previous comment, replace try/catch with `if (!TryGet)` as that's the standard pattern for TryGet methods :)

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableSqlConnection.cs
```diff
@@ -341,9 +373,13 @@ private SqlConnection OpenConnection()
                 if (_underlyingConnection.State != ConnectionState.Open)
                 {
                     _underlyingConnection.Open();
+
+                    // Asking if the server is sqldw or not
+                    IsSqlDwConnection(_underlyingConnection);
```
kevcunnane: Minor: why call it in here when you call it again in `SetDefaultAnsiSettings` just a few lines down? Is there a benefit / need to do so?

test/Microsoft.SqlTools.ServiceLayer.Test/Connection/CachedServerInfoTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        [InlineData(true)]  // is SqlDW instance
+        [InlineData(false)]   // is not a SqlDw Instance
+        public void AddOrUpdateIsSqlDw(bool state)
+        {
+            // Set sqlDw result into cache
+            bool isSqlDwResult;
+            CachedServerInfo.AddOrUpdateCache("testDataSource", state, CachedServerInfo.CacheVariable.IsSqlDw);
+
+            // Expect the same returned result
+            CachedServerInfo.TryGetIsSqlDw("testDataSource", out isSqlDwResult);
```
kevcunnane: please check return values once this returns bool

test/Microsoft.SqlTools.ServiceLayer.Test/Connection/CachedServerInfoTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            // Assert cache is set on a per connection basis
+            Assert.Equal(isSqlDwResult, state);
+            Assert.Equal(isSqlDwResult2, !state);
+
+        }
+
+        [Fact]
+        public void AskforSqlDwBeforeCached()
+        {
+            bool exceptionThrown = false;
```
kevcunnane: And please rewrite this per change to TryGet method

src/Microsoft.SqlTools.ServiceLayer/Utility/CommandOptions.cs
```diff
@@ -32,6 +32,10 @@ public CommandOptions(string[] args)
                             case "-enable-logging":
                                 EnableLogging = true;
                                 break;
+                            case "-locale":
```
kevcunnane: Minor: this is for a separate review :-)

kburtram:
:shipit:

PR #222, opened 2017-02-01
Fix issue where calling "use database" can fail to change the database

coquagli:
Fixes [#645](https://github.com/Microsoft/vscode-mssql/issues/645). 

The issue was that the `ChangeConnectionDatabaseContext` callback in [ConnectionService](https://github.com/Microsoft/sqltoolsservice/blob/dev/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs#L942) was sometimes called while the query was still running, which would make `ChangeDatabase()` throw: "There is already an open DataReader associated with this Command which must be closed first". This Exception was being caught and ignored.

If we postpone the callback until after the query completes, the DataReader from the previous query will be closed. This, however, does not prevent the case where a new query begins execution before the `ChangeConnectionDatabaseContext` for the previous query completes. Thoughts on ignoring this case?

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
@@ -64,6 +64,11 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
         private IFileStreamFactory streamOutputFactory;
 
         /// <summary>
+        /// Name of the new database if the database name was changed in the query
+        /// </summary>
+        private string newDatabaseName;
+
+        /// <summary>
         /// ON keyword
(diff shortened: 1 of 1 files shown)


benrr101:
> This, however, does not prevent the case where a new query begins execution before the ChangeConnectionDatabaseContext for the previous query completes

I believe this isn't an issue. We can't begin executing a new query (or batch) until the first one completes execution. Therefore, the only way that we could get start execution of a new query before the ChangeConnectionDatabaseContext finishes is if that method is truly async (and not awaited).

coquagli:
Is it possible for `OnInfoMessage` to get called after the query completes? If so, I may need more changes.

PR #223, opened 2017-02-02
Inter-Service API for executing queries

benrr101:
Adding new methods for executing queries from other services (such as the upcoming edit data service). The code is written to avoid duplicating logic by using lambdas to perform custom logic.
Additionally, the service host protocol has been slightly modified to split the IMessageSender into IEventSender and IRequestSender. This allows us to use either a ServiceHost or any RequestContext<T> to send events. It becomes very convenient to use another service's request context to send the events for query execution.

**Breaking Change**: This removes the messages property for query dispose results and instead elects to use error for any errors encountered during query disposal. A result is only used when something is successful.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
@@ -332,6 +332,11 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
         public async Task InterServiceExecuteQuery(ExecuteRequestParamsBase executeParams, IEventSender eventSender,
             Func<Task> queryCreatedAction, Func<string, Task> failureAction)
         {
+            Validate.IsNotNull(nameof(executeParams), executeParams);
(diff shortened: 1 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -338,17 +316,75 @@ public void InitializeService(ServiceHost serviceHost)
...
+                {
+                    await failureAction(SR.QueryServiceRequestsNoQuery);
+                    return;
+                }
+
+                // Cleanup the query
+                result.Dispose();
+
+                // Success
+                await successAction();
```
coquagli: To prevent an `await null` we should either check for null here or in `InterServiceExecuteQuery`. Likewise for `failureAction`.

benrr101: Good call on that. I didn't think of that idea. For now I'm just going to throw on null.

PR #226, opened 2017-02-06
Task: localization for the sqltoolsservice 

Raymondd:
Completes https://github.com/Microsoft/vscode-mssql/issues/641

- adds CLI option to set localization (`--locale XX`)
- adds an example of localization with a test value in `sr.es.resx`
- adds tests for both of the above features 
- integrated with srGen (fallback is set to sr.resx (the file srGen created by default))


src/Microsoft.SqlTools.ServiceLayer/Utility/CommandOptions.cs
```diff
@@ -26,18 +28,31 @@ public CommandOptions(string[] args)
                     string arg = args[i];
                     if (arg.StartsWith("--") || arg.StartsWith("-"))
                     {
+                        // Extracting arguments and properties
                         arg = arg.Substring(1).ToLowerInvariant();
-                        switch (arg)
+                        string argName = arg;
+                        string argProperty = "";
+                        int splitIndex = arg.IndexOf(' ');
```
benrr101: So this code splits arguments by " "? We would only get an argument like this if we get `sqltoolsservice.exe "some argument"`. What sort of scenario do we expect this to happen?

src/Microsoft.SqlTools.ServiceLayer/Utility/CommandOptions.cs
```diff
@@ -82,10 +102,32 @@ public string Usage
                     "Microsoft.SqlTools.ServiceLayer.exe " + Environment.NewLine +
                     "   Options:" + Environment.NewLine +
                     "        [--enable-logging]" + Environment.NewLine +
-                    "        [--help]" + Environment.NewLine,
+                    "        [--help]" + Environment.NewLine +
+                    "        [--locale]" + Environment.NewLine,
```
benrr101: Will we provide a list of supported locales?

src/Microsoft.SqlTools.ServiceLayer/Utility/CommandOptions.cs
benrr101: Ah, I see what's going on now. I think it might be a bit easier to have your argument set up like `--locale=es` which would give you a single argument for the locale. Then in the locale case you could split on = and take the second item in the array

Raymondd: The convention for dotnet command line arguments is `--myarg property`. 
In this case I am using `--locale enu`. If the argument has a space to split by, then I split it into its argument name and property, otherwise i just keep it as an argument with an empty property. We were always just using arguments with no properties. so this wasn't needed until now.

src/Microsoft.SqlTools.ServiceLayer/Utility/CommandOptions.cs
Raymondd: Right now I just have is print an exception telling you the locale is not supported to console and fall back to en-US.

PR #227, opened 2017-02-07
Check database case sensitivity for peek definition token matching

sharonravindran:
Added case sensitivity checks to token match based on database property.
Modified and added tests.


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
```diff
@@ -377,16 +388,17 @@ internal DefinitionResult GetDefinitionErrorResult(string errorMessage)
...
+        internal string GetFullObjectNameFromQuickInfo(string quickInfoText, string tokenText, StringComparison caseSensitivity)
         {
             if (string.IsNullOrEmpty(quickInfoText) || string.IsNullOrEmpty(tokenText))
             {
                 return null;
             }
             // extract full object name from quickInfo text
             string[] tokens = quickInfoText.Split(' ');
-            List<string> tokenList = tokens.Where(el => el.Contains(tokenText)).ToList();
+            List<string> tokenList = tokens.Where(el => el.IndexOf(tokenText, caseSensitivity) >= 0).ToList();
```
benrr101: It's preferable to not use IndexOf, but in this case there's not much choice. Would it possibly be suitable to do .ToLowerInvariant() or .ToUpperInvariant() and do .Contains on that?

test/Microsoft.SqlTools.ServiceLayer.Test/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -251,17 +267,17 @@ public void GetFullObjectNameFromQuickInfoWithNullStringsTest()
 
             string objectName = null;
             string quickInfoText = "table master.dbo.testTable";
-            string result = peekDefinition.GetFullObjectNameFromQuickInfo(quickInfoText, objectName);
+            string result = peekDefinition.GetFullObjectNameFromQuickInfo(quickInfoText, objectName, StringComparison.Ordinal);
```
benrr101: Ordinal or OrdinalIgnoreCase?

sharonravindran: I added a separate test for OrdinalIgnoreCase. So I'm leaving this as Ordinal which was the original test intent.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/PeekDefinition.cs
sharonravindran: Using IndexOf seems more straightforward and we can keep the case sensitivity check's code pattern consistent everywhere.

PR #229, opened 2017-02-10
TSQL Formatter Service

kevcunnane:
This is an initial PR intended to help see code coverage changes and prep for review. As this is a rather large port, I'd recommend using a viewing tool such as CodeFlow since GitHub is hiding file contents. Note that I intend to add additional service-level tests at a minimum before checking in. I will also likely fix a number of code formatting issues inherited from the internal project, and add better documentation.

Changes in this PR:
- Added a TSqlFormatterService with support for formatting document and text range inside document.
- Added Settings support for all formatting options. 
- Created a new ServiceProvider patter with use of MEF for extensible service lookup. The TSqlFormatterService is initialized using MEF extensibility, and can find all necessary TSqlFormatters and the WorkspaceService using this. We will want to apply this across the codebase eventually, removing the singleton .Instance pattern. This makes test configuration easier.
- Unit Tests and baseline test framework for the formatter code has been ported from the internal project. Will investigate coverage % and major gaps after this.


kburtram:
I see this PR includes the "_site" output folder in the docs directory.  This is the documentation build output and probably doesn't need to be checked in..it will also clean-up the PR if you remove those files.  Maybe updating the .gitignore to include _site is a good idea.  

That directory should be published to the sqltoolssdk repo to update the documentation page.

Raymondd:
Because this branch was merged in with a dev head which not longer exists (due to the rebase) we are going to have to revert the commit that brought in dev: d8cffea88beef29263098b10b3c4e40e24c05283 .

kevcunnane:
@Raymondd fixed. The branch now has the same commits as Dev

kevcunnane:
@kburtram - the inclusion of _sites folder was from Aditya's PR. Now we fixed the merge issue it no longer shows up in this review.

I think we should fix by removing them from source in a different PR

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            InitializeRequestHandlersAndServices(serviceHost, sqlToolsContext);
+
+            // Start the service only after all request handlers are setup. This is vital
+            // as otherwise the 
+            serviceHost.Start().Wait();
+            return serviceHost;
+        }
+
+        internal static void InitializeRequestHandlersAndServices(ServiceHost serviceHost, SqlToolsContext sqlToolsContext)
```
llali: >InitializeRequestHandlersAndServices [](start = 29, length = 36)

InitializeRequestHandlersAndServices and InitializeHostServices should be private

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+namespace Microsoft.SqlTools.ServiceLayer
+{
+    /// <summary>
+    /// Provides support for starting up a service host. This is a common responsibility
+    /// for both the main service program and test driver that interacts with it
+    /// </summary>
+    public static class HostLoader
+    {
+
+        internal static ServiceHost CreateAndStartServiceHost(SqlToolsContext sqlToolsContext)
```
llali: we should make sure this method only get called once.

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
kevcunnane: I can make this method private, but InitializeHostedServices is used by the TSqlFormatter unit tests. I'll update to just have this one private

---
In reply to: [101151457](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101151457) [](ancestors = 101151457)

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
kevcunnane: Will do

---
In reply to: [101151664](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101151664) [](ancestors = 101151664)

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+
+            // Grab the instance of the service host
+            ServiceHost serviceHost = ServiceHost.Instance;
+            serviceHost.Initialize();
+
+            InitializeRequestHandlersAndServices(serviceHost, sqlToolsContext);
+
+            // Start the service only after all request handlers are setup. This is vital
+            // as otherwise the
```
kburtram: >the  [](start = 28, length = 4)

sentence seems to stop abruptly

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+            serviceProvider.RegisterSingleService(QueryExecutionService.Instance);
+
+            InitializeHostedServices(serviceProvider, serviceHost);
+
+            serviceHost.InitializeRequestHandlers();
+        }
+
+        /// <summary>
+        /// Internal to support testing. Initializes <see cref="IHostedService"/> instances in the service,
+        /// an registers them for their preferred service type
```
kburtram: >an [](start = 12, length = 2)

and

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -720,7 +720,7 @@ private void RunConnectRequestHandlerTask(ConnectParams connectParams)
                     }
 
                     // open connection based on request details
-                    result = await Instance.Connect(connectParams);
```
kburtram: >Instance [](start = 35, length = 8)

why are we just removing Instance in a couple cases and not all the cases.  I'm assuming the motivation from removing instance is that someday we'll have MEF create the singleton and then Instance would be the wrong actual instance?

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -368,7 +368,9 @@ public static void SetLockAndCommandTimeout(IDbConnection conn)
             IDbCommand cmd = null;
             try
             {
-                Debug.Assert(conn.State == ConnectionState.Open, "connection passed to ExecuteReader should be open.");
+                // Skipping this debug assert as it is causing debug errors to be logged on Mac when tests with mock
+                // connections are run
+                // Debug.Assert(conn.State == ConnectionState.Open, "connection passed to ExecuteReader should be open.");
```
kburtram: >// Debug.Asser [](start = 16, length = 14)

should we just delete the assert if there aren't plans to reenable?

src/Microsoft.SqlTools.ServiceLayer/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+    
+    /// <summary>
+    /// A store for MEF exports of a specific type. Provides basic wrapper functionality around MEF to standarize how
+    /// we lookup types and return to callers.
+    /// </summary>
+    public class ExtensionStore
+    {
+        CompositionHost host;
```
kburtram: >CompositionHost [](start = 8, length = 15)

internal?

src/Microsoft.SqlTools.ServiceLayer/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+            string directory = Path.GetDirectoryName(assemblyPath);
+            return new ExtensionStore(typeof(T), (conventions) => 
+                new ContainerConfiguration().WithAssembliesInPath(directory, conventions));
+        }
+
+        public IEnumerable<T> GetExports<T>()
+        {
+            if (exports == null)
+            {
+                exports = host.GetExports(contractType).ToList();
```
kburtram: >GetExports [](start = 31, length = 10)

this can't return null?

src/Microsoft.SqlTools.ServiceLayer/Extensibility/IComposableService.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+
+namespace Microsoft.SqlTools.ServiceLayer.Extensibility
+{
+    /// <summary>
+    /// A Service that expects to lookup other services. 
+    /// </summary>
+    interface IComposableService
```
kburtram: >interface [](start = 4, length = 9)

internal?

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
kevcunnane: Yeah, will do

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kevcunnane: It could cause problems if unit testing any ConnectionService that's not the static instance. Right now, you can do this but this method will call into the static instance and register connections. That's a bug and also just bad practice (it should be calling its own Connect method).

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -0,0 +1,270 @@
...
+                options.KeywordCasing = settings.KeywordCasing;
+            }
+            return options;
+        }
+        
+        private ScriptFile GetFile(DocumentFormattingParams docFormatParams)
+        {
+            if (WorkspaceService == null)
+            {
+                // TODO throw
```
kburtram: >// TODO throw [](start = 16, length = 13)

why not just throw since the next line is going to throw anyways...or delete the if and just let the following line throw

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -0,0 +1,270 @@
...
+            }
+            return WorkspaceService.Workspace.GetFile(docFormatParams.TextDocument.Uri);
+        }
+
+        private static TextEdit PrepareEdit(ScriptFile scriptFile)
+        {
+            Position start = new Position { Line = 0, Character = 0 };
+            int lastChar = scriptFile.FileLines[scriptFile.FileLines.Count - 1].Length;
```
kburtram: >Count [](start = 69, length = 5)

do we know this will always be greater then 0?

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -0,0 +1,270 @@
...
+            Position end = new Position { Line = scriptFile.FileLines.Count -1, Character = lastChar };
+
+            TextEdit edit = new TextEdit
+            {
+                Range = new Range { Start = start, End = end }
+            };
+            return edit;
+        }
+
+        // TODO Assess whether to return a single TextEdit for the formatting, or multiple edits.
```
kburtram: >// TODO  [](start = 8, length = 8)

I think best practice with disabled code is to leave a date somewhere in the comments, so people know to delete in a couple years when cleaning-up

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -0,0 +1,270 @@
...
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex.ToString());
+            }
+        }
+
+
+
+        public string Format(TextReader input)
```
kburtram: >public [](start = 8, length = 6)

static?  same with other Format methods it looks like.

kburtram:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/Extensibility/ExtensionServiceProvider.cs
kevcunnane: private - will fix

---
In reply to: [101186997](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101186997) [](ancestors = 101186997)

src/Microsoft.SqlTools.ServiceLayer/Extensibility/ExtensionServiceProvider.cs
kevcunnane: No - it's an enumerable and returns empty enum rather than null.

---
In reply to: [101187416](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101187416) [](ancestors = 101187416)

src/Microsoft.SqlTools.ServiceLayer/Extensibility/IComposableService.cs
kevcunnane: Really should be public - will fix

---
In reply to: [101187599](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101187599) [](ancestors = 101187599)

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
kevcunnane: Yeah, will just remove and let the following line throw.

---
In reply to: [101188299](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101188299) [](ancestors = 101188299)

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
kevcunnane: In reality any file should have at least 1 non-empty line. However it's a valid point, I'll update to verify this.

---
In reply to: [101188393](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101188393) [](ancestors = 101188393)

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
kevcunnane: I'll delete.

---
In reply to: [101188515](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101188515) [](ancestors = 101188515)

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
kevcunnane: No - all rely on the DoFormat method which uses ServiceProvider to get the Formatter factories. Therefore you need these to be instance-level methods since they require the ServiceProvider to be available. 

---
In reply to: [101188584](https://github.com/Microsoft/sqltoolsservice/pull/229#discussion_r101188584) [](ancestors = 101188584)

PR #231, opened 2017-02-10
Adding TestDbResultSet for Better Mocking

benrr101:
Replacing the awful and inflexible Dictionary<string, string>[][] with a much better test data set class. This is a necessary step for making proper unit tests for the EditData service. But since it's such a large change by itself, I figured I'd cherry-pick it to be merged in separately.


Raymondd:
Because this branch was merged in with a dev head which not longer exists (due to the rebase) we are going to have to revert the commit that brought in dev:  af2de43656011996f1bd655544154ba8290d9008.

test/Microsoft.SqlTools.ServiceLayer.Test/Connection/ConnectionServiceTests.cs
```diff
@@ -61,29 +57,26 @@ public DbConnection CreateMockDbConnection(Dictionary<string, string>[][] data)
         [Fact]
         public void CanCancelConnectRequest()
         {
-            var testFile = "file:///my/test/file.sql";
+            const string testFile = "file:///my/test/file.sql";
```
benrr101: Agreed. I have some other changes I'd like to make to properly isolate and share commonly used code in the test suites. It'll be in a forthcoming PR.

PR #235, opened 2017-02-13
Fix: Adding OpenAsync to ReliableConnection

benrr101:
Adding an OpenAsync method to the ReliableConnection to allow it to properly utilize the cancellation token support of the SqlConnection. It eliminates the need for a "cancellation thread" that has a high likelihood of throwing uncaught exceptions (as per Microsoft/vscode-mssql/issues/547)


coquagli:
This looks good. We should either file a separate issue to alter `Connect()` to remove the usage of the 2 connections threads (`openTask` and `cancellationTask`), or we should do it in this PR.

benrr101:
@kevcunnane I'm thinking this is a drop-in replacement for the OpenConnection code we had before. I'm not entirely sure if we should add more tests for this or if the tests we have are sufficient?

PR #238, opened 2017-02-15
Clear and rebuild IntelliSense cache

kburtram:


kevcunnane:
Looks like the compilers on AppVeyor/Travis aren't happy with the optional arg you added in one of the files: `An expression tree may not contain a call or invocation that uses optional arguments`

kburtram:
I've fixed the broken tests and added a new test case for the "overwrite connection context" method.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -443,6 +446,51 @@ await requestContext.SendEvent(TelemetryNotification.Type, new TelemetryParams()
         }
 
         /// <summary>
+        /// Handle the rebuild IntelliSense cache notification
+        /// </summary>
+        public async Task HandleRebuildIntelliSenseNotification(ScriptFile scriptFile, EventContext eventContext)
```
kevcunnane: Minor: there's a lot of overlap between this and the UpdateLanguageServiceOnConnection method. Could we share some of the logic between them?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -230,6 +230,9 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
             // Register the file open update handler
             WorkspaceService<SqlToolsSettings>.Instance.RegisterTextDocOpenCallback(HandleDidOpenTextDocumentNotification);
 
+            // Register the IntelliSense rebuild handler
+            WorkspaceService<SqlToolsSettings>.Instance.RegisterRebuildIntelliSenseCallback(HandleRebuildIntelliSenseNotification);
```
kevcunnane: We don't need this indirection - we should just call `serviceHost.SetRequestHandler(RebuildIntelliSenseNotification.Type,, HandleRebuildIntelliSenseNotification)` instead. We already handle most other requests here (except global events that are common to the workspace and other stuff)

src/Microsoft.SqlTools.ServiceLayer/Workspace/WorkspaceService.cs
```diff
@@ -94,6 +96,13 @@ public WorkspaceService()
         public delegate Task TextDocCloseCallback(ScriptFile closedFile, EventContext eventContext);
 
         /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="openFile"></param>
+        /// <param name="eventContext"></param>
+        public delegate Task RebuildIntelliSenseCallback(ScriptFile openFile, EventContext eventContext);
```
kevcunnane: This isn't needed - per my other comment, we should handle the event directly in the LanguageService

kevcunnane:
:clock1:

PR #241, opened 2017-02-16
Edit Data Service

benrr101:
This is a very large change. I'll try to outline what's going on.

1. This adds the **EditDataService** which manages editing **Sessions**.
    1. Each session has a **ResultSet** (from the QueryExecutionService) which has the rows of the table and basic metadata about the columns 
    2. Each session also has an **IEditTableMetadata** implementation which is derived from SMO metadata which provides more in-depth and trustworthy data about the table than SqlClient alone can.
    3. Each session holds a list of **RowEditBase** abstract class implementations
        1. **RowUpdate** - Update cells in a row (generates `UPDATE` statement)
        2. **RowDelete** - Delete an entire row (generates `DELETE` statement)
        3. **RowCreate** - Add a new row (generates `INSERT INTO` statement)
    4. Row edits have a collection of **CellUpdates** that hold updates for individual cells (except for RowDelete)
        1. Cell updates are generated from text
     5. RowEditBase offers some baseline functionality
        1. Generation of `WHERE` clauses (which can be parameterized)
        2. Validation of whether a column can be updated
2. New API Actions
    1. edit/initialize - Queries for the contents of a table/view, builds SMO metadata, sets up a session
    2. edit/createRow - Adds a new RowCreate to the Session
    3. edit/deleteRow - Adds a new RowDelete to the Session
    4. edit/updateCell - Adds a CellUpdate to a RowCreate or RowUpdate in the Session
    5. edit/revertRow - Removes a RowCreate, RowDelete, or RowUpdate from the Session
    6. edit/script - Generates a script for the changes in the Session and stores to disk
    7. edit/dispose - Removes a Session and releases the query
3. Smaller updates (unit test mock improvements, tweaks to query execution service)

**There are more updates planned -- this is just to get eyeballs on the main body of code**


benrr101:
Ping!
I believe I've gotten this code to a state that is complete for the functionality it provides. Please please please take a look at it!

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -200,6 +200,43 @@ WorkspaceServicePositionColumnOutOfRange(int line) = Position is outside of colu
...
+
+EditDataUpdatePending = An update is already pending for this row and must be reverted first
+
+EditDataUpdateNotPending = Given row ID does not have pending updated
+
+EditDataObjectMetadataNotFound = Table or view metadata could not be found
+
+EditDataInvalidFormatBinary = Invalid format for binary column
+
+EditDataInvalidFormatBoolean = Boolean columns must be numeric 1 or 0, or string true or false
```
kevcunnane: Allowed values for boolean columns are 0, 1, "true", "false"

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -200,6 +200,43 @@ WorkspaceServicePositionColumnOutOfRange(int line) = Position is outside of colu
 WorkspaceServiceBufferPositionOutOfOrder(int sLine, int sCol, int eLine, int eCol) = Start position ({0}, {1}) must come before or be equal to the end position ({2}, {3})
 
 ############################################################################
+# Edit Data Service
+
+EditDataSessionNotFound = Edit session does not exist.
+
+EditDataUnsupportedObjectType(string typeName) = Database object {0} cannot be used for editing.
+
+EditDataQueryNotCompleted = Query has not completed execution
```
kevcunnane: A lot of these might be weird when shown in the front-end UI. In what scenarios will they be used? Can we get a PM to review them? I care more about ones that are likely to show up a lot vs. diagnostic / edge case error messages

kevcunnane:
> using System;

Missing copyright.

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs:1 in ecc2f51. [](commit_id = ecc2f5192778995aae2cc58fec849fe2524e2144, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -0,0 +1,328 @@
...
+        /// </summary>
+        internal ConcurrentDictionary<string, Session> ActiveSessions => editSessions.Value;
+
+        #endregion
+
+        /// <summary>
+        /// Initializes the edit data service with the service host
+        /// </summary>
+        /// <param name="serviceHost">The service host to register commands/events with</param>
+        public void InitializeService(ServiceHost serviceHost)
```
kevcunnane: You don't initialize this in Program.cs. You will need to update it there. Or as I mentioned, make this an IHostedService like the FormatterService and it'll "just work".

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -0,0 +1,328 @@
...
+                await requestContext.SendError(e.Message);
+            }
+        }
+
+        internal async Task HandleDeleteRowRequest(EditDeleteRowParams deleteParams,
+            RequestContext<EditDeleteRowResult> requestContext)
+        {
+            try
+            {
+                Session session = GetActiveSessionOrThrow(deleteParams.OwnerUri);
```
kevcunnane: All the session-dependent methods could be done from a single function - HandleSessionRequest<T>. We have a general pattern for this (see the TSqlFormatterService) as shown below. This would be a specialized version, that also takes in the session - something like below

```
        private async Task HandleRequest<T>(Func<Session,Task<T>> handler, RequestContext<T> requestContext, string requestType, Uri ownerUri)
        {
            Logger.Write(LogLevel.Verbose, requestType);

            try
            {
                Session session = GetActiveSessionOrThrow(ownerUri);
                T result = await handler(session);
                await requestContext.SendResult(result);
            }
            catch (Exception ex)
            {
                await requestContext.SendError(ex.ToString());
            }
        }
```

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -0,0 +1,328 @@
...
+            catch (Exception e)
+            {
+                await requestContext.SendError(e.Message);
+            }
+        }
+
+        internal async Task HandleInitializeRequest(EditInitializeParams initParams,
+            RequestContext<EditInitializeResult> requestContext)
+        {
+            try
```
kevcunnane: As mentioned above, I've found a handy pattern is to have 1 "HandleRequest" method that does the error handling & returning of messages over the channel. I'd suggest using it here instead of the try/catch. See TSqlFormatterService.HandleRequest<T>()

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -0,0 +1,328 @@
...
+                        OwnerUri = initParams.OwnerUri,
+                        Success = false
+                    };
+                    return requestContext.SendEvent(EditSessionReadyEvent.Type, readyParams);
+                };
+
+                // Put together a query for the results and execute it
+                ExecuteStringParams executeParams = new ExecuteStringParams
+                {
+                    Query = $"SELECT * FROM {SqlScriptFormatter.FormatMultipartIdentifier(initParams.ObjectName)}",
```
kevcunnane: >$ [](start = 28, length = 1)

Is this a new string.format syntax? We should have a plan for these new language features. They're nice, I just worry about how long it takes for people to learn these. We're often a bit conservative about how long before we adopt this kind of syntactic sugar in order to avoid .Net updates (not an issue here since it's done at compile time) but also ramp up time for people not familiar with latest & greatest.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -0,0 +1,328 @@
...
+        private void Dispose(bool disposing)
+        {
+            if (disposed)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                // TODO: Dispose objects that need disposing
```
kevcunnane: And these are...?

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Linq;
+using System.Diagnostics;
+using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
+{
+    /// <summary>
+    /// Class that provides metadata about the table or view being edited
```
kevcunnane: "Provides metadata about the table or view being edited" - class is redundant

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -0,0 +1,96 @@
...
+            if (keyColumns.Count == 0)
+            {
+                // We didn't find any explicit key columns. Instead, we'll use all columns that are
+                // trustworthy for uniqueness (usually all the columns)
+                keyColumns = columns.Where(c => c.IsTrustworthyForUniqueness).ToList();
+            }
+
+            // If a table is memory optimized it is Hekaton. If it's a view, then it can't be Hekaton
+            Table smoTable = smoObject as Table;
+            IsHekaton = smoTable != null && smoTable.IsMemoryOptimized;
```
kevcunnane: Please use IsMemoryOptimized instead of Hekaton. We should try to move away from codenames, even though it's that way in older internal code.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        /// <summary>
+        /// Whether or not the object being edited is hekaton/memory optimized
+        /// </summary>
+        public bool IsHekaton { get; }
+
+        /// <summary>
+        /// Read-only list of columns that are used to uniquely identify a row
+        /// </summary>
+        public IEnumerable<EditColumnWrapper> KeyColumns => keyColumns;
```
kevcunnane: >keyColumns [](start = 60, length = 10)

Does this really return a ReadonlyCollection? Shouldn't you call keyColumns.AsReadOnly instead? And do the same for Columns?

Also, as mentioned before: the desire to shorten things is nice, but we should agree whether we really want to do this or keep with the boring (but clear) old { get { return xyz; } } syntax.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        private readonly List<EditColumnWrapper> columns;
+        private readonly List<EditColumnWrapper> keyColumns;
+
+        /// <summary>
+        /// Constructor that extracts useful metadata from the provided metadata objects
+        /// </summary>
+        /// <param name="dbColumns">DB columns from the ResultSet</param>
+        /// <param name="smoObject">SMO metadata object for the table/view being edited</param>
+        public EditTableMetadata(IList<DbColumnWrapper> dbColumns, TableViewTableTypeBase smoObject)
```
kevcunnane: Validate not null please

src/Microsoft.SqlTools.ServiceLayer/EditData/IEditTableMetadata.cs
```diff
@@ -0,0 +1,35 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.EditData
+{
+    /// <summary>
+    /// Interface for a metadata provider to be used for edit scenarios
```
kevcunnane: An <see cref="IEditTableMetadata" /> is used in edit data scenarios, defining necessary properties to determine which columns are editable, how to handle updates to the table, and what the primary key columns are among other things.

src/Microsoft.SqlTools.ServiceLayer/EditData/IEditTableMetadata.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// The escaped name of the table that's being edited
+        /// </summary>
+        string EscapedMultipartName { get; }
+
+        /// <summary>
+        /// Whether or not this table is a hekaton table
+        /// </summary>
+        bool IsHekaton { get; }
```
kevcunnane: IsMemoryOptimized

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
```diff
@@ -0,0 +1,194 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.SqlTools.ServiceLayer.EditData.Contracts;
+using Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.EditData
+{
+    public class Session
```
kevcunnane: Document please

kevcunnane: >publ [](start = 4, length = 4)

Why internal?

kevcunnane: >publ [](start = 4, length = 4)

Why internal? Also this is formatted like a property, not a field (TitleCase instead of camelCase) - why not at a minimum make a property?

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            // Sanity check the row ID
+            if (rowId >= NextRowId || rowId < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(rowId), SR.EditDataRowOutOfRange);
+            }
+
+            // Create a new row delete update and add to cache
+            RowDelete deleteRow = new RowDelete(rowId, associatedResultSet, objectMetadata);
+            if (!EditCache.TryAdd(rowId, deleteRow))
```
kevcunnane: Doesn't this break in the following scenario:
- User adds a new row, but doesn't commit
- User then calls delete on that row (not revert)

Also fails in the scenario:
- User edits an existing row (added to the cache)
- User then calls delete on that row

These could in theory be handled by the UI code, but really we should cleanly handle here by changing the cache info, rather than throwing if something previously changed the row

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+                    return null;
+                }
+
+                // We have a reliable connection, use the underlying connection
+                sqlConn = reliableConn.GetUnderlyingConnection();
+            }
+
+            Server server = new Server(new ServerConnection(sqlConn));
+            TableViewTableTypeBase result;
+            switch (objectType.ToLowerInvariant())
```
kevcunnane: >objectType [](start = 20, length = 10)

Could/should we use an enum here?

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                case "table":
+                    result = server.Databases[sqlConn.Database].Tables[objectName];
+                    break;
+                case "view":
+                    result = server.Databases[sqlConn.Database].Views[objectName];
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(objectType), SR.EditDataUnsupportedObjectType(objectType));
```
kevcunnane: We're throwing a lot of different exception types across the codebase. I'd suggest that for external code manageability, could we consider having fewer ones? For truly unexpected behavior, throw argument exceptions or invalid operation exceptions. But for most things, have a common exception type that says "couldn't do xyz, please report error to the user"

src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCreateRowRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+//
+
+using Microsoft.SqlTools.ServiceLayer.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.EditData.Contracts
+{
+    /// <summary>
+    /// Parameters for the update cell request
+    /// </summary>
+    public class EditCreateRowParams
```
kevcunnane: Could we have a generic EditParams class with the OwnerUri and have others inherit from it? This would make handling things like the HandleSessionRequest method in the service (that I suggested) easier, since we'd know any params for an edit would have that.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -0,0 +1,192 @@
...
+        private static readonly Regex HexRegex = new Regex("0x[0-9A-F]+", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+        /// <summary>
+        /// Constructs a new cell update based on the the string value provided and the column
+        /// for the cell.
+        /// </summary>
+        /// <param name="column">Column the cell will be under</param>
+        /// <param name="valueAsString">The string from the client to convert to an object</param>
+        public CellUpdate(DbColumn column, string valueAsString)
+        {
```
kevcunnane: Please convert this 150 line constructor into many smaller methods - each handler should be its own method. This should just do the basic routing so it's clear what we're doing.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
benrr101: A lot of these definitely are edge cases and very unlikely to occur. I definitely think we should be catching these edge cases -- assuming they won't happen will lead to disaster and likewise throwing generic "failed" messages will make things a pain to track down. Maybe the wording of the errors can be changed or we can have more detailed errors for internal usage, but I think that can wait for a later PR.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEdit.cs
```diff
@@ -0,0 +1,194 @@
...
+        /// The internal ID of the row to which this edit applies, relative to the result set
+        /// </summary>
+        public long RowId { get; }
+
+        /// <summary>
+        /// The result set that is associated with this row edit
+        /// </summary>
+        public ResultSet AssociatedResultSet { get; }
+
+        public IEditTableMetadata AssociatedObjectMetadata { get; }
```
kevcunnane: Missed the doc comment for this one?

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
benrr101: It's a C# 6 feature, string interpolation. Basically a string literal that has a $ in front of the quotation mark will include anything inside `{}` in the string. It's pretty straight forward and I don't think it's unclear what's going on, especially considering this is quite similar to how Ruby, PowerSell and other languages offer string interpolation.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <summary>
+        /// Constructs an update statement to change the associated row.
+        /// </summary>
+        /// <returns>An UPDATE statement</returns>
+        public override string GetScript()
+        {
+            // Build the "SET" portion of the statement
+            IEnumerable<string> setComponents = cellUpdates.Select(kvp =>
+            {
+                string formattedColumnName = SqlScriptFormatter.FormatIdentifier(kvp.Value.Column.ColumnName);
```
kevcunnane: >kvp.Value [](start = 81, length = 9)

Please give this a useful name like "updateInfo". It's really hard when hopping into code to track down what the value in a kvp is "meant" to represent.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Execution/RequestParamTests.cs
```diff
@@ -0,0 +1,78 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+//using System;
+//using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests;
+//using Microsoft.SqlTools.ServiceLayer.SqlContext;
+//using Microsoft.SqlTools.ServiceLayer.Workspace;
```
kevcunnane: Please remove this file since it's entirely commented out. Add back in later when you have it ready to go, but no benefit in committing as-is

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Execution/ServiceIntegrationTests.cs
```diff
@@ -91,66 +91,66 @@ public void GetSqlTextFromInvalidType()
...
-            var eventSender = new EventFlowValidator<ExecuteRequestResult>().Complete().Object;
-            Func<Task> successFunc = () => Task.FromResult(0);
-
-            // If: I call the inter-service API to execute a query with a a null failure function
-            // Then: It should throw
-            await Assert.ThrowsAsync<ArgumentNullException>(
-                () => qes.InterServiceExecuteQuery(executeParams, eventSender, successFunc, null));
-        }
+        //[Fact]
+        //public async Task InterServiceExecuteNullExecuteParams()
```
kevcunnane: Why are all these tests disabled? Can we re-enable?

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -0,0 +1,328 @@
...
+using Microsoft.SqlTools.ServiceLayer.Connection;
+using Microsoft.SqlTools.ServiceLayer.EditData.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Hosting;
+using Microsoft.SqlTools.ServiceLayer.Hosting.Protocol;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+using ConnectionType = Microsoft.SqlTools.ServiceLayer.Connection.ConnectionType;
+
+namespace Microsoft.SqlTools.ServiceLayer.EditData
```
kevcunnane: Please document the public class and methods

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -0,0 +1,328 @@
...
+{
+    public class EditDataService : IDisposable
+    {
+        #region Singleton Instance Implementation
+
+        private static readonly Lazy<EditDataService> LazyInstance = new Lazy<EditDataService>(() => new EditDataService());
```
kevcunnane: Minor: do you feel like trying out the new extensibility mechanism instead of this singleton?

kevcunnane:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
benrr101: It doesn't save a ton of code, but I've coded it up. The biggest savings was in unit tests

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
benrr101: ... I guess nothing. I thought I needed to implement a IDisposable for the Session, but there isn't any resources that need disposing in there.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
benrr101: No, it doesn't actually return a ReadonlyCollection, it returns the same List we had to begin with. However, unless you attempt to cast it to a List (which wouldn't be a safe guarantee), you'll effectively have a read-only collection. An IEnumerable is read-only by the nature that it can only return an enumerator and can't do concurrent modification.

As for the new syntax, I've stayed away from certain implementations of this that make it unclear what's going on (basically if the shortened form would be >1 line). But, I wasn't the first to start using this syntax -- @coquagli used it first! Either way, I think it's fine to use this as long as we don't let it get too confusing.

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
benrr101: It's easier to validate that the correct behavior occurred and to fake improper behavior in unit tests if I make these properties internal. Since they're meant to only be used outside the class in unit tests, I feel like they should stay as member variables, but I'll make them properties as per your request.

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
benrr101: I was thinking that the UI should handle this and not offer the option to delete a row if there are changes pending for it. Sort of along the lines of "are you sure you want to exit without saving changes?" But I can easily swap this for a AddOrUpdate call.

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
benrr101: I don't think so. The value comes directly from the JSON request. Conversion to/from enums with JSON is tricky and I don't know if it works correctly with our service host. We could still use an enum, but we'd ultimately be parsing it to an enum, then doing the same switch/case for accepting valid objects. In this case it'd add complexity without giving us any more type-security.

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
benrr101: I think I can count on one hand how many types of exceptions are being thrown. Either way, I'm not sure that using less exceptions is what we'd want to do anyhow. You'd have to give me a better example of what you mean.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Execution/RequestParamTests.cs
benrr101: Oh whoops, that was meant for another change I was planning to make

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Execution/ServiceIntegrationTests.cs
benrr101: I made a change to the methods and forgot to fix the unit tests for them. They'll be fixed in the next revision

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
benrr101: Not right now. I feel like it'd be a large refactor to get everything (unit tests especially) to behave properly. I'll add it to my list of things to do, tho.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
kevcunnane: Cool. I'm OK with these - ideally we'd have a style guide saying what's standard and run with it. This is fine overall though

---
In reply to: [101876610](https://github.com/Microsoft/sqltoolsservice/pull/241#discussion_r101876610) [](ancestors = 101876610)

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
kevcunnane: Thanks

---
In reply to: [102103578](https://github.com/Microsoft/sqltoolsservice/pull/241#discussion_r102103578) [](ancestors = 102103578)

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
kevcunnane: That makes sense and it's good enough for now

---
In reply to: [102097569](https://github.com/Microsoft/sqltoolsservice/pull/241#discussion_r102097569) [](ancestors = 102097569)

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
kevcunnane: Personally I'd prefer this as I think it makes UI's job easier. Can you add a task for that?

---
In reply to: [102099712](https://github.com/Microsoft/sqltoolsservice/pull/241#discussion_r102099712) [](ancestors = 102099712)

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
kevcunnane: It's a pattern we used in DacFx. Basically to avoid generic catch(Exception) methods, and ensure that callers downstream have a sane number of exceptions to expect for most things. Argument type exceptions usually mean something catastrophic / unexpected has happened. Anyhow.. not a dealbreaker but long term I'd rather we move to some extension-specific exceptions that make it easier in general to know what to catch. If we split things up into separate DLLs, this will become more important.

---
In reply to: [102101046](https://github.com/Microsoft/sqltoolsservice/pull/241#discussion_r102101046) [](ancestors = 102101046)

PR #242, opened 2017-02-16
Fix peek definition test

sharonravindran:
Fix GetUnsupportedDefinitionErrorTest that was intermittently failing on Jenkins.

The test was failing on the await for UpdateLanguageServiceOnConnection. I ran some builds on Jenkins for this branch and the builds pass.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/PeekDefinitionTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/PeekDefinitionTests.cs
@@ -165,6 +165,7 @@ GO";
             connectionResult.TextDocumentPosition = textDocument;
             var languageService = LanguageService.Instance;
             await languageService.UpdateLanguageServiceOnConnection(connectionResult.ConnectionInfo);
+            Thread.Sleep(2000);
             ScriptParseInfo parseInfo = languageService.GetScriptParseInfo(connectionResult.ScriptFile.ClientFilePath);
             Assert.NotNull(parseInfo);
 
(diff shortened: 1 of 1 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -165,6 +165,7 @@ public void GetTableDefinitionWithSchemaTest()
             connectionResult.TextDocumentPosition = textDocument;
             var languageService = LanguageService.Instance;
             await languageService.UpdateLanguageServiceOnConnection(connectionResult.ConnectionInfo);
+            Thread.Sleep(2000);
```
kevcunnane: Why is this required? What's causing the test to fail when you say "it's failing on the await"? A couple of things here:
- At a minimum, please comment on why it's needed in the source
- Thread.Sleep == slow tests. The best practice we should use is to listen for an event by waiting on a semaphore to be raised, but with a timeout to avoid freezing tests if the event is never called. I can't tell what you are actually waiting to be ready - what's not setup correctly? 

Here's how to raise a semaphore when something completes. It will speed up the test in the 99% case.
```

SemaphoreSlim semaphoreSlim = new SemaphoreSlim(0, 1);
serviceHostMock.Setup(/*... capture whatever event you need to listen to*/).Callback(c => semaphoreSlim.Release());

// Do the action that you need to listen for
await languageService.UpdateLanguageServiceOnConnection(connectionResult.ConnectionInfo);

// Wait on the lock for N seconds - mostly, it'll have succeeded by then. 
semaphoreSlim.Wait(TimeSpan.FromSeconds(2));

```

sharonravindran: The test is failing because of the shared LanguageService instance.  This is causing an exception in the Blocking Collection when it runs alongside a test that also uses the same instance. I'm refactoring the test to avoid this. Closing the PR until I'm done refactoring.

kevcunnane:
:shipit:

PR #245, opened 2017-02-22
Formatter and PeekDefinition Telemetry

kevcunnane:
- Add telemetry point for format requests
- Update PeekDefinition telemetry so that it captures whether the attempt succeeded and if it was connected. This will help us understand whether our current behavior of just notifying success/failure to the output window was useful or not. It will also help us understand whether we're doing a good job implementing this / if we need to improve reliability and preformance.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
@@ -73,6 +73,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Formatter
                 return FormatAndReturnEdits(docFormatParams);
             };
             await HandleRequest(requestHandler, requestContext, "HandleDocFormatRequest");
+
             DocumentStatusHelper.SendTelemetryEvent(requestContext, CreateTelemetryProps(isDocFormat: true));
         }
 
@@ -83,6 +84,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Formatter
                 return FormatRangeAndReturnEdits(docRangeFormatParams);
             };
             await HandleRequest(requestHandler, requestContext, "HandleDocRangeFormatRequest");
+
             DocumentStatusHelper.SendTelemetryEvent(requestContext, CreateTelemetryProps(isDocFormat: false));
         }
         private static TelemetryProperties CreateTelemetryProps(bool isDocFormat)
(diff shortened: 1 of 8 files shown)


src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -84,6 +83,19 @@ private WorkspaceService<SqlToolsSettings> WorkspaceService
                 return FormatRangeAndReturnEdits(docRangeFormatParams);
             };
             await HandleRequest(requestHandler, requestContext, "HandleDocRangeFormatRequest");
+            DocumentStatusHelper.SendTelemetryEvent(requestContext, CreateTelemetryProps(isDocFormat: false));
+        }
```
benrr101: very minor -- I'd like to see a extra line between these methods

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/TelemetryNotification.cs
```diff
@@ -55,5 +55,46 @@ public static class TelemetryEventNames
         /// telemetry even name for when definition is requested
         /// </summary>
         public const string PeekDefinitionRequested = "PeekDefinitionRequested";
-    } 
+
+        /// <summary>
+        /// telemetry even name for when definition is requested
```
benrr101: spelling: even**t**

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.SqlTools.ServiceLayer.Formatter.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
 using Microsoft.SqlTools.ServiceLayer.Hosting.Protocol;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices;
```
benrr101: I'm somewhat concerned that a "general purpose" telemetry event sending is being done in the LanguageService. It seems to me that this would cause a lot of tangling between services. If telemetry is something to offer up as a service for other services to consume, we should make it its own service. If there's precedent to have it in the LanguageService, we should decouple it from there and make it its own service.

kevcunnane: You're right - we should do this. However I'm not going to get that into the codebase in time for 0.3.0 release. Therefore I've created https://github.com/Microsoft/vscode-mssql/issues/712 to track this - we should tackle when we can.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/TelemetryNotification.cs
kevcunnane: Fixed

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
kevcunnane: Fixed

PR #248, opened 2017-02-22
Exposing the DisplayBitAsNumber setting

benrr101:


src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -22,7 +22,7 @@ public class QueryExecutionSettings
         /// <summary>
         /// Default number of chars to store for long text fields (de facto standard as per SSMS)
         /// </summary>
-        private const int DefaultMaxCharsToStore = 65535; // 64 KB - QE default
```
kevcunnane: I'd recomment keeping this the same as before (it's the max value for an unsigned short) and changing the test instead. This is late in the game and while odds are this won't break something, I'd prefer to be conservative

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/SettingsTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            Assert.Equal(2*1024*1024, sqlToolsSettings.QueryExecutionSettings.MaxXmlCharsToStore);
+            Assert.False(sqlToolsSettings.QueryExecutionSettings.ExecutionPlanOptions.IncludeActualExecutionPlanXml);
+            Assert.False(sqlToolsSettings.QueryExecutionSettings.ExecutionPlanOptions.IncludeEstimatedExecutionPlanXml);
+            Assert.True(sqlToolsSettings.QueryExecutionSettings.DisplayBitAsNumber);
+        }
+
+        [Fact]
+        public void ValidateSettingsParsedFromJson()
+        {
+            // NOTE: Only testing displayBitAsNumber for now because it is the only one piped through
```
kevcunnane: Any reason not to just do a single `@"...."` string instead of combining like this? It's harder to read the formatting this way

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/SettingsTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            // If: I parse the settings JSON object
+            JObject message = JObject.Parse(settingsJson);
+            JToken messageParams;
+            Assert.True(message.TryGetValue("params", out messageParams));
+            SqlToolsSettings sqlToolsSettings = messageParams.ToObject<SqlToolsSettings>();
+
+            // Then: The values defined in the JSON should propagate to the setting object
+            Assert.False(sqlToolsSettings.QueryExecutionSettings.DisplayBitAsNumber);
+        }
```
kevcunnane: I would be good to have a test verifying the settings is updated on ConfigChangeCallback events so we know it's the right version

PR #250, opened 2017-02-23
System.Composition not compatible with NetCore 1.0

kevcunnane:
- Reverting to a beta version of System.Composition that is compatible with NetCore 1.1. Otherwise we end up with a mishmash of versions which breaks Mac OpenSSL support


kevcunnane:
Merging immediately as this is a ship blocker

PR #251, opened 2017-02-23
Fix/update net core sdk

benrr101:
Updating the .NET Core SDK version in our global.json file. The version selected is the latest version available on https://www.microsoft.com/net/download/core. This should keep us from having to have manual updates to global.json for a while.

I'll add notes on how to fix problems you'll encounter after the upgrade once this is checked in.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "projects": [ "src", "test" ],
   "sdk": {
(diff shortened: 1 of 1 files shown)


benrr101:
No particular reason. I got tired of always having to update my global.json since the version I had was a lot newer than the one we had.

PR #252, opened 2017-02-23
Isolate Shared Test Code

benrr101:
This is partly an experimental pull request to make sure that automated testing is still working correctly.

The goal of this make sure that test code is correctly organized to ensure that test suites aren't dependent on each other.
* UnitTests get their own project now (renaming Microsoft.SqlTools.ServiceLayer.Test to Microsoft.SqlTools.ServiceLayer.UnitTests) which is about 90% of the changes to the files.
* IntegrationTests no longer depends on UnitTests, only Test.Common
* Any shared components from TestObjects that spins up a "live" connection has been moved to IntegrationTests Utility/LiveConnectionHelper.cs
* The dictionary-based mock file stream factory has been moved to Test.Common since it is used by UnitTests and IntegrationTests
    * Added a overload that doesn't take a dictionary for when we don't care about monitoring the storage (about 90% of the time)
* The RunIf* wrapper methods have been moved to Test.Common
* OwnerUri and StandardQuery constants have been moved to Test.Common Constants file


kburtram:
Looks like there is an AppVeyor break with this PR.

benrr101:
@kburtram looking into the logs for this, AppVeyor is trying to run `dotnet test test/Microsoft.SqlTools.ServiceLayer.Test` which no longer exists. My understanding was that if I change the appveyor.yml file to point to the correct test project to run, it would be fine. But even though I've updated that file, it still wants to run the non-existent tests. Is there something else I need to change for appveyor to update its file?

abist:
@benrr101 Go to AppVeyor -> Settings -> Tests, and then you can change the command there

kburtram:
Please also make sure this check-in doesn't break the Jenkins code coverage runs.

PR #254, opened 2017-02-24
Handle open\close document events with git URI prefix

kburtram:


src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
```diff
@@ -124,9 +141,10 @@ internal static bool IsPathInMemory(string filePath)
             // with 'inmemory'.  Untitled files which have been marked of
             // type SqlTools have a path starting with 'untitled'.
             return
-                filePath.StartsWith("inmemory") ||
-                filePath.StartsWith("tsqloutput") ||
-                filePath.StartsWith("untitled");
+                filePath.StartsWith("inmemory:") ||
+                filePath.StartsWith("tsqloutput:") ||
+                filePath.StartsWith("git:") ||
```
kburtram: add GIT into this list so that ContainsFile won't through when trying to resolve the URI

kevcunnane:
:shipit:

PR #262, opened 2017-03-01
edit/commit Command

benrr101:
The main goal of this feature is to enable a command that will
1) Generate a parameterized command for each edit that is in the session
2) Execute that command against the server
3) Update the cached results of the table/view that's being edited with the committed changes (including computed/identity columns)

There's some secret sauce in here where I cheated around worrying about gaps in the updated results. This was accomplished by implementing an IComparable for row edit objects that ensures deletes are the *last* actions to occur and that they occur from the bottom of the list up (highest row ID to lowest). Thus, all other actions that are dependent on the row ID are performed first, then the largest row ID is deleted, then next largest, etc. Nevertheless, by the end of a commit the associated ResultSet is still the source of truth. It is expected that the results grid will need updating once changes are committed.

Also worth noting, although this pull request supports a "many edits, one commit" approach, it will work just fine for a "one edit, one commit" approach.

***Note:*** There are still a few todo's that I want to clean up. Some of them are larger functionality (such as transactional support, etc) that will wait for a later PR. Some of them are tiny (move string to constants), that I'll do a quick sweep of before committing.


src/Microsoft.SqlTools.Hosting/Utility/LongList.cs
```diff
@@ -138,6 +143,34 @@ public T GetItem(long index)
...
+        {
+            if (Count <= this.ExpandListSize)
+            {
+                int i32Index = Convert.ToInt32(index);
+                shortList[i32Index] = value;
+            }
+            else
+            {
+                int iArray32Index = (int) (Count / this.ExpandListSize);
+                if (expandedList.Count > iArray32Index)
```
coquagli: Should you handle the else case? Maybe throw? Could this ever happen?

benrr101: Hmm... This was a slight modification of the code I brought over from the SSMS code. I mostly just trusted they knew what they were doing, but now that you mention it, this looks a little odd. I'll give it a once over.

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
```diff
@@ -21,13 +23,9 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
     public class Session
```
coquagli: I know this wasn't changed in this PR, but EditDataSession might be a better name for this class. It's in the EditData namespace, but still.

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
```diff
@@ -110,6 +114,29 @@ public long CreateRow()
...
+        /// <param name="connection">The connection to use for executing the query</param>
+        /// <param name="successHandler">Callback to perform when the commit process has finished</param>
+        /// <param name="errorHandler">Callback to perform if the commit process has failed at some point</param>
+        public void CommitEdits(DbConnection connection, Func<Task> successHandler, Func<Exception, Task> errorHandler)
+        {
+            Validate.IsNotNull(nameof(connection), connection);
+            Validate.IsNotNull(nameof(successHandler), successHandler);
+            Validate.IsNotNull(nameof(errorHandler), errorHandler);
+
+            // Make sure that there a commit task isn't in progress
```
coquagli: // Make sure that there isn't a commit task in progress

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
```diff
@@ -110,6 +114,29 @@ public long CreateRow()
...
+            if (CommitTask != null && !CommitTask.IsCompleted)
```
coquagli: Should this have a lock?

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
```diff
@@ -110,6 +114,29 @@ public long CreateRow()
         }
 
         /// <summary>
+        /// Commits the edits in the cache to the database and then to the associated result set of
+        /// this edit session. This is launched asynchronously.
+        /// </summary>
+        /// <param name="connection">The connection to use for executing the query</param>
+        /// <param name="successHandler">Callback to perform when the commit process has finished</param>
+        /// <param name="errorHandler">Callback to perform if the commit process has failed at some point</param>
+        public void CommitEdits(DbConnection connection, Func<Task> successHandler, Func<Exception, Task> errorHandler)
```
coquagli: I don't find the separation of CommitEdits and CommitEditsInternal particularly useful because they are both pretty short. Unless there is a test case that has a good reason to bypass the checks, I personally would merge the. More of a personal preference, so it's up to you.

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
benrr101: Much English, very grammar, wow.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -89,7 +89,7 @@ public CellUpdate(DbColumn column, string valueAsString)
         /// <summary>
         /// The column that the cell will be placed in
         /// </summary>
-        public DbColumn Column { get; }
+        public DbColumnWrapper Column { get; }
```
coquagli: I'd be in favor of moving ` #region Properties` above all functions.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -17,7 +22,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
     /// </summary>
     public sealed class RowCreate : RowEditBase
     {
-        private const string InsertStatement = "INSERT INTO {0}({1}) VALUES ({2})";
+        private const string InsertStart = "INSERT INTO {0}({1})";
 
         private readonly CellUpdate[] newCells;
```
coquagli: _newCells ?

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -34,42 +39,121 @@ public RowCreate(long rowId, ResultSet associatedResultSet, IEditTableMetadata a
...
+                command.Parameters.Add(param);
+            }
+            string joinedInColumns = string.Join(", ", inColumns);
+            string joinedOutColumns = string.Join(", ", outColumns);
+            
+            // Get the start clause
+            string start = GetTableClause();
+
+            // Put the whole #! together
+            command.CommandText = $"{start} OUTPUT {joinedOutColumns} VALUES ({joinedInColumns})";
```
coquagli: Should OUTPUT and VALUES be const variables?

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEdit.cs
```diff
@@ -58,8 +60,31 @@ protected RowEditBase(long rowId, ResultSet associatedResultSet, IEditTableMetad
         /// </summary>
         public IEditTableMetadata AssociatedObjectMetadata { get; }
 
+        /// <summary>
+        /// Sort ID for a row edit. Ensures that when a collection of RowEditBase objects are
+        /// sorted, the appropriate types are sorted to the top.
+        /// </summary>
+        protected abstract int SortId { get; }
```
coquagli: Since the rest of this file is so organized, you might as well stick this in Abstract Methods and rename it to something like "Abstract" or "Abstract Entities".

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
```diff
@@ -39,13 +39,79 @@ public RowUpdate(long rowId, ResultSet associatedResultSet, IEditTableMetadata a
         }
 
         /// <summary>
+        /// Sort order property. Sorts to same position as RowCreate
+        /// </summary>
+        protected override int SortId => 1;
```
coquagli: Is there a reason RowCreate and RowUpdate are both 1? Would there be any motivation to always do one before the other?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsWriterBase.cs
```diff
@@ -76,6 +76,15 @@ public int WriteRow(StorageDataReader dataReader)
         public abstract void WriteRow(IList<DbCellValue> row, IList<DbColumnWrapper> columns);
 
         /// <summary>
+        /// Not implemented, do not use.
+        /// </summary>
+        [Obsolete]
+        public void Seek(long offset)
+        {
+            throw new InvalidOperationException("SaveAs writers are meant to be written once contiguously.");
```
coquagli: NotImplementedException ?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -70,25 +65,31 @@ public class ResultSet : IDisposable
...
+        /// <summary>
         /// The special action which applied to this result set
         /// </summary>
         private readonly SpecialAction specialAction;
 
+        /// <summary>
+        /// Total number of bytes written to the file. Used to jump to end of the file for append
+        /// scenarios. Internal for unit test validation.
+        /// </summary>
+        internal long totalBytesWritten;
```
coquagli: I think this gets set to 0 automatically, but I would be in favor of explicitly setting it to 0.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -354,6 +370,50 @@ public Task<ExecutionPlan> GetExecutionPlan()
...
         /// <summary>
+        /// Removes a row from the result set cache
+        /// </summary>
+        /// <param name="internalId">Internal ID of the row</param>
+        public void RemoveRow(long internalId)
+        {
+            // Make sure that the results have been read
+            if (!hasBeenRead)
+            {
+                throw new InvalidOperationException();
```
coquagli: Should there be a string comment provided here?

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Execution/RequestParamTests.cs
```diff
@@ -0,0 +1,78 @@
+﻿//
```
coquagli: Dead code?

test/Microsoft.SqlTools.ServiceLayer.Test/EditData/RowEditBaseTests.cs
```diff
@@ -178,6 +270,18 @@ public override EditUpdateCellResult SetCell(int columnId, string newValue)
             {
                 throw new NotImplementedException();
             }
+
+            public override Task ApplyChanges(DbDataReader reader)
```
coquagli: Are these functions used?

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
benrr101: Yeah, I think I could do it as you suggested. The downsides to it are mostly that I'd end up having code nested inside the construction of a Task. While not terrible, I also followed this pattern for query execution, so I think I'll just stay with it as-is.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
benrr101: So am I 😄 
But Kevin said a while back that the convention here is to do member variables -> constructor -> properties -> public methods -> private methods.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
benrr101: I didn't think we ever followed that standard?

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
benrr101: The format string was originally a const, but I rewrote it once I started using the OUTPUT clauses. Since we're only using the format string once, I can't really justify making it a const. But, then again, I'm not sure what the guidance is for using or not using a const.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEdit.cs
benrr101: Eh... I think organizing by what the "thing" is is better -- so since it's a property, it should go in the properties section.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
benrr101: I really only needed deletes to go last. I haven't yet found a reason to do row create or row update first, so I just set them to the same values. We can revisit if we find it to be problematic.

test/Microsoft.SqlTools.ServiceLayer.Test/EditData/RowEditBaseTests.cs
benrr101: No, but they need to be there since those are the abstract methods an implementation of RowEditBase expects.

coquagli: Oops. I missed line 203.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsWriterBase.cs
benrr101: I would go the NotImplemented route, but that is considered by ReSharper to be in the same boat as a "todo" comment.

I like your comment about the interfaces, but I'm not exactly sure how I'd do it. The ResultSet class would need to get a writer that implements both the Seek method and the write row methods.

test/Microsoft.SqlTools.ServiceLayer.Test/QueryExecution/Execution/RequestParamTests.cs
benrr101: Ah dang! I keep accidentally adding that file.

src/Microsoft.SqlTools.ServiceLayer/EditData/Session.cs
coquagli: After a second look, I agree that this is fine.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
coquagli: I've definitely been asked to add _ in a PR before, but I guess if we don't follow that standard it's not an issue.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
coquagli: `InsertStart` is also only used once (but it does have the extra `{0}({1}`). I feel that values like this are traditionally const, regardless of the number of times they are used. 

It's also literally a string, so either way it's not a huge issue.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEdit.cs
coquagli: Oops. For some reason I thought it was just dangling in no `region`.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
coquagli: Sounds good. Just making sure.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
benrr101: Good point. I'll move them to consts anyways.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -203,13 +235,51 @@ public EditUpdateCellResult UpdateCell(long rowId, int columnId, string newValue
...
+            // doesn't already exist in the cache
+            RowEditBase editRow = EditCache.GetOrAdd(rowId, key => new RowUpdate(rowId, associatedResultSet, objectMetadata));
 
             // Pass the call to the row update
             return editRow.SetCell(columnId, newValue);
         }
 
         #endregion
 
+        #region Private Helpers
```
abist: In a previous CR earlier, Kevin suggested not putting private regions if it's just one function. It made looking up private functions harder than usual.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/RowDeleteTests.cs
```diff
@@ -5,9 +5,12 @@
```
abist: need the MIT license

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -313,14 +327,20 @@ public Task<ExecutionPlan> GetExecutionPlan()
...
         /// </summary>
+        /// <param name="dbDataReader">The data reader for getting results from the db</param>
         /// <param name="cancellationToken">Cancellation token for cancelling the query</param>
-        public async Task ReadResultToEnd(CancellationToken cancellationToken)
+        public async Task ReadResultToEnd(DbDataReader dbDataReader, CancellationToken cancellationToken)
         {
+            // Sanity check to make sure we got a reader
+            Validate.IsNotNull(nameof(dbDataReader), SR.QueryServiceResultSetReaderNull);
+
             try
```
abist: We don't want to deal with caught exceptions (if there are any) here?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -526,14 +587,45 @@ private SpecialAction ProcessSpecialAction()
...
+        {
+            Validate.IsNotNull(nameof(dbDataReader), dbDataReader);
+            if (!hasBeenRead)
+            {
+                throw new InvalidOperationException(SR.QueryServiceResultSetNotRead);
+            }
+            if (!dbDataReader.HasRows)
+            {
+                // @TODO Move to constants file
+                throw new InvalidOperationException("Cannot add row to result buffer, data reader does not contain rows");
```
abist: I think we missed this comment here.

benrr101: Good catch

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
benrr101: No, if this exception is thrown, it will bubble up and we'll handle it there

benrr101: Although there was a nasty bug here where I was checking if the error string was null...

PR #263, opened 2017-03-02
Fix namespaces for code moved to separate class libraries

kburtram:


kburtram:
In addition to changing namespaces this PR also moves ServiceHost from the shared Hosting library into both credentials and servicelayer libraries since this class should be customized per executable service.

This is a refactoring only change without any "functional" change.

benrr101:
I was concerned for a second that you were giving up on having the Hosting library be shared. But I see that this is just duplicating the ServiceHost class. I agree, I think that class was built to have some level of customizability.
👍 as long as the CI builds are working!

PR #265, opened 2017-03-04
Block Initializing Edit Sessions When In Progress

benrr101:
One issue I ran into was that during unit testing, it's not possible to completely wait until the edit session has been created. This is due to there not being a handle to wait for initialization to finish. This manifests itself as a race condition where it's possible to initialize a session while the same initialization is occurring. To fix this, I introduced a owner URI to TaskCompletionSource map. TaskCompletionSource is created when starting initialization (atomically!) and disposed when the initialization finishes (successful or otherwise). The TaskCompletionSource can be awaited in the unit tests to allow us to properly test the initialize request. We can also use the same mechanism to block execution of initialization while another one is in progress for the same owner uri


kburtram:
>                     Query = $"SELECT * FROM {SqlScriptFormatter.FormatMultipartIdentifier(initParams.ObjectName)}",

typically it's best practice not to SELECT * in an application but to SELECT the specific columns you need..this make the app more resilient to future changes. 

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs:214 in 2a96980. [](commit_id = 2a96980311c6100285680d29bb7d5f17046a61fd, deletion_comment = False)

benrr101:
@kburtram I definitely agree with your comment that we shouldn't use `SELECT *` when running the query for the edit table. This is definitely a larger change that I'd like to take on in this code review. I've created https://github.com/Microsoft/carbon/issues/203 to track this.

PR #266, opened 2017-03-04
edit/createRow Default Values

benrr101:
Returns strings for the default value of a column when a new row is created. The values that could come back:
* `null` when there isn't a default for the column
* a string when there is a default for the column
* a placeholder (currently <TBD>) when the column cannot be updated


src/Microsoft.SqlTools.ServiceLayer/EditData/EditColumnWrapper.cs
```diff
@@ -19,11 +19,23 @@ public class EditColumnWrapper
         public DbColumnWrapper DbColumn { get; set; }
 
         /// <summary>
+        /// If set, this is a string representation of the default value. If set to null, then the
```
kburtram: >If set to null [](start = 74, length = 14)

I suppose this logic works fine when the default value is null?

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -200,6 +200,8 @@ EditDataScriptFilePathNull = An output filename must be provided
 
 EditDataCommitInProgress = A commit task is in progress. Please wait for completion.
 
+EditDataComputedColumnPlaceholder = <TBD>
```
kburtram: ><TBD> [](start = 36, length = 5)

We should probably find a better placeholder than this IMO.  Is this what SSDT or SSMS uses?

benrr101: SSMS - Doesn't show defaults. Just fills all fields in with "NULL".
SSDT - Doesn't show defaults. Just fills all fields in with "NULL".
JetBrains - Uses "<Generated>" as the place holder. Fields with defaults are shown as "<Default>"

benrr101: I'm open for suggestions for what to use here. Though I think that using "NULL" isn't the best idea.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditColumnWrapper.cs
benrr101: The logic basically generates a list of display values for the new row in the UI.
* No default is returned as `null`
* `NULL` default is returned as "NULL" (as per how SMO returns the value)
* Literal defaults are unwrapped from what SMO gives us and returned (eg, `((8))` => 8, `N'stuff'` => stuff)
* Computed columns are returned as a placeholder

PR #267, opened 2017-03-06
Provide connection options to host process

kburtram:


kburtram:
Ping?  Could someone take a look at this PR?

src/Microsoft.SqlTools.ServiceLayer/ServiceHost.cs
```diff
@@ -171,6 +172,140 @@ public void RegisterInitializeTask(InitializeCallback callback)
...
+        internal async Task HandleCapabilitiesRequest(
+            CapabilitiesRequest initializeParams, 
+            RequestContext<CapabilitiesResult> requestContext)            
+        {
+            await requestContext.SendResult(
+                new CapabilitiesResult
+                {
+                    Capabilities = new DmpServerCapabilities
+                    {
+                        ProtocolVersion = "1.0",
```
coquagli: I would personally be in favor of making these `readonly`/`const` at the top of the file, but it's obviously not 100% necessary.

src/Microsoft.SqlTools.Hosting/Hosting/Contracts/ConnectionProviderOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+        public static readonly string SpecialValueAuthType = "authType";
+        public static readonly string SpecialValueUserName = "userName";
+        public static readonly string SpecialValuePasswordName = "password";
+
+        public string Name { get; set; }
+
+        public string DisplayName { get; set; }
+
+        /// <summary>
+        /// Type of the parameter.  Can be either string, number, or category.
```
coquagli: Should the setter enforce the allowed values?

kburtram: Good idea.  I'll add that logic in follow-up change.

PR #268, opened 2017-03-06
edit/revertCell

benrr101:
Adding a new request type to the edit data API: "edit/revertCell" which reverts a single cell in a row. For update and create edits, it removes the cell from the list of pending updates and returns the original value (or null). If there isn't a pending edit to remove, the original value (or null) is returned anyways.


src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -79,6 +79,7 @@ public void InitializeService(ServiceHost serviceHost)
             serviceHost.SetRequestHandler(EditDeleteRowRequest.Type, HandleDeleteRowRequest);
             serviceHost.SetRequestHandler(EditDisposeRequest.Type, HandleDisposeRequest);
             serviceHost.SetRequestHandler(EditInitializeRequest.Type, HandleInitializeRequest);
+            serviceHost.SetRequestHandler(EditRevertCellRequest.Type, HandleRevertCellRequest);
```
kburtram: do we have a use case for this message type?  I was suspecting that the grid would do reverts on the row level instead of the cell level.

benrr101: I'm thinking this is the behavior for if you hit <kbd>esc</kbd> when you're typing a value into a cell. If you revert the entire row, this could be annoying if you need to switch to another tab but are in the middle of changing a row. But this is something we can discuss.

PR #274, opened 2017-03-09
NULL column validation when updating cells

benrr101:
Validates whether or not a cell can be set to `NULL` based on whether or not the column allows null values. If the column does not allow `NULL` and the user tried to set it to `NULL` it will throw an exception.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
@@ -199,7 +199,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
         private void ProcessNullValue()
         {
             // Make sure that nulls are allowed if we set it to null
-            if (Column.AllowDBNull.HasTrue())
+            if (!Column.AllowDBNull.HasTrue())
             {
                 throw new InvalidOperationException(SR.EditDataNullNotAllowed);
             }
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -197,6 +196,18 @@ private void ProcessBooleanCell(string valueAsString)
             ValueAsString = Value.ToString();
         }
 
+        private void ProcessNullValue()
+        {
+            // Make sure that nulls are allowed if we set it to null
+            if (!Column.AllowDBNull.HasTrue())
+            {
+                throw new InvalidOperationException(SR.EditDataNullNotAllowed);
```
coquagli: Where does this get caught?

benrr101: This would be caught in the edit/updateCell handler and the exception message would be returned to the user via an error.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -197,6 +196,18 @@ private void ProcessBooleanCell(string valueAsString)
...
+            }
+
+            Value = DBNull.Value;
```
coquagli: I assume the -0.02% drop is from these 2 lines not getting hit.

PR #275, opened 2017-03-09
Adding validation of timespan columns w/in 24hrs

benrr101:
Very small feature add to make sure that values entered for `TIME` columns are < 24hrs. If the value is >=24hrs when setting the column, an exception will be thrown that the edit/updateCell handler will catch and convert into an error on the JSON RPC side.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
@@ -196,7 +196,6 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
             ValueAsString = Value.ToString();
         }
 
-<<<<<<< HEAD
         private void ProcessTimespanColumn(string valueAsString)
         {
             TimeSpan ts = TimeSpan.Parse(valueAsString, CultureInfo.CurrentCulture);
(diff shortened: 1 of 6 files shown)


test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/CellUpdateTests.cs
```diff
@@ -159,6 +159,17 @@ public void BoolInvalidFormatTest()
         }
 
         [Theory]
+        [InlineData("24:00:00")]
+        [InlineData("105:00:00")]
+        public void TimeSpanTooLargeTest(string value)
```
coquagli: If you're feeling really ambitious you could make a test that verifies that timespans within the allowed range don't throw.

benrr101: That's already there as part of the "RoundTripTest" right below this test 😃

coquagli: Oops.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -196,6 +196,7 @@ private void ProcessBooleanCell(string valueAsString)
             ValueAsString = Value.ToString();
         }
 
+<<<<<<< HEAD
```
coquagli: Heads up for this.

PR #276, opened 2017-03-10
removed mono dependency

abist:
Possible fix for failing builds on Travis for OSX platform.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
   - linux
   - osx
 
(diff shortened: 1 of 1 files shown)


abist:
This change brought down the time on Travis CI to run tests on Travis from 19 min 30 sec to 10 min 20 sec.

benrr101:
We shaved 3min from the mac build times! Whoo hoo!
:shipit:

PR #278, opened 2017-03-13
Changing CSV encoding to UTF16LE w/o BOM

benrr101:
Small tweak to use Unicode (UTF-16LE w/o BOM) encoding for writing CSV files. My research suggests that this format will open in Excel 2016 on both OSX and Windows. This also allows for encoding the full range of UTF-16 characters that N* columns support.


benrr101:
I'm going to back out this pull request for now. It turns out UTF-16LE w/o BOM isn't supported by VS Code.

PR #279, opened 2017-03-14
Save as excel

wjzhou:
This adds the save as excel function to the backend.

To test this code before frontend change in vscode-mssql repository, you may hijack the SaveAsCsv request by change the file [QueryExecution/Contracts/SaveResultsRequest.cs](https://github.com/wjzhou/sqltoolsservice/blob/feature/SaveAsExcel/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/SaveResultsRequest.cs#L131-L141) to

``` csharp
    public class SaveResultsAsCsvRequest
    {
        public static readonly
            RequestType<SaveResultsAsCsvRequestParams, SaveResultRequestResult> Type =
            RequestType<SaveResultsAsCsvRequestParams, SaveResultRequestResult>.Create("query/saveCsv2");
    }

    /// <summary>
    /// Request type to save results as Excel
    /// </summary>
    public class SaveResultsAsExcelRequest
    {
        public static readonly
            RequestType<SaveResultsAsExcelRequestParams, SaveResultRequestResult> Type =
            RequestType<SaveResultsAsExcelRequestParams, SaveResultRequestResult>.Create("query/saveCsv");
    }
```

In order to be usable in dotnet core now and avoid large external dependency, this implements a basic Excel exporter based on ECMA-376 in [SaveAsExcelFileStreamWriterHelper](https://github.com/wjzhou/sqltoolsservice/blob/feature/SaveAsExcel/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs.)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsJsonFileStreamWriter.cs
```diff
@@ -78,16 +78,24 @@ public override void WriteRow(IList<DbCellValue> row, IList<DbColumnWrapper> col
             jsonWriter.WriteEndObject();
         }
 
+        private bool disposed = false;
         /// <summary>
         /// Disposes the writer by closing up the array that contains the row objects
         /// </summary>
-        public new void Dispose()
+        protected override void Dispose(bool disposing)
```
benrr101: Thanks for improving this dispose implementation!

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriter.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            sheet.AddRow();
+            for (int i = startIndex; i < endIndex; ++i)
+            {
+                object o = row[i].RawObject;
+                if (o != null) {
+                    TypeCode type = Type.GetTypeCode(o.GetType());
+                    switch (type)
+                    {
+                        case TypeCode.Byte:
```
benrr101: Technically, we'll never get `SByte`, `UInt16`, `UInt32`, `UInt64` returned from the SqlClient, so I think we can omit these cases.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriter.cs
```diff
@@ -0,0 +1,114 @@
...
+                        case TypeCode.SByte:
+                        case TypeCode.Int16:
+                        case TypeCode.UInt16:
+                        case TypeCode.Int32:
+                        case TypeCode.UInt32:
+                        case TypeCode.Int64:
+                        case TypeCode.UInt64:
+                            sheet.AddCell(o);
+                            continue;
+                        case TypeCode.DateTime:
```
benrr101: Will `TimeSpan` need their own special case here? We'll get a TimeSpan object back for `TIME` columns.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriter.cs
```diff
@@ -0,0 +1,114 @@
...
+                    sheet.AddCell(columns[i].ColumnName);
+                }
+                headerWritten = true;
+            }
+
+            sheet.AddRow();
+            for (int i = startIndex; i < endIndex; ++i)
+            {
+                object o = row[i].RawObject;
+                if (o != null) {
```
benrr101: Rather than nest all this code inside a if not null, can we invert this? Have a check for null that outputs the display value then continues. Then have the switch that has a default case what just outputs the display value. The other cases could have breaks rather than continue statements.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+ */
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Xml;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
+{
+    public sealed class SaveAsExcelFileStreamWriterHelper : IDisposable
```
benrr101: Please provide class-level block comments for public classes. I think in this case we can move the big block comment at the top of the file down here.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+                }
+                AddCellDateTimeInternal(excelDate, style);
+            }
+
+            /// <summary>
+            /// Write a object cell
+            /// </summary>
+            /// The program will try to output number/datetime, otherwise, call the ToString 
+            /// <param name="o"></param>
+            public void AddCell(object o)
```
benrr101: It seems this code is duplicated in the SaveAsExcelFileStreamWriter. Can we eliminate on or the other?

Looking at the code I like the implementation of it here better. Maybe we can tweak this to accept a DbCellValue?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+            {
+                if (o == null)
+                {
+                    AddCellEmpty();
+                    return;
+                }
+                switch (Type.GetTypeCode(o.GetType()))
```
benrr101: Same notes regarding the types we don't need to process from above apply here.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+                _writer.WriteValue(number);
+                _writer.WriteEndElement();
+
+                _writer.WriteEndElement();
+            }
+
+            // The excel epoch is 1/1/1900, but it has 1/0/1900 and 2/29/1900
+            // which is equal to set the epoch back two days to 12/30/1899
+            // new DateTime(1899,12,30).Ticks
+            private const long ExcelEpochTick = 599264352000000000L;
```
benrr101: Our coding standard defines files be laid out like this:
1. constants
2. member variables
3. constructors
4. properties *(don't ask me why 😛)*
5. public methods
6. private methods

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+            private char[] _currReference = new char[3 + 7]; //maximal XFD1048576
+            private int _currReferenceRowLength;
+            private int _currReferenceColumnLength;
+            private XmlWriter _writer;
+
+            public void AssureColumnReference()
+            {
+                if (_currColumn == 0)
+                {
+                    throw new ExporterException("AddRow must be called before AddCell");
```
benrr101: We'll need to move these exception messages to our string constants files for localization, but I won't foist that hassle on you.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+
+                }
+                if (_currColumn > 16384)
+                {
+                    throw new ExporterException("max column number is 16384, see https://support.office.com/en-us/article/Excel-specifications-and-limits-1672b34d-7043-467e-8e27-269d656771c3");
```
benrr101: Can we get away with using InvalidOperationException in these cases rather than define a new exception type?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+                }
+            }
+
+            public void WriteAndIncreaseColumnReference()
```
benrr101: Please add block comments for any public methods

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+                _writer.WriteStartAttribute("r");
+                _writer.WriteValue(_currRow);
+                _writer.WriteEndAttribute();
+
+                ResetColumnReference(); //This need to be called before the increase
+
+                _currRow++;
+            }
+
+            public bool PenddingRowEndTag()
```
benrr101: Personally, I'd move this to a property.
Also, there's a typo: `PendingRowEndTag`

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelperTests/[Content_Types].xml
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Comment (benrr101) [inline test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelperTests/[Content_Types].xml]: I'm not entirely certain, but do these files need to be committed? Or are they leftovers from running the tests?

If the latter, you can utilize the MemoryFileSystem in Tests.Common to store any generated files in memory rather than storing them on the disk and cleaning up afterwards.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamFactory.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        #region Properties
+
+        /// <summary>
+        /// Settings for query execution
+        /// </summary>
+        public QueryExecutionSettings QueryExecutionSettings { get; set; }
+
+        /// <summary>
+        /// Parameters for the save as CSV request
```
benrr101: Might want to replace instances of CSV with Excel in this file

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+        public sealed class ExcelSheet : IDisposable
+        {
+            /// <summary>
+            /// Start a new row
+            /// </summary>
+            public void AddRow()
+            {
+                EndRowIfNeeded();
+
+                _referenceManager.AssureRowReference();
```
benrr101: I know it's a pain, but our coding style is to have private member variables be `camelCased` without the `_` prefix.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelperTests.cs
```diff
@@ -0,0 +1,220 @@
...
+                sheet.AddRow();
+                sheet.AddCell(new DateTime(1900, 2, 28));
+                sheet.AddCell(new DateTime(1900, 3, 1));
+                sheet.AddCell(new DateTime(1, 1, 1, 15, 00, 00));
+            }
+        }
+        Regex contentRemoveLinebreakLeadingSpace = new Regex(@"\r?\n\s*");
+        private void ContentMatch(string fileName)
+        {
+            string referencePath = Path.Combine("QueryExecution",
```
kevcunnane: This is failing in [Travis builds](https://travis-ci.org/Microsoft/sqltoolsservice/jobs/211004702) because of a file not found error. This happens for file lookups since the path isn't the same as when run locally. The best solution is follow the pattern used by other tests with baseline files - `FormatterUnitTestBase.TestLocationDirectory` is a good start. Steps to follow:
- Move your `DataStorage` folder with test files under `Microsoft.SqlTools.ServiceLayer.Test.Common\TestData`
- Get the path to the DataStorage folder like so:
```

        private static string TestLocationDirectory
        {
            get
            {
                return Path.Combine(RunEnvironmentInfo.GetTestDataLocation(), "DataStorage");
            }
        }
```
- Change your code to 
```
string referencePath = Path.Combine(TestLocationDirectory, "SaveAsExcelFileStreamWriterHelperTests", fileName);
```

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+                _writer.WriteEndElement();
+
+                _writer.WriteEndElement();
+            }
+            /// <summary>
+            /// Write a DateTime cell.
+            /// </summary>
+            /// <param name="dateTime">Datetime</param>
+            /// <remark>
+            /// If the DateTime do not have date part, it will be written as datetime and show as time only
```
kevcunnane: Minor: `do not have` -> `does not have`

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
kevcunnane: Please move the private fields to the top of the file. Our coding convention is to put all private fields there.

benrr101:
One other quick note, please take a look at https://travis-ci.org/Microsoft/sqltoolsservice/jobs/211004702 and make sure that the unit tests pass before we can check them in. If you need assistance working out our CI system (which can be annoying at times), I can help you out.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+        public void Dispose()
+        {
+            WriteMinimalTemplate();
+            zipArchive.Dispose();
+        }
+
+
+        internal class ReferenceManager
+        {
+            public ReferenceManager(XmlWriter writer)
```
kevcunnane: Please move constructor below the private fields (another coding convention).

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,583 @@
...
+                string rowReference = _currRow.ToString();
+                _currReferenceRowLength = rowReference.Length;
+                rowReference.CopyTo(0, _currReference, 3, rowReference.Length);
+            }
+
+            public void AssureRowReference()
+            {
+                if (_currRow > 1048576)
+                {
+                    throw new ExporterException("max row number is 1048576, see https://support.office.com/en-us/article/Excel-specifications-and-limits-1672b34d-7043-467e-8e27-269d656771c3");
```
kevcunnane: We should put this string in SR.strings so that it can be localized. @benrr101 can you help with this after checkin? @wjzhou you can also do this but it does require Mono to be installed on your machine. In that case, add a new entry to SR.strings, save & run `.\build.cmd` which should add this to the SR.designer.cs file so you can reference it in your code.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
kevcunnane: typo: `PenddingRowEndTag` -> `PendingRowEndTag`

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelperTests.cs
```diff
@@ -0,0 +1,220 @@
...
+        {
+            var xmlWriter = _xmlWriterMock.Object;
+            var manager = new SaveAsExcelFileStreamWriterHelper.ReferenceManager(xmlWriter);
+
+            var ex = Assert.Throws<SaveAsExcelFileStreamWriterHelper.ExporterException>(
+                () => manager.WriteAndIncreaseColumnReference());
+            Assert.Contains("AddRow must be called before AddCell", ex.Message);
+        }
+
+        //This test is too slow
```
kevcunnane: I would suggest removing the commented out code - it sounds like you don't intend to add back

wjzhou:
Thank you for the fast reviewing. I'm working on the changes now.

The failed tests are because of I access a folder in the test folder
``` csharp
string referencePath = Path.Combine("QueryExecution",
                "DataStorage",
                "SaveAsExcelFileStreamWriterHelperTests",
                fileName);
```

These tests work on my computer but not on the CI server. (Probably because of the relative path)

Background on that folder: I compare the generated xlsx(zip) with a known good folder. This helped when I changed the implementation from string writer to xmlwriter. 
I keep the reference files in a folder, such that, I may compress the folder as zip and confirm these reference files are still good. 
 
But it seems the CI server doesn't like my way of  accessing the reference folder. 

What is your recommendation? Do I have to move these to c# file?

kevcunnane:
I added a comment on how to fix this. We have a number of tests that also red from a "known good" file so you can reuse the same pattern. here's the exact steps:

This is failing in Travis builds because of a file not found error. This happens for file lookups since the path isn't the same as when run locally. The best solution is follow the pattern used by other tests with baseline files - `FormatterUnitTestBase.TestLocationDirectory` is a good start. Steps to follow:

- Move your DataStorage folder with test files under `Microsoft.SqlTools.ServiceLayer.Test.Common\TestData`
- Get the path to the DataStorage folder like so:
```
        private static string TestLocationDirectory
        {
            get
            {
                return Path.Combine(RunEnvironmentInfo.GetTestDataLocation(), "DataStorage");
            }
        }
```
- Change your code to
```
string referencePath = Path.Combine(TestLocationDirectory, "SaveAsExcelFileStreamWriterHelperTests", fileName);
```

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -269,6 +270,7 @@ public void Dispose()
 
             private XmlWriter writer;
```
kevcunnane: Minor: we should again move these private fields to the top of the class definition

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -461,28 +463,19 @@ public void WriteAndIncreaseRowReference()
...
-            /// Check if there is a open row tag
-            /// </summary>
-            /// <returns></returns>
-            public bool PenddingRowEndTag()
-            {
-                return currRow != 1;
-            }
         }
 
         private ZipArchive zipArchive;
```
kevcunnane: Minor: again please move all private fields to the top of the class

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/SaveResultsRequest.cs
```diff
@@ -117,7 +128,17 @@ public class SaveResultsAsCsvRequest
     {
         public static readonly
             RequestType<SaveResultsAsCsvRequestParams, SaveResultRequestResult> Type =
-            RequestType<SaveResultsAsCsvRequestParams, SaveResultRequestResult>.Create("query/saveCsv");
+            RequestType<SaveResultsAsCsvRequestParams, SaveResultRequestResult>.Create("query/saveCsv2");
```
benrr101: I think this was an oversight during testing. This should be query/saveCsv, right?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/SaveResultsRequest.cs
```diff
@@ -117,7 +128,17 @@ public class SaveResultsAsCsvRequest
...
+    }
+
+    /// <summary>
+    /// Request type to save results as Excel
+    /// </summary>
+    public class SaveResultsAsExcelRequest
+    {
+        public static readonly
+            RequestType<SaveResultsAsExcelRequestParams, SaveResultRequestResult> Type =
+            RequestType<SaveResultsAsExcelRequestParams, SaveResultRequestResult>.Create("query/saveCsv");
```
benrr101: And this should be query/saveExcel, right?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -0,0 +1,688 @@
+﻿
```
benrr101: One thing we wanted double check on, if this code was written by you, we'll need the MIT license header like in SaveAsExcelFileStreamWriter.cs. If the code wasn't written by you (it came from a sample or is a copy of a library), we'll need to make sure we list the source and the license it was from in here.
Sorry for another change, but it's one of the licensing hurdles we need to make sure everything is legal.

MyRefreshDllsForTestRun.cmd
```diff
@@ -0,0 +1,17 @@
...
+SET _PerfTestSourceLocation="%WORKINGDIR%\test\Microsoft.SqlTools.ServiceLayer.PerfTests\bin\%_BuildConfiguration%\netcoreapp1.0\win7-x64\publish"
+SET _ServiceSourceLocation="%WORKINGDIR%\src\Microsoft.SqlTools.ServiceLayer\bin\%_BuildConfiguration%\netcoreapp1.0\win7-x64\publish"
+
+
+
+dotnet publish %WORKINGDIR%\test\Microsoft.SqlTools.ServiceLayer.PerfTests -c %_BuildConfiguration%
+dotnet publish %WORKINGDIR%\src\Microsoft.SqlTools.ServiceLayer -c %_BuildConfiguration%
+		
+XCOPY /i /E /y %_PerfTestSourceLocation% "%_TargetLocation%\Tests"
+XCOPY /i /E /y %_ServiceSourceLocation% "C:\Users\wzhou\Source\external\vscode-mssql\sqltoolsservice\0.3.0-alpha.9\Windows"
```
benrr101: You probably don't want to check this file in... 😄

wjzhou:
Sorry for the mess of the partial changes. This is my first pull request to a public repository. In my workplace, we need to notify the reviewer (if any) to redo the review process.

MyRefreshDllsForTestRun.cmd
wjzhou: Sorry, I was in a panic mode when I saw you were reviewing the partial commits. Don't want to waste your time. Will remove that

benrr101:
Ah, understood, it's no problem!
I think we generally make commits locally and then push when we have enough stuff that satisfies the review comments. But, then again, we're still kinda new to doing our work on GitHub/in public, and since this is our first community pull request, it's a learning experience for us, too!
We're still super grateful for your contributions!

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/SaveResultsRequest.cs
wjzhou: Sorry, removing these now.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
wjzhou: Understand. 
I wrote all the code freshly, thus there should not be any licence problem.

All the code are written based on the ECMA-376, with some ideas from an unzipped empty excel generated xlsx file. I never reverse-engineering the Excel.

Background: I was looking for for a good library for exporting query result for dotnet core, but can't find one. One day, I opened up a xlsx file and found it is not that complex. I think I can create one within 300 lines. Thus, this code.

I have a question, may I reuse some of these code and make a general export to excel library later? I will licence the library as MIT.

wjzhou:
Hi @benrr101 ,
I think the code is ready for another round of review. Could you take another look? 

So far, it seems to be working well. I have tried the generated file in Excel, Google sheet and LibreOffice.
It seems all the bool, DateTime, Timespan, numbers, strings are formatted as expected. And I also tried a 300,000 rows excel sheet which took about 20 seconds, not as fast as I prefered but worked.

Future work: The code needs more integration tests similar to the json and csv writer.  I can work on these, but I assume you are more familiar with the code base, and can write better tests than I do. 


Sorry for the bad first round, that was more like a proof of concept style coding. I was expecting to get the answer of weather should I continue or not. Thank you for the detailed review.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
benrr101: Yep, I consulted with our legal folks real quick to make sure I fully understood our CLA. You will still own the copyright of the code you contribute to our project. You would be able to use your code however you like, without any need to attribute Microsoft. So to sum it up, yes, you can totally reuse your code to make a general export to Excel library.

PR #280, opened 2017-03-14
Initial metadata and scripting services

kburtram:


src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// <summary>
+        /// Create a SqlConnection to use for querying metadata
+        /// </summary>
+        internal static SqlConnection OpenMetadataConnection(ConnectionInfo connInfo)
+        {
+            try
+            {                 
+                // increase the connection timeout to at least 30 seconds and and build connection string
+                // enable PersistSecurityInfo to handle issues in SMO where the connection context is lost in reconnections
+                int? originalTimeout = connInfo.ConnectionDetails.ConnectTimeout;
```
kburtram: I'm going to do a follow-up PR to refactor connection management so that we share the metadata connection with the language service connection.

kburtram:
I'm going to push this in to unblock UI work.  If there are follow-ups please leave them against the commit and email me.  I'll pick up those changes in the next iteration of these service changes.

PR #281, opened 2017-03-15
fixed peek definition tokenizing

abist:
Fixes [peek definition tokenizing](https://github.com/Microsoft/vscode-mssql/issues/769).

The issue here was that the parser wasn't correctly getting the correct token position, thus failing Peek Definition if the cursor was right in the starting or the end of the object we want to look up. For example, Peek Definition would fail if, let's say, the cursor were anywhere outside this range - sys.d **[atabase]** s

According to the new way PeekDefinition works, instead of returning an error if an object isn't found, we first look at the children and parent tokens and return the definition of those. If nothing is defined, then we return the error.


kevcunnane:
@abist, could you clarify what this is actually fixing? Neither the issue nor this PR describe what the impact will be. For example: "fixes issue where Peek Definition fails to find the correct token position, resulting in failure to return results...." would be my guess? But this helps us prioritize reviewing and signoff since we know what we're looking for. Please also make sure this description is included on checkin so we can look back at commit history and see what was done. Thanks!

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ScriptDocumentInfo.cs
```diff
@@ -129,5 +129,30 @@ internal static Token GetToken(ScriptParseInfo scriptParseInfo, int startLine, i
             }
             return null;
         }
+
+        /// <summary>
+        /// Returns the token that is used for Peek Definition objects
+        /// </summary>
+        internal static Token GetPeakDefinitionToken(ScriptParseInfo scriptParseInfo, int startLine, int startColumn)
```
kevcunnane: Typo: Peak -> Peek

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -814,12 +814,17 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
                         bindingTimeout: LanguageService.PeekDefinitionTimeout,
                         bindOperation: (bindingContext, cancelToken) =>
                         {
-                            string schemaName = this.GetSchemaName(scriptParseInfo, textDocumentPosition.Position, scriptFile);
+                            // Handle tokenizing for Peek Definition 
+                            Position tokenPosition = new Position();
```
kevcunnane: 2 suggestions:
- Move the Position creation up to before `Token selectedToken = ScriptDocumentInfo.GetPeekDefinitionToken) and use its line and character values. This minimizes the times we need to convert the line/character counts by adding 1 to them. Every time we do this, we risk an error later since we're tracking in multiple places
- Use the constructor pattern as follow since it's a clean way to create the object :
```
Position tokenPosition = new Position
{
    Line = textDocumentPosition.Position.Line + 1,
    Character = textDocumentPosition.Position.Character + 1
};
```

abist:
@kevcunnane Agreed, I updated the description.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -808,18 +808,23 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
                         bindOperation: (bindingContext, cancelToken) =>
                         {
-                            string schemaName = this.GetSchemaName(scriptParseInfo, textDocumentPosition.Position, scriptFile);
+                            string schemaName = this.GetSchemaName(scriptParseInfo, tokenPosition, scriptFile);
                             // Script object using SMO
                             PeekDefinition peekDefinition = new PeekDefinition(bindingContext.ServerConnection, connInfo);
                             return peekDefinition.GetScript(
                                 scriptParseInfo.ParseResult, 
-                                textDocumentPosition.Position, 
+                                tokenPosition,
```
kevcunnane: Isn't this an error? `PeekDefinition.GetScript` already adjusts line & character by 1. You need to alter that method to no longer do so.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -808,18 +808,23 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
             {
                 try
                 {
+                    // Handle tokenizing for Peek Definition 
+                    Position tokenPosition = new Position();
```
kevcunnane: We have an internal class called BufferPosition, which tends to use the VSCode Position values but 1-indexed. Perhaps for clarity, we should change the code to use that? Otherwise it's kind of confusing since we're adjusting this but nobody would know why.

abist: The arguments strictly require the Position class, so I'm not sure if we can use the BufferPosition class here.

kevcunnane: I'm suggesting changing the PeekDefinition class to use BufferPosition, since the pattern we have is that this uses Position's values + 1. That way it's clear what the expected input is. Otherwise, if someone else uses this class they might happily send in an unmodified Position object.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -795,7 +795,7 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
             }
 
             // Get token from selected text
-            Token selectedToken = ScriptDocumentInfo.GetToken(scriptParseInfo, textDocumentPosition.Position.Line + 1, textDocumentPosition.Position.Character);
+            Token selectedToken = ScriptDocumentInfo.GetPeakDefinitionToken(scriptParseInfo, textDocumentPosition.Position.Line + 1, textDocumentPosition.Position.Character + 1);
```
benrr101: `Peek` instead of `Peak
Gold Peak Tea

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -796,76 +842,127 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
-            string tokenText = TextUtilities.RemoveSquareBracketSyntax(selectedToken.Text);
 
-            if (scriptParseInfo.IsConnected && Monitor.TryEnter(scriptParseInfo.BuildingMetadataLock))
+            DefinitionResult lastResult = null;
+
+            //try children tokens first
+            foreach (var i in selectedToken.Item1.ToList())
             {
-                try
+                var token = selectedToken.Item1.Pop();
```
kevcunnane: Please define what Item1 is. Assign to a clear, named variable.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -796,76 +842,127 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
+                // Strip "[" and "]"(if present) from the token text to enable matching with the suggestions.
+                // The suggestion title does not contain any sql punctuation
+                string tokenText = TextUtilities.RemoveSquareBracketSyntax(token.Text);
+                textDocumentPosition.Position.Line = token.StartLocation.LineNumber;
+                textDocumentPosition.Position.Character = token.StartLocation.ColumnNumber;
+                if (scriptParseInfo.IsConnected && Monitor.TryEnter(scriptParseInfo.BuildingMetadataLock))
```
kevcunnane: Since you'll move the "not connected" error up, please add a separate error for when you fail to enter the monitor. Probably state "Timeout out waiting to query metadata from the server"

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -796,76 +842,127 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
                         Locations = null
                     };
                 }
-                finally
+            }
+
+            // then check the parents 
+            foreach (var i in selectedToken.Item2.ToList())
+            {
+                var token = selectedToken.Item2.Dequeue();
```
kevcunnane: What is Item2? What does it represent.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -796,76 +842,127 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
+                // Strip "[" and "]"(if present) from the token text to enable matching with the suggestions.
+                // The suggestion title does not contain any sql punctuation
+                string tokenText = TextUtilities.RemoveSquareBracketSyntax(token.Text);
```
kevcunnane: This is common logic - use a helper method

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -796,76 +842,127 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
+                textDocumentPosition.Position.Line = token.StartLocation.LineNumber;
+                textDocumentPosition.Position.Character = token.StartLocation.ColumnNumber;
+                if (scriptParseInfo.IsConnected && Monitor.TryEnter(scriptParseInfo.BuildingMetadataLock))
```
kevcunnane: Since this is the exact same logic, refactor to have 1 method, and just call with the relevant tokens

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -796,76 +842,127 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
+        }
+   
+        /// <summary>
+        /// Wrapper around find token method 
+        /// </summary>
+        /// <param name="scriptParseInfo"></param>
+        /// <param name="startLine"></param>
+        /// <param name="startColumn"></param>
+        /// <returns> token index</returns>
+        private int FindTokenWrapper(ScriptParseInfo scriptParseInfo, int startLine, int startColumn)
```
kevcunnane: Minor: should this be a global utility? Does any other scenario need this? Also I'd rename to "FindTokenWithCorrectOffset" or similar. "Wrapper" doesn't make clear why this is.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ScriptDocumentInfo.cs
```diff
@@ -129,5 +131,60 @@ internal static Token GetToken(ScriptParseInfo scriptParseInfo, int startLine, i
...
         }
+
+        /// <summary>
+        /// Returns the token that is used for Peek Definition objects
+        /// </summary>
+        internal static Tuple<Stack<Token>, Queue<Token>> GetPeekDefinitionTokens(ScriptParseInfo scriptParseInfo, int startLine, int startColumn)
+        {
+            Stack<Token> childrenTokens = new Stack<Token>();
+            Queue<Token> parentTokens = new Queue<Token>();
+            if (scriptParseInfo != null && scriptParseInfo.ParseResult != null && scriptParseInfo.ParseResult.Script != null && scriptParseInfo.ParseResult.Script.Tokens != null)
```
kevcunnane: Minor: put each if check on its own line

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ScriptDocumentInfo.cs
```diff
@@ -129,5 +131,60 @@ internal static Token GetToken(ScriptParseInfo scriptParseInfo, int startLine, i
...
+            {
+                var tokenIndex = scriptParseInfo.ParseResult.Script.TokenManager.FindToken(startLine, startColumn);
+                if (tokenIndex >= 0)
+                {
+                    // return the current token and the ones to its right
+                    // until we hit a whitespace token
+                    int currentIndex = 0;
+                    foreach (var token in scriptParseInfo.ParseResult.Script.Tokens)
+                    {
+
```
kevcunnane: Minor: remove unnecessary whitespace

src/Microsoft.SqlTools.ServiceLayer/Workspace/Contracts/TextDocument.cs
```diff
@@ -156,7 +156,36 @@ public class Position
...
+        public int Character { get; set; }
+
+        /// <summary>
+        /// Overrides the base equality method
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public override bool Equals(object obj)
+        {
+            if (obj == null || GetType() != obj.GetType())
```
kevcunnane: What if someone has a subclass of position? I think IsInstanceOf or similar (we have cases of this in the project) work better. You could also just do obj as Position and check if null

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -691,8 +692,84 @@ public void GetDatabaseWithQueryConnectionTest()
...
+        /// Get Definition for a object with no definition. Expect a error result
+        /// </summary>
+        [Fact]
+        public async void GetDefinitionFromChildrenAndParents()
+        {
+            string queryString = "select * from master.sys.objects";
 
+            // place the cursor on every token
+            //cursor on objects
+            TextDocumentPosition objectDocument = new TextDocumentPosition
```
kevcunnane: Use a helper method to make these all 1-line definitions

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -691,8 +692,84 @@ public void GetDatabaseWithQueryConnectionTest()
...
-        }
+        }
+
+        /// <summary>
+        /// Get Definition for a object with no definition. Expect a error result
+        /// </summary>
+        [Fact]
+        public async void GetDefinitionFromChildrenAndParents()
+        {
+            string queryString = "select * from master.sys.objects";
```
kevcunnane: At a minimum please add procedure-specific tests, and tests with and without escaping of brackets on each one. Ideally make it easy to define multiple inputs so future bug regressions can be tested (e.g. add in a framework that takes JSON file input defining the query & positions to test from, and what's expected to be returned at each). This 2nd ask is a stretch goal though - can ship without it.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -820,6 +820,54 @@ internal CompletionItem ResolveCompletionItem(CompletionItem completionItem)
             return result;
         }
 
+        private DefinitionResult GetDefinitionFromTokenList(TextDocumentPosition textDocumentPosition, List<Token> tokenList,
+                ScriptParseInfo scriptParseInfo, ScriptFile scriptFile, ConnectionInfo connInfo)
+        {
+
```
kevcunnane: Minor: minimize whitespace lines

PR #283, opened 2017-03-16
feature/edit/subset

benrr101:
This change adds the edit/subset command that returns a subset of rows with pending edits applied to it.

This change also adds some very important bug fixes to the LongList class that basically allow it to work properly.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
@@ -86,6 +86,22 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
         #region Properties
 
         /// <summary>
+        /// Converts the cell update to a DbCellValue
+        /// </summary>
+        public DbCellValue AsDbCellValue
+        {
+            get
+            {
+                return new DbCellValue
+                {
+                    DisplayValue = ValueAsString,
+                    IsNull = Value == DBNull.Value,
+                    RawObject = Value
+                };
+            }
+        }
+
+        /// <summary>
         /// The column that the cell will be placed in
         /// </summary>
         public DbColumnWrapper Column { get; }
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
@@ -393,7 +393,8 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
         }
 
         /// <summary>
-        /// Retrieves the requested subset of rows from the requested result set. Intended to
+        /// Retrieves the requested subset of rows from the requested result set. Intended to be
+        /// called by another service.
         /// </summary>
         /// <param name="subsetParams">Parameters for the subset to retrieve</param>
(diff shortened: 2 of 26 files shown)


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -86,6 +86,22 @@ public CellUpdate(DbColumnWrapper column, string valueAsString)
         #region Properties
 
         /// <summary>
+        /// Converts the cell update to a DbCellValue
+        /// </summary>
+        public DbCellValue AsDbCellValue
```
kburtram: the naming of this property suggests it should probably be a method (i.e. "As" is a verb)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -422,6 +392,28 @@ public void InitializeService(ServiceHost serviceHost)
             }
         }
 
+        /// <summary>
+        /// Retrieves the requested subset of rows from the requested result set. Intended to
```
kburtram: >Intended to [](start = 82, length = 11)

comment seems to cut off

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -422,6 +392,28 @@ public void InitializeService(ServiceHost serviceHost)
...
+        /// </summary>
+        /// <param name="subsetParams">Parameters for the subset to retrieve</param>
+        /// <returns>The requested subset</returns>
+        /// <exception cref="ArgumentOutOfRangeException">The requested query does not exist</exception>
+        public async Task<ResultSetSubset> InterServiceResultSubset(SubsetParams subsetParams)
```
kburtram: >InterService [](start = 43, length = 12)

I'm not quite clear what "InterService" means in this context.  Perhaps a different name (or completed comments :) ) may make it more intuitive.

kburtram:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
benrr101: Not to sound grammar obsessed, but in this case "as" is used as a preposition :)
As such, I think it works fine as-is. (Yes, pun intended)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
benrr101: I have a tendency to do this. Will fix.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
benrr101: That was the bit of the comment that was left off, lol...

Basically I picked the name "interservice" to indicate that it's meant to be called by another service (in this case the EditDataService). This bypasses the request context and implementation that assumes its being called by the JSON RPC. We also use the same methods from the JSON RPC request handlers, so I'm trying to keep the implementations here DRY.

PR #285, opened 2017-03-20
Added options to connection parameter

llali:
- Added a generic list of options to connection parameters. 
- Added some missing options provider options
- Added tests to verify the new connection parameter

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
@@ -15,6 +15,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
     /// </remarks>
     public class ConnectionDetails : ConnectionSummary
     {
+        public ConnectionDetails()
+        {
+            Options = new Dictionary<string, object>();
+        }
+
         /// <summary>
         /// Gets or Sets the connection options
         /// </summary>
@@ -27,11 +32,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
         public string Password {
             get
             {
-                return getOptionValue<string>("password");
+                return GetOptionValue<string>("password");
             }
             set
             {
-                setOptionValue("password", value);
+                SetOptionValue("password", value);
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -14,119 +16,447 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
...
+                return getOptionValue<string>("typeSystemVersion");
+            }
+
+            set
+            {
+                setOptionValue("typeSystemVersion", value);
+            }
+        }
+
+        private T getOptionValue<T>(string name)
```
kburtram: In C# naming convention methods should start with a capital letter.  So this should be GetOptionValue and SetOptionValue.

kburtram: I wonder if the return type should be a nullable type so that we can return null if the value isn't in the map?  Otherwise, caller can't tell difference between "not found item" and "item has default value for type".  You don't need to make the change, just wondering if this is desired handling of not found items.

llali: if the method is called with nullable type then the default is null. For example GetOptionValue<int?>('some param') returns null if the option is not set

src/Microsoft.SqlTools.ServiceLayer/ServiceHost.cs
```diff
@@ -190,15 +190,15 @@ public void RegisterInitializeTask(InitializeCallback callback)
...
+        internal static ConnectionProviderOptions BuildConnectionProviderOptions()
         {
             return new ConnectionProviderOptions
             {
                 Options = new ConnectionOption[]
                 {
                     new ConnectionOption
                     {
-                        Name = "Server Name",
+                        Name = "server",
```
kburtram: If we are fixing the Name field we should add an explicit DisplayName field for the UI to show.  By default the DMP handlers will mirror Name into DisplayName is there is no DisplayName provided.

src/Microsoft.SqlTools.ServiceLayer/ServiceHost.cs
```diff
@@ -245,62 +245,142 @@ private static ConnectionProviderOptions BuildConnectionProviderOptions()
...
                     },
                     new ConnectionOption
                     {
-                        Name = "Context Connection",
+                        Name = "contextConnection",
+                        ValueType = ConnectionOption.ValueTypeString
+                    },
+                    new ConnectionOption
+                    {
+                        Name = "port",
```
kburtram: should we also add default values for fields that have known default values?  (e.g. 1433)

PR #287, opened 2017-03-21
Update cake to use JSON.Net 9.0.1 to fix build break

kburtram:
Cake commands are failing in Jenkins builds with the below error.  This PR fixes that error and also makes the LOC packages compatible with newer nuget version by adding version to filename.

An error occurred while downloading package 'Newtonsoft.Json.10.0.1 : ' from source 'https://www.nuget.org/api/v2'.
NuGet exited with 1
Could not find any assemblies compatible with .NETFramework,Version=v4.5.
Error: Failed to install addin 'Newtonsoft.Json'.


kburtram:
This is a build script update to fix a build break so I'll automerge.  Ping me directly with any feedback.

PR #288, opened 2017-03-21
Remove SELECT * from edit/initialize Query

benrr101:
This change removes the `SELECT *` query generation during the edit/initialize request. Although it should be simple, it's actually complex.

_Edit(Table|Column)Metadata_
Originally the EditMetadataFactory class was written with the assumption that we had both SqlClient metadata and SMO metadata, since both were available when we used the metadata factory, after querying the table/view. Now that we need SMO metadata *before* we execute the query, we need the metadata factory to generate Edit(Table|Column)Metadata classes without having the SqlClient metadata. Thus, the Edit(Table|Column)Metadata classes were refactored to have a Basic and Extended properties set. When initially creating them, only properties from SMO metadata will be needed. When SqlClient metadata is available, the metadata can be `.Extend()`ed to include properties from SqlClient.

_Async SMO Metadata Lookup_
Since SMO actions can take serious time, I decided to make sure that the SMO metadata lookup happened asynchronously. To do this without odd callbacks, I wrapped all the initialization logic in a Task in the Session. The edit data service would then fire-and-forget the Session initialize session. The task inside the Session is exposed for awaiting in unit tests. And edit/initialize is blocked on whether or not there is a Session in the active sessions list (ie, if a session is initialized it *must* be disposed even if it fails)

_Cheating Callback-Based Async_
To simplify the flow of the initialization process, I cheated query execution to make sure it return a query at the end. The cheating was done by wrapping the query execution in a TaskCompletionSource, setting a result in all of the query callbacks, and awaiting the TaskCompletionSource at the end of the method. 

Small tweak, I added JetBrains' Project Rider project folders to the .gitignore. I also tweaked it as per the .gitignore plugin's suggestions where it was a clear improvement.


src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -313,52 +271,64 @@ private EditSession GetActiveSessionOrThrow(string ownerUri)
...
-                // Create the editSession and add it to the sessions list
-                EditSession editSession = new EditSession(resultSet, metadata);
-                if (!ActiveSessions.TryAdd(initParams.OwnerUri, editSession))
-                {
-                    throw new InvalidOperationException("Failed to create edit editSession, editSession already exists.");
-                }
-                readyParams.Success = true;
-            }
-            catch (Exception)
+            // Execute the
```
kevcunnane: Please complete or remove this comment.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -313,52 +271,64 @@ private EditSession GetActiveSessionOrThrow(string ownerUri)
...
         }
 
-        private void CompleteInitializeWaitHandler(string ownerUri, bool result)
+        private static Task SendSessionReadyEvent(IEventSender eventSender, string ownerUri, bool success,
+            string message)
         {
-            // If there isn't a wait handler, just ignore it
-            TaskCompletionSource<bool> initializeWaiter;
-            if (ownerUri != null && InitializeWaitHandles.TryRemove(ownerUri, out initializeWaiter))
+            var esrp = new EditSessionReadyParams
```
kevcunnane: Please avoid acronyms - `sessionReadyParams` is probably a good name for this.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -150,6 +195,11 @@ public EditCreateRowResult CreateRow()
         /// <param name="errorHandler">Callback to perform if the commit process has failed at some point</param>
         public void CommitEdits(DbConnection connection, Func<Task> successHandler, Func<Exception, Task> errorHandler)
         {
+            if (!IsInitialized)
```
kevcunnane: Minor: since you check for this in every method, why not have a 1-line check by creating a `ThrowIfNotInitialized` method and calling it. We do this in other codebases as it shortens up methods.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -378,5 +485,26 @@ public EditUpdateCellResult UpdateCell(long rowId, int columnId, string newValue
...
+        private string ConstructInitializeQuery()
+        {
+            // Using the columns we know, put together a query for the rows in the table
+            var columns = objectMetadata.Columns.Select(col => col.EscapedName);
+            var columnClause = string.Join(", ", columns);
+
+            return $"SELECT ${columnClause} FROM ${objectMetadata.EscapedMultipartName}";
+        }
+
+        public class EditSessionQueryExecutionState
```
kevcunnane: This is public and inside a public class, right? 2 questions:
- Should it be an inner class, or a standalone one?
- Can you document the class since it's public? Thanks!

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -44,25 +44,59 @@ public IEditTableMetadata GetObjectMetadata(DbConnection connection, DbColumnWra
...
+                EditColumnMetadata column = new EditColumnMetadata
+                {
+                    DefaultValue = defaultValue,
+                    EscapedName = SqlScriptFormatter.FormatIdentifier(smoColumn.Name),
+                    Ordinal = i,
+                };
+                editColumns.Add(column);
+            }
+
+            // If a table is memory optimized it is Hekaton. If it's a view, then it can't be Hekaton
```
kevcunnane: I'd rewrite this as `Only tables can be memory-optimized`.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.xlf
```diff
@@ -561,6 +561,36 @@
         <target state="new">NULL is not allowed for this column</target>
         <note></note>
       </trans-unit>
+      <trans-unit id="EditDataObjectNotFound">
+        <source>Table or view requested to edit could not be found</source>
```
kevcunnane: Minor: I'd suggest `requested for edit` might be better

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/SessionTests.cs
```diff
@@ -253,26 +278,260 @@ public static IEnumerable<object> RowIdOutOfRangeData
...
+            // ... The success handler should not have been called at all
+            successHandler.Verify(f => f(), Times.Never);
+        }
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("It fail.")]
+        public async Task InitializeQueryFailReturnNull(string message)
+        {
+            // Setup:
```
kevcunnane: Minor: it seems like overall setup for these methods is nearly identical. Might a single private Setup method save some space and tighten up the test code?

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
benrr101: I generally think of nested classes as being only for the benefit of the class its nested in. There's no use for this class outside of an EditSession, so I think it makes sense for it to be a nested class. If you feel strongly about this, I can change it.

As for documenting, I'll definitely do that

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/SessionTests.cs
benrr101: You're right for the most part. However, for each test, we swap out a different component in the setup (eg in InitializeQueryFailReturnNull vs InitializeQueryFailException, we use a different mock query runner delegate). At that point, even though they're the same sort of setup steps, a  private setup method would likely reduce the code by only a line or two.

PR #291, opened 2017-03-22
fixed an issue with setting a boolean option to string

llali:
In connection details boolean options can be set to string "True" or "False" but when parsing back the value should be converted to boolean correctly

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
@@ -3,7 +3,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
 using System.Collections.Generic;
+using System.Globalization;
+using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -446,9 +446,16 @@ private T GetOptionValue<T>(string name)
             if (Options != null && Options.ContainsKey(name))
             {
                 object value = Options[name];
-                if(value != null && ( typeof(T) == typeof(int) || typeof(T) == typeof(int?)))
+                if (value != null && (typeof(T) != value.GetType()))
                 {
-                    value = System.Convert.ToInt32(value);
+                    if (typeof(T) == typeof(int) || typeof(T) == typeof(int?))
```
benrr101: Would it work if we just used Convert.ChangeType ? This way we can add a lot of type support without a lot of code additions

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -446,9 +446,16 @@ private T GetOptionValue<T>(string name)
...
+                    {
+                        value = System.Convert.ToInt32(value);
```
benrr101: The use of `System` is unnecessary

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
llali: ChangeType was not working to specifically converting to int32

benrr101: I'm not sure how, convert.changetype takes the type as it's second parameter, so you should be able to specify `int` as the type to convert to.

llali: it cannot convert logn to int, it fails, I tried that already

PR #292, opened 2017-03-22
Ensuring DisplayValue is set to a value when null

benrr101:
Make sure that a DbCellValue always has a DisplayValue. When the RawValue of a DbCellValue is null, set the DisplayValue to "NULL", which is configured in the string constants file

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
@@ -205,6 +205,14 @@ namespace Microsoft.SqlTools.ServiceLayer
             }
         }
 
+        public static string QueryServiceCellNull
+        {
+            get
+            {
+                return Keys.GetString(Keys.QueryServiceCellNull);
+            }
+        }
+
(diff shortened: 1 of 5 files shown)


coquagli:
Is this really the only thing that needs to change to make the value `"NULL"`?

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
```diff
@@ -205,6 +205,14 @@ public static string QueryServiceColumnNull
             }
         }
 
+        public static string QueryServiceCellNull
+        {
+            get
+            {
+                return Keys.GetString(Keys.QueryServiceCellNull);
```
llali: Did you add it here to localize it?

benrr101: Pretty much so that we always return the localized null string and so that we always return *something* for the display value of a dbcellvalue.

Though the code that actually changes behavior is in the "Changes that *actually do something!*" v2 commit... whoops

PR #294, opened 2017-03-24
Important bug fixes for edit/commit (extended properties)

benrr101:
Important bug fixes for edit/commit
* Ensures that metadata is "extended" before creating a new row edit object
* Adds the commit handler to edit data service initialization calls
* Unit tests for the associated changes


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEdit.cs
```diff
@@ -38,6 +38,11 @@ public abstract class RowEditBase : IComparable<RowEditBase>
         /// <param name="associatedMetadata">Metadata provider for the object to edit</param>
         protected RowEditBase(long rowId, ResultSet associatedResultSet, EditTableMetadata associatedMetadata)
         {
+            if (!associatedMetadata.HasExtendedProperties)
+            {
+                throw new ArgumentException(SR.EditDataMetadataNotExtended);
```
Raymondd: Is this caught somewhere and a failure returned?

benrr101: Yes, this will be caught by the Handle*Requests (in our case HandleUpdateCellRequest) and returned as an error.

PR #295, opened 2017-03-27
Adding decoding of multipart identifiers, default schema workaround

benrr101:
This change adds a couple things

_Multipart Identifier Decoding_
The ability to decode a multipart identifier (with or without escaping) has been added to the SqlScriptFormatter utility class. This code is utilized to split a table name provided to the edit/initialize request into schema and table name.

_Default Schema Workaround_
The code that retrieves the SMO metadata objects originally used the `[]` operator to access the objects. Due to a bug(?) in SMO, this results in problems when loading tables without a default schema (in our case if you're logged in as SA). Using the metadata object constructors gets around this issue, we are explicitly using them.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/IEditMetadataFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/IEditMetadataFactory.cs
@@ -16,9 +16,13 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
         /// Generates a edit-ready metadata object
         /// </summary>
         /// <param name="connection">Connection to use for getting metadata</param>
-        /// <param name="objectName">Name of the object to return metadata for</param>
+        /// <param name="objectNamedParts">
+        /// The multipart namefor the object split and unwrapped. At most two components can be
+        /// provided (schema, table/view name). At minimum table/view name can be provided, and
+        /// default schema will be used for schema name.
+        /// </param>
         /// <param name="objectType">Type of the object to return metadata for</param>
         /// <returns>Metadata about the object requested</returns>
-        EditTableMetadata GetObjectMetadata(DbConnection connection, string objectName, string objectType);
+        EditTableMetadata GetObjectMetadata(DbConnection connection, string[] objectNamedParts, string objectType);
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/IEditMetadataFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/IEditMetadataFactory.cs
@@ -17,7 +17,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
         /// </summary>
         /// <param name="connection">Connection to use for getting metadata</param>
         /// <param name="objectNamedParts">
-        /// The multipart namefor the object split and unwrapped. At most two components can be
+        /// The multipart name for the object split and unwrapped. At most two components can be
         /// provided (schema, table/view name). At minimum table/view name can be provided, and
         /// default schema will be used for schema name.
         /// </param>
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
@@ -11,6 +11,7 @@ using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlServer.Management.Smo;
 using Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection;
 using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.EditData
 {
@@ -23,11 +24,21 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
         /// Generates a edit-ready metadata object using SMO
         /// </summary>
         /// <param name="connection">Connection to use for getting metadata</param>
-        /// <param name="objectName">Name of the object to return metadata for</param>
+        /// <param name="objectNamedParts">Split and unwrapped name parts</param>
         /// <param name="objectType">Type of the object to return metadata for</param>
         /// <returns>Metadata about the object requested</returns>
-        public EditTableMetadata GetObjectMetadata(DbConnection connection, string objectName, string objectType)
+        public EditTableMetadata GetObjectMetadata(DbConnection connection, string[] objectNamedParts, string objectType)
         {
+            Validate.IsNotNull(nameof(objectNamedParts), objectNamedParts);
+            if (objectNamedParts.Length <= 0)
+            {
+                throw new ArgumentNullException(nameof(objectNamedParts), "A object name must be provided");
+            }
+            if (objectNamedParts.Length > 2)
+            {
+                throw new InvalidOperationException("Explicitly specifying server or database is not supported");
+            }
+
             // Get a connection to the database for SMO purposes
             SqlConnection sqlConn = connection as SqlConnection;
             if (sqlConn == null)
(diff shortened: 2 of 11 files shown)


src/Microsoft.SqlTools.ServiceLayer/EditData/IEditMetadataFactory.cs
```diff
@@ -16,9 +16,13 @@ public interface IEditMetadataFactory
         /// Generates a edit-ready metadata object
         /// </summary>
         /// <param name="connection">Connection to use for getting metadata</param>
-        /// <param name="objectName">Name of the object to return metadata for</param>
+        /// <param name="objectNamedParts">
+        /// The multipart namefor the object split and unwrapped. At most two components can be
```
coquagli: name for

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -23,11 +24,21 @@ public class SmoEditMetadataFactory : IEditMetadataFactory
...
+        /// <param name="objectNamedParts">Split and unwrapped name parts</param>
         /// <param name="objectType">Type of the object to return metadata for</param>
         /// <returns>Metadata about the object requested</returns>
-        public EditTableMetadata GetObjectMetadata(DbConnection connection, string objectName, string objectType)
+        public EditTableMetadata GetObjectMetadata(DbConnection connection, string[] objectNamedParts, string objectType)
         {
+            Validate.IsNotNull(nameof(objectNamedParts), objectNamedParts);
+            if (objectNamedParts.Length <= 0)
+            {
+                throw new ArgumentNullException(nameof(objectNamedParts), SR.EditDataMetadataObjectNameRequired);
```
coquagli: Do we have a better message in SR to provide with this Exception? E.g. "Array cannot be empty"

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -46,22 +57,23 @@ public EditTableMetadata GetObjectMetadata(DbConnection connection, string objec
...
             Server server = new Server(new ServerConnection(sqlConn));
+            Database db = new Database(server, sqlConn.Database);
             TableViewTableTypeBase smoResult;
             switch (objectType.ToLowerInvariant())
             {
                 case "table":
-                    smoResult = server.Databases[sqlConn.Database].Tables[objectName];
+                    smoResult = objectNamedParts.Length == 1
+                        ? new Table(db, objectNamedParts[0])                        // No schema provided
+                        : new Table(db, objectNamedParts[1], objectNamedParts[0]);  // Schema provided
```
coquagli: If we are expecting things in the array to be in a certain order, it might be good to mention that in the function header.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs
```diff
@@ -260,6 +337,12 @@ private static string FormatBinary(object value)
             return "0x" + BitConverter.ToString(bytes).Replace("-", string.Empty);
         }
 
+        private static bool HasNextCharacter(string haystack, char needle, int position)
```
coquagli: Although the analogy is pretty funny, it might be good to rename these params.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs
```diff
@@ -176,6 +176,83 @@ public static string UnwrapLiteral(string literal)
...
+                    }
+                }
+                else
+                {
+                    switch (iChar)
+                    {
+                        case '[':
+                            if (bracketsClosed)
+                            {
+                                throw new FormatException();
```
coquagli: A message might be good here, considering this is a pretty specific syntax error.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs
```diff
@@ -176,6 +176,83 @@ public static string UnwrapLiteral(string literal)
             return literal;
         }
 
+        public static string[] DecodeMultipartIdenfitier(string multipartIdentifier)
```
coquagli: Might be good to provide the exceptions with messages.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs
```diff
@@ -176,6 +176,83 @@ public static string UnwrapLiteral(string literal)
...
+            StringBuilder sb = new StringBuilder();
+            List<string> namedParts = new List<string>();
+            bool insideBrackets = false;
+            bool bracketsClosed = false;
+            for (int i = 0; i < multipartIdentifier.Length; i++)
+            {
+                char iChar = multipartIdentifier[i];
+                if (insideBrackets)
+                {
+                    if (iChar == ']')
```
coquagli: Do we need to escape `[` instead of `]`? Looking here under "Using Wildcard Characters As Literals":
https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/SqlScriptFormatterTests.cs
```diff
@@ -308,6 +310,55 @@ public void GuidTest()
...
+                yield return new object[] {"mixed.[escaped]", new[] {"mixed", "escaped"}};
+                yield return new object[] {"[escaped].mixed", new[] {"escaped", "mixed"}};
+            }
+        }
+
+        [Theory]
+        [InlineData("[bracket]closed")]
+        [InlineData("[bracket][closed")]
+        [InlineData(".stuff")]
+        [InlineData(".")]
```
coquagli: [InlineData("stuff.")]
[InlineData("[bracket")]

Are those negative cases?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/SqlScriptFormatterTests.cs
```diff
@@ -308,6 +310,55 @@ public void GuidTest()
...
+                yield return new object[] {"[multi].[escaped].[split]", new[] {"multi", "escaped", "split"}};
+                yield return new object[] {"[escaped]]characters]", new[] {"escaped]characters"}};
+                yield return new object[] {"[multi]]escaped]]chars]", new[] {"multi]escaped]chars"}};
+                yield return new object[] {"[multi]]]]chars]", new[] {"multi]]chars"}};
+                yield return new object[] {"unescaped]chars", new[] {"unescaped]chars"}};
+                yield return new object[] {"multi]unescaped]chars", new[] {"multi]unescaped]chars"}};
+                yield return new object[] {"multi]]chars", new[] {"multi]]chars"}};
+                yield return new object[] {"[escaped.dot]", new[] {"escaped.dot"}};
+                yield return new object[] {"mixed.[escaped]", new[] {"mixed", "escaped"}};
+                yield return new object[] {"[escaped].mixed", new[] {"escaped", "mixed"}};
```
coquagli:
```
yield return new object[] {"multi.[[].chars", new[] {"multi" ,"[", "chars"}};
```
Is that a legit positive case? Do we need other cases with `[`?

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs
benrr101: This is actually a very popular naming convention in PHP: http://php.net/manual/en/function.strpos.php

coquagli: Didn't know that! Feel free to keep it as-is, then.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs
benrr101: No need to escape `[` when used inside square brackets. The syntax you listed only matters for values where quasi-regex patterns are being written (ie, in a `LIKE` clause).

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/SqlScriptFormatterTests.cs
benrr101: `SELECT [timestamp]. FROM [identitytest]` => `Incorrect syntax near the keyword 'FROM'.`
`SELECT [timestamp FROM [identitytest]` => `Invalid column name 'timestamp FROM [identitytest'.`

Yes, these would be legit negative cases

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/SqlScriptFormatterTests.cs
benrr101: Although I wouldn't recommend it `SELECT dbo.[[].weird FROM [[]` totally works. I will add a case for that.

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
benrr101: Agreed. This is really just temporary stuff since the code will likely be blown away with the metadata service @kburtram is working on.

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
benrr101: I think the SR.EditDataMetadataObjectNameRequired is a bit clearer message for the user.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatter.cs
coquagli: Gotcha.

PR #296, opened 2017-03-27
Add table and view metadata messges

kburtram:


src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoMetadataFactory.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <param name="objectName">Name of the object to return metadata for</param>
+        /// <param name="objectType">Type of the object to return metadata for</param>
+        /// <returns>Metadata about the object requested</returns>
+        TableMetadata GetObjectMetadata(DbConnection connection, string schemaName, string objectName, string objectType);
+    }
+
+    /// <summary>
+    /// Factory that generates metadata using a combination of SMO and SqlClient metadata
+    /// </summary>
+    public class SmoMetadataFactory : IMetadataFactory
```
kburtram: this code is adapted from the edit table metadata functionality.  I'm going to remove the edit table metadata code and have that component reuse this code as a separate change since they slightly different requirements and I'd like to review the modifications to the edit data scenario independent of the initial metadata service changes.

src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/TableMetadataRequest.cs
```diff
@@ -0,0 +1,33 @@
...
+        public string OwnerUri { get; set; }        
+
+        public string Schema { get; set; }  
+
+        public string ObjectName { get; set; }  
+    }
+
+    public class TableMetadataResult
+    {
+        public ColumnMetadata[] Columns { get; set; }
```
benrr101: I believe this should be returning a TableMetadata object

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoMetadataFactory.cs
benrr101: I think ideally, the MetadataService should expose a method for retrieving table metadata. This way the EditDataService doesn't need to take a hard dependency on a class provided inside another service's namespace. Or to put it another way, the functionality that a service needs should be discoverable via MEF. Since the MEF objects are all services, in order to stick with precedent, we'd want the functionality to be accessible via the service.

The QueryExecutionService provides a decent example of what I'm suggesting. The base functionality of executing a query and disposing of a query is provided via InterService* methods. When a different service needs access to query handling (ie, EditDataService) it calls the InterServiceExecuteQuery method on the QueryExecutionService. If a JSON RPC request to execute a query is received, it's handled by HandleExecuteQueryRequest which calls InterServiceExecuteQuery.

I'm guessing that a change like I'm suggesting is further in the pipeline, but I want to make sure we're on the same page for providing access to metadata to multiple services in the tools service.

src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/ColumnMetadata.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        /// <summary>
+        /// Extracts extended column properties from the database columns from SQL Client
+        /// </summary>
+        /// <param name="dbColumn">The column information provided by SQL Client</param>
+        public void Extend(DbColumnWrapper dbColumn)
+        {
+            Validate.IsNotNull(nameof(dbColumn), dbColumn);
+
+            // DbColumn = dbColumn;
```
abist: we can get rid of this comment

PR #297, opened 2017-03-29
Implicit Revert Row Cleanup Logic

benrr101:
This change enhances the way that edit/updateCell and edit/revertCell operations are performed. 

## **THE API BREAKING CHANGES**:
* edit/updateCell now returns an EditCell (a DbCellValue with a dirty flag) and a row dirty flag.
* edit/revertCell now returns an EditCell (a DbCellValue with a dirty flag) and a row dirty flag.

If by setting the value of a cell via edit/updateCell the row no longer has any edits (an "implicit revert"), the entire row's edit will be removed from the cache. Additionally, if by requesting edit/revert all the pending edits for a row are removed, the entire row's edit will be removed from the cache. This will prevent issues where committing will generate an invalid script because it has no pending changes.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
@@ -1,0 +1,24 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.EditData.Contracts
+{
+    public class EditCell : DbCellValue
+    {
+        public EditCell(DbCellValue dbCellValue, bool isDirty)
+        {
+            IsDirty = isDirty;
+
+
+            DisplayValue = dbCellValue.DisplayValue;
+            IsNull = dbCellValue.IsNull;
+            RawObject = dbCellValue.RawObject;
+        }
+
+        public bool IsDirty { get; set; }
+    }
+}
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
@@ -13,6 +13,11 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.Contracts
     public class EditCell : DbCellValue
     {
         /// <summary>
+        /// Default, parameterless constructor to make sure that JSON serializing is happy
+        /// </summary>
+        public EditCell() {}
+
+        /// <summary>
         /// Constructs a new EditCell based on a DbCellValue
         /// </summary>
         /// <param name="dbCellValue">The DbCellValue that will be enhanced</param>
@@ -20,7 +25,6 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.Contracts
         public EditCell(DbCellValue dbCellValue, bool isDirty)
         {
             IsDirty = isDirty;
-
 
             DisplayValue = dbCellValue.DisplayValue;
             IsNull = dbCellValue.IsNull;
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
@@ -7,8 +7,16 @@ using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
 
 namespace Microsoft.SqlTools.ServiceLayer.EditData.Contracts
 {
+    /// <summary>
+    /// Cell that wraps info from <see cref="DbCellValue"/> for edit purposes
+    /// </summary>
     public class EditCell : DbCellValue
     {
+        /// <summary>
+        /// Constructs a new EditCell based on a DbCellValue
+        /// </summary>
+        /// <param name="dbCellValue">The DbCellValue that will be enhanced</param>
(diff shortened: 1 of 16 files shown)


src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Constructs a new EditCell based on a DbCellValue
+        /// </summary>
+        /// <param name="dbCellValue">The DbCellValue that will be enhanced</param>
+        /// <param name="isDirty">Whether or not the edit cell is dirty</param>
+        public EditCell(DbCellValue dbCellValue, bool isDirty)
+        {
+            IsDirty = isDirty;
+
```
coquagli: Extra newline here.

src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Cell that wraps info from <see cref="DbCellValue"/> for edit purposes
+    /// </summary>
+    public class EditCell : DbCellValue
+    {
+        /// <summary>
+        /// Constructs a new EditCell based on a DbCellValue
+        /// </summary>
+        /// <param name="dbCellValue">The DbCellValue that will be enhanced</param>
+        /// <param name="isDirty">Whether or not the edit cell is dirty</param>
+        public EditCell(DbCellValue dbCellValue, bool isDirty)
```
coquagli: I feel like it's super weird to pass the parent class into the constructor of the child class.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -319,8 +319,17 @@ public string RevertCell(long rowId, int columnId)
                 throw new ArgumentOutOfRangeException(nameof(rowId), SR.EditDataUpdateNotPending);
             }
 
+            // Update the row
+            EditRevertCellResult revertResult = pendingEdit.RevertCell(columnId);
+            if (!revertResult.IsRowDirty)
```
coquagli: This if statement can probably be refactored into a separate method

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -214,10 +216,8 @@ public override EditUpdateCellResult SetCell(int columnId, string newValue)
             // Put together a result of the change
             EditUpdateCellResult eucr = new EditUpdateCellResult
```
coquagli: `editUpdateCellResult` might be a clearer name here

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
```diff
@@ -167,15 +167,21 @@ public override string GetScript()
         /// </summary>
         /// <param name="columnId">Ordinal of the column to revert</param>
         /// <returns>The value that was </returns>
-        public override string RevertCell(int columnId)
+        public override EditRevertCellResult RevertCell(int columnId)
         {
             Validate.IsWithinRange(nameof(columnId), columnId, 0, associatedRow.Count - 1);
 
             // Remove the cell update
+            // NOTE: This is best effort. The only way TryRemove can fail is if it is already
```
coquagli: This block could also be put into a mehtod

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -319,8 +319,17 @@ public string RevertCell(long rowId, int columnId)
...
+            {
+                // Make an attempt to remove the clean row edit. If this fails, it'll be handled on commit attempt.
```
coquagli: Perhaps in this comment mention the method where commit handles this. I'm kinda iffy commenting this as code often gets refactored while comments do not, but it could potentially be helpful.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
benrr101: Eh...... I don't see the benefit of doing that here. It's not repeated anywhere and it's only 2 lines

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
benrr101: Fun fact is that it the code doesn't actually do this yet 👍

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
coquagli: Other than variable names, 177-183 and 208-214 are identical.

benrr101: Except that they define two different class instances...

src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            IsDirty = isDirty;
+
+            DisplayValue = dbCellValue.DisplayValue;
```
kevcunnane: 2 points:
- Please validate dbCellValue arg is not null
- Often we implement a `CopyTo` method on a class like DbCellValue if you want to copy over some data from it to this new subclass. You could put that in and call it from this constructor, meaning that the parent class will know what should be copied instead of this subclass knowing.

PR #299, opened 2017-03-29
Task removing unused constants

Raymondd:
Addresses https://github.com/Microsoft/vscode-mssql/issues/800

Removes localizable constants with no references from our project. 

```
ErrorUnexpectedCodeObjectType 
EditDataObjectNotFound
EditDataColumnUpdateNotPending
EditDataInitializeInProgress
SqlScriptFormatterMultipartDecodeFail
```


kevcunnane:
It seems like an SR-related issue failed during test runs. Could you run tests locally, fix and update the PR?

benrr101:
Yep, the SrTests.cs file in UnitTests/Utility directly references a large swath of the constants defined in SR.cs. If you removed any of those, that unit test will fail. So, the fix would be to remove the constants in question from that test.

PR #302, opened 2017-04-01
Returning EditCell in EditRows

benrr101:
Instead of returning DbCellValues inside an EditRow, we should be returning EditCells. This way we can preserve dirty state when scrolling.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
@@ -130,9 +130,13 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
         public override EditRow GetEditRow(DbCellValue[] cachedRow)
         {
             // Iterate over the new cells. If they are null, generate a blank value
-            DbCellValue[] editCells = newCells.Select(cell => cell == null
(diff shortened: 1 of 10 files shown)


benrr101:
NOTE: This change is backwards compatible with existing implementation b/c EditCell inherits from DbCellValue

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -130,9 +130,13 @@ public override DbCommand GetCommand(DbConnection connection)
         public override EditRow GetEditRow(DbCellValue[] cachedRow)
         {
             // Iterate over the new cells. If they are null, generate a blank value
-            DbCellValue[] editCells = newCells.Select(cell => cell == null
-                    ? new DbCellValue {DisplayValue = string.Empty, IsNull = false, RawObject = null}
-                    : cell.AsDbCellValue)
+            EditCell[] editCells = newCells.Select(cell =>
+                {
+                    DbCellValue dbCell = cell == null
+                        ? new DbCellValue {DisplayValue = string.Empty, IsNull = false, RawObject = null}
```
coquagli: Is it correct for `IsNull = false` here?

benrr101: Yes, if we return null, it indicates that the value is null and should apply the null styling.

PR #304, opened 2017-04-03
Send Error Object on SendError

benrr101:
This change ensures that when calling `requestContext.SendError` you are only able to supply parameters that match the language service beta protocol expected Error object. In other words, you have to provide an error message and optionally and error code.

# **BREAKING API CHANGES**
This will break displaying errors in Microsoft/vscode-mssql. I will be making changes to properly handle the error object shortly.

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
@@ -38,13 +38,14 @@ namespace Microsoft.SqlTools.Hosting.Protocol
                 eventParams);
         }
 
-        public virtual async Task SendError(string errorMessage, int errorCode = 0)
+        public virtual async Task SendError(string errorMessage, int errorCode = 0, object data = null)
         {
             // Build the error message
             ErrorContract error = new ErrorContract
             {
                 Message = errorMessage,
-                Code = errorCode
+                Code = errorCode,
+                Data = data
             };
 
             // Send the message
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
@@ -3,9 +3,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
 using System.Threading.Tasks;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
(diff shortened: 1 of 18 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -898,36 +898,36 @@ internal DefinitionResult GetDefinition(TextDocumentPosition textDocumentPositio
...
-                    Locations = null
-                };
+            if (scriptParseInfo.IsConnected)
+            {
+                //try children tokens first
+                Stack<Token> childrenTokens = selectedToken.Item1;
+                List<Token> tokenList = childrenTokens.ToList();
+                DefinitionResult childrenResult = GetDefinitionFromTokenList(textDocumentPosition, tokenList, scriptParseInfo, scriptFile, connInfo);
+
+                // if the children peak definition returned null then
```
kevcunnane: >peak [](start = 35, length = 4)

Very minor: peak -> peek

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
```diff
@@ -5,6 +5,7 @@
 
 using System.Threading.Tasks;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using ErrorContract = Microsoft.SqlTools.Hosting.Contracts.Error;
```
pensivebrian: >ErrorContract [](start = 6, length = 13)

IMHO, I find this a little confusing since we're special case this one contract type...I prefer we leave it as Error.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
```diff
@@ -37,13 +38,22 @@ public RequestContext(Message requestMessage, MessageWriter messageWriter)
                 eventParams);
         }
 
-        public virtual async Task SendError(object errorDetails)
+        public virtual async Task SendError(string errorMessage, int errorCode = 0, object data = null)
```
pensivebrian: >public virtual async Task SendError(string errorMessage, int errorCode = 0, object data = null) [](start = 8, length = 95)

Can we add an overload that takes an exception? Seems like a common use case. This overload would:
1) Set the message to exception.Message
2) Set the data to the exception.ToString()
3) Maybe, if a SqlException or Win32Exception, set the error code

pensivebrian:
:shipit:

PR #307, opened 2017-04-05
Connect with different properties should actually change context

kevcunnane:
- Up to now, calling Connect for a previously-connected URI would disconnect, then reconnect ot the original (not new) target. WIth these changes we handle changes to database name or other key properties by updating the ConnectionInfo and connecting to the new target
- Some interesting scenarios are raised by our API, notably that an empty database name maps to the default DB (which we know nothing about). This limits the new feature such that only if the DB Name is specified, we'll change the connection. Hence 2 calls to an empty DB will not result in a DB change.

Additional changes:
- After discussion with Ben, we're simplifying the cancellation logic. He had made changes to support this, so the main update is that we dispose the token in the final block after its last use (hence avoiding a disposed exception) and clean up the number of Waits required since we already have async cancellation support
- Factored some logic such that the OnConnection callback isn't invoked until after we've updated the database name in the GetConnectionCompleteParams method. Again, this supports reporting the actual DB name instead of leaving it blank for default DB requests.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
@@ -101,10 +101,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
 
         public bool HasConnectionType(string connectionType)
         {
-            if (connectionType == null)
-            {
-                connectionType = ConnectionType.Default;
-            }
+            connectionType = connectionType ?? ConnectionType.Default;
             return ConnectionTypeToConnectionMap.ContainsKey(connectionType);
         }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
@@ -99,6 +99,15 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             }
         }
 
+        public bool HasConnectionType(string connectionType)
+        {
+            if (connectionType == null)
+            {
+                connectionType = ConnectionType.Default;
+            }
+            return ConnectionTypeToConnectionMap.ContainsKey(connectionType);
+        }
+
         /// <summary>
         /// The count of DbConnectioninstances held by this ConnectionInfo 
         /// </summary>
@@ -155,7 +164,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                 DbConnection connection;
                 ConnectionTypeToConnectionMap.TryRemove(type, out connection);
             }
-        }
-
+        } 
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -385,10 +385,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                     {
                         cancelTupleToCancellationTokenSourceMap.TryRemove(cancelKey, out sourceValue);
                     }
-                    if (source != null)
-                    {
-                        source.Dispose();
-                    }
+                    source?.Dispose();
                 }
             }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -188,22 +188,23 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             // If there is no ConnectionInfo in the map, create a new ConnectionInfo, 
             // but wait until later when we are connected to add it to the map.
             ConnectionInfo connectionInfo;
+            bool connectionChanged = false;
             if (!ownerToConnectionMap.TryGetValue(connectionParams.OwnerUri, out connectionInfo))
             {
                 connectionInfo = new ConnectionInfo(ConnectionFactory, connectionParams.OwnerUri, connectionParams.Connection);
             }
-
-            // Resolve if it is an existing connection
-            // Disconnect active connection if the URI is already connected for this connection type
-            DbConnection existingConnection;
-            if (connectionInfo.TryGetConnection(connectionParams.Type, out existingConnection))
+            else if (IsConnectionChanged(connectionParams, connectionInfo))
             {
-                var disconnectParams = new DisconnectParams()
-                {
-                    OwnerUri = connectionParams.OwnerUri, 
-                    Type = connectionParams.Type
-                };
-                Disconnect(disconnectParams);
+                // We are actively changing the connection information for this connection. We must disconnect
+                // all active connections, since it represents a full context change
+                connectionChanged = true;
+            }
+
+            DisconnectExistingConnectionIfNeeded(connectionParams, connectionInfo, disconnectAll: connectionChanged);
+
+            if (connectionChanged)
+            {
+                connectionInfo = new ConnectionInfo(ConnectionFactory, connectionParams.OwnerUri, connectionParams.Connection);
             }
 
             // Try to open a connection with the given ConnectParams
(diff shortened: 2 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -99,6 +99,15 @@ public ICollection<string> AllConnectionTypes
             }
         }
 
+        public bool HasConnectionType(string connectionType)
+        {
+            if (connectionType == null)
```
benrr101: I'd use the null coalescing operator here, but this works, too

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -99,6 +99,15 @@ public ICollection<string> AllConnectionTypes
             }
         }
 
+        public bool HasConnectionType(string connectionType)
+        {
+            if (connectionType == null)
+            {
+                connectionType = ConnectionType.Default;
+            }
+            return ConnectionTypeToConnectionMap.ContainsKey(connectionType);
```
benrr101: Is this a ConcurrentDictionary we're operating on? If so, I'd highly recommend that getting a specific connection type be done via a TryGet instead of a ContainsKey and Get that way we perform the check and get in an atomic operation.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -188,44 +188,78 @@ public ConnectionCompleteParams ValidateConnectParams(ConnectParams connectionPa
...
-
-            // If this is the first connection for this URI, add the ConnectionInfo to the map
-            if (!ownerToConnectionMap.ContainsKey(connectionParams.OwnerUri))
+            }
+
+            // If this is the first connection for this URI, add the ConnectionInfo to the map
+            bool addToMap = connectionChanged || !ownerToConnectionMap.ContainsKey(connectionParams.OwnerUri);
+            if (addToMap)
             {
                 ownerToConnectionMap[connectionParams.OwnerUri] = connectionInfo;
```
benrr101: This, in particular, isn't safe since the contains key and add to map actions aren't performed in an atomic block. I'd suggest using a try add, since that will not add if the existing item already exists.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -304,46 +338,21 @@ private ConnectionCompleteParams GetConnectionCompleteParams(string connectionTy
...
-                using (source = new CancellationTokenSource())
+                source = new CancellationTokenSource();
+                // Locking here to perform two operations as one atomic operation
+                lock (cancellationTokenSourceLock)
                 {
-                    // Locking here to perform two operations as one atomic operation
-                    lock (cancellationTokenSourceLock)
+                    // If the URI is currently connecting from a different request, cancel it before we try to connect
+                    CancellationTokenSource currentSource;
+                    if (cancelTupleToCancellationTokenSourceMap.TryGetValue(cancelKey, out currentSource))
```
benrr101: I'm pretty sure we could spruce up this code by replacing it with a AddOrUpdate call, but that might be going above and beyond for this code review.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -304,46 +338,21 @@ private ConnectionCompleteParams GetConnectionCompleteParams(string connectionTy
...
-                    });
-
-                    // Open the connection
-                    await Task.WhenAny(openTask, cancellationTask).Unwrap();
-                    source.Cancel();
+                    cancelTupleToCancellationTokenSourceMap[cancelKey] = source;
                 }
+                    
+                // Open the connection
+                await connection.OpenAsync(source.Token);
```
benrr101: 'Bout dang time we started utilizing this properly 👍

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -376,6 +385,10 @@ private ConnectionCompleteParams GetConnectionCompleteParams(string connectionTy
                     {
                         cancelTupleToCancellationTokenSourceMap.TryRemove(cancelKey, out sourceValue);
                     }
+                    if (source != null)
```
benrr101: I *might* be wrong here, but we could tighten this up by using `source?.Dispose()`

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
```
benrr101: AFAIK, this isn't necessary.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -485,6 +486,33 @@ private void SetOptionValue<T>(string name, T value)
             {
                 Options.Add(name, value);
             }
-        }
+        }
+
+        public bool IsComparableTo(ConnectionDetails other)
```
benrr101: Why not implement IComparable<ConnectionDetails> or IEquatable?

kevcunnane: I originally did. However, it's incorrect to say it's the same and this shouldn't be used in dictionary / list handling of the objects. We ignore a number of properties and fudge the treatment of `DatabaseName` due to the fact an empty name could mean any DB due to SQL's default user DB preferences.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
kevcunnane: Thanks - leftover from when I was using a different approach. Will remove

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kevcunnane: 👍

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kevcunnane: Yep I'll leave this and fix the other calls

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kevcunnane: This isn't actually a ConcurrentDictionary right now. You're suggesting we change it to be one?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
kevcunnane: Fixed

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
kevcunnane: So in this case we're not doing a Get. We check if it has the key. If so, we later on do a disconnect (which removes the key) and replace this ConnectionInfo with a new one, thus discarding all results. I'm struggling to see a way to do that without a method like this. Sometimes, you need to check contents of the object.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
benrr101: I think this is definitely something we'd want to make a concurrent dictionary. We definitely have the potential for multiple requests to do things with the connections. However, since that's a bigger change than you'd like in this PR, we can do that in another review (or as part of the rethinking of the connection stuff)

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
benrr101: Yeah, I figured it was something like this. It's not a direct 1:1 comparison of the two objects, so it's not quite accurate to call it IEquatable. And since there's no real ordering that can be done with these, it doesn't make sense to use IComparable. Good to know we're on the same page with this.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
benrr101: So one of the really nice features of ConcurrentDictionary is that it supports having a function called to update an item in the dictionary. So if you're doing something like you say (check for item with given connection type, if it exists disconnect it and add a new one) you can do it pretty neatly with

```
string connType = connectionType ?? ConnectionType.Default;
ConnectionInfo newCi = new ConnectionInfo(); //whatever you do to create one
ConnectionTypeToConnectionMap.AddOrUpdate(connType, newCi, (key, oldCi) => {
  oldCi.Disconnect(); // or whatever else you need to do to it
  return newCi;
});
```

benrr101: And then it's all nicely atomic and you don't need to worry about race conditions.
BUT, if ConnectionTypeToConnectionMap isn't a ConcurrentDictionary, I'll concede on this request for now.

kevcunnane: Yep. We can reassess in the future. This on is a concurrent dictionary, but the entire logic is in a different class (where it should be). We need to take a bigger look at the whole thing. For instance, we'd probably need to add a method here that lets you operate on the map by passing in a function. Then in the calling class refactor to use this approach. For this PR that's too much. But... we need to fix in a separate PR and actually improve the design in the ConnectionService.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kevcunnane: Agreed. I can't see doing it in this PR though, as it'd smash 2 separate tasks together. Let's track and do a proper connection service fix in a future sprint.

PR #308, opened 2017-04-05
Include Internal Row ID with DbCellValue

benrr101:
Update to include the Row ID with a DbCellValue. This will be super useful to us for the purposes of managing individual rows via slick grid.

Although this changes the API for edit/subset and query/subset, it is effectively backwards compatible since it's adding a parameter, not removing anything.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
@@ -30,13 +30,15 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
 
         #region Member Variables
 
+        private delegate FileStreamReadResult ReadMethod(long fileOffset, long rowId, DbColumnWrapper column);
+
         private byte[] buffer;
 
         private readonly QueryExecutionSettings executionSettings;
 
         private readonly Stream fileStream;
 
-        private readonly Dictionary<Type, Func<long, DbColumnWrapper, FileStreamReadResult>> readMethods;
+        private readonly Dictionary<Type, ReadMethod> readMethods;
 
         #endregion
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -183,17 +186,21 @@ private void AssureBufferLength(int newBufferLength)
...
+        /// <param name="rowId">Internal ID of the row to set on all cells in this row</param>
         /// <param name="convertFunc">Function to use to convert the buffer to the target type</param>
         /// <param name="isNullFunc">
         /// If provided, this function will be used to determine if the value is null
         /// </param>
         /// <param name="toStringFunc">Optional function to use to convert the object to a string.</param>
         /// <typeparam name="T">The expected type of the cell. Used to keep the code honest</typeparam>
         /// <returns>The object, a display value, and the length of the value + its length</returns>
-        private FileStreamReadResult ReadCellHelper<T>(long offset, Func<int, T> convertFunc, Func<int, bool> isNullFunc = null, Func<T, string> toStringFunc = null)
+        private FileStreamReadResult ReadCellHelper<T>(long offset, long rowId,
```
coquagli: Minor: If you're going to put parameters on multiple lines, it's probably a good idea to put all all of them on new lines. E.g.: 
```
private FileStreamReadResult ReadCellHelper<T>(
   long offset, 
   long rowId,
   Func<int, T> convertFunc, 
   Func<int, bool> isNullFunc = null,
   Func<T, string> toStringFunc = null)
{		          

```

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -255,13 +255,14 @@ public Task<ResultSetSubset> GetSubset(long startRow, int rowCount)
                     {
                         // Iterate over all the rows and process them into a list of string builders
                         // ReSharper disable once AccessToDisposedClosure   The lambda is used immediately in string.Join call
-                        IEnumerable<string> rowValues = fileOffsets.Select(rowOffset => fileStreamReader.ReadRow(rowOffset, Columns)[0].DisplayValue);
+                        IEnumerable<string> rowValues = fileOffsets.Select(rowOffset => fileStreamReader.ReadRow(rowOffset, 0, Columns)[0].DisplayValue);
```
coquagli: Just want to verify that 0 is the correct thing to be passing here? Seems weird.

benrr101: In this case we really don't care what the rowId is, it is overridden when we put together the new, single DbCellValue for the execution plan.

PR #316, opened 2017-04-14
Add scripting API implemented by the SqlScriptPublishModel

pensivebrian:
Update the ScriptingService to expose new scripting JSON-RPC APIs that use the SqlScriptPublishModel for script generation.

The SqlScriptPublishModel is the model behind the SSMS scripting wizard. To enable scripting for CLI tools, we've ported SqlScriptPublishModel to .NET Core. The SqlScriptPublishModel wraps the SMO scripting APIs for .sql script generation.

- Added three new methods to the ScriptingService: ScriptingRequest, ScriptingListObjectsRequest, ScriptingCancelRequest. 
- Generating scripts are long running operations, so the ScriptingRequest and ScriptingListObjectsRequest kick off a long running scripting task and return immediately. 
- Long running scripting task reports progress and completion, and can be cancelled by a ScriptingCancelRequest request.
- Bumped the SMO nuget package to 140.17049.0. This new version contains a signed SSMS_Rel build of SMO with the SqlScriptPublishModel.
- For testing, adding the Northwind database schema

TODO (in later pull requests)
- Integrate the new ScriptingService APIs with the ConnectionService
- Integrate with the metadata support recently added


MrMeemus:
@MrMeemus is added to the review. #Closed

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.Designer.cs
```diff
@@ -258,47 +258,281 @@ public class sr {
...
+        public static string EditDataSessionNotInitialized {
+            get {
+                return ResourceManager.GetString("EditDataSessionNotInitialized", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///    Looks up a localized string similar to TIME column values must be between 00:00:00.0000000 and 23:59:59.9999999.
+        /// </summary>
+        public static string EditDataTimeOver24Hrs {
```
kevcunnane: >EditDataTimeOver24Hrs [](start = 29, length = 21)

The resx is in sync with this?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.SqlServer.Management.Sdk.Sfc;
+using Microsoft.SqlServer.Management.SqlScriptPublish;
+using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    public static class ScriptingExtensionMethods
```
kevcunnane: Please make internal or document any public classes / methods. Thanks!

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+//
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    public static class ScriptingObjectMatchProcessor
```
kevcunnane: Please make internal or document

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingOperation.cs
```diff
@@ -0,0 +1,21 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    public abstract class ScriptingOperation : IDisposable
```
kevcunnane: Please document this, seems important

kevcunnane:
>         internal static LanguageService LanguageServiceInstance

Minor: we should update this to use the IHostedService concept in the future, instead of having all these static instance references. Otherwise you can get bugs when running tests since they all share this. If it's an IHostedService it doesn't need to have its own .Instance, and all services are given by the ServiceProvider so you get them "for free" and it's easier to test with mocks

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs:65 in 4a2a965. [](commit_id = 4a2a9659cb9c419e912c2260afdfaf14ba62aba3, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -73,13 +79,104 @@ internal static LanguageService LanguageServiceInstance
...
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Await.Warning", "CS4014:Await.Warning", Justification = "Not using await for ScriptingOperation.Execute() since this a long running operation.")]
+        public async Task HandleScriptExecuteRequest(ScriptingParams parameters, RequestContext<ScriptingResult> requestContext)
+        {
+            try
+            {
+                ScriptingOperation operation = new ScriptingScriptOperation(parameters, requestContext);
+                this.ActiveOperations[operation.OperationId] = operation;
+                await requestContext.SendResult(new ScriptingResult { OperationId = operation.OperationId });
+#pragma warning disable 4014
+                operation.Execute().ContinueWith((t) => this.ActiveOperations.TryRemove(operation.OperationId, out operation));
```
kevcunnane: Per offline discussion, our best practice is to a) put this in its own method similar to ConnectionService.RunConnectRequestHandlerTask, and b) kick off this request before doing SendResult (e.g. move this line above `equestContext.SendResult`. Code would look similar to
```
Task.Run(async () => 
            {
                try
                {
                    await operation.Execute();
                    this.ActiveOperations.TryRemove(operation.OperationId, out operation);
                }
                catch (Exception ex)
                {
                    // either send event with failure status, or call SendError if these fail
                    await ServiceHost.SendEvent(ConnectionCompleteNotification.Type, result);
                }
            });
```

Thanks!

kevcunnane: For others reference: If you kick off a task and don't await it or check the IsFaulted value in the continuation method, any exception will bubble out of the loop and take down the service. That's why we wrap everything in a try/catch loop or await. I think that ContinueWith still requires checking faulted state but that may have been fixed in newer .Net versions

---
In reply to: [111651714](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111651714) [](ancestors = 111651714)

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingErrorEvent.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public string Message { get; set; }
+
+        public string DiagnosticMessage { get; set; }
+    }
+
+    /// <summary>
+    /// Event sent to indicate a scripting database operation has encountered an error.
+    /// </summary>
+    public class ScriptingErrorEvent
```
kevcunnane: Please check with Ben, this is a nice practice / looks good but I know he's been doing work in the error handling area and it'd be great if we align this so it's consistent.

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingEventParams.cs
```diff
@@ -0,0 +1,12 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting.Contracts
+{
+    public abstract class ScriptingEventParams
```
kevcunnane: Please doc

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingListObjectsEvent.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting.Contracts
+{
+    /// <summary>
+    /// Parameters to be sent to a indicate a list objects operation has completed.
+    /// </summary>
+    public class ScriptingListObjectsCompleteParameters : ScriptingEventParams
+    {
+        public List<ScriptingObject> DatabaseObjects { get; set; }
```
kevcunnane: Please doc

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingObject.cs
```diff
@@ -0,0 +1,92 @@
...
+        ///     SqlAssembly,
+        ///     DdlTrigger,
+        ///     Synonym,
+        ///     XmlSchemaCollection,
+        ///     Schema,
+        ///     PlanGuide,
+        ///     UserDefinedType,
+        ///     UserDefinedAggregate,
+        ///     FullTextCatalog,
+        ///     UserDefinedTableType
```
kevcunnane: Will there be more in the future? We have a bunch of other types in the ObjectExplorer tree, can they all be added in here / should they be?

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+{
+    /// <summary>
+    /// Database scripting options
+    /// </summary>
+    public class ScriptOptions
+    {
+        /// <summary>
+        /// Generate ANSI padding statements
+        /// </summary>
+        public string ScriptAnsiPadding { get; set; }
```
kevcunnane: Aren't these bools? Why string?

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingProgressNotificationEvent.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting.Contracts
+{
+    /// <summary>
+    /// Parameters to indicate the script operation has resolve the objects to be scripted.
+    /// </summary>
+    public class ScriptingProgressNotificationParams : ScriptingEventParams
+    {
+        public ScriptingObject ScriptingObject { get; set; }
```
kevcunnane: Please doc

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingRequest.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting.Contracts
+{
+    /// <summary>
+    /// Parameters for a script request
+    /// </summary>
+    public class ScriptingParams
+    {
+        public string FilePath { get; set; }
```
kevcunnane: Please doc

kevcunnane:
@llali is added to the review. #Closed

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -0,0 +1,132 @@
...
+                throw new ArgumentException("Property scriptingObject.Name must have a value", "scriptingObject");
+            }
+
+            if (string.IsNullOrWhiteSpace(scriptingObject.Type))
+            {
+                throw new ArgumentException("Property scriptingObject.Type must have a value", "scriptingObject");
+            }
+
+            string urn = string.Format(
+                "Server[@Name='{0}']/Database[@Name='{1}']/{2}[@Name='{3}' {4}]",
```
kevcunnane: >Server[@Name='{0}']/Database[@Name='{1}']/{2}[@Name='{3}' {4}]", [](start = 17, length = 64)

@llali - it's worth looking at this for ObjectExplorer. This is how you use URN instead of the direct SMO object model. We should evaluate this in the future for PERF reasons, particularly as we implement filtering since this has good filter support. I'd be interested to see if this has significantly better perf in the long-run vs. using the object model. Note that Brian has done the hard work of adding InternalsVisibleTo for our project to the SMO DLLs that needed updating :)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -0,0 +1,132 @@
...
+                scriptingObject.Type,
+                scriptingObject.Name,
+                scriptingObject.Schema != null ? string.Format("and @Schema = '{0}'", scriptingObject.Schema) : string.Empty);
+
+            return new Urn(urn);
+        }
+
+        public static ScriptingObject ToScriptingObject(this Urn urn)
+        {
+            if (urn == null)
```
kevcunnane: >urn [](start = 16, length = 3)

Please use Validate.IsNotNull(nameof(urn), urn) as this is our standard pattern elsewhere

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        public static Urn ToUrn(this ScriptingObject scriptingObject, string server, string database)
+        {
+            if (scriptingObject == null)
+            {
+                throw new ArgumentNullException("scriptingObjectc");
+            }
+
+            if (string.IsNullOrWhiteSpace(server))
```
kevcunnane: Please use our Validate class for all these checks. If you need custom error messages and expect them to be user-visible, we should a) extend that class to accept custom error messages, and b) add those as localizable strings as all user-visible strings should be localizable

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingOptions.cs
pensivebrian: These can be null if they aren't specified in the request. I can explore making them Nullable<bool>, I just wasn't sure if the JSON-RPC layer handles this, but I'll look into it.

---
In reply to: [111652013](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111652013) [](ancestors = 111652013)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        private int totalScriptedObjectCount = 0;
+
+        private ScriptingListObjectsParams Parameters { get; set; }
+
+        private RequestContext<ScriptingListObjectsResult> RequestContext { get; set; }
+
+        public ScriptingListObjectsOperation(ScriptingListObjectsParams parameters, RequestContext<ScriptingListObjectsResult> requestContext)
+        {
+            if (parameters == null)
```
kevcunnane: Please use the Validate class for all these

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
```diff
@@ -0,0 +1,219 @@
...
+                    OperationId = this.OperationId,
+                    DatabaseObjects = databaseObjects,
+                    Count = databaseObjects.Count,
+                };
+
+                this.SendJsonRpcEventAsync(ScriptingListObjectsCompleteEvent.Type, eventParameters);
+            }
+            catch (Exception e)
+            {
+                if (e.IsOperationCanceledException())
```
kevcunnane: Minor: Normally this would be its own catch(OperationCanceledException) clause above the general catch method?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
```diff
@@ -0,0 +1,219 @@
...
+                e.ContinueScripting = false;
+                throw new OperationCanceledException(this.cancellation.Token);
+            }
+
+            if (e.Completed)
+            {
+                this.scriptedObjectCount += 1;
+            }
+
+            // TODO: Handle the e.Error case.
```
kevcunnane: >. [](start = 33, length = 1)

Do we have an issue tracking this?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        public static IEnumerable<ScriptingObject> Match(
+            ScriptingObject includeCriteria,
+            ScriptingObject excludeCriteria,
+            IEnumerable<ScriptingObject> candidates)
+        {
+            return Match(
+                includeCriteria == null ? new ScriptingObject[0] : new[] { includeCriteria },
+                excludeCriteria == null ? new ScriptingObject[0] : new[] { excludeCriteria },
+                candidates);
```
kevcunnane: >candidates [](start = 16, length = 10)

Please use Validate class here

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            IEnumerable<ScriptingObject> matchedObjects = new List<ScriptingObject>();
+
+            if (includeCriteria != null && includeCriteria.Count() > 0)
+            {
+                foreach (ScriptingObject scriptingObjectCriteria in includeCriteria)
+                {
+                    IEnumerable<ScriptingObject> matches = MatchCriteria(scriptingObjectCriteria, candidates);
+                    matchedObjects = matchedObjects.Union(matches);
```
kevcunnane: >matchedObjects [](start = 20, length = 14)

This results in multiple enumerations if it's not a list. A nicer way would be to declare `matchedObjects = candidates;` first, then just have the foreach loop with no `Count() > 0` check

kevcunnane: Actually, I see why this is done the way it is. I think the right thing is to check .Any(), which only accesses the 1st element.

---
In reply to: [111654045](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111654045) [](ancestors = 111654045)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+            IEnumerable<ScriptingObject> matchedObjects = databaseObjects;
+
+            if (!string.IsNullOrWhiteSpace(criteria.Type))
+            {
+                matchedObjects = matchedObjects.Where(o => string.Equals(criteria.Type, o.Type, StringComparison.OrdinalIgnoreCase));
+            }
+
+            if (!string.IsNullOrWhiteSpace(criteria.Schema))
+            {
+                if (criteria.Schema.Equals("*", StringComparison.OrdinalIgnoreCase))
```
kevcunnane: This "*" should be a constant with a clear name such as WILDCARD

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            if (!string.IsNullOrWhiteSpace(criteria.Name))
+            {
+                if (criteria.Name.Equals("*", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Don't filter any objects
+                }
+                else if (criteria.Name.EndsWith("*", StringComparison.OrdinalIgnoreCase))
+                {
+                    matchedObjects = matchedObjects.Where(o => o.Name.StartsWith(
```
kevcunnane: >matchedObjects [](start = 20, length = 14)

This is repeated logic. It would be nice to move into a common method.

kevcunnane:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
pensivebrian: Fixed.

---
In reply to: [111650593](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111650593) [](ancestors = 111650593)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
pensivebrian: Fixed

---
In reply to: [111650616](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111650616) [](ancestors = 111650616)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
pensivebrian: I adopted the pattern you mentioned so the scripting service methods should now be consistent.

---
In reply to: [111651826](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111651826) [](ancestors = 111651826,111651714)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
pensivebrian: This extension method checks inner exceptions - SMO wraps all exceptions in it's own exception class.

---
In reply to: [111653257](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111653257) [](ancestors = 111653257)

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.Designer.cs
pensivebrian: I've sync'd the latest code from dev. It seems the checkin run srgen and include the latest sr.Designer.cs

---
In reply to: [111650561](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111650561) [](ancestors = 111650561)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
pensivebrian: I updated all null checks to use the Validate class. I think the default messages should be fine.

---
In reply to: [111653103](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111653103) [](ancestors = 111653103)

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingObject.cs
pensivebrian: These are dynamically discovered by SMO, so this list may change. I'll add comment about this.

---
In reply to: [111651987](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111651987) [](ancestors = 111651987)

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingErrorEvent.cs
pensivebrian: Will do.

---
In reply to: [111651923](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111651923) [](ancestors = 111651923)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
pensivebrian: [https://github.com/Microsoft/sql-xplat-cli/issues/42](https://github.com/Microsoft/sql-xplat-cli/issues/42)

---
In reply to: [111653975](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111653975) [](ancestors = 111653975)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
pensivebrian: Yep, good idea.

---
In reply to: [111654162](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r111654162) [](ancestors = 111654162)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -73,13 +79,143 @@ internal static LanguageService LanguageServiceInstance
...
+                    await operation.Execute();
+                }
+                catch (Exception e)
+                {
+                    await operation.RequestContext.SendError(e);
+                }
+                finally
+                {
+                    ScriptingOperation temp;
+                    this.ActiveOperations.TryRemove(operation.OperationId, out temp);
```
benrr101: Doesn't this just remove the operation every time you start it up, regardless of if it fails?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -73,13 +79,143 @@ internal static LanguageService LanguageServiceInstance
...
+                }
+            });
+        }
+
+        /// <summary>
+        /// Runs the async task to script objects.
+        /// </summary>
+        private void RunScriptTask(ScriptingScriptOperation operation)
+        {
+            Task.Run(async () =>
```
benrr101: I'm not quite sure the benefit of wrapping this in a Task.Run... If all the operation.Executes don't await anything, then there's no need to make .Execute async. Then you don't need to wrap the startup operation in a Task.

benrr101: But I'm also curious since this pattern is seemingly reused several times, why isn't this method used more than once? Or more concisely, what makes this private method different than the other implementations of the same pattern.

pensivebrian: You're right, the await is not needed. I removed the awaits and made Execute synchronous.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -73,13 +79,143 @@ internal static LanguageService LanguageServiceInstance
...
+        /// <summary>
+        /// Runs the async task to list scriptable objects.
+        /// </summary>
+        private void RunListObjectsTask(ScriptingListObjectsOperation operation)
+        {
+            Task.Run(async () =>
+            {
+                try
+                {
+                    this.ActiveOperations[operation.OperationId] = operation;
```
pensivebrian: A new operation is instantiated per request, and each has a unique operation id so this should never happen. I added debug assert to make this more explicit.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
pensivebrian: Correct.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
pensivebrian: Good point. I refactored to single method reused by the request handlers.

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingCancelRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// Parameters to cancel a scripting request.
+    /// </summary>
+    public class ScriptingCancelParams
+    {
+        /// <summary>
+        /// Gets or sets the operation id of the scripting operation to cancel.
+        /// </summary>
+        public string OperationId { get; set; }
```
pensivebrian: Yep, nice catch.

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingObject.cs
```diff
@@ -0,0 +1,93 @@
...
+        public override string ToString()
+        {
+            string objectName = string.IsNullOrEmpty(this.Schema)
+                ? this.Name
+                : this.Schema + "." + this.Name;
+
+            return objectName;
+        }
+
+        public override int GetHashCode()
```
pensivebrian: Nope, I don't use ReSharper :)

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingCancelEvent.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Parameters sent when a scripting operation has been canceled.
+    /// </summary>
+    public class ScriptingCancelParameters : ScriptingEventParams { }
+
+    /// <summary>
+    /// Event sent to indicate that a scripting operation has been canceled.
+    /// </summary>
+    public class ScriptingCancelEvent
```
pensivebrian: Makes sense to combine completed and cancel if that's the pattern used by services.

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingRequest.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the file path used when writing out the script.
+        /// </summary>
+        public string FilePath { get; set; }
+
+        /// <summary>
+        /// Gets or sets connection string of the target database the scripting operation will run against.
+        /// </summary>
+        public string ConnectionString { get; set; }
```
pensivebrian: I'm going to leave it as is since it's only defined in two request types. If we start seeing it in more request types, I'll make this change.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -0,0 +1,116 @@
...
+
+                Logger.Write(
+                    LogLevel.Verbose,
+                    string.Format(
+                        "Loaded SMO urn object count {0} for type {1}, urns: {2}",
+                        objectType,
+                        databaseObjectsOfType.Count(),
+                        string.Join(", ", databaseObjectsOfType.Select(p => p.Value))));
+
+                foreach (KeyValuePair<string, string> databaseObjectOfType in databaseObjectsOfType)
```
pensivebrian: Fixed.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+
+            return databaseObjects;
+        }
+
+        /// <summary>
+        /// Returns true if the passed exception or any inner exception is an OperationCanceledException instance.
+        /// </summary>
+        public static bool IsOperationCanceledException(this Exception e)
```
pensivebrian: Moved.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -0,0 +1,326 @@
...
+
+                publishModel = BuildPublishModel();
+                publishModel.ScriptItemsCollected += this.OnPublishModelScriptItemsCollected;
+                publishModel.ScriptProgress += this.OnPublishModelScriptProgress;
+                publishModel.ScriptError += this.OnPublishModelScriptError;
+
+                ScriptOutputOptions outputOptions = new ScriptOutputOptions
+                {
+                    SaveFileMode = ScriptFileMode.Overwrite,
+                    SaveFileType = ScriptFileType.Unicode,
```
pensivebrian: This is UTF-16.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -0,0 +1,326 @@
...
+                }
+            }
+        }
+
+        private SqlScriptPublishModel BuildPublishModel()
+        {
+            SqlScriptPublishModel publishModel = new SqlScriptPublishModel(this.Parameters.ConnectionString);
+            PopulateAdvancedScriptOptions(publishModel.AdvancedOptions);
+
+            bool hasIncludeCriteria = this.Parameters.IncludeObjectCriteria != null && this.Parameters.IncludeObjectCriteria.Any();
```
pensivebrian: I think it's readable the way it is and I'm okay with multiple null/Any calls on the collection. The key optimization begin made is avoiding the call to ScriptingObjectMatchProcessor.Match if possible since it needs to load all of the objects from the database. I added a comment to make that more explicit.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,131 @@
...
+            Validate.IsNotNull("candidates", candidates);
+
+            IEnumerable<ScriptingObject> matchedObjects = new List<ScriptingObject>();
+
+            if (includeCriteria != null && includeCriteria.Any())
+            {
+                foreach (ScriptingObject scriptingObjectCriteria in includeCriteria)
+                {
+                    IEnumerable<ScriptingObject> matches = MatchCriteria(scriptingObjectCriteria, candidates);
+                    matchedObjects = matchedObjects.Union(matches);
```
pensivebrian: No craziness for me. To be frank, I don't even know how I would accomplish that with an aggregate.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,131 @@
...
+                }
+                if (property.EndsWith(Wildcard, StringComparison.OrdinalIgnoreCase))
+                {
+                    matchedObjects = candidates.Where(o => propertySelector(o).StartsWith(
+                        propertySelector(o).Substring(0, propertySelector(o).Length - 1),
+                        StringComparison.CurrentCultureIgnoreCase));
+                }
+                else
+                {
+                    matchedObjects = matchedObjects.Where(o => string.Equals(property, propertySelector(o), StringComparison.CurrentCultureIgnoreCase));
```
pensivebrian: I was wonder this myself. I think Invariant more correct. Good call.

pensivebrian: Looks like .net core doesn't have invariant, so I'll go with ordinal.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        private int scriptedObjectCount = 0;
+
+        private int totalScriptedObjectCount = 0;
+
+        private ScriptingListObjectsParams Parameters { get; set; }
+
+        public RequestContext<ScriptingListObjectsResult> RequestContext { get; private set; }
+
+        public ScriptingListObjectsOperation(ScriptingListObjectsParams parameters, RequestContext<ScriptingListObjectsResult> requestContext)
```
pensivebrian: I'll use the event pattern like the example given.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
```diff
@@ -0,0 +1,212 @@
...
+                    Count = databaseObjects.Count,
+                };
+
+                this.SendJsonRpcEventAsync(ScriptingListObjectsCompleteEvent.Type, eventParameters);
+            }
+            catch (Exception e)
+            {
+                if (e.IsOperationCanceledException())
+                {
+                    this.SendJsonRpcEventAsync(
```
pensivebrian: Yep.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingListObjectsOperation.cs
```diff
@@ -0,0 +1,212 @@
...
+                ScriptingObject = e.Urn.ToScriptingObject(),
+                Status = e.Completed ? "Completed" : "Progress",
+                CompletedCount = this.scriptedObjectCount,
+                TotalCount = this.totalScriptedObjectCount,
+            };
+
+            this.SendJsonRpcEventAsync(ScriptingProgressNotificationEvent.Type, eventParams);
+        }
+
+        public override void Cancel()
```
pensivebrian: Fixed.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -0,0 +1,326 @@
...
+
+        private bool disposed = false;
+
+        private int scriptedObjectCount = 0;
+
+        private int totalScriptedObjectCount = 0;
+
+        private ScriptingParams Parameters { get; set; }
+
+        public RequestContext<ScriptingResult> RequestContext { get; private set; }
```
pensivebrian: Fixed.

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        /// <summary>
+        /// Continue to script if an error occurs. Otherwise, stop.
+        /// </summary>
+        public string ContinueScriptingOnError { get; set; }
+
+        /// <summary>
+        /// Convert user-defined data types to base types.
+        /// </summary>
+        public string ConvertUDDTToBaseType { get; set; }
```
pensivebrian: I'm using the same names used by SMO. I'm my mind, this makes it easier to reason over all of options. I'm going to leave as-is for now unless folks feel strongly.

---
In reply to: [112277263](https://github.com/Microsoft/sqltoolsservice/pull/316#discussion_r112277263) [](ancestors = 112277263)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatchProcessor.cs
```diff
@@ -0,0 +1,131 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    /// <summary>
+    /// Implementes the matchin logic to filter scripting objects based on
```
kevcunnane: Minor typos: `Implements the matching logic`

PR #317, opened 2017-04-17
OpenCover vBump

benrr101:
This pull request is mostly for test purposes. OpenCover has given us a new version that claims to resolve our random failure issues. I'll run the build 10 or so times to validate that it's working.

Note that this solution is temporary. This solution has their test nuget package directly checked in. Ideally, when they officially release the nuget package, we can replace that with the official package.


abist:
The coverage already looks good, hopefully it stays like this.

benrr101:
I wanted to run the appveyor build 10x to validate that it's always working, but I think it's good enough to at least preserve the existing build quality.

PR #318, opened 2017-04-18
fixed bug with batch separators

abist:
Fixes [637](https://github.com/Microsoft/carbon/issues/637).

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/BatchParser/BatchParserWrapper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/BatchParser/BatchParserWrapper.cs
@@ -140,47 +140,64 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
             int count = 0;
             int offset = 0;
             bool foundAllOffsets = false;
-            using (StringReader reader = new StringReader(content)) {
+            int maxStartLine = GetMaxStartLine(positions);
+            using (StringReader reader = new StringReader(content))
+            {
+                // go until we have found offsets for all batches
                 while (!foundAllOffsets)
-                 {
-                    for (int i = 0; i <= GetMaxStartLine(positions); i++)
+                {
+                    // go until the last start line of the batches
+                    for (int i = 0; i <= maxStartLine ; i++)
                     {
-                        int ch;
-                        while (true)
-                        {
-                            if (positions[count].Item1 == i) {
-                                count++;
-                                offsets.Add(offset);
-                                if (count == positions.Count) 
-                                {
-                                    foundAllOffsets = true;
-                                    break;
-                                }
-                            }
-                            ch = reader.Read();
-                            if (ch == -1) // EOF do nothing
-                            {
-                                break;
-                            }
-                            else if (ch == 10 /* for \n */) // End of line increase and break
-                            {
-                                offset++;
-                                break;
-                            }
-                            else // regular char just increase
-                            {
-                                offset++;
-                            }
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/BatchParser/BatchParserWrapper.cs
```diff
@@ -111,6 +117,67 @@ public sealed class BatchParserWrapper : IDisposable
             return batchDefinitionList;
         }
 
+        private static int GetMaxStartLine(IList<Tuple<int, int>> positions) {
```
kburtram: '{' should be on a separate line by itself in C#... please fix all instances in this PR

src/Microsoft.SqlTools.ServiceLayer/BatchParser/BatchParserWrapper.cs
```diff
@@ -111,6 +117,73 @@ public sealed class BatchParserWrapper : IDisposable
...
+        /// Gets offsets for all batches
+        /// </summary>
+        private static List<int> GetOffsets(string content, IList<Tuple<int, int>> positions) 
+        {
+
+            List<int> offsets = new List<int>();
+            int count = 0;
+            int offset = 0;
+            bool foundAllOffsets = false;
+            using (StringReader reader = new StringReader(content)) {
```
kburtram: there are still a couple '{' misplaced here and on line 151 for example.  Also '{' on 145 is misaligned.  Please double check formatting for this change.

src/Microsoft.SqlTools.ServiceLayer/BatchParser/BatchParserWrapper.cs
```diff
@@ -111,6 +117,73 @@ public sealed class BatchParserWrapper : IDisposable
...
+                while (!foundAllOffsets)
+                 {
+                    for (int i = 0; i <= GetMaxStartLine(positions); i++)
+                    {
+                        int ch;
+                        while (true)
+                        {
+                            if (positions[count].Item1 == i) {
```
kburtram: personally I think this function is a bit too complicated to easily understand.  For example, there are 3 nested loops, one of the loops is "infinite" condition, the inner loop has multiple break statements, various counters are being incremented conditionally, and there's a lack of comments explaining what's going on.

I'd recommend to try to split this method into a few smaller methods, clean up the loops to avoid the complicated branching, add detailed comments, and add sufficient test cases to make sure this method has near-100% code coverage.

src/Microsoft.SqlTools.ServiceLayer/BatchParser/BatchParserWrapper.cs
```diff
@@ -111,6 +117,73 @@ public sealed class BatchParserWrapper : IDisposable
...
+        {
+
+            List<int> offsets = new List<int>();
+            int count = 0;
+            int offset = 0;
+            bool foundAllOffsets = false;
+            using (StringReader reader = new StringReader(content)) {
+                while (!foundAllOffsets)
+                 {
+                    for (int i = 0; i <= GetMaxStartLine(positions); i++)
```
kburtram: GetMaxStartLine is being called within a loop with the same parameter.  Should you call it once outside the loop and cache the return value?

abist: Ah, that's a good call. Not sure how I missed this.

PR #322, opened 2017-04-19
Fix the OE service where returns database as the root node for database connection

AbbiePetcht:


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -26,12 +26,26 @@ public class ObjectExplorerServiceTests
...
+            using (SqlTestDb testDb = SqlTestDb.CreateNew(TestServerType.OnPrem, false, databaseName, query, uri))
+            {
+                var session = await CreateSession(null, uri);
+                await CreateSessionAndServerNode(testDb.DatabaseName, session);
+                CancelConnection(uri);
+            }
+        }
+
+        [Fact]
+        public async void CreateSessionAndExpandOnTheDatabaseShouldReturnDatabaseObjects()
```
llali: none of these tests pass master or empty string as database name. can you please add a test to verify that. the code coverage is dropped in this pr so it might be a bug or missing tests

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -78,7 +78,13 @@ public ObjectExplorerServiceTests()
         public async Task CreateSessionRequestReturnsSuccessAndNodeInfo()
```
llali: please also add test for when the database is specified or empty

llali:
>                 Server server = ServerCreator.Create(connection);

I'm pretty sure this also has to change for database instance. the query context for database and setting the parent

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/ServerNode.cs:168 in a9acbee. [](commit_id = a9acbee361dfb09a1856b01e825e75201fb2ffaa, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/ServerNode.cs
```diff
@@ -47,10 +47,19 @@ public ServerNode(ConnectionCompleteParams connInfo, IMultiServiceProvider servi
 
             this.context = new Lazy<SmoQueryContext>(() => CreateContext(serviceProvider));
 
-            NodeValue = connectionSummary.ServerName;
+            if (IsMasterConnection())
```
llali: I'm pretty sure there's something wrong with this approach. this root code is not really database instance node and there are some logic in the database node needed for expanding and scripting. The context for server node and database node are different and database has a smo object because it is scriptable. the tests should verify that

llali:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/ServerNode.cs
```diff
@@ -70,6 +79,12 @@ internal SmoServerCreator ServerCreator
             }
         }
 
+        internal bool IsMasterConnection()
```
kburtram: we probably should check if the use is connecting to any of the system tables. e.g., master, tempdb, msdb or model.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerUtils.cs
```diff
@@ -71,6 +72,15 @@ public static TreeNode FindNode(TreeNode node, Predicate<TreeNode> condition, Pr
                 }
             }
             return null;
+        }
+
+        public static bool IsSystemDatabaseConnection(string databaseName)
```
llali: please add comment for public methods

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -346,8 +346,19 @@ public ObjectExplorerSession(string uri, TreeNode root, IMultiServiceProvider se
...
             public static ObjectExplorerSession CreateSession(ConnectionCompleteParams response, IMultiServiceProvider serviceProvider)
             {
-                TreeNode serverNode = new ServerNode(response, serviceProvider);
-                return new ObjectExplorerSession(response.OwnerUri, serverNode, serviceProvider, serviceProvider.GetService<ConnectionService>());
+                TreeNode rootNode = new ServerNode(response, serviceProvider);
+                var session = new ObjectExplorerSession(response.OwnerUri, rootNode, serviceProvider, serviceProvider.GetService<ConnectionService>());
+                if (!ObjectExplorerUtils.IsSystemDatabaseConnection(response.ConnectionSummary.DatabaseName))
+                {
+                    var children = rootNode.Expand();
+                    var databasesRoot = children.FirstOrDefault(x => x.NodeTypeId == NodeTypes.Databases);
```
llali: this code assuming the databases are in a folder under server. if the structure change the code will break. you might want to traverse the tree to find the database node

AbbiePetcht: Will write the comment on assumption. If we change the structure, it will break the OE tests

---
In reply to: [112340512](https://github.com/Microsoft/sqltoolsservice/pull/322#discussion_r112340512) [](ancestors = 112340512)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ObjectExplorerServiceTests.cs
AbbiePetcht: Already added it in integration test

---
In reply to: [112104491](https://github.com/Microsoft/sqltoolsservice/pull/322#discussion_r112104491) [](ancestors = 112104491)

Raymondd:
Appveyor says the build passed, but it looks like a test is still failing: https://ci.appveyor.com/project/kburtram/sqltoolsservice/build/1.0.911/tests

PR #323, opened 2017-04-19
Fixed the issue with nodeType set to null for many sql objects in object explorer 

llali:
- Set the node type to the node name from the xml instead of null. 
- Forgot to remove "instance" from so many places in previous PR so did that in this PR


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeGenerator.tt
```diff
@@ -179,6 +179,12 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 				{
 					WriteLine("            var child = new SmoTreeNode();");
 					WriteLine("            child.IsAlwaysLeaf = true;");
+					if(type.EndsWith("s"))
```
kburtram: are we sure that none of the types well need to support will end with 's'?  Also, it looks like we're missing a space between f and ( in the "if (" statement.

llali: this logic is for the node types that don't have specific class for creating node, Folders, tables, views they have specific class but column, triggers ...  don't have it

---
In reply to: [112316790](https://github.com/Microsoft/sqltoolsservice/pull/323#discussion_r112316790) [](ancestors = 112316790)

PR #326, opened 2017-04-20
Fix: Removing race condition from treeNodeGenerator test

Raymondd:
A race condition was causing this test to occasionally fail due to another asynchronous test that changed locale. Using SR instead of the hardcoded constant should prevent this race condition.


llali:
the change looks good but why the code coverage is %10 lower? can you make sure nothing else is broken?

Raymondd:
looks like a code coverage bug. None of the file I changed are impacting coverage.

PR #328, opened 2017-04-25
Implemented the oe filtering

llali:
- Filtering SQL object based on the server version it's valid for
- Filtering SQL objects based on the filters assigned to each type 

 TODO:
 
- Add more tests
- Fix the filters for functions, system views, system data types


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/SqlServerType.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// Creates a server type from the server version
+        /// </summary>
+        public static SqlServerType CalculateServerType(ServerInfo serverInfo)
+        {
+            SqlServerType serverType = SqlServerType.Unknown;
+            string serverVersion = serverInfo.ServerVersion;
+
+            if (serverInfo.IsCloud)
+            {
+                if (serverVersion.StartsWith("11", StringComparison.Ordinal))
```
kevcunnane: Good news: Azure v11 is no more! We only have to handle AzureV12 now. This will simplify the logic somewhat. In theory there might be a future version of Azure that's different enough to re-introduce this logic, but I don't see that happening for a while.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -1,133 +1,169 @@
...
+    public class SmoChildFactoryBase : ChildFactory
+    {
+        public override IEnumerable<string> ApplicableParents()
+        {
+            return null;
+        }
+        
+        public override IEnumerable<TreeNode> Expand(TreeNode parent)
+        {
+            //parent.BeginChildrenInit();
```
kevcunnane: We should probably remove the commented out parts. My memory is it was SSDT presentation-related so isn't vital for us.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -1,133 +1,169 @@
...
+                }
+            }
+
+            return filterTheNode;
+        }
+
+        private string GetProperyFilter(IEnumerable<NodeFilter> filters, Type querierType, ValidForFlag validForFlag)
+        {
+            string filter = "";
+            if (filters != null)
```
kevcunnane: minor: please use `string.Empty`.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.cs
```diff
@@ -19,15 +21,33 @@ internal partial class SqlDatabaseQuerier: SmoQuerier
...
             if (parentServer != null)
             {
                 var retValue = parentServer.Databases;
-                if(retValue != null)
+                bool hasFilter = !string.IsNullOrEmpty(filter);
+                HashSet<string> urns = null;
+                if (hasFilter)
                 {
-                    return new SmoCollectionWrapper<Database>(retValue);
+                    string urn = $"{parentServer.Urn.ToString()}/Database" + filter;
```
kevcunnane: This is a great change. Nice job using SFC to get efficient querying while still returning usable SMO objects so we can cleanly manage the code later on.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -120,154 +138,212 @@ private void ExpandAndVerifyDatabaseNode(string databaseName, ObjectExplorerSess
...
                 var databaseNodeInfo = await ExpandServerNodeAndVerifyDatabaseHierachy(testDb.DatabaseName, session);
-                await ExpandTree(databaseNodeInfo, session);
-                CancelConnection(uri);
+                await FindNodeByLabel(databaseNodeInfo, session, SR.SchemaHierarchy_Tables);
             }
+            CancelConnection(uri);
+
         }
 
         //[Fact]
```
kevcunnane: Will these tests be enabled in the future?

llali: Yes I'll enable them when I can figure out why they are crashing when running all together. I'll address your comments in a new PR.

PR #329, opened 2017-04-26
Initial admin and DR services.

kburtram:
This is an in-progress merge to share early code with others in feature area.  There are additional iterations planned to complete out this scenario.


kburtram:
I'm going to add some tests to avoid impacting code coverage from this PR.

src/Microsoft.SqlTools.ServiceLayer/TaskServices/Contracts/TaskInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.SqlTools.ServiceLayer.TaskServices.Contracts
+{
+    public enum TaskState
+    {
+        NotStarted = 0,
+        Running = 1,
+        Complete = 2
```
llali: what about failed?

kburtram: This PR is just setting up placeholder for next step of work.  I think there are other completed states like failed, aborted, etc. that we will need to communicate.  We can refactor when we do the actual task management implementation in the next round of PRs.

PR #330, opened 2017-04-26
fix peek definition integration tests

abist:
Changed test connection method so that it doesn't use the same URI every time for every connection.


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/PeekDefinitionTests.cs
```diff
@@ -698,113 +698,122 @@ public void GetDatabaseWithQueryConnectionTest()
...
+            connInfo.RemoveAllConnections();
+            var bindingQueue = new ConnectedBindingQueue();
+            bindingQueue.AddConnectionContext(connInfo);
+            scriptFile.Contents = queryString;
+
+            var service = new LanguageService();
+            service.RemoveScriptParseInfo(OwnerUri);
+            service.BindingQueue = bindingQueue;
+            await service.UpdateLanguageServiceOnConnection(connectionResult.ConnectionInfo);
+            Thread.Sleep(2000);
```
kburtram: can we do anything better than sleeping?  this approach seems to cause intermittent failures when things timeout.

abist: I followed some other tests that did the same thing. I assumed this is the way to go. The failures were caused by reusing the same ownerURI for every test that uses live connection. I think the new changes should fix that problem.

PR #331, opened 2017-04-27
fixed db trigger, system types and oe tests

llali:
- Fixed the system types folders
- Fixed database trigger 
- Added a simple test for OE to verify all object types so hopefully will be faster to CI builds


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodeFilter.cs
```diff
@@ -76,6 +78,22 @@ public string ToPropertyFilterString()
...
+
+        public static string ConcatProperties(IEnumerable<NodeFilter> filters)
+        {
+            string filter = "";
+            var list = filters.ToList();
+            for (int i = 0; i < list.Count; i++)
+            {
+                var value = list[i];
+                
+                string orPrefix = i == 0 ? "" : "and";
```
kburtram: should this be named andPrefix?  (also, as a minor nit, I typically use string.Empty instead of "" literal for readability).

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestData/ObjectExplorer/Baselines/AllSqlObjects.txt
```diff
@@ -0,0 +1,136 @@
+NodeType: Table Label: HumanResources.Employee
```
kburtram: I wonder if we should stop using AdventureWorks for testing, and perhaps create a simpler database schema with 1 single item for each object type (i.e. 1 table, 1 view, 1 sproc, etc.)...not related to this PR, but it may help with some of the test duration issues we're seeing.

llali: That's exactly what I did here, and I think OE is the only that's using adventure works

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodeFilter.cs
llali: I'll fix this in next PR

PR #333, opened 2017-04-29
Changes to sqltoolsservice to allow empty password for SqlLogin

huypha:
Changes to allow empty passwords. This is paired with a PR to the vscode-mssql extension.


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectParamsExtensions.cs
```diff
@@ -30,15 +30,11 @@ public static bool IsValid(this ConnectParams parameters, out string errorMessag
             }
             else if (string.IsNullOrEmpty(parameters.Connection.AuthenticationType) || parameters.Connection.AuthenticationType == "SqlLogin")
             {
-                // For SqlLogin, username/password cannot be empty
+                // For SqlLogin, username cannot be empty
```
llali: you should probably make sure password is not set to required in capabilities service (used by carbon)

huypha: I'll take a look at that separately in a different PR

PR #334, opened 2017-05-01
Support connecting with a connection string

MattIrv:
This is part of the changes for https://github.com/Microsoft/vscode-mssql/issues/786

- Add a ConnectionString property to ConnectionDetails, and use it to connect in place of the other parameters, if given
- Add a test that attempts to connect using a connection string

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -277,12 +277,26 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                 // Update with the actual database name in connectionInfo and result
                 // Doing this here as we know the connection is open - expect to do this only on connecting
                 connectionInfo.ConnectionDetails.DatabaseName = connection.Database;
-                response.ConnectionSummary = new ConnectionSummary
+                if (!string.IsNullOrEmpty(connectionInfo.ConnectionDetails.ConnectionString))
                 {
-                    ServerName = connectionInfo.ConnectionDetails.ServerName,
-                    DatabaseName = connectionInfo.ConnectionDetails.DatabaseName,
-                    UserName = connectionInfo.ConnectionDetails.UserName,
-                };
+                    // If the connection was set up with a connection string, use the connection string to get the details
+                    var connectionString = new SqlConnectionStringBuilder(connection.ConnectionString);
+                    response.ConnectionSummary = new ConnectionSummary
+                    {
+                        ServerName = connectionString.DataSource,
+                        DatabaseName = connectionString.InitialCatalog,
+                        UserName = connectionString.UserID
+                    };
+                }
+                else
+                {
+                    response.ConnectionSummary = new ConnectionSummary
+                    {
+                        ServerName = connectionInfo.ConnectionDetails.ServerName,
+                        DatabaseName = connectionInfo.ConnectionDetails.DatabaseName,
+                        UserName = connectionInfo.ConnectionDetails.UserName
+                    };
+                }
 
                 response.ConnectionId = connectionInfo.ConnectionId.ToString();
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -821,7 +821,16 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         /// <param name="connectionDetails"></param>
         public static string BuildConnectionString(ConnectionDetails connectionDetails)
         {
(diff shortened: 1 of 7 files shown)


MattIrv:
There are some whitespace changes in the diff because some lines in the files I edited previously had Windows line endings. To view the diff without them, visit https://github.com/Microsoft/sqltoolsservice/pull/334/files?w=1

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -1219,15 +1219,5 @@ public void ConnectParamsAreInvalidIfConnectionIsNull()
 
             Assert.NotEmpty(connectionResult.ConnectionId);
         }
-
-        [Fact]
-        public async Task ConnectionWithBadConnectionStringFails()
```
abist: I think this is still a valid test that checks the format of a connection string, isn't it?

MattIrv: That functionality is already tested by BuildConnectionStringWithInvalidAuthType, which is why I removed this test.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -485,33 +501,33 @@ private void SetOptionValue<T>(string name, T value)
...
+
+        public bool IsComparableTo(ConnectionDetails other)
+        {
+            if (other == null)
+            {
+                return false;
+            }
+
+            if (!string.Equals(ServerName, other.ServerName)
+                || !string.Equals(AuthenticationType, other.AuthenticationType)
```
benrr101: Generally speaking, this string comparison should be done with `==` unless there's a need for case insensitive comparison.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -5,7 +5,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
+using System.Globalization;
```
benrr101: I don't see anything that requires Globalization. Let's double check that it's needed before adding.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -1210,5 +1210,14 @@ public void ConnectParamsAreInvalidIfConnectionIsNull()
             await Assert.ThrowsAsync<InvalidOperationException>(
                 () => service.GetOrOpenConnection(TestObjects.ScriptUri, ConnectionType.Query));
         }
+
+        [Fact]
+        public async Task ConnectionWithConnectionStringSucceeds()
```
benrr101: I think since it's possible to have both a connection string and the piece-meal connection properties, we should have a test for that, to make sure the expected behavior is produced.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
MattIrv: This is unrelated to my change except that my change restored a unix line ending on this line.

Globalization is used near the bottom of the file by some code that isn't part of my change.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
MattIrv: These are unrelated to my changes but were included because the line ending changes. In any case I went ahead and replaced with ==

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
MattIrv: added in the new commit

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -294,7 +294,7 @@ private ConnectionCompleteParams GetConnectionCompleteParams(string connectionTy
                     {
                         ServerName = connectionInfo.ConnectionDetails.ServerName,
                         DatabaseName = connectionInfo.ConnectionDetails.DatabaseName,
-                        UserName = connectionInfo.ConnectionDetails.UserName
+                        UserName = connectionInfo.ConnectionDetails.UserName,
```
abist: extra comma in the end

PR #336, opened 2017-05-08
Add command line option to specify the logging directory

pensivebrian:
The mssql-scripter command line tool installs the sqltoolsservice to a system directory. This is problematic for logging, which logs to the same directory as the sqltoolsservice executable, which means sudo/elevated permissions are need to turn on sqltoolsservice logging.

This PR adds a command line option to specify the logging using the --log-dir argument, so that sqltoolsservice logging set to a directory that does not require elevated permissions.

Other changes include:

- A fix for command line parsing bug where we in-correctly parse arguments that requires a value. Only the -locale parameter was impacted.
- Wrapping main in a try catch so that we can log the exception

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Utility/Logger.cs
+++ b/src/Microsoft.SqlTools.Hosting/Utility/Logger.cs
@@ -79,9 +79,17 @@ namespace Microsoft.SqlTools.Utility
             {
                 if (!Directory.Exists(logDir))
                 {
-                    Directory.CreateDirectory(logDir);
+                    try
+                    {
+                        Directory.CreateDirectory(logDir);
+                    }
+                    catch (Exception)
+                    {
+                        // Creating the log directory is a best effort operation, so ignore any failures.
+                    }
                 }
             }
+
 
             // get a unique number to prevent conflicts of two process launching at the same time
             int uniqueId;
--- a/src/Microsoft.SqlTools.Hosting/Utility/Logger.cs
+++ b/src/Microsoft.SqlTools.Hosting/Utility/Logger.cs
@@ -73,6 +73,16 @@ namespace Microsoft.SqlTools.Utility
 
             Logger.isInitialized = true;
 
+            // Create the log directory
+            string logDir = Path.GetDirectoryName(logFilePath);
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/Utility/CommandOptions.cs
```diff
@@ -102,33 +104,44 @@ public string Usage
...
-                // Setting our internal SR culture to our global culture
-                SR.Culture = CultureInfo.CurrentCulture;
+                this.LoggingDirectory = Path.GetFullPath(loggingDirectory);
             }
-            catch(Exception ex)
+            catch (Exception ex)
             {
-                // Warn user of invalid locale, and fall back to english 
+                // Warn user of invalid logging directory, and fall back to the default
                 Console.WriteLine(ex);
```
pensivebrian: >Console.WriteLine(ex); [](start = 16, length = 22)

I'm thinking I should remove this and log a warning instead. By writing to stdout like this, we're going to confuse any client expecting json-rpc.

pensivebrian: I removed the try/catch to let the exception bubble up to the outer try/catch in the constructor.

---
In reply to: [115342948](https://github.com/Microsoft/sqltoolsservice/pull/336#discussion_r115342948) [](ancestors = 115342948)

src/Microsoft.SqlTools.Hosting/Utility/Logger.cs
```diff
@@ -73,6 +73,16 @@ public static class Logger
 
             Logger.isInitialized = true;
 
+            // Create the log directory
+            string logDir = Path.GetDirectoryName(logFilePath);
+            if (!string.IsNullOrWhiteSpace(logDir))
+            {
+                if (!Directory.Exists(logDir))
+                {
+                    Directory.CreateDirectory(logDir);
```
kburtram: should this be in a try/catch block?  or are exceptions already handled\recovered higher up?

pensivebrian: Putting it in a try catch block makes sense to make this a best effort operation.

PR #337, opened 2017-05-08
Fix scripting data in the scripting service

pensivebrian:
A regression was introduced in the scripter service refactoring where data is no longer scripted. This PR fixes the issue, and updates the tests to catch this in the future.

The issue is in the getter for SqlScriptPublishModel.AdvancedOptions, there is some strange logic which will cause the SqlScriptPublishModel.AdvancedOptions to get reset and lose all values based the ordering of when SqlScriptPublishModel.ScriptAllObjects is set.  In the scripting service refactoring, we started to hit this reset of the AdvanceOptions, which would lose the option to script data, along with other options.

To workaround this, we initialize with SqlScriptPublishModel.ScriptAllObjects to true, and then set all SqlScriptPublishModel.AdvancedOptions values.  Then, we set SqlScriptPublishModel.ScriptAllObjects, and avoid calling the SqlScriptPublishModel.AdvancedOptions getter.

Also included in this PR are misc scripting service changes:
- Add a sequence number field to the scripting operation events
- Add an error message and error details to scripting progress events
- Expect a null scripting option parameter
- Correctly set the exception message and details for json-rpc events
- More logging
- Comments


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -138,39 +133,66 @@ public override void Execute()
...
         {
+            this.SetCommomEventProperties(parameters);
             this.ProgressNotification?.Invoke(this, parameters);
         }
 
+        private void SetCommomEventProperties(ScriptingEventParams parameters)
+        {
+            parameters.OperationId = this.OperationId;
+            parameters.SequenceNumber = this.eventSequenceNumber;
+            this.eventSequenceNumber += 1;
```
kburtram: ++this.eventSequenceNumber :)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -280,23 +305,24 @@ private void OnPublishModelScriptError(object sender, ScriptEventArgs e)
...
                 LogLevel.Verbose,
                 string.Format(
-                    "Sending scripting error progress event, Urn={0}, OperationId={1}, Completed={2}, Error={3}",
+                    "Sending scripting error progress event, Urn={0}, OperationId={1}, Sequence={2}, Completed={3}, Error={4}",
                     e.Urn,
                     this.OperationId,
+                    this.eventSequenceNumber,
                     e.Completed,
-                    e.Error));
+                    e?.Error?.ToString()));
```
kburtram: ?. syntax returns null if the object is null, right?  Does passing nulls to string.Format work as expected?

pensivebrian: It works since I' using more than 3 arguments which invokes a particular overload of string.format. I'll update to use the null coalescing operator to make this more clear. 

---
In reply to: [115635528](https://github.com/Microsoft/sqltoolsservice/pull/337#discussion_r115635528) [](ancestors = 115635528)

PR #341, opened 2017-05-09
Merge create db work in progress

kburtram:
Pushing the create db work in-progress to avoid having so many changes in flight.


kburtram:
This is work in progress, so I'll just merge "as-is" please leave feedback on the follow-up incremental changes if there is feedback.

PR #344, opened 2017-05-12
Remove unused code and fix test break

kburtram:


kburtram:
This is strictly deleting and moving code, no functional change, so automerging.

PR #345, opened 2017-05-12
Delete more unneeded sql manager UI code

kburtram:


kburtram:
More (mostly) non-functional clean-ups...automerging.  Email directly if there is review feedback.

PR #349, opened 2017-05-12
Add EXCEPT to default keyword completion list

kburtram:


kburtram:
This is trivial change so I'll merge it now.

PR #352, opened 2017-05-16
fixed the bug with not closing data reader for oe

llali:
- OE was failing sometimes with this exception: 
There is already an open DataReader associated with this Command which must be closed first.
so probably either the data reader was not closed or two expand request was getting processed at the same time, adding the lock and disposing the data reader seems to fix it

-Trigger has null subtype/NodeStatus


AbbiePetcht:
:shipit:

PR #359, opened 2017-05-25
Ensure connection open for OE queries and fix connection disposal

kevcunnane:
- Dispose connection in Metadata service, to ensure we cleanly dispose and don't rely on garbage colleciton
- Fixed issue where if the connection was closed, expanding databases in the Server would fail. This is because SMO doesn't always reopen the connection, certainly not for Server level queries. The solution is to always check if open and reopen.
- Added unit tests for this, which required mocking the relevant IsOpen / OpenConnection methods. Refactored SMO wrapper calls into a dedicated class file to handle this

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
@@ -74,8 +74,10 @@ namespace Microsoft.SqlTools.ServiceLayer.Metadata
                 var metadata = new List<ObjectMetadata>();
                 if (connInfo != null) 
                 {                    
-                    SqlConnection sqlConn = OpenMetadataConnection(connInfo);
-                    ReadMetadata(sqlConn, metadata);
+                    using (SqlConnection sqlConn = OpenMetadataConnection(connInfo))
+                    {
+                        ReadMetadata(sqlConn, metadata);
+                    }
                 }
 
                 await requestContext.SendResult(new MetadataQueryResult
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -74,8 +74,10 @@ public void InitializeService(ServiceHost serviceHost)
                 var metadata = new List<ObjectMetadata>();
                 if (connInfo != null) 
                 {                    
-                    SqlConnection sqlConn = OpenMetadataConnection(connInfo);
-                    ReadMetadata(sqlConn, metadata);
+                    using (SqlConnection sqlConn = OpenMetadataConnection(connInfo))
```
kburtram: >using ( [](start = 20, length = 7)

oops :)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoWrapper.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
+{
+    /// <summary>
+    /// Internal for testing purposes only. This class provides wrapper functionality
+    /// over SMO objects in order to facilitate unit testing
+    /// </summary>
+    internal class SmoWrapper
+    {
+        public virtual Server Create(SqlConnection connection)
```
kburtram: >Create [](start = 30, length = 6)

maybe CreateServer is a better name, since it's possible we may want to create other SMO object types eventually?

kevcunnane: Good point, will do

PR #361, opened 2017-05-26
Avoid throwing exception in LangageService for OE connections

kevcunnane:
The LanguageService was throwing multiple exceptions when trying to do a File.Open on the ObjectExplorer and Connection URIs. The fix was to add correct handling and only support relative paths, file:, untitled:, and tsqloutput: URI types since these represent actual files. Anything else will result in the parsing & binding being skipped.


src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
```diff
@@ -259,6 +298,25 @@ internal string ResolveRelativeScriptPath(string baseFilePath, string relativePa
...
 
+        private bool IsNonFileUri(string path)
+        {
+            try
+            {
+
+                string scheme = GetScheme(path);
+                if (scheme != null && scheme.Length > 0 && !scheme.Equals("file"))
+                {
+                    return !fileUriSchemes.Contains(scheme);;
```
llali: extra ;

llali: >!fileUriSchemes.Contains(scheme) [](start = 27, length = 32)

!fileUriSchemes.Contains(scheme) already covers   !scheme.Equals("file")

src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
```diff
@@ -142,17 +153,41 @@ private string ResolveFilePath(string filePath)
...
+            }
+            catch(Exception)
+            {
+                // Intentionally skipping if there is a parse error
+            }
+
+            return false;
+        }
+
+        public static string GetScheme(string uri)
```
llali: just a suggestion ... these new methods are all for parsing a file path and can be all moved to a new class so the tests can only focus those methods.

src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
```diff
@@ -259,6 +298,25 @@ internal string ResolveRelativeScriptPath(string baseFilePath, string relativePa
             return combinedPath;
         }
 
+        private bool IsNonFileUri(string path)
+        {
+            try
+            {
+
+                string scheme = GetScheme(path);
+                if (scheme != null && scheme.Length > 0 && !scheme.Equals("file"))
```
kburtram: `scheme != null && scheme.Length > 0` could be shortened to `!string.IsNullOrEmpty(scheme)`

kburtram:
@kevcunnane there's a test fail in AppVeyor.  Please check it's not related to this change.

kevcunnane:
@benrr101 fully agree. Karl and I both discussed this and we need to refactor long term. We should be able to create specific connections first, and optionally kick off / allow for LanguageService / Metadata Services to be spun up for these. It's a refactor we'll need to take on at some point.

src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
kevcunnane: For now given time constraints will leave as-is

src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
kevcunnane: Thanks, will make these changes

PR #362, opened 2017-05-26
Edit data schema and SQLAuth support

kevcunnane:
- Fix non-dbo schema support, adding in specific parameter for schema. If this isn't specified, the code will still fall back to splitting up the objectName if it's a multi-part identifier. This ensures that input from action bar or CLI scenarios can still work since we'll accept multi-part names there
- Fix failure to edit data on Azure. This was failing as the SMO query to get the table info failed when cloning SqlConnection. In .Net Core it currently loses the passwor unless PersistSecurity = true.


llali:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -58,6 +58,7 @@ public EditTableMetadata GetObjectMetadata(DbConnection connection, string[] obj
             // Connect with SMO and get the metadata for the table
             Server server = new Server(new ServerConnection(sqlConn));
             Database db = new Database(server, sqlConn.Database);
+            // Database db = new Database(server, sqlConn.Database);
```
kburtram: >// Database db = new Da [](start = 12, length = 23)

dead code

src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
```diff
@@ -287,17 +293,26 @@ public static string GetScheme(string uri)
...
+        private bool IsNonFileUri(string path)
+        {
+            string scheme = GetScheme(path);
+            if (!string.IsNullOrEmpty(scheme))
+            {
+                return !fileUriSchemes.Contains(scheme); ;
+            }
+            return false;
+        }
+        private bool IsUntitled(string path)
```
kburtram: >private [](start = 8, length = 7)

nit: I think we're generally adding a blank line in between methods

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
kevcunnane: Sorry - missed during merge. Will remove

PR #364, opened 2017-05-27
Fix nullref in backup from not providing SFC XML innerDoc

kburtram:


kburtram:
auto-merging to catch the next build.  please send me review feedback directly.

PR #369, opened 2017-06-06
handling offline and inassisable db in oe

llali:
- Checking if db is accessible and if not returning an error 
- Update Smo nuget package because one of the smo method was renamed
- moved the smo properties to TreeNodeDefinition.xml to be able to add filtering.


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -130,19 +130,32 @@ FOR CREATE_DATABASE
...
                 Assert.False(tableChildren.Any(t => t.Label == "dbo.t1"));
 
             });
         }
 
         [Fact]
+        public async void ExpandOfflineDatabaseShouldReturnError()
+        {
+            var query = "ALTER DATABASE {0} SET OFFLINE WITH ROLLBACK IMMEDIATE";
+            string databaseName = "master";
```
kburtram: will this test leave master offline when if it fails or stops in the middle?  if so we should consider using a different test database.

llali: master won't be offline, I create a test db and make that offline and I delete it at the end. Will add comment in the test to explain

PR #370, opened 2017-06-08
Get Database Info

anthonydresser:
Adds get database info to admin service


src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/DatabaseInfoWrapper.cs
```diff
@@ -0,0 +1,67 @@
...
+//
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.Admin.Contracts 
+{
+    public class DatabaseInfoWrapper
+    {
+        public string name;
```
kevcunnane: A couple of things:
- We don't use public fields, we always use properties instead. So this should become `public string Name { get; set; }` for example
- Normally, we leave the business logic out of contracts classes so they're "pure". I'd suggest having a Utils method that converts DatabaseInfo to this wrapper type
- I'd also have a generic method that works like `GetValueAsString` and handles the TryGetValue and setting of the value so it's easier to assign the fields.

src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/DatabaseInfoWrapper.cs
```diff
@@ -0,0 +1,67 @@
...
+        public string owner;
+        public string collation;
+        public string recoveryModel;
+        public string databaseState;
+        
+        public DatabaseInfoWrapper(DatabaseInfo info)
+        {
+            object placeholder = null;
+            
+            info.Options.TryGetValue(AdminServicesProviderOptionsHelper.Name, out placeholder);
```
kevcunnane: As mentioned above, when you move this to a utils class I'd make this a method instead of repeating the code for each field.

src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/DatabaseInfoWrapper.cs
```diff
@@ -0,0 +1,67 @@
...
+            
+            if (placeholder != null)
+            {
+                name = (string) placeholder;
```
kevcunnane: This can cause ClassCastExceptions. Instead please use `placeholder as string`. If it's incompatible it'll be null.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Admin/DatabaseAdminTests.cs
```diff
@@ -85,5 +85,25 @@ private LiveConnectionHelper.TestConnectionResult GetLiveAutoCompleteTestObjects
             requestContext.VerifyAll();
         }
 
+        /// <summmary>
+        /// Get database info test
+        /// </summary>
+        // [Fact]
```
kevcunnane: Why is the test commented out? Will you be filling in the real logic in a future iteration?

anthonydresser: I followed the tests previously in the file, didn't realize commenting out the Fact attribute comments out the test.

src/Microsoft.SqlTools.ServiceLayer/Admin/Common/Utils.cs
```diff
@@ -260,6 +262,40 @@ public static string MakeSqlBracket(string s)
...
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static T TryAndGetValueAsType<T>(Dictionary<string, object> dict, string value) where T : class
+        {
+            object placeholder = null;
+            dict.TryGetValue(value, out placeholder);
+            if (placeholder != null)
+            {
+                return placeholder as T;
+            } else {
```
kevcunnane: else (and following `{`) each get their own newline in C#

src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/DatabaseInfoWrapper.cs
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.Admin.Contracts 
+{
+    public class DatabaseInfoWrapper
```
kevcunnane: Minor: please add some doc comments since it's a public class

src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/GetDatabaseInfoRequest.cs
```diff
@@ -0,0 +1,26 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Admin.Contracts
+{
+    public class GetDatabaseInfoParams
```
kevcunnane: Again minor: forgot to request doc comments on these public types

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Admin/DatabaseAdminTests.cs
kburtram: the other tests were failing in some of the CI runs and were commented out until the create db feature is done.  If this test is reliable you should leave it enabled.

---
In reply to: [120791129](https://github.com/Microsoft/sqltoolsservice/pull/370#discussion_r120791129) [](ancestors = 120791129)

src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/GetDatabaseInfoRequest.cs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /// <summary>
+    /// Get database info request mapping
+    /// </summary>
+    public class GetDatabaseInfoRequest
+    {
+        public static readonly
+            RequestType<GetDatabaseInfoParams, GetDatabaseInfoResponse> Type =
+                RequestType<GetDatabaseInfoParams, GetDatabaseInfoResponse>.Create("admin/getDatabaseInfo");
```
kburtram: we haven't been using camel casing for rpc messages so far, so we should keep them all lower-case unless we plan to change all message to camel case.

src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/DatabaseInfoWrapper.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.Admin.Contracts 
+{
+    /// <summary>
+    /// Wrapper object for the database info
+    /// </summary>
+    public class DatabaseInfoWrapper
```
kburtram: >DatabaseInfoWrapper [](start = 17, length = 19)

do we need a wrapper, or should we add properties to DatabaseInfo that read\write to the dictionary...the second approach is what we do in ConnectionDetails for example and it seems to work fine (and avoids having another class).

src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/GetDatabaseInfoRequest.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// <summary>
+    /// Response object for get database info
+    /// </summary>
+    public class GetDatabaseInfoResponse
+    {
+        /// <summary>
+        /// The object containing the database info
+        /// </summary>
+        public DatabaseInfoWrapper Result { get; set; }
```
kburtram: >DatabaseInfoWrapper [](start = 15, length = 19)

this should return a dictionary of values similar to the other messages..having hard-code values will make the messages not be compatible across different providers (i.e., postgres will have different properties).

PR #373, opened 2017-06-09
Added missing properties for databases and server

anthonydresser:
Added Last backup date and last log backup date, and machine name


src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
```diff
@@ -730,6 +744,8 @@ public bool HasSameValueAs(DatabaseData other)
             {
                 bool result =
                     (this.recoveryModel == other.recoveryModel) &&
+                    (DateTime.Compare(this.lastBackupDate, other.lastBackupDate) == 0) &&
```
benrr101: It isn't necessary to use the `DateTime.Compare`, you should be able to just use `==`

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
```diff
@@ -730,6 +744,8 @@ public bool HasSameValueAs(DatabaseData other)
...
+                    (DateTime.Compare(this.lastLogBackupDate, other.lastLogBackupDate) == 0) &&
```
benrr101: Also, all these parens are extraneous, but I can see the desire to stick with the existing pattern

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
```diff
@@ -892,6 +908,40 @@ public virtual RecoveryModel RecoveryModel
...
+        public DateTime LastBackupDate
+        {
+            get
+            {
+                return this.currentState.lastBackupDate;
+            }
+            set
+            {
+                this.currentState.lastBackupDate = value;
+                this.NotifyObservers();
```
benrr101: I think ideally if this is notifying observers when a property is being updated, it should be implementing the INotifyPropertyChanged interface. But, I also assume this is a huge change, so I guess we can leave it be.

anthonydresser: Yeah I was mostly just doing what other properties were doing.

src/Microsoft.SqlTools.ServiceLayer/Admin/AdminServicesProviderOptionsHelper.cs
```diff
@@ -32,6 +32,9 @@ public class AdminServicesProviderOptionsHelper
         internal const string DatabaseContainmentType = "databaseContainmentType";
         internal const string DatabaseState = "databaseState";
         internal const string RecoveryModel = "recoveryModel";
+        internal const string CompatabilityLevel = "compatabilityLevel";
```
kevcunnane: compatibility

PR #375, opened 2017-06-10
Add LanguageFlavorNotification handling and refactor LanguageService

kevcunnane:
- Added new notification handler for language flavor changed
- Skip processing of documents if a different flavor is used. This can be used in the extension to support ignoring non-MSSQL file types
- Refactored the LanguageService so that it no longer relies on so many intertwined static calls, which meant it was impossible to test without modifying the static instance. This will help with test reliability in the future, and prep for replacing the instance with a service provider.

Will work to add additional unit tests in a future iteration

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/LanguageFlavorChange.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/LanguageFlavorChange.cs
@@ -1,0 +1,42 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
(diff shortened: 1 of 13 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/LanguageFlavorChange.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        /// <summary>
+        /// A URI identifying the affected resource         
+        /// </summary>
+        public string Uri { get; set; }
+
+        /// <summary>
+        /// The primary language
+        /// </summary>
+        public string Language { get; set; }
```
MattIrv: Does language need to be part of these notifications or could we just assume that it's always SQL?

kevcunnane: I would prefer not to assume this. I would like to keep this generic enough that others could use in the future (even though for now, this is the most likely option)

PR #377, opened 2017-06-12
fixed couple of issues in task service

llali:


test/Microsoft.SqlTools.ServiceLayer.UnitTests/TaskServices/TaskServiceTests.cs
```diff
@@ -57,7 +57,7 @@ public void NewTaskShouldSendNotification()
             });
            
             serviceHostMock.Verify(x => x.SendEvent(TaskCreatedNotification.Type,
-                It.Is<TaskInfo>(t => t.TaskId == sqlTask.TaskId.ToString())), Times.Once());
+                It.Is<TaskInfo>(t => t.TaskId == sqlTask.TaskId.ToString() && t.ProviderName == "MSSQL")), Times.Once());
```
hyoshis: why is this necessary?

llali: checking that provider name is mssql? It should always be mssql so just checking that

PR #381, opened 2017-06-13
Register backup to task service for execution

hyoshis:
Register backup to task service.
Added unit tests.


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -110,9 +111,21 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    TaskMetadata metadata = new TaskMetadata();
+                    metadata.ServerName = connInfo.ConnectionDetails.ServerName;
+                    metadata.DatabaseName = connInfo.ConnectionDetails.DatabaseName;                    
+                    metadata.Name = "Backup Database";
+                    metadata.Description = "Backup Database";
+                    metadata.IsCancelable = true;
+
+                    // create backup task and perform
+                    SqlTask sqlTask = TaskService.Instance.TaskManager.CreateTask(metadata, Instance.BackupTask);
+                    sqlTask.Run();
```
llali: that is my fault to not have a sync method to call, it's not a good practice to call an async method but not wait for it. is it possible for you to add a sync method in Task service and call that? or create a sync method here and then call task.Run?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -163,7 +176,63 @@ internal void PerformBackup()
         internal BackupConfigInfo GetBackupConfigInfo(string databaseName)
         {
             return this.backupUtilities.GetBackupConfigInfo(databaseName);
-        }   
-        
+        }
+
+        public async Task<TaskResult> BackupTask(SqlTask sqlTask)
```
llali: add comment for public method

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -110,9 +111,21 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    // create task metadata
+                    TaskMetadata metadata = new TaskMetadata();
+                    metadata.ServerName = connInfo.ConnectionDetails.ServerName;
+                    metadata.DatabaseName = connInfo.ConnectionDetails.DatabaseName;                    
+                    metadata.Name = "Backup Database";
+                    metadata.Description = "Backup Database";
+                    metadata.IsCancelable = true;
+
+                    // create backup task and perform
+                    SqlTask sqlTask = TaskService.Instance.TaskManager.CreateTask(metadata, Instance.BackupTask);
```
llali: >TaskManager [](start = 59, length = 11)

SqlTaskManager.Instance.CreateTask don't call TaskService here

llali:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -163,7 +176,68 @@ internal void PerformBackup()
...
+                    catch (Exception ex)
+                    {
+                        result.TaskStatus = SqlTaskStatus.Failed;
+                        result.ErrorMessage = ex.Message;
+                    }
+
+                    return result;
+                });
+
+                Task<TaskResult> completedTask = await Task.WhenAny(backupTask, cancelTask);
```
llali: > Task<TaskResult> completedTask [](start = 15, length = 31)

this seems wrong, when you await, you should not get back task

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -163,7 +176,68 @@ internal void PerformBackup()
...
+        /// <summary>
+        /// Create a backup task for execution and cancellation
+        /// </summary>
+        /// <param name="sqlTask"></param>
+        /// <returns></returns>
+        internal async Task<TaskResult> BackupTask(SqlTask sqlTask)
+        {                           
+            return await await Task.Factory.StartNew(async () =>
+            {
+                sqlTask.AddMessage("In progress", SqlTaskStatus.InProgress, true);
```
llali: this block is returing task, no need to put it inside another Factory.StartNew

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: The signature of WhenAny is:
public static Task<Task<TResult>> WhenAny<TResult>(params Task<TResult>[] tasks);

That's why when it awaits, it gets Task<TResult>

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -163,7 +176,70 @@ internal void PerformBackup()
...
+        /// </summary>
+        /// <param name="sqlTask"></param>
+        /// <returns></returns>
+        internal async Task<TaskResult> BackupTask(SqlTask sqlTask)
+        {
+            sqlTask.AddMessage("In progress", SqlTaskStatus.InProgress, true);
+            Task<TaskResult> completedTask = await Task.WhenAny(PerformTask(), CancelTask(sqlTask));
+            sqlTask.AddMessage("Finished", completedTask.Result.TaskStatus);
+            return completedTask.Result;
+        }
```
llali: please add tests

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/BackupTests.cs
```diff
@@ -3,14 +3,62 @@
...
+        };
+
+        [Fact]
+        public async Task VerifyCreateAndRunningBackupTask()
+        {
+            using (SqlTaskManager manager = new SqlTaskManager())
+            {
+                var mockUtility = new Mock<IBackupUtilities>();
+                DisasterRecoveryService service = new DisasterRecoveryService(mockUtility.Object);
+                SqlTask sqlTask = manager.CreateTask(this.taskMetaData, service.BackupTask);
```
llali: why don't you call the service to create the task? that's part of the verification too right? that HandleRequest in the service creates a new SQlTask

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -16,11 +16,12 @@
...
 namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery
 {
-    public class BackupUtilities
+    public class BackupUtilities: IBackupUtilities
     {
         private CDataContainer dataContainer;
         private ServerConnection serverConnection;
-        private CommonUtilities backupRestoreUtil = null;        
+        private CommonUtilities backupRestoreUtil = null;
+        private Backup bk = null;
```
kburtram: >bk [](start = 23, length = 2)

please use a more complete name for this variable that follows code base conventions.  For example, "backup"

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -110,9 +119,21 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    // initialize backup
                     DisasterRecoveryService.Instance.InitializeBackup(helper.DataContainer, sqlConn);
                     DisasterRecoveryService.Instance.SetBackupInput(backupParams.BackupInfo);
-                    DisasterRecoveryService.Instance.PerformBackup();
+
+                    // create task metadata
+                    TaskMetadata metadata = new TaskMetadata();
+                    metadata.ServerName = connInfo.ConnectionDetails.ServerName;
+                    metadata.DatabaseName = connInfo.ConnectionDetails.DatabaseName;                    
+                    metadata.Name = "Backup Database";
```
kburtram: >"Backup Database" [](start = 36, length = 17)

we should read this from a resource file.  Please see the SR gen classes for this.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -163,7 +184,75 @@ internal void PerformBackup()
...
+        }
+
+        /// <summary>
+        /// Create a backup task for execution and cancellation
+        /// </summary>
+        /// <param name="sqlTask"></param>
+        /// <returns></returns>
+        internal async Task<TaskResult> BackupTask(SqlTask sqlTask)
+        {
+            sqlTask.AddMessage("In progress", SqlTaskStatus.InProgress, true);
```
kburtram: >"In progress" [](start = 31, length = 13)

this string should come from resource file

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -163,7 +184,75 @@ internal void PerformBackup()
...
+            });
+        }
+
+        private async Task<TaskResult> CancelTask(SqlTask sqlTask)
+        {
+            // create a task for backup cancellation request
+            return await Task.Factory.StartNew(() =>
+            {
+                TaskResult result = new TaskResult();
+                while (true)
```
kburtram: > while (true) [](start = 15, length = 13)

we should use cancellation tokens and wait for the token to be signaled.  We can't check-in a spin loop like this.

kburtram:
> using Microsoft.SqlTools.ServiceLayer.Admin;

this file needs MIT header

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/IBackupUtilities.cs:1 in 7c2fbf8. [](commit_id = 7c2fbf84bd9f6c5782e172ca4d22fd2095d27514, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/IBackupUtilities.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.SqlTools.ServiceLayer.Admin;
+using Microsoft.SqlTools.ServiceLayer.DisasterRecovery.Contracts;
+using System.Data.SqlClient;
+
+namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery
+{
+    public interface IBackupUtilities
```
kburtram: >IBackupUtilities [](start = 21, length = 16)

please add C# document comments for this interface and every method in the interface...our public interfaces need to be documented.

src/Microsoft.SqlTools.ServiceLayer/TaskServices/SqlTask.cs
```diff
@@ -96,6 +96,13 @@ public SqlTask(TaskMetadata taskMetdata, Func<SqlTask, Task<TaskResult>> testToR
             });
         }
 
+        public void Run()
```
kburtram: >Run [](start = 20, length = 3)

what's the point of this change?  Also function is missing C# XML comments.

src/Microsoft.SqlTools.ServiceLayer/TaskServices/SqlTask.cs
```diff
@@ -96,6 +96,13 @@ public SqlTask(TaskMetadata taskMetdata, Func<SqlTask, Task<TaskResult>> testToR
...
+        {
+            RunAsync().ContinueWith(task =>
```
kburtram: >ContinueWith [](start = 23, length = 12)

we probably don't need the continuewith if we aren't doing anything in the callback.

llali: my fault here,  please change it to Task.Run(() => RunAsync());

---
In reply to: [122303762](https://github.com/Microsoft/sqltoolsservice/pull/381#discussion_r122303762) [](ancestors = 122303762)

llali: The reason is I wanted to add a Sync method because I was getting warnings when calling the Async method inside another async but was not waiting for it. this method runs the task doesn't wait for the result  

---
In reply to: [122304875](https://github.com/Microsoft/sqltoolsservice/pull/381#discussion_r122304875) [](ancestors = 122304875,122303762)

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: added

src/Microsoft.SqlTools.ServiceLayer/TaskServices/SqlTask.cs
hyoshis: changed.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryStrings.resx
```diff
@@ -0,0 +1,129 @@
...
+        <xsd:choice maxOccurs="unbounded">
+          <xsd:element name="metadata">
+            <xsd:complexType>
+              <xsd:sequence>
+                <xsd:element name="value" type="xsd:string" minOccurs="0" />
+              </xsd:sequence>
+              <xsd:attribute name="name" use="required" type="xsd:string" />
+              <xsd:attribute name="type" type="xsd:string" />
+              <xsd:attribute name="mimetype" type="xsd:string" />
+              <xsd:attribute ref="xml:space" />
```
kburtram: can we add the resources with the rest of the sql tools service resources, and use SR gen like the rest of the code?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/BackupTests.cs
hyoshis: I can create integration test to call HandleRequest. For unit test though, I wanted to check that BackupTask takes care of execution and cancellation.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -11,14 +11,17 @@
...
 
 namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery
 {
     public class DisasterRecoveryService
     {
         private static readonly Lazy<DisasterRecoveryService> instance = new Lazy<DisasterRecoveryService>(() => new DisasterRecoveryService());
         private static ConnectionService connectionService = null;
-        private BackupUtilities backupUtilities;
+        private IBackupUtilities backupUtilities;
+        private ManualResetEvent manualResetEvent = new ManualResetEvent(initialState: false);
```
kburtram: >manualResetEvent [](start = 33, length = 16)

probably a better name would be something like "taskCompletedEvent" or something more specific about what the event is

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -166,7 +188,82 @@ internal void PerformBackup()
...
+
+        private async Task<TaskResult> CancelTask(SqlTask sqlTask)
+        {
+            // create a task for backup cancellation request
+            return await Task.Factory.StartNew(() =>
+            {
+                TaskResult result = new TaskResult();
+
+                CancellationToken token = sqlTask.TokenSource.Token;
+                WaitHandle[] waitHandles = new WaitHandle[2]
```
kburtram: >waitHandles [](start = 29, length = 11)

don't you need to wait for these handles somewhere?

kburtram:
@hyoshis please also check why the CI builds are failing on this PR.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -33,9 +37,9 @@ public class BackupUtilities
         /// UI input values
         /// </summary>
         private BackupInfo backupInfo;
-        public BackupComponent backupComponent { get; set; }
-        public BackupType backupType { get; set; } // 0 for Full, 1 for Differential, 2 for Log
-        public BackupDeviceType backupDeviceType { get; set; }
+        private BackupComponent backupComponent { get; set; }
```
kburtram: >{ get; set; } [](start = 48, length = 13)

typically private values just use fields, instead of properties.  Is there a reason to use auto-properties here instead of just using fields?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -16,11 +16,15 @@
 
 namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery
 {
-    public class BackupUtilities
+    /// <summary>
+    /// This class implements backup operations
+    /// </summary>
+    public class BackupUtilities: IBackupUtilities
```
kburtram: >: [](start = 32, length = 1)

nit: missing ' ' before : character

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -136,6 +147,109 @@ public BackupConfigInfo GetBackupConfigInfo(string databaseName)
         }
 
         /// <summary>
+        /// Execute backup
+        /// </summary>
+        public void PerformBackup()
+        {
+            this.backup = new Backup();
```
kburtram: >this.backup = new Backup(); [](start = 12, length = 27)

so we can only have 1 backup running at a time per instance of this class?  should have some logic to enforce this?  i.e. check if a backup is currently running and either exit the method or raise exception or abort operation..etc.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -136,6 +147,109 @@ public BackupConfigInfo GetBackupConfigInfo(string databaseName)
...
+            this.SetBackupProps();
+            backup.Database = this.backupInfo.DatabaseName;
+            backup.Action = this.backupActionType;
+            backup.Incremental = this.IsBackupIncremental;
+            if (backup.Action == BackupActionType.Files)
+            {
+                IDictionaryEnumerator IEnum = this.backupInfo.SelectedFileGroup.GetEnumerator();
```
kburtram: >IEnum [](start = 38, length = 5)

variables should use camelCase names...also this name isn't particularly descriptive.   Maybe something like fileGroupEnum or something like that.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -136,6 +147,109 @@ public BackupConfigInfo GetBackupConfigInfo(string databaseName)
...
+                        System.Globalization.CultureInfo.InvariantCulture);
+                    if (CurrentKey.IndexOf(",", StringComparison.Ordinal) < 0)
+                    {
+                        // is a file group
+                        backup.DatabaseFileGroups.Add(CurrentValue);
+                    }
+                    else
+                    {
+                        // is a file
+                        int Idx = CurrentValue.IndexOf(".", StringComparison.Ordinal);
```
kburtram: >Idx [](start = 28, length = 3)

variable names should be camelCase...so idx is better here.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -136,6 +147,109 @@ public BackupConfigInfo GetBackupConfigInfo(string databaseName)
...
+            backup.Database = this.backupInfo.DatabaseName;
+            backup.Action = this.backupActionType;
+            backup.Incremental = this.IsBackupIncremental;
+            if (backup.Action == BackupActionType.Files)
+            {
+                IDictionaryEnumerator IEnum = this.backupInfo.SelectedFileGroup.GetEnumerator();
+                IEnum.Reset();
+                while (IEnum.MoveNext())
+                {
+                    string CurrentKey = Convert.ToString(IEnum.Key,
```
kburtram: >CurrentKey [](start = 27, length = 10)

...camelCase variables names please...I won't list any more, but please update all variables to follow naming conventions for this code base.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
```diff
@@ -136,6 +147,109 @@ public BackupConfigInfo GetBackupConfigInfo(string databaseName)
...
+                    {
+                        // is a file
+                        int Idx = CurrentValue.IndexOf(".", StringComparison.Ordinal);
+                        CurrentValue = CurrentValue.Substring(Idx + 1, CurrentValue.Length - Idx - 1);
+                        backup.DatabaseFiles.Add(CurrentValue);
+                    }
+                }
+            }
+
+            bool bBackupToUrl = false;
```
kburtram: >bBackupToUrl [](start = 17, length = 12)

please don't use hungarian notation in variable names..so this should be backupToUrl

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -166,7 +197,83 @@ internal void PerformBackup()
...
+        /// <returns></returns>
+        internal async Task<TaskResult> BackupTask(SqlTask sqlTask)
+        {
+            sqlTask.AddMessage(SR.Task_InProgress, SqlTaskStatus.InProgress, true);
+            Task<TaskResult> performTask = this.PerformTask();
+            Task<TaskResult> cancelTask = this.CancelTask(sqlTask);
+            Task<TaskResult> completedTask = await Task.WhenAny(performTask, cancelTask);
+            if (completedTask == performTask)
+            {
+                this.backupCompletedEvent.Reset();
```
kburtram: >Reset [](start = 42, length = 5)

should this be a call to Set() here?  Also, there doesn't seem to be anywhere that we Reset() the event when starting a second backup...so it looks like this class is for a single use.  If that is the intended design we need to enforce the contract by throwing when PerformBackup is called more than once.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/BackupTests.cs
```diff
@@ -3,14 +3,62 @@
...
+                {
+                    Assert.Equal(sqlTask.TaskStatus, SqlTaskStatus.Succeeded);                 
+                });
+
+                await taskToVerify;
+            }
+        }
+
+        [Fact]
+        public async Task CancelBackupTask()
```
kburtram: >CancelBackupTask [](start = 26, length = 16)

it seems like there have been a couple bugs in the implementation for cancellation in this and the previous iteration.  Could you please look at this test and perhaps add additional test methods to cover those scenarios?  I would have expected tests to fail since the eventing and waiting isn't correct in the cancellation code path.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
hyoshis: Let me think about it. I'll make changes regarding it as separate PR.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupUtilities.cs
hyoshis: Sorry about that. I went through all the imported code and updated casing now.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: Yes, I changed it to Set(). Right now this is for single use, I'll think about the design and make changes as separate PR.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/BackupTests.cs
hyoshis: When I switched to use cancellation token's waithandle, there was slight delay between calling cancel and getting signal. So what happend in the test is that performBackup, which is empty method, was finished before cancel was called. I fixed the test by creating stub perform operation which sleeps for few seconds while waiting for cancel.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
kburtram: could you open an issue to track fixing this? currently the class seems like it can be easily used in a way that will cause it to not work correctly.

---
In reply to: [122376955](https://github.com/Microsoft/sqltoolsservice/pull/381#discussion_r122376955) [](ancestors = 122376955)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/BackupUtilitiesStub.cs
```diff
@@ -0,0 +1,61 @@
...
+        public void SetBackupInput(BackupInfo input)
+        {
+        }
+
+        /// <summary>
+        /// Execute backup
+        /// </summary>
+        public void PerformBackup()
+        {
+            Thread.Sleep(2000);
```
kburtram: >2000 [](start = 25, length = 4)

you might see if a shorter delay is sufficient for this test case.  Perhaps 200ms or 500ms, just so the tests run a little faster.

kburtram:
@hyoshis please see latest iteration comments, and please open Issues for the couple things you mentioned you want to do as separate PRs.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/BackupUtilitiesStub.cs
hyoshis: I changed it to 500ms.

hyoshis:
I created https://github.com/Microsoft/carbon/issues/1204 (one backup per DisasterRecovery service instance) for follow up.

PR #383, opened 2017-06-14
Adds a execute and return result message

anthonydresser:
Adds a new rpc message that will execute a query string and return the results rather than sending them in a notification later on.

This is useful for when we want to execute a query internally without needing to handle all the communication infrastructure required by query/executeString.

Still not sold on the name `query/executeandreturnresult` let me know if anyone can think of a better name.


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteAndReturnResultRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests
+{
+    /// <summary>
+    /// Parameters for executing a query from a document open in the workspace
+    /// </summary>
+    public class ExecuteAndReturnResultParams
```
kevcunnane: How about we call this `SimpleExecute` instead of `ExecuteAndReturnResult`? Something shorter at least?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteAndReturnResultRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <summary>
+        /// 2D array of the cell values requested from result set
+        /// </summary>
+        public DbCellValue[][] Rows { get; set; }
+    }
+
+    public class ExecuteAndReturnResultRequest
+    {
+        public static readonly
+            RequestType<ExecuteAndReturnResultParams, ExecuteAndReturnResultResult> Type =
```
kevcunnane: Is this going to block the message channel waiting for a response? I'm worried this isn't truly async, do we need to split into a `ExecuteRequest` that kicks off and an `ExecutionComplete` notification?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -169,6 +170,59 @@ public void InitializeService(ServiceHost serviceHost)
...
+            Query.QueryAsyncEventHandler queryComplete = async q =>
+            {
+                int rowCount = 0;
+                try
+                {
+                    rowCount = Convert.ToInt32(q.Batches[0].ResultSets[0].RowCount);
+                }
+                catch(OverflowException e)
+                {
+                    await requestContext.SendError(e);
```
kevcunnane: I would recommend checking batch count explicitly first, and sending a clear message over the channel instead of the try\catch exception pattern. That pattern is discouraged in C# since first-chance exceptions can slow debugging (and it's just not something C# seems to like). 

Also, I'd ask what should happen when we have more than 1 batch? In that case should something be done differently?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteAndReturnResultRequest.cs
kevcunnane: Or given we're limited to a single resultset, how about `RetrieveResultset`?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -169,6 +170,59 @@ public void InitializeService(ServiceHost serviceHost)
         }
 
         /// <summary>
+        /// Handles a request to execute a string and return the result
+        /// </summary>
+        internal Task HandleExecuteAndReturnResultRequest(ExecuteAndReturnResultParams executeParams,
+            RequestContext<ExecuteAndReturnResultResult> requestContext)
+        {
+            Func<string, Task> queryCreateFailureAction = message => requestContext.SendError(message);
+            ExecuteStringParams newParams = new ExecuteStringParams
```
kevcunnane: how about `executeStringParams` for clarity here instead of `newParams`

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -169,6 +170,59 @@ public void InitializeService(ServiceHost serviceHost)
...
+                await requestContext.SendResult(result);
+            };
+
+            // handle sending error back when query fails
+            Query.QueryAsyncErrorEventHandler queryFail = async (q, e) =>
+            {
+                await requestContext.SendError(e);
+            };
+
+            return InterServiceExecuteQuery(newParams, requestContext, null, queryCreateFailureAction, queryComplete, queryFail);
```
kevcunnane: - Should we be doing cleanup after this is done? It seems like in this case we're assuming client-side caching, so shouldn't we dispose the resultset etc? 

- I am guessing you may run multiple different queries on the same channel. If this is the case, how do we ensure they don't conflict? I'm not sure how the query service manages this - is there a queue?

Also what if the size is larger than some set limit, should we handle that at all?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultOnlyContext.cs
```diff
@@ -0,0 +1,25 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.SqlTools.Hosting.Protocol;
+using System.Threading.Tasks;
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
+{
+    public class ResultOnlyContext<TResult> : IEventSender
```
kevcunnane: Please add doc comments

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
```diff
@@ -432,6 +432,21 @@ public void GetSqlTextFromInvalidType()
             Assert.Equal(1, queryService.ActiveQueries.Count);
         }
 
+        [Fact]
```
kevcunnane: Can we have some true unit tests also?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteAndReturnResultRequest.cs
anthonydresser: The asynchronous of this request is given by the asynchronous of the entire query execution (see **https://github.com/Microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs#L565**). I don't believe the message channel ever waits for a message. I believe the synchronous problem we were having before was because the handlers were not asynchronous, not because the message channel was waiting on a response. Having a separate notification complete pretty much defeats the purpose of having a simple "execute this query" and we may as well just use the query string execution message.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteAndReturnResultRequest.cs
anthonydresser: I like Simpleexecute, I wasn't sure what to call this so was hoping for suggestions.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
anthonydresser: I wanted this to be the simplest query execution possible so i wanted it to be limited to 1 batch 1 resultset, having multiple batches and results world complicate the response more than what I was looking for. But if we think this is something that should be done I can implement something like that.

On the topic of this try catch, @benrr101 can explain a bit more, but basically there is a worry that if the client requests a rowcount that is too large it could take up all the memory, so limiting this value to a int makes sense. I meant this try catch to be the point that it checks if there are too many rows to safely load into memory. Although I agree the error sent back should be better. 

The other way I guess this could be done to avoid try catch is to explicit check if the long is greater than 2^31 - 1 and send back an error if it is.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
anthonydresser: I should look into disposing the results, I hadn't thought of that.

The size being larger than some set limit is what the try catch on the convert to int was above.

kevcunnane: Cool - I'd just advise commenting on why you're doing it that way since it's not immediately obvious.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/ExecuteAndReturnResultRequest.cs
kevcunnane: Cool, this is a constant cause of worry so as long as this is the case, I'm happy with this implementation.

benrr101: Yep, @anthonydresser sums it up pretty well

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
benrr101: I'd go with the 2nd option here, though I'd probably use `int.MaxInt` instead 😄 

As @anthonydresser were discussing originally, an improvement here would be to have a AllRows IEnumerable in the result set, but it runs the risk of those who call it to do something like `.ToList()` on it and end up loading all the results into memory anyways. But the implementation here is the quickest/easiest to get the functionality we're going after.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultOnlyContext.cs
```diff
@@ -8,6 +8,11 @@
 
 namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
 {
+    /// <summary>
+    /// Implementation of IEventSender that swollows events without doing anything with them.
```
kevcunnane: typo `swollows` -> `swallows`?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
```diff
@@ -433,16 +433,15 @@ public void GetSqlTextFromInvalidType()
         }
 
         [Fact]
-        public async Task QueryExecuteAndReturnQueryTest()
+        public async Task SimpleExecuteTest()
```
kevcunnane: We really need a test to verify multiple requests can be handled...

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -165,7 +166,85 @@ public void InitializeService(ServiceHost serviceHost)
...
+                return requestContext.SendError(SR.QueryServiceQueryInvalidOwnerUri);
+            }
+
+            if (connInfo.ConnectionDetails.MultipleActiveResultSets == null || connInfo.ConnectionDetails.MultipleActiveResultSets == false) {
+                // if multipleActive result sets is not allowed, don't specific a connection and make the ownerURI the true owneruri
+                connInfo = null;
+                executeStringParams.OwnerUri = executeParams.OwnerUri;
+            }
+
+            Func<string, Task> queryCreateFailureAction = message => requestContext.SendError(message);
```
anthonydresser: I get syntax errors when I try that.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -165,7 +166,85 @@ public void InitializeService(ServiceHost serviceHost)
...
+             ExecuteStringParams executeStringParams = new ExecuteStringParams
+            {
+                Query = executeParams.QueryString,
+                // generate guid as the owner uri to make sure every query is unique
+                OwnerUri = Guid.NewGuid().ToString()
+            };
+
+            // get connection
+            ConnectionInfo connInfo;
+            if (!ConnectionService.TryFindConnection(executeParams.OwnerUri, out connInfo))
```
anthonydresser: whats the difference? the connection finding in CreateQuery uses tryfindconnection, so I just copied that.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -166,51 +166,76 @@ public void InitializeService(ServiceHost serviceHost)
...
+            }
+
+            if (connInfo.ConnectionDetails.MultipleActiveResultSets == null || connInfo.ConnectionDetails.MultipleActiveResultSets == false) {
+                // if multipleActive result sets is not allowed, don't specific a connection and make the ownerURI the true owneruri
+                connInfo = null;
+                executeStringParams.OwnerUri = executeParams.OwnerUri;
+            }
+
+            Func<string, Task> queryCreateFailureAction = message => requestContext.SendError(message);
+
```
anthonydresser: fixed

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
anthonydresser: fixed

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -197,45 +197,48 @@ public void InitializeService(ServiceHost serviceHost)
...
             Func<string, Task> queryCreateFailureAction = message => requestContext.SendError(message);
 
-           
-
             ResultOnlyContext<SimpleExecuteResult> newContext = new ResultOnlyContext<SimpleExecuteResult>(requestContext);
 
             // handle sending event back when the query completes
             Query.QueryAsyncEventHandler queryComplete = async q =>
             {
+                Query trashQ;
```
kevcunnane: Minor: some internal tools like Policheck don't like words like trash. Might consider renaming to removedQuery

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -396,6 +421,7 @@ public void InitializeService(ServiceHost serviceHost)
         /// Callback to call when query has completed execution with errors. May be <c>null</c>.
         /// </param>
         public async Task InterServiceExecuteQuery(ExecuteRequestParamsBase executeParams, 
+            ConnectionInfo connInfo,
```
anthonydresser: Fixed

PR #385, opened 2017-06-16
Fixed typo in sr.strings

kevcunnane:
- added newline to separate 2 properties and ran SrGen to regenerate the loc files


llali:
:shipit:

PR #386, opened 2017-06-16
Fix DMP message name to use correct casing

kburtram:


kburtram:
@fixing casing type..nothing really to review, so merging.

PR #387, opened 2017-06-18
Reset event signal in backup service

hyoshis:
Call reset method of MRE otherwise it will stay as signaled preventing the next backup execution.


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -212,7 +212,10 @@ internal BackupConfigInfo GetBackupConfigInfo(string databaseName)
             Task<TaskResult> completedTask = await Task.WhenAny(performTask, cancelTask);
             if (completedTask == performTask)
             {
+                // Send event signal to allow the waiting CancelTask to proceed 
                 this.backupCompletedEvent.Set();
+                // Reset the event signal
+                this.backupCompletedEvent.Reset();
```
kburtram: you should use an auto-reset event if you want this pattern.

hyoshis:
Closing this PR since this is fixed/addressed in different PR.

PR #395, opened 2017-06-23
LanguageService must send result or intellisense hangs

kevcunnane:
- Unless the language services methods return results for requests, the VSCode language service protocol will never send a response up to its higher-level code. This means with intellisense off, it appears to hang instead of saying "No results found". This is clearly sub-optimal and if any other extension wants to provide SQL suggestions, it would break them from sending results
- Minor refactor to fully remove Instance-field references in the code

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -389,14 +389,14 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             ReferencesParams referencesParams,
             RequestContext<Location[]> requestContext)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -369,14 +389,14 @@ private static TelemetryProperties CreatePeekTelemetryProps(bool succeeded, bool
             ReferencesParams referencesParams,
             RequestContext<Location[]> requestContext)
         {
-            await Task.FromResult(true);
+            await requestContext.SendResult(null);;
```
llali: please remove the extra ;

llali:
:shipit:

PR #396, opened 2017-06-24
Fix backup service to have multiple backup instances

hyoshis:
- Fix backup service to have multiple backup instances
- Changed to use AutoResetEvent for cancellation wait
- Added a test for multiple backup operations.


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -179,56 +164,66 @@ internal static SqlConnection GetSqlConnection(ConnectionInfo connInfo)
...
-            this.backupUtilities.PerformBackup();
+            BackupOperation backupOperation = new BackupOperation();
+            backupOperation.Initialize(dataContainer, sqlConnection);
+            backupOperation.SetBackupInput(input);
+            return backupOperation;
         }
-
-        internal BackupConfigInfo GetBackupConfigInfo(string databaseName)
+                
+        internal void PerformBackup(BackupOperation backupOperation)
```
kburtram: do we need this method?  it looks like it's just directly calling another method.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -179,56 +164,66 @@ internal static SqlConnection GetSqlConnection(ConnectionInfo connInfo)
...
         /// <param name="sqlTask"></param>
         /// <returns></returns>
-        internal async Task<TaskResult> BackupTask(SqlTask sqlTask)
+        internal async Task<TaskResult> BackupTaskAsync(SqlTask sqlTask)
         {
             sqlTask.AddMessage(SR.Task_InProgress, SqlTaskStatus.InProgress, true);
-            Task<TaskResult> performTask = this.PerformTask();
-            Task<TaskResult> cancelTask = this.CancelTask(sqlTask);
+
+            IBackupOperation backupOperation = sqlTask.TaskMetadata.Data as IBackupOperation;
```
kburtram: we should null check backupOperation since this is best practice anytime using 'as' cast.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -179,56 +164,66 @@ internal static SqlConnection GetSqlConnection(ConnectionInfo connInfo)
...
+            AutoResetEvent backupCompletedEvent = new AutoResetEvent(initialState: false);
+            Task<TaskResult> performTask = this.PerformTaskAsync(backupOperation);
+            Task<TaskResult> cancelTask = this.CancelTaskAsync(backupOperation, sqlTask.TokenSource.Token, backupCompletedEvent);
```
kburtram: nit: we're typically not using 'this.' prefix when calling methods..only when access fields or properties per typical .Net code guidelines.

hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -100,15 +90,11 @@ public void InitializeService(ServiceHost serviceHost)
             {
                 DatabaseTaskHelper helper = AdminService.CreateDatabaseTaskHelper(connInfo, databaseExists: true);
                 SqlConnection sqlConn = GetSqlConnection(connInfo);
-                if (sqlConn != null)
+                if ((sqlConn != null) && !connInfo.IsSqlDW)
```
llali: does back up work for Azure?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: This is for testing purpose. I'll add comment here.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: No. We do not show backup task for azure connection.

llali: so we should add the check here too maybe

---
In reply to: [124131526](https://github.com/Microsoft/sqltoolsservice/pull/396#discussion_r124131526) [](ancestors = 124131526)

hyoshis: Added check for azure connection.

PR #398, opened 2017-06-28
Support "SQL" settings in addition to MSSQL

kevcunnane:
- Handles having 2 separate configuration definitions and merging / treating them as 1 throughout the app
- If a settings group such as Intellisense is defined on mssql, it will override any generic SQL properties
- Retains backwards compatibility with existing settings.


src/Microsoft.SqlTools.ServiceLayer/SqlContext/SqlToolsSettings.cs
```diff
@@ -103,19 +144,116 @@ public bool IsIntelliSenseEnabled
...
 
+    public interface ISqlToolsSettingsValues
+    {
+        IntelliSenseSettings IntelliSense { get; set; }
+        QueryExecutionSettings QueryExecutionSettings { get; set; }
+        FormatterSettings Format { get; set; }
+        ObjectExplorerSettings ObjectExplorer { get; set; }
+    }
+
+    public class CompoundToolsSettingsValues: ISqlToolsSettingsValues
```
benrr101: I think these classes need their own files. It's starting to get cluttered in here.

PR #399, opened 2017-06-30
Fix CancelMultipleBackup test failure

hyoshis:
CancelMultipleBackupTasks test has been intermittently failing in lab runs.
Added semaphore for cancellation tests to make it more reliable.


hyoshis:
I'll update it to use semaphore.

PR #403, opened 2017-07-07
Feature/restore db

llali:
- Moved RestoreDatabaseTaskDataObject and RestoreUtil from ssms to be able to run restore operation the same way ssms does

- Added a service class to handle the restore operations. It can create a restore plan and execute it

- Added some tests to verify restore but still have to add more

-Implemented the service handlers and contracts


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -101,6 +110,82 @@ public void InitializeService(ServiceHost serviceHost)
...
+                        TaskMetadata metadata = new TaskMetadata();
+                        metadata.ServerName = connInfo.ConnectionDetails.ServerName;
+                        metadata.DatabaseName = connInfo.ConnectionDetails.DatabaseName;
+                        metadata.Name = SR.Backup_TaskName;
+                        metadata.IsCancelable = true;
+                        metadata.Data = restoreDataObject;
+
+
+                        // create backup task and perform
+                        SqlTask sqlTask = SqlTaskManager.Instance.CreateTask(metadata, this.restoreDatabaseService.RestoreTaskAsync, restoreDatabaseService.CancelTaskAsync);
```
kburtram: >CreateTask [](start = 66, length = 10)

maybe we should have another method like RunTask, so the explicit Run() call isn't required?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -101,6 +110,82 @@ public void InitializeService(ServiceHost serviceHost)
...
+                        sqlTask.Run();
+                        response.TaskId = sqlTask.TaskId.ToString();
+                    }
+                    else
+                    {
+                        response.ErrorMessage = "Failed to create restore task";
```
kburtram: >"Failed to create restore task" [](start = 48, length = 31)

should the error messages come from resource strings?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -163,6 +248,37 @@ internal static SqlConnection GetSqlConnection(ConnectionInfo connInfo)
             return null;
         }
 
+        public bool IsBackupRestoreOperationSupported(RestoreParams restoreParams, out ConnectionInfo connectionInfo)
```
kburtram: >public [](start = 8, length = 6)

does this method need to be public?  looks like it could be "private static"?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/Contracts/RestoreRequest.cs
```diff
@@ -0,0 +1,128 @@
...
+        public static readonly
+            RequestType<RestoreParams, RestoreResponse> Type =
+                RequestType<RestoreParams, RestoreResponse>.Create("disasterrecovery/restore");
+    }
+
+    public class RestorePlanRequest
+    {
+        public static readonly
+            RequestType<RestoreParams, RestorePlanResponse> Type =
+                RequestType<RestoreParams, RestorePlanResponse>.Create("disasterrecovery/restorePlan");
```
kburtram: >restorePlan [](start = 89, length = 11)

please keep all DMP message names all lower case.  this is the convention that we're following

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseService.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            if (restoreDataObject != null)
+            {
+                var backupTypes = restoreDataObject.GetBackupSetInfo();
+                return backupTypes.Any(x => x.BackupType.StartsWith(RestoreConstants.TypeFull));
+            }
+            return false;
+        }
+
+        public RestoreDatabaseTaskDataObject CreateRestoreDatabaseTaskDataObject(RestoreParams restoreParams)
```
kburtram: >CreateRestoreDatabaseTaskDataObject [](start = 45, length = 35)

please add C# XML method doc comments for public methods

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseService.cs
```diff
@@ -0,0 +1,217 @@
...
+using Microsoft.SqlTools.ServiceLayer.Connection;
+using Microsoft.SqlTools.ServiceLayer.DisasterRecovery.Contracts;
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+
+namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery.RestoreOperation
+{
+    /// <summary>
+    /// Includes method to all restore operations
+    /// </summary>
+    public class RestoreDatabaseService
```
kburtram: >RestoreDatabaseService [](start = 17, length = 22)

maybe this should be named something like RestoreDatabaseHelper since Service has a specific meaning in this app (it's registered as a service on startup).  If this isn't acting like the other services in this regard let's please use a different suffix.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseService.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// <summary>
+        /// Async task to cancel restore
+        /// </summary>
+        public async Task<TaskResult> CancelTaskAsync(SqlTask sqlTask)
+        {
+            RestoreDatabaseTaskDataObject restoreDataObject = sqlTask.TaskMetadata.Data as RestoreDatabaseTaskDataObject;
+            TaskResult taskResult = null;
+
+
+            if (restoreDataObject != null)
```
kburtram: >restoreDataObject [](start = 16, length = 17)

do you also need to null check restoreDataObject.RestorePlan or is that property always non-null?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseService.cs
```diff
@@ -0,0 +1,217 @@
...
+                response.ErrorMessage = "Failed to create restore database plan";
+            }
+            return response;
+
+        }
+
+        /// <summary>
+        /// Returns true if the restoring the restoreDataObject is supported in the service
+        /// </summary>
+        private bool CanRestore(RestoreDatabaseTaskDataObject restoreDataObject)
```
kburtram: >private [](start = 8, length = 7)

looks like this method could be static

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreUtil.cs
```diff
@@ -0,0 +1,738 @@
...
+using Microsoft.SqlServer.Management.Smo;
+
+namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery.RestoreOperation
+{
+    public class RestoreUtil
+    {
+        public RestoreUtil(Server server)
+        {
+            this.server = server;
+            this.excludedDB = new List<string> { "master", "tempdb" };
```
kburtram: >"tempdb" [](start = 59, length = 8)

do we also need to exclude either msdb or model?  or those are ok for restore?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreUtil.cs
```diff
@@ -0,0 +1,738 @@
...
+        }
+
+        /// <summary>
+        /// Current sql server instance
+        /// </summary>
+        public readonly Server server;
```
kburtram: >public [](start = 8, length = 6)

we should only have properties be public.  fields should be private

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreUtil.cs
```diff
@@ -0,0 +1,738 @@
...
+                }
+            }
+            return databaseNames;
+        }
+
+        internal DateTime GetServerCurrentDateTime()
+        {
+            DateTime dt = DateTime.MinValue;
+            /*
+            string query = "SELECT GETDATE()";
```
kburtram: dead code here and below.  If you don't need that code could you please delete..if it's being temporarily disabled could you please add a comment why it's disabled with a date and your alias, so we can keep track of when to reenable

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreUtil.cs
```diff
@@ -0,0 +1,738 @@
...
+                                databaseNames.Add(dbName);
+                            }
+                        }
+                    }
+                }
+            }
+            return databaseNames;
+        }
+
+        public bool StrEqual(string strA, string strB)
```
kburtram: > public bool StrEqual [](start = 7, length = 21)

seems like this isn't a very use method in my opinion, rather just call compare directly where needed.  If this method isn't called outside this class could you make it private to not encourage it being used else where?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/TaskServices/DatabaseOperationStub.cs
```diff
@@ -31,15 +31,15 @@ public void FailTheOperation()
...
             return await Task.Factory.StartNew(() =>
             {
                 while (!IsStopped)
                 {
                     //Just keep running
-                    if (cancellationTokenSource.Token.IsCancellationRequested)
+                    if (sqlTask.TaskStatus == SqlTaskStatus.Canceled)
                     {
-                        throw new OperationCanceledException();
+                        //throw new OperationCanceledException();
```
kburtram: > //throw new Opera [](start = 23, length = 18)

dead code..is this intentional?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/RestoreDatabaseServiceTests.cs
```diff
@@ -0,0 +1,213 @@
...
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlTools.ServiceLayer.Test.Common;
+using Microsoft.SqlTools.ServiceLayer.UnitTests;
+using Moq;
+using Xunit;
+using static Microsoft.SqlTools.ServiceLayer.IntegrationTests.Utility.LiveConnectionHelper;
+
+namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.DisasterRecovery
+{
+    public class RestoreDatabaseServiceTests : ServiceTestBase
```
kburtram: >RestoreDatabaseServiceTests [](start = 17, length = 27)

could you please make sure the checked-in BAK files are as small as possible? the general guidance is to avoid checking in binaries into a Git repo when possible.  we don't want to get a large cache of test assets checked-in.

src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
```diff
@@ -19,7 +19,7 @@
   </ItemGroup>
   <ItemGroup>
 	<PackageReference Include="System.Data.SqlClient" Version="4.4.0-preview1-25305-02" />
-	<PackageReference Include="Microsoft.SqlServer.Smo" Version="140.2.1" />
+	<PackageReference Include="Microsoft.SqlServer.Smo" Version="140.2.2" />
```
kburtram: >140.2.2" [](start = 62, length = 8)

I see an "add" for the new nuget package but I don't see a delete for the older version...could you please delete the older nuget package?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreUtil.cs
llali: this code is copied from ssms so I'm going to change it now unless we know there's a bug in ssms

---
In reply to: [126552863](https://github.com/Microsoft/sqltoolsservice/pull/403#discussion_r126552863) [](ancestors = 126552863)

PR #405, opened 2017-07-11
Add advanced options to backup service

hyoshis:
Added advanced options to backup service
Added integration test


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation.cs
```diff
@@ -139,11 +140,12 @@ public void SetBackupInput(BackupInfo input)
         /// <returns></returns>
         public BackupConfigInfo GetBackupConfigInfo(string databaseName)
```
kburtram: >GetBackupConfigInfo [](start = 32, length = 19)

nit: typically methods that create new objects are named CreateXXXX and methods that retrieve existing things are GetXXXX.  Methods that either create if not existing or get existing are GetOrCreateXXXX.  I think all or most the GetXXXX methods in this class would be better named as Create methods

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -806,6 +806,8 @@ prototype_file_noApplicableFileGroup = No Applicable Filegroup
 ############################################################################
 # Backup Service
 Backup_TaskName = Backup Database
+Backup_ServerCertificate = ServerCertificate;
```
kburtram: >ServerCertificate [](start = 27, length = 17)

Are these the correct string values?  I'd expect at least the words to have spaces between them for text we show to users..i.e., .these strings don't appear to require localization.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupTests.cs
```diff
@@ -18,11 +18,21 @@ namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.DisasterRecovery
 {
     public class BackupTests
     {
+        // Query format to create master key and certificate for backup encryption
+        private const string CreateCertificateQueryFormat = @"USE master;
+IF NOT EXISTS(SELECT * FROM sys.symmetric_keys WHERE symmetric_key_id = 101)
+CREATE MASTER KEY ENCRYPTION BY PASSWORD = 'Yukon900';
```
kburtram: >Yukon900 [](start = 44, length = 8)

we shouldn't put passwords in the source code..and it's highly likely this won't be the correct password in most environments.

kburtram:
>         ///[Fact]

should you reenable this test now assuming you've verified its stable again?

---
Refers to: test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupTests.cs:62 in f5067f8. [](commit_id = f5067f829629e7dcd5e22f95771afe3882c06f1e, deletion_comment = False)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupTests.cs
```diff
@@ -63,17 +73,11 @@ public void CreateBackupTest()
...
-            backupInfo.BackupPathList = new List<string>(new string[] { backupPath });
-            backupInfo.BackupsetName = "default_backup";
-            backupInfo.BackupType = (int)BackupType.Full;
-            backupInfo.DatabaseName = databaseName;
-            backupInfo.SelectedFileGroup = null;
-            backupInfo.SelectedFiles = "";
+
+            BackupInfo backupInfo = createBackupInfo(databaseName,
+                BackupType.Full,
+                new List<string>(new string[] { backupPath }),
```
kburtram: >new string[] { backupPath  [](start = 33, length = 26)

why create an array and then throw it away to construct a list...seems like a minor abuse of the memory heap...I think "new List<string>() { backupPath };" works fine here.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupTests.cs
```diff
@@ -94,5 +98,90 @@ public void CreateBackupTest()
...
+            testDb.RunQuery(createCertificateQuery);
+            var liveConnection = LiveConnectionHelper.InitLiveConnectionInfo(databaseName);
+
+            // Initialize backup service
+            DatabaseTaskHelper helper = AdminService.CreateDatabaseTaskHelper(liveConnection.ConnectionInfo, databaseExists: true);
+            SqlConnection sqlConn = DisasterRecoveryService.GetSqlConnection(liveConnection.ConnectionInfo);
+
+            // Get default backup path
+            BackupConfigInfo backupConfigInfo = DisasterRecoveryService.Instance.GetBackupConfigInfo(helper.DataContainer, sqlConn, sqlConn.Database);
+            string backupPath = backupConfigInfo.DefaultBackupFolder + "\\" + databaseName + ".bak";
```
kburtram: >"\\" [](start = 71, length = 4)

you should prefer Path.Combine to manually building paths...or at least use Path.PathSeparator instead of "\\"

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation.cs
hyoshis: k

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupTests.cs
hyoshis: oops that's right. changed.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupTests.cs
hyoshis: k

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupTests.cs
hyoshis: I changed to use generated password instead.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
hyoshis: These strings are from SSMS, but I also don't think these need to be localized. I removed the strings, and instead changed to pass both the encryptor type and name.

hyoshis:
Changed to randomly generate master key password in the test.

kburtram:
> using System;

please add MIT copyright header to this file

---
Refers to: test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs:1 in 640509a. [](commit_id = 640509a284bc3038da18d0d0be20f2351c4b7fe1, deletion_comment = False)

PR #409, opened 2017-07-12
Adding cancellation preemption to query execution

benrr101:
This change adds some checks of cancellation tokens during the process of executing a query. Ideally this improve perf a bit when cancelling queries and help ensure that queries are cancelled.

Additionally a superfluous ConfigureAwait during batch execution is being removed. This may help with https://github.com/Microsoft/vscode-mssql/issues/879


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -359,6 +362,9 @@ public Task<ExecutionPlan> GetExecutionPlan()
                     Columns = dataReader.Columns;
                     while (await dataReader.ReadAsync(cancellationToken))
                     {
+                        // Make sure we haven't cancelled
+                        cancellationToken.ThrowIfCancellationRequested();
```
kevcunnane: Minor: this one probably isn't needed since it's just returned from a call that should be handling cancellation checks.

PR #410, opened 2017-07-12
Fix/include dependencies fix

MrMeemus:
Updated SMO and SqlScriptPublishModel nuget package which contains fixes:
  - UTF-8 default encoding for the output file.
  - Embedded resources handling when in dot net core (Exposed when scripting dependencies).

Scripting Service Fix:
  - null exception in property selector when matching objects.
  - Duplicates being scripted when scripting dependencies due to missing server name in URN.


pensivebrian:
Missing the SMO package? I see a delete, but not an add.

---
In reply to: [314908013](https://github.com/Microsoft/sqltoolsservice/pull/410#issuecomment-314908013) [](ancestors = 314908013)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
...
-                publishModel.SelectedObjects.Add(scriptingObject.ToUrn(database));
+                publishModel.SelectedObjects.Add(scriptingObject.ToUrn(server, database));
             }
 
             return publishModel;
         }
 
+        private string GetServerNameFromLiveInstance(string connectionString)
+        {
+            SqlConnection connection = new SqlConnection(connectionString);
```
pensivebrian: >SqlConnection connection = new SqlConnection(connectionString); [](start = 12, length = 63)

Using statement

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
...
+            SqlCommand cmd = null;
```
pensivebrian: Using statement:: using(SqlCommand cmd = connection.CreateCommand()

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
...
+                Logger.Write(LogLevel.Error, string.Format("Error during SQL query for server name. Message: {0}", e.Message));
+            }
+            finally
+            {
+                if (cmd != null)
+                {
+                    cmd.Dispose();
+                }
+                connection.Close();
+            }
```
pensivebrian: We can remove moving to using statements which cleans things up.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
...
+            try {
+                connection.Open();
+            
+                cmd = connection.CreateCommand();
+                cmd.CommandText = "SELECT @@servername";
+
+                object retValue = cmd.ExecuteScalar();  
+                serverName = retValue as string;
+            }
+            catch(Exception e)
```
pensivebrian: >( [](start = 17, length = 1)

nit: space

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
...
+            {
+                Logger.Write(LogLevel.Error, string.Format("Error during SQL query for server name. Message: {0}", e.Message));
+            }
```
pensivebrian: So we're swallowing the exception. Should we just let it bubble up? Or rethrow?

MrMeemus: we should rethrow or else the user won't know, good catch!

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -75,7 +75,6 @@ public override void Execute()
                 ScriptOutputOptions outputOptions = new ScriptOutputOptions
                 {
                     SaveFileMode = ScriptFileMode.Overwrite,
-                    SaveFileType = ScriptFileType.Unicode,          // UTF-16
```
pensivebrian: Can you add a comment above that SMO is hard coded to produce a UTF8 file.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
                     "Scripting object count {0}, objects: {1}",
                     selectedObjects.Count(),
                     string.Join(", ", selectedObjects)));
-
+            
+            // Get the server name explicitly after connecting to the server.
+            // This is required to catch the actual server name when targeting sql server on docker in linux.
```
pensivebrian: > in linux [](start = 99, length = 9)

Let's remove..they could be using macOS :)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
...
             return publishModel;
         }
 
+        private string GetServerNameFromLiveInstance(string connectionString)
+        {
+            SqlConnection connection = new SqlConnection(connectionString);
+            SqlCommand cmd = null;
+            string serverName = string.Empty;
+
+            try {
```
pensivebrian: > { [](start = 15, length = 2)

nit: brace on newline

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +217,52 @@ private SqlScriptPublishModel BuildPublishModel()
...
+            {
+                if (cmd != null)
+                {
+                    cmd.Dispose();
+                }
+                connection.Close();
+            }
+
+            return serverName;
+        }
```
pensivebrian: >} [](start = 8, length = 1)

nit: newline after curly brace

pensivebrian:
:clock1:

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -89,7 +89,7 @@ public static List<ScriptingObject> GetDatabaseObjects(this SqlScriptPublishMode
         /// <param name="scriptingObject">The scripting object instance.</param>
         /// <param name="database">The name of the database referenced by the Urn.</param>
         /// <returns>The Urn instance.</returns>
-        public static Urn ToUrn(this ScriptingObject scriptingObject, string database)
+        public static Urn ToUrn(this ScriptingObject scriptingObject, string server, string database)
```
pensivebrian: >server [](start = 77, length = 6)

Need to add a Validate.IsNotNullOrWhitespaceString("server", server);

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingObjectMatcher.cs
```diff
@@ -226,9 +226,11 @@ private static IEnumerable<ScriptingObject> MatchCriteria(string property, Func<
                 }
                 if (property.EndsWith(Wildcard, StringComparison.OrdinalIgnoreCase))
                 {
-                    matchedObjects = candidates.Where(o => propertySelector(o).StartsWith(
-                        propertySelector(o).Substring(0, propertySelector(o).Length - 1),
-                        StringComparison.OrdinalIgnoreCase));
+                    matchedObjects = candidates.Where(
+                        o => 
+                            propertySelector(o) != null && 
+                            propertySelector(o).StartsWith(propertySelector(o).Substring(0, propertySelector(o).Length - 1),
```
pensivebrian: >propertySelector(o).Substring(0, propertySelector(o).Length - 1), [](start = 59, length = 65)

nit: can you move this to a newline such that it's clear both parameters are being passed to .StartsWith()

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +218,46 @@ private SqlScriptPublishModel BuildPublishModel()
...
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                Logger.Write(LogLevel.Error, string.Format("Error during SQL query for server name. Message: {0}", e.Message));
+                throw e;
+            }           
+
+            return serverName;
```
pensivebrian: >                     [](start = 30, length = 20)

nit: extra spaces

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +218,46 @@ private SqlScriptPublishModel BuildPublishModel()
...
+
+                        object retValue = cmd.ExecuteScalar();  
+                        serverName = retValue as string;
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                Logger.Write(LogLevel.Error, string.Format("Error during SQL query for server name. Message: {0}", e.Message));
+                throw e;
```
pensivebrian: >e [](start = 22, length = 1)

The 'throw e' statement will reset the callstack. I think you just want 'throw'. However, I think we should remove the try catch - the call stack will be pretty clear in the log so I don't think we need to log a message.

MrMeemus: sure will do!

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
pensivebrian: Eeek. The comment was fine...I was just referring to removing the "in linux" part of the comment.

---
In reply to: [127084230](https://github.com/Microsoft/sqltoolsservice/pull/410#discussion_r127084230) [](ancestors = 127084230)

pensivebrian: Can you restore the comment?

---
In reply to: [127087410](https://github.com/Microsoft/sqltoolsservice/pull/410#discussion_r127087410) [](ancestors = 127087410,127084230)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +218,40 @@ private SqlScriptPublishModel BuildPublishModel()
...
             }
 
             return publishModel;
         }
 
+        private string GetServerNameFromLiveInstance(string connectionString)
+        {
+            string serverName = string.Empty;
+            try
+            {
```
pensivebrian: You can remove the try {}

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +218,40 @@ private SqlScriptPublishModel BuildPublishModel()
...
-                publishModel.SelectedObjects.Add(scriptingObject.ToUrn(database));
+                publishModel.SelectedObjects.Add(scriptingObject.ToUrn(server, database));
             }
 
             return publishModel;
         }
 
+        private string GetServerNameFromLiveInstance(string connectionString)
+        {
+            string serverName = string.Empty;
```
pensivebrian: >string serverName = string.Empty; [](start = 12, length = 33)

nit: you can declare and assign this variable in one statement....removes one line.

pensivebrian: provided you move the return into the using statement, which I think makes sense.

---
In reply to: [127088389](https://github.com/Microsoft/sqltoolsservice/pull/410#discussion_r127088389) [](ancestors = 127088389)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -218,16 +218,40 @@ private SqlScriptPublishModel BuildPublishModel()
...
+                    using(SqlCommand cmd = connection.CreateCommand())
+                    {
+                        cmd.CommandText = "SELECT @@servername";
+
+                        object retValue = cmd.ExecuteScalar();  
+                        serverName = retValue as string;
+                    }
+                }
+            }   
+            return serverName;
```
pensivebrian: >                     [](start = 30, length = 20)

nit: still have trailing spaces

pensivebrian:
LGTM

PR #417, opened 2017-07-17
Use proper SecureString constructor to avoid empty password issue

MattIrv:
AdminService's CreateDatabaseTaskHelper method was failing with empty passwords because we used an unsupported SecureString constructor that does not work with empty strings. I've refactored it to use the documented constructor for SecureString objects.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
@@ -175,25 +175,22 @@ namespace Microsoft.SqlTools.ServiceLayer.Admin
         internal static DatabaseTaskHelper CreateDatabaseTaskHelper(ConnectionInfo connInfo, bool databaseExists = false)
         {
             XmlDocument xmlDoc = CreateDataContainerDocument(connInfo, databaseExists);
-            char[] passwordArray = connInfo.ConnectionDetails.Password.ToCharArray();
             CDataContainer dataContainer;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
```diff
@@ -175,25 +175,22 @@ internal static DatabaseInfo GetDatabaseInfo(ConnectionInfo connInfo)
...
-                    }
+                var passwordSecureString = new System.Security.SecureString();
+                foreach (char c in connInfo.ConnectionDetails.Password) {
+                    passwordSecureString.AppendChar(c);
                 }
+                dataContainer = new CDataContainer(
+                    CDataContainer.ServerType.SQL,
+                    connInfo.ConnectionDetails.ServerName,
+                    false,
+                    connInfo.ConnectionDetails.UserName,
```
kevcunnane: question: if password is empty do we even need to do this? Can we instead pass in null or a blank securestring? It's minor I know.

MattIrv: This code ends up passing in a blank SecureString, since it will never enter the foreach loop above if the password is empty.

We might be able to get away with passing null instead of creating a SecureString for empty passwords, but I don't know what assumptions the code further down the stack makes about whether it is null. I think it's safer to just use a blank SecureString, which clearly represents an empty string.

PR #419, opened 2017-07-19
Update SMO to pickup latest fixes

kburtram:
Update SMO for latest fixes.  Specifically this contains a fix to SMO LoginSecure property to no longer throw on non-windows platforms.


kburtram:
This is a vbump-only PR so I'll automerge.

PR #420, opened 2017-07-20
Adding more features to restore operation

llali:
- Using a generic format similar to connection details for restore request to be more generic for different database types 

 - Returning the list of database names extracted from the backup files in the plan response
 - Returning the backup sets details in the response
 - Supporting multiple backup files
 - Sending percent complete messages to client
 - Filtering the backup set based on what is selected in client
 - Creating a session for new restore plan to use when executing the restore plan
 - Fixing a bug with restore task was getting canceled after it was complete 
 - Added more tests for restore scenarios


AbbiePetcht:
:shipit:

PR #421, opened 2017-07-26
Multiple Connection Simple Execute

anthonydresser:
Changes simple execute to open a new connection on every request and close it when it is finished. I'm not convinced this is a good solution, but it does work, hoping someone else will have a better idea.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
@@ -179,95 +179,102 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
         internal async Task HandleSimpleExecuteRequest(SimpleExecuteParams executeParams,
             RequestContext<SimpleExecuteResult> requestContext)
         {
-            string randomUri = Guid.NewGuid().ToString();
-            ExecuteStringParams executeStringParams = new ExecuteStringParams
+            try
             {
-                Query = executeParams.QueryString,
-                // generate guid as the owner uri to make sure every query is unique
-                OwnerUri = randomUri
-            };
-
-            // get connection
-            ConnectionInfo connInfo;
-            if (!ConnectionService.TryFindConnection(executeParams.OwnerUri, out connInfo))
-            {
-                await requestContext.SendError(SR.QueryServiceQueryInvalidOwnerUri);
-                return;
-            }
-            
-            ConnectParams connectParams = new ConnectParams
-            {
-                OwnerUri = randomUri,
-                Connection = connInfo.ConnectionDetails,
-                Type = ConnectionType.Default
-            };
-
-            await ConnectionService.Connect(connectParams);
-
-            ConnectionInfo newConn;
-            ConnectionService.TryFindConnection(randomUri, out newConn);
-
-            Func<string, Task> queryCreateFailureAction = message => requestContext.SendError(message);
-
-            ResultOnlyContext<SimpleExecuteResult> newContext = new ResultOnlyContext<SimpleExecuteResult>(requestContext);
-
-            // handle sending event back when the query completes
-            Query.QueryAsyncEventHandler queryComplete = async q =>
-            {
-                try
+                string randomUri = Guid.NewGuid().ToString();
+                ExecuteStringParams executeStringParams = new ExecuteStringParams
                 {
-                    // check to make sure any results were recieved
-                    if (q.Batches.Length == 0 || q.Batches[0].ResultSets.Count == 0) 
-                    {
-                        await requestContext.SendError(SR.QueryServiceResultSetHasNoResults);
-                        return;
-                    } 
+                    Query = executeParams.QueryString,
+                    // generate guid as the owner uri to make sure every query is unique
+                    OwnerUri = randomUri
+                };
 
-                    var rowCount = q.Batches[0].ResultSets[0].RowCount;
-                    // check to make sure there is a safe amount of rows to load into memory
-                    if (rowCount > Int32.MaxValue) 
-                    {
-                        await requestContext.SendError(SR.QueryServiceResultSetTooLarge);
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -187,14 +190,21 @@ public void InitializeService(ServiceHost serviceHost)
...
+                Connection = connInfo.ConnectionDetails,
+                Type = ConnectionType.Default
+            };
 
-            if (connInfo.ConnectionDetails.MultipleActiveResultSets == null || connInfo.ConnectionDetails.MultipleActiveResultSets == false) {
-                // if multipleActive result sets is not allowed, don't specific a connection and make the ownerURI the true owneruri
-                connInfo = null;
-                executeStringParams.OwnerUri = executeParams.OwnerUri;
-            }
+            await ConnectionService.Connect(connectParams);
```
kevcunnane: Can this throw? Generally we're wrapping every JSON-RPC message handler with a try/catch where any error results in a requestContext.SendError being sent over the channel. I think we need it here as there's plenty going on that could throw.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -187,14 +190,21 @@ public void InitializeService(ServiceHost serviceHost)
             ConnectionInfo connInfo;
             if (!ConnectionService.TryFindConnection(executeParams.OwnerUri, out connInfo))
```
kevcunnane: Minor: I was confused by use of `executeParams.OwnerUri` vs `executeStringParams.OwnerUri`. One is the original request URI, the other is a random one used to ensure we have a dedicated connection. Given this, I'd suggest using a variable called `randomUri` or `newUri` instead of `executeStringParams.OwnerUri`.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -218,6 +232,10 @@ public void InitializeService(ServiceHost serviceHost)
                 {
                     await requestContext.SendError(SR.QueryServiceResultSetTooLarge);
                     ActiveQueries.TryRemove(executeStringParams.OwnerUri, out removedQuery);
+                    ConnectionService.Disconnect(new DisconnectParams(){
```
kevcunnane: Why not make this a finally or something instead? You always want to disconnect, so let's ensure in the code that it happens no matter what the result of the query.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
```diff
@@ -431,15 +432,14 @@ public void GetSqlTextFromInvalidType()
                 .Complete();
             await queryService.HandleSimpleExecuteRequest(queryParams, efv.Object);
 
-            Query q;
-            queryService.ActiveQueries.TryGetValue(Constants.OwnerUri, out q);
-
-            // wait on the task to finish
+            Query q = queryService.ActiveQueries.Values.First();
```
kevcunnane: It would be nice to have a test that actually verifies we can run 2 queries against the same URI using this approach.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
anthonydresser: resolved

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
anthonydresser: resolved

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -173,82 +176,105 @@ public void InitializeService(ServiceHost serviceHost)
...
+                            Rows = subset.Rows
+                        };
+                        await requestContext.SendResult(result);
+                    } 
+                    finally 
+                    {
+                        Query removedQuery;
+                        // remove the active query since we are done with it
+                        ActiveQueries.TryRemove(randomUri, out removedQuery);
+                        ConnectionService.Disconnect(new DisconnectParams(){
```
benrr101: I think you'll want to disconnect in all scenarios, in other words, I'd move this to the outside try/catch/finally block.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -173,82 +176,105 @@ public void InitializeService(ServiceHost serviceHost)
...
                 
-                SubsetParams subsetRequestParams = new SubsetParams
+                ConnectParams connectParams = new ConnectParams
                 {
-                    OwnerUri = executeStringParams.OwnerUri,
-                    BatchIndex = 0,
-                    ResultSetIndex = 0,
-                    RowsStartIndex = 0,
-                    RowsCount = Convert.ToInt32(rowCount)
+                    OwnerUri = randomUri,
```
benrr101: By connecting with a random URI, doesn't this incur a lot of overhead by kicking off intellisense discovery and whatnot?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
anthonydresser: Good question @kburtram would know more.

PR #425, opened 2017-08-02
filtering backup sets

llali:
- Filtering the backup sets based on what's selected in client before restoring
- Modifying the list of selected backup sets based on the type of the backup sets that are selected (Full backup should always be selected, if a log is not selected, any logs after that cannot be selected)
-Added tests to verify recovering database scenario (Creating full, diff, and logs backups and restoring the database using the generated backup files)


AbbiePetcht:
:shipit:

AbbiePetcht:
>                     GroupName = "Restore options"

can you add defaultValue for all service option?

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOptionsHelper.cs:105 in 560d21e. [](commit_id = 560d21e0a3fb8f429364be3efbd598c2968a5a82, deletion_comment = False)

PR #426, opened 2017-08-04
Integrate generate script with task service

hyoshis:
Iter2:
- addressed pr comments
- added unit/integration tests
- generate script for execution
----
Integrate generate script with task service
- each task defines its execution mode: execute or script
- whenever script content is generated and added to the task, the service will send notification to carbon.


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -230,18 +230,33 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
                         BackupOperation backupOperation = DisasterRecoveryService.Instance.SetBackupInput(helper.DataContainer, sqlConn, backupParams.BackupInfo);
+                        SqlTask sqlTask = null;
 
                         // create task metadata
                         TaskMetadata metadata = new TaskMetadata();
                         metadata.ServerName = connInfo.ConnectionDetails.ServerName;
-                        metadata.DatabaseName = connInfo.ConnectionDetails.DatabaseName;
-                        metadata.Name = SR.Backup_TaskName;
-                        metadata.IsCancelable = true;
+                        metadata.DatabaseName = connInfo.ConnectionDetails.DatabaseName;
```
llali: please remove the extra spaces

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -230,18 +230,33 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
-                        // create backup task and perform
-                        SqlTask sqlTask = SqlTaskManager.Instance.CreateTask(metadata, Instance.BackupTaskAsync);
-                        sqlTask.Run();
+
+                        if (backupParams.IsScripting)
+                        {
+                            metadata.Name = string.Format("{0} {1}", SR.BackupTaskName, SR.ScriptTaskName);
+                            metadata.TaskExecutionMode = TaskExecutionMode.Script;
+                            metadata.IsCancelable = false;
+                            sqlTask = SqlTaskManager.Instance.CreateTask(metadata, Instance.BackupScriptTaskAsync);
```
llali: I changed this method to get a cancelMthod also, if you use that, you can remove some of logic from you cancel method, it's up to you but might make the code cleaner

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -230,18 +230,33 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+                        }
+                        else
+                        {
+                            metadata.Name = SR.BackupTaskName;
+                            metadata.TaskExecutionMode = TaskExecutionMode.Execute;
+                            metadata.IsCancelable = true;
```
llali: what does  is cancelable = false do? does it send back the info to client?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -149,6 +150,28 @@ public BackupConfigInfo CreateBackupConfigInfo(string databaseName)
...
+            StringBuilder sb = new StringBuilder();            
+            SqlExecutionModes oldExecutionMode = this.dataContainer.Server.ConnectionContext.SqlExecutionModes;
+            this.dataContainer.Server.ConnectionContext.SqlExecutionModes = SqlExecutionModes.CaptureSql;
+            this.dataContainer.Server.ConnectionContext.CapturedSql.Clear();
+
+            this.PerformBackup();
+
+            foreach (String s in this.dataContainer.Server.ConnectionContext.CapturedSql.Text)
+            {
+                sb.AppendLine(s);
```
llali: I think this will add \n as end of line, you should do append and add the Environment End of  line. or test the result in mac

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/Contracts/BackupRequest.cs
```diff
@@ -12,6 +12,8 @@ public class BackupParams
         public string OwnerUri { get; set; }
 
         public BackupInfo BackupInfo { get; set; }
+
+        public bool IsScripting { get; set; }
```
llali: add comment for public properties

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/Contracts/BackupRequest.cs
hyoshis: k

llali:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
hyoshis: changed to use append and then environment end of line.

hyoshis:
I found out IsCancelable is not being used. For now I set it as true for backup scripting as well. Given that the scripting will be always fast, a user won't be able to cancel the script anyway. But eventually, we could use IsCancelable property for operation we don't want to allow cancellation.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/IBackupOperation.cs
```diff
@@ -38,7 +38,12 @@ public interface IBackupOperation
         /// Execute backup
         /// </summary>
         void PerformBackup();
```
kevcunnane: Minor: why not have a generic task API instead?
- `Execute(IExecuteParams config)`
- `Cancel`
- `OnCanceled` event
- `OnScriptCreated` event
- `OnComplete` event

`IExecuteParams`: options to script, execute, or both?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -230,21 +230,31 @@ public void InitializeService(IProtocolEndpoint serviceHost)
                     if ((sqlConn != null) && !connInfo.IsSqlDW && !connInfo.IsAzure)
                     {
                         BackupOperation backupOperation = DisasterRecoveryService.Instance.SetBackupInput(helper.DataContainer, sqlConn, backupParams.BackupInfo);
+                        SqlTask sqlTask = null;
```
kevcunnane: Why are these static methods that call into a static service? This makes it near-impossible to reliably unit test or integration test, since other runs will impact your own one. You can't parallelize, state errors will break unrelated tests, etc.
- At a minimum, stop having these be static methods and just call SetBackupInput in these methods. I recently did this for the language service

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -230,21 +230,31 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
-                        // create backup task and perform
-                        SqlTask sqlTask = SqlTaskManager.Instance.CreateTask(metadata, Instance.BackupTaskAsync);
-                        sqlTask.Run();
+                        metadata.IsCancelable = true;
+
+                        if (backupParams.IsScripting)
+                        {
+                            metadata.Name = string.Format("{0} {1}", SR.BackupTaskName, SR.ScriptTaskName);
+                            metadata.TaskExecutionMode = TaskExecutionMode.Script;
+                            sqlTask = SqlTaskManager.Instance.CreateTask(metadata, Instance.BackupScriptTaskAsync);
```
kevcunnane: again, you should never directly refer to the .Instance of other services as it blocks all unit testing. Create a local property "TaskManagerInstance" and allow tests to override its value (so you can pass in a mock).
Ideally, use the newer service implementation that takes in a ServiceProvider so you never need to do manually override, instead tests pass in a different service provider with mocks.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -230,21 +230,31 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
-                        metadata.Name = SR.Backup_TaskName;
-                        metadata.IsCancelable = true;
                         metadata.Data = backupOperation;
-                        
-                        // create backup task and perform
-                        SqlTask sqlTask = SqlTaskManager.Instance.CreateTask(metadata, Instance.BackupTaskAsync);
-                        sqlTask.Run();
+                        metadata.IsCancelable = true;
+
+                        if (backupParams.IsScripting)
```
kevcunnane: Again, why not have the backup params align more closely with a generic task API? So this is just a flavor and you don't need to remember to convert these things?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -230,21 +230,31 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+                        {
+                            metadata.Name = string.Format("{0} {1}", SR.BackupTaskName, SR.ScriptTaskName);
+                            metadata.TaskExecutionMode = TaskExecutionMode.Script;
+                            sqlTask = SqlTaskManager.Instance.CreateTask(metadata, Instance.BackupScriptTaskAsync);
+                            sqlTask.Run();
+                        }
+                        else
+                        {
+                            metadata.Name = SR.BackupTaskName;
```
kevcunnane: Didn't we discuss defaulting to "Execute and generate script" in this case? E.g. if you're executing, why not still script first then execute so users can review what happened?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -332,106 +342,122 @@ internal void PerformBackup(BackupOperation backupOperation)
...
+                        taskResult.TaskStatus = SqlTaskStatus.Succeeded;
+                        sqlTask.AddScript(taskResult.TaskStatus, script);
+                    }
+                    catch (Exception ex)
+                    {
+                        taskResult.TaskStatus = SqlTaskStatus.Failed;
+                        taskResult.ErrorMessage = ex.Message;
+                        if (ex.InnerException != null)
+                        {
+                            taskResult.ErrorMessage += System.Environment.NewLine + ex.InnerException.Message;
```
kevcunnane: Is this standard? Do we have a pattern for how to send on detailed inner exception info already from the connection service?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -332,106 +342,122 @@ internal void PerformBackup(BackupOperation backupOperation)
...
-                }
-
-                sqlTask.AddMessage(completedTask.Result.TaskStatus == SqlTaskStatus.Failed ? completedTask.Result.ErrorMessage : SR.Task_Completed,
-                                   completedTask.Result.TaskStatus);
-                taskResult = completedTask.Result;
+                    try
+                    {
+                        script = backupOperation.ScriptBackup();
+                        taskResult.TaskStatus = SqlTaskStatus.Succeeded;
+                        sqlTask.AddScript(taskResult.TaskStatus, script);
```
kevcunnane: This should be separated from task completion. As mentioned, I'd recommend:
- Separate messages for script generation vs. task completion
- Separate messages for task execution start / in progress / complete vs. task completion.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -332,106 +342,122 @@ internal void PerformBackup(BackupOperation backupOperation)
...
-        
+
         /// <summary>
-        /// Create a backup task for execution and cancellation
+        /// Create a backup script task
         /// </summary>
         /// <param name="sqlTask"></param>
         /// <returns></returns>
-        internal async Task<TaskResult> BackupTaskAsync(SqlTask sqlTask)
+        internal async Task<TaskResult> BackupScriptTaskAsync(SqlTask sqlTask)
```
kevcunnane: As discussed, there is no need / sense to make this 2 different methods for scripting and execution. Combine into one, have a config that lets you script, then execute, then send the result.

A good pattern from SSDT:
- Create an `EnumerateSteps` method that yield returns tasks to be executed in sequence. This would build up a set of steps (e.g. `verifyInputs`, `createPreview`, `createScript`, `executeTask`) that should be run
- Have you main method just iterate over these. For each:
  - Check cancellation status and send cancelled notification / complete the task if cancelled
  - Execute the current step in a try/catch loop
  - If an exception occurred, send notification that the task failed
  - Otherwise, continue the loop until done
Once all steps are complete, send the `TaskComplete` notification and clean up

If you do it this way, you can have identical task service implementation for Backup, Restore and most other tasks. All get the same basic config (e.g. to do scripting, execution etc.) and you just need to specialize the actions. Backup can be made much smaller and we'll reduce the code overall so that new tasks are really cheap to add.

src/Microsoft.SqlTools.ServiceLayer/TaskServices/TaskExecutionMode.cs
```diff
@@ -0,0 +1,16 @@
...
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.TaskServices
+{
+    /// <summary>
+    /// Task execution mode
+    /// </summary>
+    public enum TaskExecutionMode
+    {
+        Execute,
```
kevcunnane: Please document behavior for each enum value

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -282,24 +306,28 @@ public void PerformBackup()
...
-                }             
+                    Restore restore = new Restore();
+                    restore.Devices.AddRange(this.backup.Devices);
+                    restore.Database = this.backup.Database;
+
+                    string errorMessage = null;
+                    restore.SqlVerifyLatest(this.dataContainer.Server, out errorMessage);
+                    if (errorMessage != null)
+                    {
+                        throw new Exception(errorMessage);
```
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: I found that in SmoChildFactoryBase.cs, it sets the mssage like:

string error = string.Format(CultureInfo.InvariantCulture, "Failed expanding oe children. parent:{0} error:{1} inner:{2} stacktrace:{3}", 
parent.GetNodePath(), ex.Message, ex.InnerException != null ? ex.InnerException.Message : "", ex.StackTrace);

Should I change to follow this style?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -441,7 +469,7 @@ private int GetDeviceType(string deviceName)
                 }
             }
             catch
-            {                
+            {
```
hyoshis: I've ported over this code. I changed to throw exception.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -149,6 +150,29 @@ public BackupConfigInfo CreateBackupConfigInfo(string databaseName)
         }
 
         /// <summary>
+        /// Generate script for backup
+        /// </summary>
+        public string ScriptBackup()
```
hyoshis: Yes. The high level call wraps it in try block for execution.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -149,6 +150,29 @@ public BackupConfigInfo CreateBackupConfigInfo(string databaseName)
...
+        {
+            StringBuilder sb = new StringBuilder();
+            SqlExecutionModes oldExecutionMode = this.dataContainer.Server.ConnectionContext.SqlExecutionModes;
+            this.dataContainer.Server.ConnectionContext.SqlExecutionModes = SqlExecutionModes.CaptureSql;
+            this.dataContainer.Server.ConnectionContext.CapturedSql.Clear();
+
+            this.PerformBackup();
+
+            foreach (String s in this.dataContainer.Server.ConnectionContext.CapturedSql.Text)
```
hyoshis: k good point. I moved resetting context to finally block.

src/Microsoft.SqlTools.ServiceLayer/TaskServices/TaskExecutionMode.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: changed to existing pattern

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/IBackupOperation.cs
hyoshis: Changed to generic task API.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
kevcunnane: You're correct (apologies for lack of clarity). You can just have each step be a function call, no need for individual tasks in this case.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -62,6 +64,22 @@ internal static ConnectionService ConnectionServiceInstance
...
+            {
+                if (sqlTaskManagerInstance == null)
+                {
+                    sqlTaskManagerInstance = SqlTaskManager.Instance;
+                }
+                return sqlTaskManagerInstance;
+            }
+            set
+            {
+                sqlTaskManagerInstance = value;
```
llali: >sqlTaskManagerInstance [](start = 16, length = 22)

if you're using this for tests, you should not make it static

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -310,128 +338,119 @@ private bool IsBackupRestoreOperationSupported(RestoreParams restoreParams, out
...
-                    backupOperation.PerformBackup();
-                    result.TaskStatus = SqlTaskStatus.Succeeded;
+                    try
+                    {
+                        backupOperation.Execute(sqlTask.TaskMetadata.TaskExecutionMode);
+                        result.TaskStatus = SqlTaskStatus.Succeeded;
+                        if (!String.IsNullOrEmpty(backupOperation.ScriptContent))
+                        {
+                            // Send generated script to client
+                            sqlTask.AddScript(result.TaskStatus, script);
```
kevcunnane: This is nice. One minor thing (not required) - it's a good case for just hooking up an event on the BackupOperation that calls sqlTask.AddScript. That way you don't need property checks / more knowledge about internals than is necessary. Instead you just act as a middleman and connect the operation to the task service

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -148,52 +151,62 @@ public BackupConfigInfo CreateBackupConfigInfo(string databaseName)
...
             this.SetBackupProps();
 
-            if (this.backup.Action == BackupActionType.Files)
+            try
             {
-                IDictionaryEnumerator filegroupEnumerator = this.backupInfo.SelectedFileGroup.GetEnumerator();
-                filegroupEnumerator.Reset();
-                while (filegroupEnumerator.MoveNext())
+                if (this.backup.Action == BackupActionType.Files)
                 {
```
kevcunnane: Minor: all these giant if blocks really should be their own methods. I know you probably just ported this code over, but makes it hard to read since the method is 150 lines long

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
```diff
@@ -53,7 +53,7 @@ public class BackupServiceTests
                 OwnerUri = liveConnection.ConnectionInfo.OwnerUri
             };
 
-            await DisasterRecoveryService.HandleBackupConfigInfoRequest(dbParams, requestContext.Object);
+            await DisasterRecoveryService.Instance.HandleBackupConfigInfoRequest(dbParams, requestContext.Object);
```
kevcunnane: >DisasterRecoveryService.Instance [](start = 18, length = 32)

Even for integration tests, using a `new DisasterRecoveryService()` instead of the instance is preferred. As mentioned, if there is shared state in the future this can mess up tests that are independent of each other.

kevcunnane:
:shipit:

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
hyoshis: I changed it to not be static. this is not used by tests btw.

PR #427, opened 2017-08-04
Update multiple Simple execute test for reliablility

anthonydresser:
Instead of assuming all queries will be there to await in the test, I just await on the queries that are there. This removes the index out of bounds possibility.


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultOnlyContext.cs
```diff
@@ -15,10 +15,10 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
     /// </summary>
     public class ResultOnlyContext<TResult> : IEventSender
     {
-        private readonly RequestContext<TResult> origContext;
+        private readonly RequestContext<TResult> OrigContext;
```
llali: why did you rename this?

anthonydresser: Its unrelated, I just noticed it wasn't consistent with the rest of our c# code

PR #429, opened 2017-08-07
supporting restore from db

llali:
- To be able to restore from a database, client has to set RestoreRequest.Options["ReadHeaderFromMedia"] to true and set the RestoreRequest.Options["SourceDatabaseName"] to the database name to restore from.
- Added a new request type to get restore config info. The new request is to get information about the server that user is going to restore to.  To get the list of databases that can be used for restoring from, client has to get restoreConfigInfo first.

- Added test to verify restore from database and verify getting restore config info

- fixed the database name in restore task info 

-Removed the code that was for using existing sessions for restore. The logic was not working correctly and most of the times I had to create the session so I removed the code to think about it more later. We might not need to create session at all because every time restore plan is called, a new restore plan has to be created 

- Adding the options values that was in the request to the restore plan response to send

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
@@ -90,6 +90,15 @@ namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery.RestoreOperation
         }
 
         /// <summary>
+        /// Returns the db names that have backupsets
+        /// </summary>
+        /// <returns></returns>
+        public List<String> GetDatabaseNamesWithBackupSets()
+        {
+            return Util.GetSourceDbNames();
+        }
+
+        /// <summary>
         /// Current sqlserver instance
         /// </summary>
         public Server Server;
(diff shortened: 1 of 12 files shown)


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -703,7 +786,7 @@ public void UpdateRestorePlan(bool relocateAllFiles = false)
                     this.dbFiles = this.GetDbFiles();
                     UpdateDBFilesPhysicalRelocate();
 
-                    if (relocateAllFiles)
+                    if (RelocateAllFiles)
```
AbbiePetcht: >R [](start = 24, length = 1)

relocatedAllFiles

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/RestoreOptionsHelperTests.cs
```diff
@@ -0,0 +1,275 @@
...
+            restoreParams.Options[RestoreOptionsHelper.KeepReplication] = true;
+
+            bool actual = RestoreOptionsHelper.GetOptionValue<bool>(RestoreOptionsHelper.KeepReplication, options, restoreDatabaseTaskDataObject.Object);
+            bool expected = (bool)options[RestoreOptionsHelper.KeepReplication].DefaultValue;
+
+            Assert.Equal(actual, expected);
+        }
+
+        [Fact]
+        public void KeppReplicationShouldSetToValueInRequestGivenRecoveryStateWithRecovery()
```
AbbiePetcht: >Kep [](start = 20, length = 3)

Keep

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/RestoreOptionsHelperTests.cs
```diff
@@ -0,0 +1,275 @@
...
+            Mock<IRestoreDatabaseTaskDataObject> restoreDatabaseTaskDataObject = CreateRestoreDatabaseTaskDataObject(optionValues);
+
+            Dictionary<string, RestorePlanDetailInfo> result = RestoreOptionsHelper.CreateRestorePlanOptions(restoreDatabaseTaskDataObject.Object);
+            Assert.NotNull(result);
+            VerifyOptions(result, optionValues);
+            Assert.True(result[RestoreOptionsHelper.KeepReplication].IsReadOnly);
+        }
+
+        [Fact]
+        public void KeppReplicationShouldSetToDefaultValueGivenRecoveryStateWithNoRecovery()
```
AbbiePetcht: >Kep [](start = 20, length = 3)

Keep

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/RestoreOptionsHelperTests.cs
```diff
@@ -0,0 +1,275 @@
...
+            Mock<IRestoreDatabaseTaskDataObject> restoreDatabaseTaskDataObject = CreateRestoreDatabaseTaskDataObject(optionValues);
+
+            Dictionary<string, RestorePlanDetailInfo> result = RestoreOptionsHelper.CreateRestorePlanOptions(restoreDatabaseTaskDataObject.Object);
+            Assert.NotNull(result);
+            VerifyOptions(result, optionValues);
+            Assert.False(result[RestoreOptionsHelper.StandbyFile].IsReadOnly);
+        }
+
+        [Fact]
+        public void KeppReplicationShouldNotBeReadOnlyGivenRecoveryStateWithNoRecovery()
```
AbbiePetcht: >Kep [](start = 20, length = 3)

Keep

AbbiePetcht:
:shipit:

PR #434, opened 2017-08-15
Changes to add serialization as part of the features metadata

huypha:
Added serialization into the features metadata in order to allow extensions to specify if they use their own serializer.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
@@ -11,18 +11,25 @@ namespace Microsoft.SqlTools.ServiceLayer.Utility
 {
     public class FeaturesMetadataProviderHelper
     {
-        public static FeatureMetadataProvider[] CreateFratureMetadataProviders()
+        public static FeatureMetadataProvider[] CreateFeatureMetadataProviders()
         {
-            List<FeatureMetadataProvider> featues = new List<FeatureMetadataProvider>();
+            List<FeatureMetadataProvider> features = new List<FeatureMetadataProvider>();
 
-            featues.Add(new FeatureMetadataProvider
+            features.Add(new FeatureMetadataProvider
             {
                 FeatureName = "Restore",
                 Enabled = true,
                 OptionsMetadata = RestoreOptionsHelper.CreateRestoreOptions()
             });
 
-            return featues.ToArray();
+            features.Add(new FeatureMetadataProvider
+            {
+                FeatureName = "serializationService",
+                Enabled = true,
+                OptionsMetadata = null
+            });
+
+            return features.ToArray();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
```diff
@@ -11,18 +11,25 @@ namespace Microsoft.SqlTools.ServiceLayer.Utility
...
             {
                 FeatureName = "Restore",
                 Enabled = true,
                 OptionsMetadata = RestoreOptionsHelper.CreateRestoreOptions()
             });
 
-            return featues.ToArray();
+            features.Add(new FeatureMetadataProvider
+            {
+                FeatureName = "serializationService",
```
MattIrv: nit: We should stick to a consistent naming pattern for these. Based off of "Restore" above, I'd capitalize this as "SerializationService" or just call it "Serialization"

src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
```diff
@@ -11,18 +11,25 @@ namespace Microsoft.SqlTools.ServiceLayer.Utility
...
+                Enabled = true,
+                OptionsMetadata = null
```
MattIrv: Should there be options here? What is this useful for if not to provide options?

src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
llali: +

---
In reply to: [133314893](https://github.com/Microsoft/sqltoolsservice/pull/434#discussion_r133314893) [](ancestors = 133314893)

llali:
can you please add some comment explaining what the change is for?

src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
huypha: Discussed in person. Mainly want this as a simple way for sqltoolsservice (or any extension) to say that it already has its own serialization feature.

src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
```diff
@@ -11,18 +11,25 @@ namespace Microsoft.SqlTools.ServiceLayer.Utility
...
                 Enabled = true,
                 OptionsMetadata = RestoreOptionsHelper.CreateRestoreOptions()
             });
 
-            return featues.ToArray();
+            features.Add(new FeatureMetadataProvider
+            {
+                FeatureName = "serializationService",
+                Enabled = true,
+                OptionsMetadata = new SqlTools.Hosting.Contracts.ServiceOption[0]
```
abist: Just import `Microsoft.SqlTools.Hosting.Contracts` at the top, instead of instantiating here.

src/Microsoft.SqlTools.ServiceLayer/Utility/FeaturesMetadataProviderHelper.cs
huypha: Discussed in person. Keeping as is to match other services

PR #437, opened 2017-08-16
Update to SMO built with .Net 2.0 RTM SDK

kburtram:
This change picks up SMO built with .Net 2.0 RTM SDK and removes the dependency on the custom DataSet assembly.


src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
```diff
@@ -19,7 +19,7 @@
   </ItemGroup>
   <ItemGroup>
 	<PackageReference Include="System.Data.SqlClient" Version="4.4.0-preview1-25305-02" />
-	<PackageReference Include="Microsoft.SqlServer.Smo" Version="140.2.3" />
+	<PackageReference Include="Microsoft.SqlServer.Smo" Version="140.2.4" />
 	<PackageReference Include="Microsoft.SqlServer.Management.SqlScriptPublishModel" Version="140.2.0" />
```
pensivebrian: This SqlScriptPublishModel nuget has dependency on the SMO version. Just wanted to make sure the nuspec was updated when bumping the SMO version.

pensivebrian:
Finally on the supported DataSet! Nice to see!

MrMeemus:
@kburtram did you pick up SMO from SSMS Main or Rel? Just curious because we have a few changes that were never RI'd to rel.

src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
kburtram: oops, it looks like I didn't bump this instance of the SqlScriptPublishModel package reference.  It doesn't appear to have been properly revev'ed for at least a couple update iterations

kburtram:
@MrMeemus these bits came from SSMS_Main.

PR #440, opened 2017-08-21
integrate backup operation with new scriptable task

llali:
Modified backup operation to inherit SmoScriptableTaskOperation so that the scripting part covered by the base class. Removed the extra code in backup service which is already in SQL task

 Also fixed the restore option CloseExistingConnections to be readonly if database connections cannot be deleted

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
@@ -57,6 +57,8 @@ namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery.RestoreOperation
         string DefaultTargetDbName { get; }
         string TargetDatabaseName { get; set; }
 
+        bool CanDropExistingConnections { get; }
+
 
     }
     /// <summary>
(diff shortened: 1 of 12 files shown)


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -57,6 +57,8 @@ public interface IRestoreDatabaseTaskDataObject
         string DefaultTargetDbName { get; }
         string TargetDatabaseName { get; set; }
 
+        bool CanDropExistingConnections { get; }
```
kburtram: >CanDropExistingConnections [](start = 13, length = 26)

I'm a little confused why this is "CanDrop..." vs. "Drop....".  Is there situations were the user is no allowed to drop existing connections and that's what this flag is being used to indicate?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -1243,6 +1244,21 @@ internal bool DbFilesLocationAreValid()
         }
 
         /// <summary>
+        /// Returns true if can close eixisting connections for give database
+        /// </summary>
+        public bool CanDropExistingConnections
+        {
+            get
+            {
+                if(RestorePlan != null && RestorePlanner != null)
```
kburtram: >f( [](start = 17, length = 2)

nit: missing ' ' between 'f' and '('

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
llali: This property returns if it's possible to drop the connections for a database which is restoring to. 
I use the property in the interface here for testing but basically I use this property to see if user can modify the restore option (CloseConnections)

PR #441, opened 2017-08-22
added a binding queue for oe to force one operation on a node at a time

llali:
- Forcing one operation on an OE node at a time by using abinding queue 
- Fixed the bug in carbon "Add precision and scale to object explorer"
- Setting the node subtype to "temporal" for temporal tables in OE node.


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTableCustomNode.cs
```diff
@@ -29,6 +29,26 @@ public override string GetNodeCustomName(object smoObject, SmoQueryContext smoCo
...
+        public override string GetNodeSubType(object context)
+        {
+            try
+            {
+                Table table = context as Table;
+                if (table != null && table.TemporalType != TableTemporalType.None)
+                {
+                    return "Temporal";
+                }
+               // return string.Empty;
```
kburtram: dead code

PR #442, opened 2017-08-22
Feature/serialization exe

huypha:
Changes to be able to create a separate executable for a serialization service.


src/Microsoft.SqlTools.Serialization/Program.cs
```diff
@@ -0,0 +1,55 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.SqlTools.Credentials;
+using Microsoft.SqlTools.Credentials.Utility;
```
kburtram: why is serialization taking a dependency on Credentials assembly?  Serialization shouldn't use Credentials HostLoader unless we're planning to expose credentials APIs via serialization utility (I don't think we are).

src/Microsoft.SqlTools.Credentials/HostLoader.cs
```diff
@@ -18,9 +19,9 @@ public static class HostLoader
         private static object lockObject = new object();
         private static bool isLoaded;
 
-        internal static CredentialsServiceHost CreateAndStartServiceHost(SqlToolsContext sqlToolsContext)
+        public static UtilityServiceHost CreateAndStartServiceHost(SqlToolsContext sqlToolsContext)
```
MattIrv: This can go back to internal now that the common code has been pulled out

src/Microsoft.SqlTools.Credentials/Utility/CommandOptions.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.SqlTools.Credentials.Utility
     /// <summary>
     /// The command-line options helper class.
     /// </summary>
-    internal class CommandOptions
+    public class CommandOptions
```
MattIrv: These back to internal as well

src/Microsoft.SqlTools.Hosting/Hosting/Contracts/DbCellValue.cs
```diff
@@ -0,0 +1,50 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.Hosting.Hosting.Contracts
```
MattIrv: You're putting this here because it's shared by serialization and SQLToolsService? I don't think Hosting really makes sense as a place for these sort of contracts. I'd suggest putting it in one and importing it from the other, unless we're trying to avoid that for some reason?

src/Microsoft.SqlTools.Serialization/HostLoader.cs
```diff
@@ -0,0 +1,82 @@
+﻿//
```
MattIrv: This entire file appears to be mostly (entirely?) copy/pasted from the Credentials Service. Can we put it in Hosting in order to share it?

src/Microsoft.SqlTools.Serialization/Serialization.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.SqlTools.Hosting.Hosting.Contracts;
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.Serialization
```
MattIrv: Probably best for consistency with the other code if we keep this in a Contracts namespace

src/Microsoft.SqlTools.Serialization/Serialization.cs
```diff
@@ -0,0 +1,31 @@
...
+        public bool IsLast { get; set; }
+
+        public SaveResultsInfo(string saveFormat, 
+            string savePath, 
+            DbCellValue[][] rows, 
+            bool isLast)
+        {
+            this.SaveFormat = saveFormat;
+            this.SavePath = savePath;
+            this.Rows = Rows;
```
MattIrv: lower case rows

src/Microsoft.SqlTools.Serialization/SerializationService.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// </summary>
+        private SerializationService()
+        {
+        }
+
+        public void InitializeService(IProtocolEndpoint serviceHost)
+        {
+            // Register request and event handlers with the Service Host
+            serviceHost.SetRequestHandler(SaveAsRequest.Type, HandleSaveAsRequest);
+
```
MattIrv: extra line

src/Microsoft.SqlTools.Serialization/SerializationService.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                return SaveAsAsync(resultsInfo, requestContext);
+            };
+
+            await HandleRequest(doSave, requestContext, "HandleSaveAsRequest");
+        }
+
+        public async Task<SaveResultRequestResult> SaveAsAsync(SaveResultsInfo resultsInfo, RequestContext<SaveResultRequestResult> requestContext)
+        {
+            // TODO: Refactor currently available serialization code in sqltools to be utilized here
```
MattIrv: A GitHub task number included in this comment might be helpful

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -28,6 +28,7 @@
   <ItemGroup>
 	<ProjectReference Include="../Microsoft.SqlTools.Hosting/Microsoft.SqlTools.Hosting.csproj" />
 	<ProjectReference Include="../Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj" />
+	<ProjectReference Include="..\Microsoft.SqlTools.Serialization\Microsoft.SqlTools.Serialization.csproj" />
```
MattIrv: The includes above use `/` and this one uses `\`. Does it make a difference? Good to be consistent in any case

sqltoolsservice.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26510.0
+VisualStudioVersion = 15.0.26730.10
```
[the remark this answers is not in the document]
huypha: This changed automatically since I'm using the latest version of Visual Studio 2017. As far as I know, there shouldn't be any negative effects from this since the MinimumVisualStudioVersion is still the same

src/Microsoft.SqlTools.Hosting/Hosting/Contracts/DbCellValue.cs
huypha: Under Kevin's suggestion, I separated out these couple of contracts out into their own dll. We can discuss what the best option is later.

src/Microsoft.SqlTools.Serialization/HostLoader.cs
huypha: Karl suggested that I created a separate hostloader.cs for serialization. It's what we currently do for credentials and servicelayer as well. Only real difference is that each one initializes different services in InitializeRequestHandlersAndServices(). We can probably figure out a way to refactor this to be shared code if we think it's important.

src/Microsoft.SqlTools.Hosting/Hosting/Contracts/DbCellValue.cs
kevcunnane: So will that change be in a new iteration?

src/Microsoft.SqlTools.Serialization/Microsoft.SqlTools.Serialization.csproj
```diff
@@ -15,8 +15,12 @@
...
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.SqlTools.Credentials\Microsoft.SqlTools.Credentials.csproj" />
+    <ProjectReference Include="..\Microsoft.SqlTools.Contracts\Microsoft.SqlTools.Contracts.csproj" />
     <ProjectReference Include="..\Microsoft.SqlTools.Hosting\Microsoft.SqlTools.Hosting.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Contracts\" />
+  </ItemGroup>
```
MattIrv: Why is this folder include necessary? Isn't everything you need included by the ProjectReference above?

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
MattIrv: Don't see this addressed yet

src/Microsoft.SqlTools.Serialization/Utility/CommandOptions.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                for (int i = 0; i < args.Length; ++i)
+                {
+                    string arg = args[i];
+                    if (arg.StartsWith("--") || arg.StartsWith("-"))
+                    {
+                        switch (arg)
+                        {
+                            case "-enable-logging":
+                                EnableLogging = true;
```
llali: please add -logDir also

src/Microsoft.SqlTools.Serialization/Microsoft.SqlTools.Serialization.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <PackageId>MicrosoftSqlToolsSerialization</PackageId>
```
kevcunnane: As this is an exe, lets consider using simpler names. `SqlSerializationService` may be better. We want to align on better names for all our exes in the future

src/Microsoft.SqlTools.Serialization/Microsoft.SqlTools.Serialization.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <Company>Microsoft</Company>
+    <Product>SqlTools Serialization Service</Product>
```
kevcunnane: `SQL Tools Serialization Service

src/Microsoft.SqlTools.Serialization/Serialization.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.SqlTools.Hosting.Contracts;
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.Serialization
+{
+    public class SaveResultsInfo
```
kevcunnane: Isn't this a contract? You should have this in a `Contracts` folder with a `Microsoft.SqlTools.Serialization.Contracts` namespace. Also, I'd recommend putting this alongside all the other contracts, e.g. those used in query execution. You're likely going to need this for your handlers which can all be in the ServiceLayer codebase. All this executable DLL should be for is registering the service, everything else can be in a core location.

src/Microsoft.SqlTools.Serialization/Serialization.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public string SaveFormat { get; set; }
+        public string SavePath { get; set; }
+        public DbCellValue[][] Rows { get; set; }
+        public bool IsLast { get; set; }
+
+        public SaveResultsInfo(string saveFormat,
```
kevcunnane: All public classes and their properties / methods must be documented

src/Microsoft.SqlTools.Serialization/SerializationService.cs
```diff
@@ -0,0 +1,96 @@
...
+using Microsoft.SqlTools.Hosting.Protocol;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.Serialization
+{
+    /// <summary>
+    /// Service responsible for securing credentials in a platform-neutral manner. This provides
+    /// a generic API for read, save and delete credentials
+    /// </summary>
+    public class SerializationService
```
kevcunnane: For newer services, please consider using the HostedService pattern and avoid singleton instance definitions. See `TSqlFormatterService` for an example. You can just change it to 

```
[Export(typeof(IHostedService))]
public class SerializationService: HostedService<SerializationService>, IComposableService
```

src/Microsoft.SqlTools.Serialization/Utility/CommandOptions.cs
```diff
@@ -0,0 +1,89 @@
...
+//
+
+using System;
+
+namespace Microsoft.SqlTools.Serialization.Utility
+{
+    /// <summary>
+    /// The command-line options helper class.
+    /// </summary>
+    internal class CommandOptions
```
kevcunnane: Please refactor the original CommandOptions so we don't have 3 almost duplicate classes (this is a classic "Rule of 3" case where the 3rd time around you make sure to stop duplicating. Seems like the only difference is the name of the app in Usage, and that's not even correct for any of our exes since the .Net Core 2.0 move.

src/Microsoft.SqlTools.Serialization/Utility/Serialization.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.SqlTools.Hosting.Contracts;
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.Serialization.Utility
+{
+    public class SaveResultsInfo
```
kevcunnane: This is a duplicate class? See previous comment, shouldn't be here at all

src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditCell.cs
```diff
@@ -3,7 +3,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+using Microsoft.SqlTools.Hosting.Contracts;
```
kevcunnane: As mentioned please undo the namespace change and you can undo most of these

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -94,6 +95,9 @@ private static void InitializeRequestHandlersAndServices(ServiceHost serviceHost
             DisasterRecoveryService.Instance.InitializeService(serviceHost);
             serviceProvider.RegisterSingleService(DisasterRecoveryService.Instance);
 
+            SerializationService.Instance.InitializeService(serviceHost);
```
kevcunnane: Why are you initializing this here? Isn't this the one place you don't want the serialization service? I thought you'd go through the existing path for requests to the SqlToolsService since it doesn't have to pipe data out and back into itself?

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -28,6 +28,7 @@
   <ItemGroup>
 	<ProjectReference Include="../Microsoft.SqlTools.Hosting/Microsoft.SqlTools.Hosting.csproj" />
 	<ProjectReference Include="../Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj" />
+	<ProjectReference Include="../Microsoft.SqlTools.Serialization/Microsoft.SqlTools.Serialization.csproj" />
```
kevcunnane: Isn't this the exact wrong way around? Serialization will rely on the ServiceLayer code. The service layer shouldn't rely on that other exe, it's where all the core code is.

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
huypha: If we don't need to initialize the serialization service here, then I'll go ahead and remove the initialization.

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
huypha: I included the dependency to serialization here since I was initializing serialization within the host loader for the serialization service. Since I'll be removing that initialization, I can remove service layer's dependency on serialization. Once the dependency on serialization is moved, I can move the contracts I moved to hosting back to the service layer. This was mainly done earlier since trying to access the contracts in the service layer in the serialization project would cause a circular dependency. Since service layer will no longer have a dependency on serialization, then I will be free to import contracts from service layer into serialization.

src/Microsoft.SqlTools.Serialization/Utility/CommandOptions.cs
huypha: Ended up refactoring CommandOptions by making it accept an additional argument to pass in the exe name. Moved CommandOptions to Hosting utilities since keeping it in e.g. service layer would create a circular dependency between credentials and service layer. Created a subclass of CommandOptions in service layer that overrode SetLocale. This was necessary to set the internal SR culture for service layer for service layer tests. I didn't want to do this in the base class since we would need to introduce an additional service layer dependency in hosting, which would definitely not be a good idea.

src/Microsoft.SqlTools.Serialization/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+                // read command-line arguments
+                CommandOptions commandOptions = new CommandOptions(args, "Microsoft.SqlTools.Serialization.exe");
+                if (commandOptions.ShouldExit)
+                {
+                    return;
+                }
+
+                // turn on Verbose logging during early development
+                // we need to switch to Normal when preparing for public preview
+                Logger.Initialize(minimumLogLevel: LogLevel.Verbose, isEnabled: commandOptions.EnableLogging);
```
llali: you should set the log dir using the LoggingDirectory from commandOptions

src/Microsoft.SqlTools.Serialization/Program.cs
```diff
@@ -14,7 +15,7 @@ namespace Microsoft.SqlTools.Serialization
     /// </summary>
     internal class Program
     {
-        /// <summary>
+        /// <summary>s
```
MattIrv: extra s here

src/Microsoft.SqlTools.Credentials/Program.cs
```diff
@@ -22,15 +24,21 @@ internal static void Main(string[] args)
...
             {
                 // read command-line arguments
-                CommandOptions commandOptions = new CommandOptions(args);
+                CommandOptions commandOptions = new CommandOptions(args, "MicrosoftSqlToolsCredentials.exe");
                 if (commandOptions.ShouldExit)
                 {
                     return;
                 }
 
+                string logFilePath = "sqltools";
```
llali: this should be "credentials" not sqltools

src/Microsoft.SqlTools.Serialization/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+            try
+            {
+                // read command-line arguments
+                CommandOptions commandOptions = new CommandOptions(args, "Microsoft.SqlTools.Serialization.exe");
+                if (commandOptions.ShouldExit)
+                {
+                    return;
+                }
+
+                string logFilePath = "sqltools";
```
llali: change this to serialization so that it generates different log file as sqltools

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
```diff
@@ -64,13 +65,13 @@ public CommandOptions(string[] args)
             {
                 if (!string.IsNullOrEmpty(ErrorMessage) || ShouldExit)
                 {
-                    Console.WriteLine(Usage);
+                    Console.WriteLine((string) Usage);
```
abist: I'd use the `.ToString()` method instead of casting it into a string

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/Serialization.cs
```diff
@@ -0,0 +1,52 @@
+﻿using Microsoft.SqlTools.Hosting.Protocol.Contracts;
```
abist: Need copyright header at the top of the file

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -11,6 +11,7 @@
 using System.Linq;
```
abist: Copyright header

abist:
A lot of the files are missing the copyright header. Please copy them and put them in each file.

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
huypha: I'll remove the cast since it's unnecessary.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/Serialization.cs
```diff
@@ -0,0 +1,57 @@
...
+            this.Rows = Rows;
+            this.IsLast = isLast;
+        }
+    }
+
+    public class SaveAsRequest
+    {
+        public static readonly
+            RequestType<SaveResultsInfo, SaveResultRequestResult> Type =
+            RequestType<SaveResultsInfo, SaveResultRequestResult>.Create("query/saveAs");
```
llali: type name should be all lowercase

src/Microsoft.SqlTools.Serialization/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        /// <summary>
+        /// Main entry point into the Serialization Service Host
+        /// </summary>
+        internal static void Main(string[] args)
+        {
+            try
+            {
+                // read command-line arguments
+                CommandOptions commandOptions = new CommandOptions(args, "Microsoft.SqlTools.Serialization.exe");
```
llali: why the name of the exe for serialization is different than other two? for credentials it's "MicrosoftSqlToolsCredentials.exe"

huypha: That's the name of the exe generated when I run 'dotnet build -r win7-x64' from src/Microsoft.SqlTools.Serialization. Yeah, it would be ideal if the generated name was the same format. Do you know how the exe name is chosen? I tried doing a search for 'Microsoft.SqlTools.Serialization' but couldn't find anything relevant

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/Serialization.cs
huypha: It looks like 'Type' is consistently capitalized across any query/* requests we have in service layer. If you mean 'query/saveAs' should be in all lowercase, it looks like we use camelcase the majority of the time. (e.g. in BatchEvents.cs in sqltoolsservice). I'll leave it as is for now. Let me know if I'm misunderstanding

src/Microsoft.SqlTools.Serialization/Program.cs
huypha: So I'll change the name to "SqlSerializationService.exe" based off of a previous suggestion by Kevin here: https://github.com/Microsoft/sqltoolsservice/pull/442/files/25ff42794b3730543210e8ebf0318f93bf8cedea#diff-6089a6649676ec48e22c0ad42a2e8834

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
```diff
@@ -94,20 +95,25 @@ public CommandOptions(string[] args)
...
                 var str = string.Format("{0}" + Environment.NewLine +
-                    "Microsoft.SqlTools.ServiceLayer.exe " + Environment.NewLine +
+                    ServiceName + " " + Environment.NewLine +
                     "   Options:" + Environment.NewLine +
                     "        [--enable-logging]" + Environment.NewLine +
                     "        [--log-dir **] (default: current directory)" + Environment.NewLine +
                     "        [--help]" + Environment.NewLine +
                     "        [--locale **] (default: 'en')" + Environment.NewLine,
-                    ErrorMessage);
+                    (object) ErrorMessage);
```
MattIrv: Is `(object)` necessary? It wasn't there before

huypha: whoops, not sure how that got introduced. Will remove

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
```diff
@@ -70,7 +71,7 @@ public CommandOptions(string[] args)
             }
         }
 
-        internal string ErrorMessage { get; private set; }
+        public string ErrorMessage { get; private set; }
```
kevcunnane: Minor: please document these public fields.

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
```diff
@@ -144,5 +141,16 @@ private void SetLocale(string locale)
                 // into the CultureInfo constructor.
             }
         }
+
+        public void LocaleSetter(string locale)
```
kevcunnane: Is this for testing? Consider making internal. At a minimum I'd suggest changing to `SetLocalUnsafe` or something, this isn't a great name.

src/Microsoft.SqlTools.Serialization/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        /// <summary>
+        /// Main entry point into the Serialization Service Host
+        /// </summary>
+        internal static void Main(string[] args)
+        {
+            try
+            {
+                // read command-line arguments
+                CommandOptions commandOptions = new CommandOptions(args, "SqlSerializationService.exe");
```
kevcunnane: Minor: I'd suggest making the exe name and the log file path as constants for easier readibility later. It's fairly standard in C#

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -6,6 +6,7 @@
 using System;
 using System.Globalization;
 using System.Text.RegularExpressions;
+using Microsoft.SqlTools.Hosting.Contracts;
```
kevcunnane: Are these needed any more? I think with the undo of your rename you can undo these changes?

src/Microsoft.SqlTools.ServiceLayer/Program.cs
```diff
@@ -4,9 +4,9 @@
 using System;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
 using Microsoft.SqlTools.ServiceLayer.SqlContext;
-using Microsoft.SqlTools.ServiceLayer.Utility;
 using Microsoft.SqlTools.Utility;
 using System.IO;
+using Microsoft.SqlTools.ServiceLayer.Utility;
```
kevcunnane: Please organize usings, it looks like this should be back where it was before.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -11,6 +11,7 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.SqlTools.Hosting.Contracts;
```
kevcunnane: Again, please undo all these using statements unless they're really needed.

src/Microsoft.SqlTools.ServiceLayer/Utility/ServiceLayerCommandOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Globalization;
+using Microsoft.SqlTools.Hosting.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Utility
+{
+    class ServiceLayerCommandOptions : CommandOptions
+    {
+        private static string ServiceLayerServiceName = "MicrosoftSqlToolsServiceLayer.exe";
```
kevcunnane: Is this class really needed? Can't you just change where the SqlToolsServiceLayer code uses CommandOptions and pass in the exe name?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ServiceHost/SrTests.cs
```diff
@@ -2,6 +2,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Security.Authentication.ExtendedProtection;
```
kevcunnane: Please remove this, it looks like you added it by accident.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/CommandOptionsTests.cs
```diff
@@ -3,9 +3,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-using Microsoft.SqlTools.ServiceLayer.Utility;
 using System.IO;
+using Microsoft.SqlTools.Hosting.Utility;
 using Xunit;
+using Microsoft.SqlTools.ServiceLayer.Utility;
```
kevcunnane: Again, please organize usings.

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
huypha: I separated this out mainly for code re-use in ServiceLayerCommandOptions as well. It sets the global culture, which is used in string formatting for messages. 'Locale' itself is used in service layer tests, otherwise is unused. I'll add in a summary stating so and make it internal.

src/Microsoft.SqlTools.ServiceLayer/Utility/ServiceLayerCommandOptions.cs
huypha: I added this class mainly because within SetLocale of this class, we set an internal culture (Sr.Culture = CultureInfo.CurrentCulture) specifically used in service layer tests. I don't need to / am not able to when handling the locale for command option usage in credentials and serialization since SR is internal to the service layer and is used in service layer specific tests.

PR #445, opened 2017-08-28
Skip formatting if the language flavor shows this is not an MSSQL doc

kevcunnane:


src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -103,6 +110,14 @@ private static TelemetryProperties CreateTelemetryProps(bool isDocFormat)
...
         {
+            if (docFormatParams == null
+                || docFormatParams.TextDocument == null
+                || docFormatParams.TextDocument.Uri == null
+                || ShouldSkipFormatting(docFormatParams.TextDocument.Uri))
+            {
+                return Array.Empty<TextEdit>();
+            }
+
             return await Task.Factory.StartNew(() =>
```
benrr101: What's the benefit of starting up a new thread for something that returns immediately? Why not do a `Task.FromResult()` sort of thing?

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -103,6 +110,14 @@ private static TelemetryProperties CreateTelemetryProps(bool isDocFormat)
 
         private async Task<TextEdit[]> FormatRangeAndReturnEdits(DocumentRangeFormattingParams docFormatParams)
         {
+            if (docFormatParams == null
+                || docFormatParams.TextDocument == null
+                || docFormatParams.TextDocument.Uri == null
+                || ShouldSkipFormatting(docFormatParams.TextDocument.Uri))
+            {
+                return Array.Empty<TextEdit>();
```
benrr101: I'm not very familiar with this way of instantiating an empty array. Can't you do it with empty `[]`?

kevcunnane: You can do it with `new TextEdit[0]`. I just prefer this way as it has nice clarity in my opinion.

src/Microsoft.SqlTools.ServiceLayer/Formatter/TSqlFormatterService.cs
kevcunnane: I'll put the other code inside the taskfactory, but the purpose is that these are event handlers for message channel notifications. Per previous discussions we don't want to hold these up - formatting the code is usually fast but for a large file could take a second. Why hold the channel up for that?

PR #446, opened 2017-08-29
Task/script refactor

abist:
Changes in this PR -

1. We now use the mssql-scripter from sqltoolsservice for scripting. Got rid of old scripting service code in sqltoolsservice.
2. Peek definition is now generalized and now run via mssql-scripter.
3. SqlScriptPublishModel is now updated with a new SMO nuget package to support scripting system objects.
4. Select statements are now generated via the service rather than having hard-coded select statements.


src/Microsoft.SqlTools.ServiceLayer/Scripting/Scripter.cs
```diff
@@ -31,218 +32,229 @@ private void Initialize()
...
         /// <returns>String collection of scripts</returns>
-        internal StringCollection GetViewScripts(string objectName, string schemaName, ScriptingOptions scriptingOptions = null)
+        internal bool GetViewScripts(string objectName, string schemaName, string filePath)
         {
             try
             {
                 View smoObject = string.IsNullOrEmpty(schemaName) ? new View(this.Database, objectName) : new View(this.Database, objectName, schemaName);
                 smoObject.Refresh();
-                return (scriptingOptions == null) ? smoObject.Script() : smoObject.Script(scriptingOptions);
+                return true;
```
kburtram: >true [](start = 23, length = 4)

don't we need to call some methods to actually generate the view script?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -273,24 +273,28 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
                 string schemaName,
                 string objectType)
         {
-            StringCollection scripts = sqlScriptGetter(objectName, schemaName, null);
+            bool scriptSuccess = false;
```
kburtram: >bool scriptSuccess = false; [](start = 12, length = 27)

I would just declare this variable where it's first used on line #281..there's no need for it to be in scope as false at this point that I can see.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -221,8 +226,7 @@ public void Dispose()
                 disposed = true;
             }
         }
```
kburtram: nit: why is the blank line between the methods being deleted?  and while you're making changes here could you fix the indention on line #229.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -234,24 +238,24 @@ public void Dispose()
...
-                    break;
-                case("Database"):
-                    results = scripter.GetDatabaseScripts(metadata.Name, metadata.Schema);
-                    break;
-                default:
-                    results = null;
-                    break;
+                // case ("Table"):
+                //     results = scripter.GetTableScripts(metadata.Name, metadata.Schema, null);
+                //     break;
```
kburtram: dead code? please either remove, reenable, or leave a comment why the code is being disabled with the date and your alias in the comment.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -302,24 +306,24 @@ public void Dispose()
...
-                    break;
-                case("Database"):
-                    results = scripter.GetDatabaseScripts(metadata.Name, metadata.Schema, options);
-                    break;
-                default:
-                    results = null;
-                    break;
+                // case ("Table"):
+                //     results = scripter.GetTableScripts(metadata.Name, metadata.Schema, options);
+                //     break;
```
kburtram: same comment here.

src/Microsoft.SqlTools.ServiceLayer/Scripting/Contracts/ScriptingRequest.cs
```diff
@@ -5,6 +5,7 @@
 
 using System.Collections.Generic;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
```
kburtram: why is this new using required in this file?  I don't see any changes in the PR that appear to necessitate this.

src/Microsoft.SqlTools.ServiceLayer/Scripting/Scripter.cs
abist: This was my bad, a few files for the peek definition changes slid into this PR. This whole file will be deleted.

kburtram:
@abist please get the code coverage number back to at least the current baseline prior to merging.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -453,7 +431,51 @@ internal IEnumerable<Declaration> GetCompletionsForToken(ParseResult parseResult
...
+
+            List<ScriptingObject> objList = new List<ScriptingObject>();
+            List<string> inclList = new List<string>();
+            inclList.Add(objectType);
+            objList.Add(obj);
+
+            // create parameters for the scripting operation
+            ScriptingParams parameters = new ScriptingParams {
+                FilePath = tempFileName,
+                ConnectionString = ConnectionService.BuildConnectionString(this.connectionInfo.ConnectionDetails),
```
kevcunnane: So this builds a completely new connection to a Server and does scripting from that? This is going to be really slow for some scenarios, e.g. Azure.
- We should definitely look at optimizing this in the future. This is a great candidate for reuse of the ObjectExplorerService's Server object. We should be able to get the object from that. Special casing script as from the OE viewlet is one option, but I'd suggest we go further and figure out a way to share a connection for all these operations so the metadata is warmed up and ready.
- For now this is OK, but I see the need to make this right in the future so that perf is acceptable.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -186,24 +190,21 @@ private void SendEvent<TParams>(IEventSender requestContext, EventType<TParams>
         /// </summary>
         private void RunTask<T>(RequestContext<T> context, ScriptingOperation operation)
         {
-            Task.Run(() =>
```
kevcunnane: Please undo this. If you need a sync method, create a separate sync method for this operation

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -93,8 +93,7 @@ internal static LanguageService LanguageServiceInstance
         /// <param name="context"></param>
         public void InitializeService(ServiceHost serviceHost)
         {
-            serviceHost.SetRequestHandler(ScriptingScriptAsRequest.Type, HandleScriptingScriptAsRequest);
```
kevcunnane: This is still needed by VSCode-MSSQL, please do not remove as it's a breaking change. Given the other discussion points we should consider at  minimum adding a timeout to the ScriptAsRequest so that it does timeout like before, or adding a new / updated method that returns immediately with a "Script Queued" result and then notifies when scripting is complete or failed.

Otherwise we're going to hold up the entire pipeline and no other queries will be processable in this time.

abist: VSCode-MSSQL doesn't actually use this request type, so I'm letting it be removed.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -268,31 +263,44 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
...
             string tempFileName = (schemaName != null) ? Path.Combine(this.tempPath, string.Format("{0}.{1}.sql", schemaName, objectName))
                                                 : Path.Combine(this.tempPath, string.Format("{0}.sql", objectName));
 
-            if (scripts != null)
+            ScriptingScriptOperation operation = InitScriptOperation(objectName, schemaName, objectType, tempFileName);
+            operation.Execute();
+
+            // If the scripting was successful, we should see a file with the scripts
+            bool objectFound = false;
+            int lineNumber = 0;
```
kburtram: >lineNumber [](start = 16, length = 10)

why lineNumber and lineCount variables in same function?  I guess lineNumber is something like lineNumberOfCreateStatement or something like that.  Please use a better name that is ambiguous with other variables in function

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -268,31 +263,44 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
...
+            if (File.Exists(tempFileName))
```
kburtram: >ile.Exists(tempFileName) [](start = 17, length = 24)

do we need to clean up the temp file after we're done with it?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -268,31 +263,44 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
...
+                    string line;
+                    int lineCount = 0;
+                    string createSyntax = null;
+                    if (objectScriptMap.ContainsKey(objectType))
                     {
-                        string createSyntax = string.Format("CREATE {0}", objectType);
-                        if (script.IndexOf(createSyntax, StringComparison.OrdinalIgnoreCase) >= 0)
+                        createSyntax = string.Format("CREATE {0}", objectScriptMap[objectType]);
+                    }
+                    while (objectFound && (line = scriptFile.ReadLine()) != null)
```
kburtram: >objectFound &&  [](start = 27, length = 15)

objectFound is initialized to false on line #278 and I don't see it get set any where in this method.  How is this loop condition ever true?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -268,31 +263,44 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
...
+                        createSyntax = string.Format("CREATE {0}", objectScriptMap[objectType]);
+                    }
+                    while (objectFound && (line = scriptFile.ReadLine()) != null) 
+                    {
+                        if (line.IndexOf(createSyntax, StringComparison.OrdinalIgnoreCase) >= 0 &&
+                            line.IndexOf(objectName, StringComparison.OrdinalIgnoreCase) >=0)
                         {
-                            scriptFile.WriteLine(script);
-                            lineNumber = GetStartOfCreate(script, createSyntax);
+                            lineNumber = lineCount;
```
kburtram: >lineNumber = lineCount; [](start = 28, length = 23)

why not break the loop once you find the line?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -268,31 +263,44 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
...
                         }
+                        lineCount++;
                     }
                 }
+            }
+            if (objectFound)
```
kburtram: >objectFound [](start = 16, length = 11)

always false?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -453,7 +446,71 @@ internal IEnumerable<Declaration> GetCompletionsForToken(ParseResult parseResult
...
+        /// </summary>
+        /// <param name="objectName"></param>
+        /// <param name="schemaName"></param>
+        /// <param name="objectType"></param>
+        /// <param name="tempFileName"></param>
+        /// <returns></returns>
+        internal ScriptingScriptOperation InitScriptOperation(string objectName, string schemaName, string objectType, string tempFileName)
+        {            
+            // object that has to be scripted
+            ScriptingObject obj = new ScriptingObject {
```
kburtram: >ScriptingObject [](start = 38, length = 15)

Any reason we don't reuse the objectmetadata class here?  We should generally use existing classes for a particular concept.  ObjectMetadata is for exactly the idea you're representing here.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -453,7 +446,71 @@ internal IEnumerable<Declaration> GetCompletionsForToken(ParseResult parseResult
...
+			    ScriptCreateDrop = "ScriptCreate",
+			    TypeOfDataToScript = "SchemaOnly",
+			    ScriptStatistics = "ScriptStatsNone",
+			    TargetDatabaseEngineEdition = GetTargetDatabaseEngineEdition(),
+			    TargetDatabaseEngineType = GetTargetDatabaseEngineType(),
+			    ScriptCompatibilityOption = GetScriptCompatibilityOption(),
+                IncludeIfNotExists = true
+		    };
+
+            List<ScriptingObject> objList = new List<ScriptingObject>();
```
kburtram: >objList [](start = 34, length = 7)

we should prefer to not use shortened names list this.  Let's user objectList and includedList...assuming that is what the shortened names is supposed to mean.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -453,7 +446,71 @@ internal IEnumerable<Declaration> GetCompletionsForToken(ParseResult parseResult
...
+        {
+            DatabaseEngineEdition dbEngineEdition = this.serverConnection.DatabaseEngineEdition;
+            string dbEngineEditionString = targetDatabaseEngineEditionMap[dbEngineEdition];
+            return (dbEngineEditionString != null) ? dbEngineEditionString : "SqlServerEnterpriseEdition";
+        }
+
+        internal string GetScriptCompatibilityOption()
+        {
+            int serverVersion = this.serverConnection.ServerVersion.Major;
+            string dbEngineTypeString = serverVersionMap[serverVersion];
```
kburtram: >serverVersionMap[serverVersion] [](start = 40, length = 31)

any need to check the map contains this value prior to indexing, or we know the map will always contain all possible values?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -133,14 +132,21 @@ public void InitializeService(ServiceHost serviceHost)
         {
             try
             {
+                ConnectionInfo connInfo;
+                ScriptingService.ConnectionServiceInstance.TryFindConnection(
+                    parameters.ConnectionString,
+                    out connInfo);
+                if (connInfo != null)
+                {
+                    parameters.ConnectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
```
kburtram: >parameters.ConnectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails); [](start = 20, length = 98)

could you add a comment why we need to rebuild the connection string?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -133,14 +132,21 @@ public void InitializeService(ServiceHost serviceHost)
         {
             try
             {
+                ConnectionInfo connInfo;
+                ScriptingService.ConnectionServiceInstance.TryFindConnection(
+                    parameters.ConnectionString,
```
kburtram: >parameters.ConnectionString [](start = 20, length = 27)

this should be looking up connections via a URI.  Are we overloading connection string parameter?  I'd prefer to add an other parameter for owneruri and make it optional

kburtram:
Please add tests to cover the new functionality added in this PR.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
abist: Map contains all the possible values.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
abist: The function does take an ownerUri, and in this case its the connectionString. Is that what you asked, or did I not understand your question?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
abist: We have a special ScriptingObject class in sqltoolsservice, with method overrides for normal object base functions (such as equality etc). I assumed we should use the defined ScriptingObject class to be robust. Please let me know if you think otherwise.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
abist: Done

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -268,32 +267,48 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
...
-
-                    foreach (string script in scripts)
+                    string line;
+                    int lineCount = 0;
+                    string createSyntax = null;
+                    if (objectScriptMap.ContainsKey(objectType))
+                    {
+                        createSyntax = string.Format("CREATE {0}", objectScriptMap[objectType]);
+                    }
+                    while (objectFound && (line = scriptFile.ReadLine()) != null)
```
abist: >objectFound && [](start = 27, length = 14)

don't need to check objectFound since we break after setting it to true

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -268,32 +267,48 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
...
-                int lineNumber = 0;
-                using (StreamWriter scriptFile = new StreamWriter(File.Open(tempFileName, FileMode.Create, FileAccess.ReadWrite)))
+                using (StreamReader scriptFile = new StreamReader(tempFileName))
                 {
-
-                    foreach (string script in scripts)
+                    string line;
+                    int lineCount = 0;
+                    string createSyntax = null;
+                    if (objectScriptMap.ContainsKey(objectType))
```
abist: >if (objectScriptMap.ContainsKey(objectType)) [](start = 20, length = 44)

what should the else clause behavior be in this method..and please look for options to simplify the code.  perhaps though use of functions and reduce condition\looping nesting.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -453,7 +453,353 @@ internal IEnumerable<Declaration> GetCompletionsForToken(ParseResult parseResult
...
+        /// </summary>
+        /// <param name="objectName"></param>
+        /// <param name="schemaName"></param>
+        /// <param name="objectType"></param>
+        /// <param name="tempFileName"></param>
+        /// <returns></returns>
+        internal ScriptingScriptOperation InitScriptOperation(string objectName, string schemaName, string objectType, string tempFileName)
+        {            
+            // object that has to be scripted
+            ScriptingObject obj = new ScriptingObject {
```
abist: >{ [](start = 54, length = 1)

{ go on separate line in C#

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -453,7 +453,353 @@ internal IEnumerable<Declaration> GetCompletionsForToken(ParseResult parseResult
...
+        }
+
+        internal static class ScriptingUtils
+        {
+            /// <summary>
+            /// Quote the name of a given sql object.
+            /// </summary>
+            /// <param name="sqlObject">object</param>
+            /// <returns>quoted object name</returns>
+            internal static string QuoteObjectName(string sqlObject)
```
abist: >QuoteObjectName [](start = 35, length = 15)

please add unit test for this method

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
abist: connection string has a specific format that is well know.  this property should not contain input that doesn't follow that format.  if we need to pass a owneruri please define an option owneruri parameter

---
In reply to: [141198831](https://github.com/Microsoft/sqltoolsservice/pull/446#discussion_r141198831) [](ancestors = 141198831)

abist: also, since we are changing method formats for JSON-RPC DMP, please verify these changes work with mssql-scripter

---
In reply to: [141472723](https://github.com/Microsoft/sqltoolsservice/pull/446#discussion_r141472723) [](ancestors = 141472723,141198831)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Microsoft.SqlTools.ServiceLayer.IntegrationTests.csproj
```diff
@@ -27,11 +27,12 @@
     <ProjectReference Include="..\Microsoft.SqlTools.ServiceLayer.UnitTests\Microsoft.SqlTools.ServiceLayer.UnitTests.csproj" />
   </ItemGroup>
   <ItemGroup>
+	  <PackageReference Include="System.Net.Http" Version="4.3.1" />
```
abist: ><PackageReference Include="System.Net.Http" Version="4.3.1" /> [](start = 3, length = 62)

why do we need this?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Scripting/ScriptingServiceTests.cs
```diff
@@ -11,6 +11,7 @@
 using Moq;
 using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Connection;
 using System.Threading.Tasks;
```
abist: please sort these...system usings first and the rest alphabetically

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -39,6 +39,8 @@ public ScriptingScriptOperation(ScriptingParams parameters)
 
         private ScriptingParams Parameters { get; set; }
 
+        public SqlScriptPublishModel PublishModel { get; private set; }
```
pensivebrian: >public [](start = 8, length = 6)

make this private, and introduce a new property for the script result

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -71,6 +73,7 @@ public override void Execute()
                 publishModel.ScriptItemsCollected += this.OnPublishModelScriptItemsCollected;
                 publishModel.ScriptProgress += this.OnPublishModelScriptProgress;
                 publishModel.ScriptError += this.OnPublishModelScriptError;
+                publishModel.AllowSystemObjects = true;
```
pensivebrian: >publishModel.AllowSystemObjects = true; [](start = 16, length = 39)

Please make sure this doesn't impact scripting a database by including system dependencies in the script.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -19,6 +20,9 @@
 using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
 using Microsoft.SqlTools.Utility;
 using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlServer.Management.Common;
+using System.Data.SqlClient;
+using Microsoft.SqlServer.Management.Sdk.Sfc;
```
pensivebrian: nit: sort

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -93,8 +96,7 @@ internal static LanguageService LanguageServiceInstance
         /// <param name="context"></param>
         public void InitializeService(ServiceHost serviceHost)
         {
-            serviceHost.SetRequestHandler(ScriptingScriptAsRequest.Type, HandleScriptingScriptAsRequest);
-            serviceHost.SetRequestHandler(ScriptingRequest.Type, this.HandleScriptExecuteRequest);
+            serviceHost.SetRequestHandler(ScriptingRequest.Type, HandleScriptExecuteRequest);
```
pensivebrian: >  [](start = 64, length = 1)

For consistency, add 'this' back

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -133,14 +187,31 @@ public void InitializeService(ServiceHost serviceHost)
...
+                ScriptingService.ConnectionServiceInstance.TryFindConnection(
+                    parameters.OwnerUri,
+                    out connInfo);
+                if (connInfo != null)
+                {
+                    parameters.ConnectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
+                }
 
-                RunTask(requestContext, operation);
+                // if the scripting operation is for select
```
pensivebrian: >/ if the scripting operation is for select [](start = 17, length = 42)

Please add a comment how "ScriptSelect" although async, wait for the script result to return, where as other scripting operations spawn a task for the scripting operation, but return immediately.

pensivebrian: Or make it spawn a task to be consistent.

---
In reply to: [141987833](https://github.com/Microsoft/sqltoolsservice/pull/446#discussion_r141987833) [](ancestors = 141987833)

pensivebrian:
>     public class SqlScriptPublishModelTests : IClassFixture<SqlScriptPublishModelTests.ScriptingFixture>

Add tests.

---
Refers to: test/Microsoft.SqlTools.ServiceLayer.TestDriver.Tests/SqlScriptPublishModelTests.cs:20 in eed5361. [](commit_id = eed5361085df26ea083575e4bf85ff18c6ee356b, deletion_comment = False)

PR #447, opened 2017-08-30
Fixing IsMemoryOptimized not supported issue

benrr101:
This change checks whether or not the IsMemoryOptimized property is supported before loading it. If it is not supported it defaults to false.


kburtram:
@benrr101 thanks for the quick fix 😄

PR #448, opened 2017-08-30
Create remote file browser service

hyoshis:
Iter 2:
- Changed open/expand/validate requests to be asynchronous
- Each service needs to register file validation callback if needed
- Added more tests
--------
1. Created remote file browser service which gets requests to:
- open file browser
- expand a folder node
- validate selected file paths
- close file browser
2. Added tests


src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,222 @@
...
+                ConnectionInfo connInfo;
+                this.ConnectionServiceInstance.TryFindConnection(
+                    fileBrowserParams.OwnerUri,
+                    out connInfo);
+
+                if (connInfo != null)
+                {
+                    SqlConnection sqlConn = GetSqlConnection(connInfo);
+                    FileBrowserOperation browser = new FileBrowserOperation(sqlConn, fileBrowserParams.ExpandPath, fileBrowserParams.FileFilters);
+                    browser.PopulateFileTree();
```
kburtram: >browser.PopulateFileTree(); [](start = 20, length = 27)

is it fine to run this method synchronously in an method handler?  i.e., can we guarantee it will always complete in less than 500ms?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,222 @@
...
+            RequestContext<FileBrowserExpandResponse> requestContext)
+        {
+            FileBrowserExpandResponse response = new FileBrowserExpandResponse();
+
+            try
+            {
+                FileBrowserOperation browser = this.ownerToFileBrowserMap[fileBrowserParams.OwnerUri];
+                if (browser != null)
+                {
+                    browser.ExpandSelectedNode(fileBrowserParams.ExpandPath);
```
kburtram: >ExpandSelectedNode [](start = 28, length = 18)

similar comment here?  should these methods be async and send a notification when complete to avoid blocking the dispatcher?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,222 @@
...
+
+            try
+            {
+                if (fileBrowserParams.SelectedFiles != null && fileBrowserParams.SelectedFiles.Length > 0)
+                {
+                    IFileValidator fileValidator = null;
+                    switch (fileBrowserParams.ServiceType)
+                    {
+                        case ServiceConstants.Backup:
+                        case ServiceConstants.Restore:
```
kburtram: it might be a more flexible design to have services register validator callbacks rather than hard-coding the logic for other services here?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,222 @@
...
+            {
+                // increase the connection timeout to at least 30 seconds and and build connection string
+                // enable PersistSecurityInfo to handle issues in SMO where the connection context is lost in reconnections
+                int? originalTimeout = connInfo.ConnectionDetails.ConnectTimeout;
+                bool? originalPersistSecurityInfo = connInfo.ConnectionDetails.PersistSecurityInfo;
+                connInfo.ConnectionDetails.ConnectTimeout = Math.Max(30, originalTimeout ?? 0);
+                connInfo.ConnectionDetails.PersistSecurityInfo = true;
+                string connectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
+                connInfo.ConnectionDetails.ConnectTimeout = originalTimeout;
+                connInfo.ConnectionDetails.PersistSecurityInfo = originalPersistSecurityInfo;
```
kburtram: I've seen this snippet elsewhere in the code..we should move to a central place to avoid spreading this logic too many places?

also, does file browser really needs its own connection or can't it get once from its parent operation..backup in this case?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileValidator/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,143 @@
...
+using Microsoft.SqlServer.Management.Sdk.Sfc;
+using Microsoft.SqlServer.Management.Smo;
+using System.Data.SqlClient;
+
+namespace Microsoft.SqlTools.ServiceLayer.FileBrowser.FileValidator
+{
+    /// <summary>
+    /// Validate selected file paths for backup/restore operations
+    /// </summary>
+    public class DisasterRecoveryFileValidator : IFileValidator
```
kburtram: should this logic reside in the disasterrecovery directory somehere?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/FileBrowser/FileBrowserServiceTests.cs
```diff
@@ -0,0 +1,200 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.Text;
```
kburtram: please sort the using statements per coding convention..System namespaces first alphabetically then other namespaces following alphabetically

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/FileBrowser/FileBrowserServiceTests.cs
```diff
@@ -0,0 +1,200 @@
...
+using Microsoft.SqlTools.ServiceLayer.Admin;
+using Microsoft.SqlTools.ServiceLayer.DisasterRecovery.Contracts;
+using System.IO;
+
+namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.FileBrowser
+{
+    /// <summary>
+    /// File browser service tests
+    /// </summary>
+    public class FileBrowserServiceTests
```
kburtram: >FileBrowserServiceTests [](start = 17, length = 23)

please verify we aren't bring down code coverage numbers with this change.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileValidator/DisasterRecoveryFileValidator.cs
hyoshis: moved it to disasterrecovery

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/FileBrowser/FileBrowserServiceTests.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: changed

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: no, the file browser doesn't need its own connection after my fix on backup connection. I updated it to get existing connection here.

kburtram:
@hyoshis could you please follow-up on the code coverage drop in this PR?  @abist is looking into a similar issue.  Please verify if we have some global infra issue, or if it's something in the PR.  In any case, let's not merge any PRs with the huge coverage drop until we root cause the issue.

abist:
@hyoshis @kburtram My issue was that some tests were failing but AppVeyor doesn't fail when integration tests fail so that was misleading. However, I see another issue here. I took a look at Leila's last PR, mine and Kate's and the number of tests that were run were different all 3 times. For Leila there were 1270, for mine there were 1254 and for Kate there were 1216 tests. So there's definitely a problem with infra.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: changed to async.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: changed to async

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,153 @@
...
+{
+    /// <summary>
+    /// Validate selected file paths for backup/restore operations
+    /// </summary>
+    public static class DisasterRecoveryFileValidator
+    {
+        private const string localSqlServer = "(local)";
+        private const string localMachineName = ".";
+
+        public static bool ValidatePaths(FileBrowserValidateEventArgs args)
```
llali: >args [](start = 70, length = 4)

check for null

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,153 @@
...
+            bool result = true;
+            string errorMessage = "";
+            DbConnection dbConn = null;
+            ConnectionInfo connInfo;
+
+            ConnectionService.Instance.TryFindConnection(args.OwnerUri, out connInfo);
+            if (connInfo != null)
+            {
+                connInfo.TryGetConnection(Connection.ConnectionType.Default, out dbConn);
+                SqlConnection conn = ReliableConnectionHelper.GetAsSqlConnection(dbConn);
```
llali: >conn [](start = 30, length = 4)

check for null

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,153 @@
...
+                        }
+                    }
+                }
+            }
+            else
+            {
+                result = false;
+            }
+
+            if (!string.IsNullOrEmpty(errorMessage))
```
llali: >errorMessage [](start = 38, length = 12)

you don't need to return the result if you always throw exception for invalid paths

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            return result;
+        }
+
+        #region private methods
+
+        private static bool IsPathExisting(ServerConnection serverConnection, string path, ref bool IsFolder)
+        {
+            Enumerator en = null;
+            DataSet ds = new DataSet();
```
llali: DataSet is disposable, you should dispose it when you're done with it's data

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,153 @@
...
+            ds.Locale = CultureInfo.InvariantCulture;
+            Request req = new Request();
+
+            en = new Enumerator();
+            req.Urn = "Server/File[@FullName='" + Urn.EscapeString(path) + "']";
+            ds = en.Process(serverConnection, req);
```
llali: this  line of code can cause exception, you should catch the exception

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,153 @@
...
+            return false;
+        }
+
+        private static string GetMachineName(string sqlServerName)
+        {
+            string machineName = "";
+            if (sqlServerName != null)
+            {
+                // special case (local) which is accepted SQL(MDAC) but by OS
+                if ((sqlServerName.ToLowerInvariant().Trim() == localSqlServer) || (sqlServerName.ToLowerInvariant().Trim() == localMachineName))
```
llali: >sqlServerName [](start = 21, length = 13)

this will throw exception if the string doesn't have any '//'

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,193 @@
...
+            bool clustered = false;
+
+            req.Urn = "Server/Information";
+            req.Fields = new string[] { "IsClustered", "PathSeparator", "HostPlatform" };
+
+            var pathSeparator = @"\";
+            var hostPlatform = HostPlatformNames.Windows;
+            try
+            {
+                ds = enumerator.Process(sqlConnection, req);
```
llali: ds should be disposed at the end

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// Returns the PathSeparator values of the Server.
+        /// </summary>
+        /// <returns>PathSeparator</returns>
+        internal static string GetPathSeparator(Enumerator sfcEnumerator, object sqlConnectionObject)
+        {
+            var req = new Request();
+            req.Urn = "Server";
+            req.Fields = new[] { "PathSeparator" };
+
+            DataSet ds = sfcEnumerator.Process(sqlConnectionObject, req);
```
llali: this will cause you trouble when two requests getting to this line at the same time

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,193 @@
...
+
+            return Convert.ToString(ds.Tables[0].Rows[0][0], System.Globalization.CultureInfo.InvariantCulture);
```
llali: validate tables has values before getting the data

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            isFolder = null;
+            var isValid = false;
+            var request = new Request
+            {
+                Urn = "Server/File[@FullName='" + Urn.EscapeString(path) + "']",
+                Fields = new[] { "IsFile" }
+            };
+
+            DataSet dataSet = sfcEnumerator.Process(sqlConnectionObject, request);
```
llali: should be disposed

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,193 @@
...
+                        Field = "IsFile"
+                    },
+                    new OrderBy
+                    {
+                        Field = "Name"
+                    }
+                }
+            };
+
+            DataSet dataSet = sfcEnumerator.Process(sqlConnectionObject, request);
```
llali: dispose dataset

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
hyoshis: This will only be a problem if it gets multiple requests for the same connection. But the services that use filebrowser service should pass unique connection. For example, backup creates a unique connection each time opening a new dialog and it passes the uri to the file browser. This way, the file browser operations will be executed on its dedicated connection.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
hyoshis: why? Trim() removes only whitespace.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
hyoshis: k, thanks.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
hyoshis: changed to send error message instead of throwing exception

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
hyoshis: k

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
llali: why aren't you do using(DataSet ds =...) {} ? are you sure ds is never null?


---
In reply to: [137388237](https://github.com/Microsoft/sqltoolsservice/pull/448#discussion_r137388237) [](ancestors = 137388237)

kburtram:
@hyoshis there is still a 3.5% drop in code coverage.  Our baseline is only around 65% now so we should easily be able to at least meet that bar, and hopefully start pushing the number back up to the mid-80% which is our target.  Please keep adding tests until we are at least at parity with the baseline.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            errorMessage = "";
+            bool result = true;
+            DbConnection dbConn = null;
+            ConnectionInfo connInfo;
+
+            if (args != null)
+            {
+                ConnectionService.Instance.TryFindConnection(args.OwnerUri, out connInfo);
+                SqlConnection conn = null;
```
kburtram: >SqlConnection conn = null; [](start = 16, length = 26)

I would define this variable on line #46 since it doesn't appear to be used outside that scope.  typically code is cleaner if variables are not defined until they are needed and only scoped to the block they are needed.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,172 @@
...
+                {
+                    bool isLocal = false;
+                    if (string.Compare(GetMachineName(serverConnection.ServerInstance), Environment.MachineName, StringComparison.OrdinalIgnoreCase) == 0)
+                    {
+                        isLocal = true;
+                    }
+
+                    foreach (string filePath in args.FilePaths)
+                    {
+                        bool IsFolder = false;
```
kburtram: >IsFolder [](start = 29, length = 8)

local variables should be camelCased..meaning first letter should be lower-case.  Please apply this rule to all code in the is PR.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        #region private methods
+
+        private static bool IsPathExisting(ServerConnection serverConnection, string path, ref bool isFolder)
+        {
+            Enumerator en = null;
+            DataSet ds = new DataSet();
+            ds.Locale = CultureInfo.InvariantCulture;
+            Request req = new Request();
+            en = new Enumerator();
```
kburtram: >en = new Enumerator(); [](start = 12, length = 22)

why declare this variable as null on line 120 and then init it here?  seems like we should just declare it here where we init?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            return result;
+        }
+
+        #region private methods
+
+        private static bool IsPathExisting(ServerConnection serverConnection, string path, ref bool isFolder)
+        {
+            Enumerator en = null;
+            DataSet ds = new DataSet();
```
kburtram: >new DataSet(); [](start = 25, length = 14)

why is this being initialized to a new DataSet and then immediately overwritten on line #131 before being used?

we should probably declare as null, then init on #131 then make sure to null check before display on 140.  Or better yet, declare in a using block on #131 and remove the finally clause.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,172 @@
...
+            {
+                result = false;
+            }
+
+            return result;
+        }
+
+        #region private methods
+
+        private static bool IsPathExisting(ServerConnection serverConnection, string path, ref bool isFolder)
```
kburtram: >ref [](start = 91, length = 3)

why is this ref parameter?  seems like an out parameter since we overwrite it before reading it?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,172 @@
...
+            return isExisting;
+        }
+
+        private static string GetMachineName(string sqlServerName)
+        {
+            string machineName = "";
+            if (sqlServerName != null)
+            {
+                // special case (local) which is accepted SQL(MDAC) but by OS
+                if ((sqlServerName.ToLowerInvariant().Trim() == localSqlServer) || (sqlServerName.ToLowerInvariant().Trim() == localMachineName))
```
kburtram: >sqlServerName.ToLowerInvariant().Trim() [](start = 21, length = 39)

if you save this value into a variable you don't need to call the exact same functions twice in a row..probably makes the if clause more readable too.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -89,14 +108,18 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
-            // Create respore plan
+            // Create restore plan
             serviceHost.SetRequestHandler(RestorePlanRequest.Type, HandleRestorePlanRequest);
 
-            // Create respore config
+            // Create restore config
             serviceHost.SetRequestHandler(RestoreConfigInfoRequest.Type, HandleRestoreConfigInfoRequest);
+
+            // Register file path validation callbacks
+            FileBrowserServiceInstance?.RegisterValidatePathsCallback(FileValidationServiceConstants.Backup, DisasterRecoveryFileValidator.ValidatePaths);
```
kburtram: >?. [](start = 38, length = 2)

why are you using ?. here?  should the file browser service ever be null?  that seems like an invalid precondition and we should throw in that case?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,197 @@
...
+        public string fullPath;
+    }
+
+    /// <summary>
+    /// Base class for file browser
+    /// </summary>
+    public abstract class FileBrowserBase
+    {
+        private Enumerator enumerator = null;
+        protected object sqlConnection = null;
```
kburtram: >object [](start = 18, length = 6)

why don't we know the expected type for this property?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <summary>
+        /// Separator string for components of the file path. Defaults to \ for Windows and / for Linux
+        /// </summary>
+        internal string PathSeparator { get; set; }
+
+        /// <summary>
+        /// Returns the PathSeparator values of the Server.
+        /// </summary>
+        /// <returns>PathSeparator</returns>
+        internal static string GetPathSeparator(Enumerator sfcEnumerator, object sqlConnectionObject)
```
kburtram: >object [](start = 74, length = 6)

don't we know what type of connection object we're expecting?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// </summary>
+        /// <param name="enumerator"></param>
+        /// <param name="sqlConnection"></param>
+        /// <returns></returns>
+        internal static IEnumerable<FileInfo> EnumerateDrives(Enumerator enumerator, object sqlConnection)
+        {
+            // if not supplied, server name will be obtained from urn
+            Request req = new Request();
+            int nItems;
+            int i;
```
kburtram: can you declare these variables where they are used and within the scope they are needed?  seems like the variable is being reused for different purposes in different scopes.  in that case we should declare new variables for that scope (can reuse variables if desired).

also, please rename nItems to not use Hungarian notation prefixes..maybe itemCount or something like that better matches coding conventions.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,197 @@
...
+            ds = enumerator.Process(sqlConnection, req);
+
+            if (ds.Tables != null && ds.Tables.Count > 0 && ds.Tables[0].Rows != null)
+            {
+                nItems = ds.Tables[0].Rows.Count;
+                for (i = 0; i < nItems; i++)
+                {
+                    var fileInfo = new FileInfo
+                    {
+                        fileName = "",
```
kburtram: >"" [](start = 35, length = 2)

nit: string.Empty instead of string literal

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,197 @@
...
+            bool clustered = false;
+
+            req.Urn = "Server/Information";
+            req.Fields = new string[] { "IsClustered", "PathSeparator", "HostPlatform" };
+
+            var pathSeparator = @"\";
+            var hostPlatform = HostPlatformNames.Windows;
+            try
+            {
+                ds = enumerator.Process(sqlConnection, req);
```
kburtram: >enumerator.Process [](start = 21, length = 18)

can this ever return null?  if so you should null check in the various places you assume ds won't be null in this method.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
kburtram: please fix for Leila's comments...put in a using, remove the explicit dispose, and make sure ds can't be null.  thanks!

---
In reply to: [137406481](https://github.com/Microsoft/sqltoolsservice/pull/448#discussion_r137406481) [](ancestors = 137406481,137388237)

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,197 @@
...
+
+            //return from row in ds.Tables[0].Rows.Cast<DataRow>()
+            //       let isFile = Convert.ToBoolean((object)row[1], CultureInfo.InvariantCulture)
+            //       select new FileInfo
+            //       {
+            //           path = isFile ? path : Convert.ToString((object)row[2], CultureInfo.InvariantCulture),
+            //           fileName = isFile ? Convert.ToString((object)row[0], CultureInfo.InvariantCulture) : String.Empty,
+            //           folderName = isFile ? String.Empty : Convert.ToString((object)row[0], CultureInfo.InvariantCulture),
+            //           fullPath = isFile ? Convert.ToString((object)row[2], CultureInfo.InvariantCulture) : String.Empty
+            //       };
```
kburtram: please remove dead code if not needed.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
```diff
@@ -0,0 +1,237 @@
...
+            ExpandSelectedNode(this.expandPath);
+        }
+
+        /// <summary>
+        /// Expand nodes for the selected path.
+        /// </summary>
+        public void ExpandSelectedNode(string expandPath)
+        {
+            this.expandPath = expandPath;
+            if (!String.IsNullOrEmpty(this.expandPath))
```
kburtram: >String [](start = 17, length = 6)

please use string instead of String for call static methods on primitive types per the coding convention of this repo.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
```diff
@@ -0,0 +1,237 @@
...
+            {
+                var dirs = this.expandPath.TrimEnd(this.PathSeparator[0]).Split(this.PathSeparator[0]);
```
kburtram: >[0] [](start = 69, length = 3)

why are we only trimming the first character of the pathseparator?  I guess we assume the separator will only be a single character?  then why is it a string instead of a char?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,264 @@
...
+    public sealed class FileBrowserService
+    {
+        private static readonly Lazy<FileBrowserService> LazyInstance = new Lazy<FileBrowserService>(() => new FileBrowserService());
+        public static FileBrowserService Instance => LazyInstance.Value;
+
+        // Cache file browser operations for expanding node request
+        private Dictionary<string, FileBrowserOperation> ownerToFileBrowserMap = new Dictionary<string, FileBrowserOperation>();
+        private ConnectionService connectionService = null;
+
+        private Dictionary<string, ValidatePathsCallback> validatePathsCallbackMap { get; set; }
```
kburtram: >validatePathsCallbackMap [](start = 58, length = 24)

properties should use PascalCasing where the first letter is Upper-Case...fields are camelCased..this is per standard .Net coding convention guidelines.

kburtram: ..not sure why this is a property instead of field?  and why it's initialized in constructor instead of inline here, or lazily if need be.  If there isn't a reason please refactor.

---
In reply to: [137413210](https://github.com/Microsoft/sqltoolsservice/pull/448#discussion_r137413210) [](ancestors = 137413210)

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            this.validatePathsCallbackMap = new Dictionary<string, ValidatePathsCallback>();
+        }
+
+        /// <summary>
+        /// Register validate path callback
+        /// </summary>
+        /// <param name="service"></param>
+        /// <param name="callback"></param>
+        public void RegisterValidatePathsCallback(string service, ValidatePathsCallback callback)
```
kburtram: >RegisterValidatePathsCallback [](start = 20, length = 29)

why isn't this done during construction, or during InitializeService?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,264 @@
...
+
+            await ServiceHost.SendEvent(FileBrowserOpenCompleteNotification.Type, result);
+        }
+
+        internal async Task RunFileBrowserExpandTask(FileBrowserExpandParams fileBrowserParams)
+        {
+            FileBrowserExpandCompleteParams result = new FileBrowserExpandCompleteParams();
+            try
+            {
+                FileBrowserOperation browser = this.ownerToFileBrowserMap[fileBrowserParams.OwnerUri];
```
kburtram: >ownerToFileBrowserMap[fileBrowserParams.OwnerUri] [](start = 52, length = 49)

doesn't this throw if key doesn't exist?  I thought you needed to check the Dictionary with ContainsKey before accessing via the indexer?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/Contracts/FileBrowserExpandCompleteNotification.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /// <summary>
+    /// Notification for expand completion
+    /// </summary>
+    public class FileBrowserExpandCompleteNotification
+    {
+        public static readonly
+            EventType<FileBrowserExpandCompleteParams> Type =
+            EventType<FileBrowserExpandCompleteParams>.Create("filebrowser/expandComplete");
```
kburtram: DMP method names should be all lowercase

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/Contracts/FileBrowserOpenCompleteNotification.cs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /// <summary>
+    /// Notification for completing file browser opening
+    /// </summary>
+    public class FileBrowserOpenCompleteNotification
+    {
+        public static readonly
+            EventType<FileBrowserOpenCompleteParams> Type =
+            EventType<FileBrowserOpenCompleteParams>.Create("filebrowser/openComplete");
```
kburtram: >openComplete [](start = 73, length = 12)

DMP method names should be all lowercase..please check them all for this convention.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -1815,6 +1815,14 @@
     <value>Backup Database</value>
     <comment></comment>
   </data>
+  <data name="BackupPathIsFolderError" xml:space="preserve">
+    <value>The file name specified is also a directory name: {0}</value>
```
kburtram: >The file name specified is also a directory name [](start = 11, length = 48)

This might be more clear as "The path provided specifies a directory but a file path is required: {0}" or something to that extent.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
hyoshis: it can throw exception if the provided field is not supported on the server. in this case, we're wrapping it in try block because of the "IsClustered" field. if we're using basic field such as "IsFile" we shouldn't need wrap it in try catch.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: ? this is called from InitializeService of DisasterRecoveryService

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.xlf
```diff
@@ -2250,6 +2250,21 @@
         <target state="new">scripting</target>
         <note></note>
       </trans-unit>
+      <trans-unit id="BackupPathIsFolderError">
+        <source>The provided path specifies a directory but a file path is required: {0}</source>
+        <target state="new">The file name specified is also a directory name: {0}</target>
```
hyoshis: I changed the text of this string but I only see that change in "source" and "target" still has the old string. does anyone know why?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                Urn = "Server/File[@FullName='" + Urn.EscapeString(path) + "']",
+                Fields = new[] { "IsFile" }
+            };
+
+            Enumerator en = new Enumerator();
+            bool isExisting = false;
+            isFolder = false;
+
+            using (DataSet ds = en.Process(connection, req))
```
kburtram: >DataSet ds  [](start = 19, length = 11)

Process can never return null?  otherwise, we should null check ds before using

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+
+            // we need to issue different queries to get all fixed drives on a normal server, and
+            // shared drives on a cluster
+            req.Urn = clustered ? "Server/AvailableMedia[@SharedDrive=true()]" : "Server/Drive";
+            req.Fields = new[] { "Name" };
+
+            using (DataSet ds = enumerator.Process(connection, req))
+            {
+                for (int i = 0; i < ds.Tables[0].Rows.Count; i++)
```
kburtram: >ds.Tables[0] [](start = 36, length = 12)

do we need to check that Tables has a Count > 0?  I won't comment all ever reference of DataSet instances in this review, but as a general comment please review them all to make sure we are doing appropriate null checking and index bounds checking, etc.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
kburtram: oops, yeah, this makes sense

---
In reply to: [137427538](https://github.com/Microsoft/sqltoolsservice/pull/448#discussion_r137427538) [](ancestors = 137427538)

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -0,0 +1,272 @@
...
+        internal async Task HandleFileBrowserCloseRequest(
+            FileBrowserCloseParams fileBrowserParams,
+            RequestContext<FileBrowserCloseResponse> requestContext)
+        {
+            FileBrowserCloseResponse response = new FileBrowserCloseResponse();
+            if (this.ownerToFileBrowserMap.ContainsKey(fileBrowserParams.OwnerUri))
+            {
+                this.ownerToFileBrowserMap.Remove(fileBrowserParams.OwnerUri);
+                response.Succeeded = true;
+            }
```
kburtram: >} [](start = 12, length = 1)

do we need an else clause to set Succeeded to false similar to the other handlers, or is it ok to just use the default Succeeded in this case (anyways, the default would be false for bool, but doesn't hurt to be explicit if that is the intent).

hyoshis: I added else clause to make it explicit as I did for other handlers.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserBase.cs
hyoshis: In SSMS code, it's only checking if the ds.Tables[0].Rows has items. And from my own testing, as long as we provide the supported filed types, we always get ds.Tables[0]. If there's no returned value, then it won't have any rows. However, looking at the enumerator.Process code, it could return null, so I added safe null and index bound check.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs
hyoshis: In SSMS code, it's only checking if the ds.Tables[0].Rows has items. And from my own testing, as long as we provide the supported filed types, we always get ds.Tables[0]. If there's no returned value, then it won't have any rows. However, looking at the enumerator.Process code, it could return null, so I added safe null and index bound check.

PR #449, opened 2017-08-31
Close backup connection after execution

hyoshis:
Close connection after backup execution is completed.


kburtram:
@hyoshis do we have any way to write a unit test for this behavior?  if not, we should review this code to make sure it's testable..not needed for this PR though.

PR #450, opened 2017-09-01
Adding URN field to ObjectMetadata class

benrr101:
This field isn't used in the SqlToolsService, so it's ok to just add it to match the DMP.


kburtram:
@benrr101 is there any updates on this PR?  do we still need it, or should we close it, or is it blocked, etc.?

benrr101:
I must've forgotten about this one. Yes, we should still push this since it keeps our contracts up-to-date. But it doesn't break anything by not having it. 

I'll merge after the restarted appveyor build finishes.

PR #456, opened 2017-09-12
XEvent Profiler initial event handlers

kburtram:
This PR is the first set of changes for the XEvent Profiler "punch-through".  This has support to connect to an existing session and poll for new events on the default target.  

Upcoming commits will add support for creating a session and getting smarter about polling, etc.  I'm pushing these changes separately to avoid having the PR too large for review.


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1086,5 +1086,36 @@ private void HandleDisconnectTelemetry(ConnectionInfo connectionInfo)
...
+                connInfo.ConnectionDetails.ConnectTimeout = originalTimeout;
+                connInfo.ConnectionDetails.PersistSecurityInfo = originalPersistSecurityInfo;
+
+                // open a dedicated binding server connection
+                SqlConnection sqlConn = new SqlConnection(connectionString); 
+                sqlConn.Open();
+                return sqlConn;
+            }
+            catch (Exception)
+            {
```
llali: if you don't throw the exception at least log it so we know why an open connection in connection service fails here

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
```diff
@@ -81,20 +81,7 @@ public virtual string AddConnectionContext(ConnectionInfo connInfo, bool overwri
...
-                    connInfo.ConnectionDetails.ConnectTimeout = Math.Max(DefaultMinimumConnectionTimeout, originalTimeout ?? 0);
-                    connInfo.ConnectionDetails.PersistSecurityInfo = true;
-                    string connectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
-                    connInfo.ConnectionDetails.ConnectTimeout = originalTimeout;
-                    connInfo.ConnectionDetails.PersistSecurityInfo = originalPersistSecurityInfo;
-
-                    // open a dedicated binding server connection
-                    SqlConnection sqlConn = new SqlConnection(connectionString);                    
-                    sqlConn.Open();
+                    SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo);
```
llali: >sqlConn [](start = 34, length = 7)

sqlConn can be null so we should check for null here

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kburtram: I'm just moving around existing code..but you're right I should clean it up as part of the process :)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
kburtram: it looks like throwing here is the correct thing to do so that we set IsConnected to false in the catch block.  If we null check we'd need another clause to set IsConnected=false...basically we just changed a SqlException to a NullRefException that gets handled in the same place (and added better logging in ConnectionService for the exception).

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        /// <summary>
+        /// Handle request to stop a profiling session
+        /// </summary>
+        internal async Task HandleStopProfilingRequest(StopProfilingParams parameters, RequestContext<StopProfilingResult> requestContext)
+        {
+            try
+            {
+                monitor.StopMonitoringSession(parameters.SessionId);
+                await requestContext.SendResult(new StopProfilingResult());
```
llali: not getting the result from StopMonitoringSession method?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        /// <summary>
+        /// Create a new XEvent sessions per the IXEventSessionFactory contract
+        /// </summary>
+        public IXEventSession CreateXEventSession(ConnectionInfo connInfo)
+        {
+            var sqlConnection = ConnectionService.OpenSqlConnection(connInfo);
+            SqlStoreConnection connection = new SqlStoreConnection(sqlConnection);
+            Session session = ProfilerService.GetOrCreateSession(connection, "Profiler");
```
llali: >GetOrCreateSession [](start = 46, length = 18)

this method should probably return an error message ( inside the session object or other ways) for when creating session fails. If session is returned as null, we don't know why

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,247 @@
...
+            Session session = store.Sessions["Profiler"];
+            try
+            {
+                // start the session if it isn't already running
+                if (session != null && !session.IsRunning)
+                {
+                    session.Start();
+                }
+            }
+            catch { }
```
llali: this should either throw the exception or create session with error message, otherwise client won't know why session failed

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
kburtram: I think that if something goes wrong we'll raise an exception.  The error message would be constructed in the outer-most request handler.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -0,0 +1,127 @@
...
+        public List<ProfilerEvent> FilterOldEvents(List<ProfilerEvent> events)
+        {
+            if (lastSeenEvent != null)
+            {
+                // find the last event we've previously seen
+                bool foundLastEvent = false;
+                int idx = events.Count;
+                while (--idx >= 0)
+                {
+                    if (events[idx].Equals(lastSeenEvent))
```
llali: >events [](start = 24, length = 6)

event cannot be null?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
kburtram: yeah, that try\catch can safely be removed since we check IsRunning before trying to start.  Only a race if someone else starts the session in the meantime (of course that would be an error we could safely ignore).

Doesn't really matter for this review since the entire "Create session" scenario isn't implemented and will be completely rewritten in the next PR today or tomorrow.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            lock (this.sessionsLock)
+            {
+                // start the monitoring thread 
+                if (this.processorThread == null)
+                {
+                    this.processorThread = Task.Factory.StartNew(ProcessSessions);;
+                }
+
+                if (!this.monitoredSessions.ContainsKey(session.SessionId))
```
llali: >monitoredSessions [](start = 26, length = 17)

doesn't need to be concurrentdictionary?

kburtram: no, the dictionary is already inside a lock block.  I looked at concurrentdictionary and prefer to manually handle locking myself for this use case.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        /// <summary>
+        /// Verify that a start profiling request starts a profiling session
+        /// </summary>
+        [Fact]
+        public async Task TestHandleStartAndStopProfilingRequests()
+        {
+            // initialize test state
+            ProfilerService profilerService = new ProfilerService();
+            var liveConnection = LiveConnectionHelper.InitLiveConnectionInfo("master");
```
llali: in my test I use  using (SelfCleaningTempFile queryTempFile = new SelfCleaningTempFile()) to create unique uri,  otherwise other tests might use the uri you are using and it can cause the connection to fail for your  test randomly

kburtram: @llali thanks!  I'll use that instead.

PR #457, opened 2017-09-14
Handle null passwords when creating SecureStrings in AdminService

MattIrv:
AdminService's CreateDatabaseTaskHelper method crashes with a NullReferenceException when using a connection that has a blank (null) password. This PR modifies that code to account for null passwords and tests the password handling code, which was previously untested.


abist:
Based on the names of the other Unit tests in that folder, I'd rename the file to AdminServiceTests.cs. Looks good otherwise

PR #458, opened 2017-09-18
Take column into account when getting current statement query

MattIrv:
The LanguageService code to get a statement at a given cursor position currently doesn't take the cursor column into account. If there are multiple statements on the same line as the cursor, the current code will always return the first statement instead of finding the statement that corresponds to the cursor's column.

This PR updates that logic by adding a check to see if a statement ends before the cursor or starts after the cursor whenever the end/start line is the same line as the cursor's, and don't treat it as a match if so. It also adds a test for this situation.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -1712,6 +1712,11 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                             if (statement.StartLocation.LineNumber <= parserLine 
                                 && statement.EndLocation.LineNumber >= parserLine)
                             {
+                                if (statement.EndLocation.LineNumber == parserLine && statement.EndLocation.ColumnNumber < parserColumn
+                                    || statement.StartLocation.LineNumber == parserLine && statement.StartLocation.ColumnNumber > parserColumn)
+                                {
+                                    continue;
+                                }
                                 return statement.Sql;
                             }
                         }
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -15,6 +15,7 @@ using Microsoft.SqlServer.Management.SqlParser.Binder;
 using Microsoft.SqlServer.Management.SqlParser.Common;
 using Microsoft.SqlServer.Management.SqlParser.Intellisense;
 using Microsoft.SqlServer.Management.SqlParser.Parser;
+using Microsoft.SqlServer.Management.SqlParser.SqlCodeDom;
 using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.Connection;
 using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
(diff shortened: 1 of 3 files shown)


test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
```diff
@@ -37,6 +37,24 @@ public void ExecuteDocumentStatementTest()
...
+            var workspaceService = GetDefaultWorkspaceService(query);
+            var queryService = new QueryExecutionService(null, workspaceService);
+
+            // If a line has multiple statements and the cursor is somewhere in the second one
+            var cursorColumn = statement1.Length + 1;
+            var queryParams = new ExecuteDocumentStatementParams { OwnerUri = Constants.OwnerUri, Line = 0, Column = cursorColumn };
+            var queryText = queryService.GetSqlText(queryParams);
+
+            // The text should match the second statement
+            Assert.Equal(queryText, statement2);
```
kevcunnane: A couple additional checks that should be quick to add:
- Verify index 0 returns statement 1
- Verify index statement1.Length (which is just to the left of the `;`) returns statement 1
- Verify the end of the text gives statement 2

I think this should be quick and cover edge cases that might come up if someone chooses to change it and accidentally add a regression in the future.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -3,66 +3,68 @@
...
+// //     /// <summary>
+// //     /// Unit tests for ProfilerService
+// //     /// </summary>
+// //     public class ProfilerServiceTests
+// //     {   
+// //         /// <summary>
+// //         /// Test starting a profiling session and receiving event callback
+// //         /// </summary>
+// //         /// <returns></returns>
+// //         [Fact]
```
kburtram: you can just comment out the // [Fact] line to disable the test

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1712,6 +1712,11 @@ internal string ParseStatementAtPosition(string sql, int line, int column)
                             if (statement.StartLocation.LineNumber <= parserLine 
                                 && statement.EndLocation.LineNumber >= parserLine)
                             {
+                                if (statement.EndLocation.LineNumber == parserLine && statement.EndLocation.ColumnNumber < parserColumn
```
kburtram: can also verify this works for lines such as "SELECT 1;   " with the cursor in the blank space following the ";" character?  I expect users would expect the command to work if they are on a line with a single statement, but outside what the parser strictly considers the statement.  If there are multiple statements on the line and we can determine which is the "active statement" then we should just continue.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -3,66 +3,68 @@
...
+// //             requestParams.OwnerUri = testUri;
+// //             requestParams.TemplateName = "Standard";
 
-            // wait a bit for profile sessions to be polled
-            Thread.Sleep(500);
+// //             await profilerService.HandleStartProfilingRequest(requestParams, requestContext.Object);
 
-            requestContext.VerifyAll();
+// //             // wait a bit for profile sessions to be polled
+// //             Thread.Sleep(500);
```
kburtram: probably just bumping this up to something like 2000 will allow time for the threads to start on the slow CI build machines..through probably we need to add test hooks to allow the product to notify the test that the polling thread has initialized.  in any case, I'm fine with disabling until we pickup the profiling work again.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
MattIrv: Added logic to always execute the single statement on a line if there is only one

kburtram: thanks!  looks good to me.

PR #460, opened 2017-09-19
Turn off connection pooling to avoid locking

kburtram:
Currently connection pooling is enabled on intellisense and other internal connections which causes locks to be held on databases after `Close` is called on the SqlConnection.


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1101,11 +1101,19 @@ internal static SqlConnection OpenSqlConnection(ConnectionInfo connInfo)
                 // enable PersistSecurityInfo to handle issues in SMO where the connection context is lost in reconnections
                 int? originalTimeout = connInfo.ConnectionDetails.ConnectTimeout;
                 bool? originalPersistSecurityInfo = connInfo.ConnectionDetails.PersistSecurityInfo;
+                bool? originalPooling = connInfo.ConnectionDetails.Pooling;
+
                 connInfo.ConnectionDetails.ConnectTimeout = Math.Max(30, originalTimeout ?? 0);
                 connInfo.ConnectionDetails.PersistSecurityInfo = true;
+                connInfo.ConnectionDetails.Pooling = false;
```
llali: can you please add some comment here why we cannot use pooling?

PR #464, opened 2017-09-20
Localization check-in for September update

Monasr:
Check in the localized XLIFF with 100% localization for September update


kevcunnane:
Merging as this looks good and need to get build out. I ran tests locally off this branch and it's fine so must be a weird case for why 1 test failed in the lab

PR #465, opened 2017-09-20
Use a concurrent dict when mocking file stream factory

MattIrv:
A query execution test mock for IFileStreamFactory was using a `Dictionary` instead of a `ConcurrentDictionary`, which needs to be updated now that we have multithreaded tests that use the mock.

This caused the `SimpleExecuteMultipleQueriesTest` in Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs to fail intermittently with a `NullReferenceException` when calling the dictionary's add method, but was just a test issue, not a bug in production code.


test/Microsoft.SqlTools.ServiceLayer.Test.Common/MemoryFileSystem.cs
```diff
@@ -12,17 +13,17 @@ public class MemoryFileSystem
...
-        public static IFileStreamFactory GetFileStreamFactory(Dictionary<string, byte[]> storage)
+        public static IFileStreamFactory GetFileStreamFactory(ConcurrentDictionary<string, byte[]> storage)
         {
             Mock<IFileStreamFactory> mock = new Mock<IFileStreamFactory>();
             mock.Setup(fsf => fsf.CreateFile())
                 .Returns(() =>
                 {
                     string fileName = Guid.NewGuid().ToString();
-                    storage.Add(fileName, new byte[8192]);
+                    storage.TryAdd(fileName, new byte[8192]);
```
kburtram: >TryAdd [](start = 28, length = 6)

do we need to do anything if this fails?

MattIrv: This only fails if the key already exists in the dictionary. Since we're creating a guid as the key in the line before, that will never happen. The only reason for the change is because `ConcurrentDictionary` doesn't expose a regular Add method

MattIrv:
Build was failing because some code was merged into master that added more tests that use Dictionary instead of ConcurrentDictionary. Should be fixed now

PR #467, opened 2017-09-21
More ConcurrentDictionary test changes - Fix build error from PR #456

MattIrv:


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/QueryExecution/ExecuteTests.cs
```diff
@@ -1,4 +1,5 @@
 ﻿
+using System.Collections.Concurrent;
 using System.Collections.Generic;
```
benrr101: .Generic becomes unused 😄

PR #470, opened 2017-09-28
fixed couple of issues in restore

llali:
fixed the bug with the first time creating a restore plan and sometime the plan was null

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
@@ -921,8 +921,9 @@ namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery.RestoreOperation
 
         public bool ShouldCreateNewPlan()
         {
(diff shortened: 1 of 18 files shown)


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -921,8 +921,9 @@ public void CreateNewRestorePlan()
 
         public bool ShouldCreateNewPlan()
         {
-            return string.Compare(RestorePlanner.DatabaseName, this.RestoreParams.GetOptionValue<string>(RestoreOptionsHelper.SourceDatabaseName), StringComparison.InvariantCultureIgnoreCase) != 0 ||
+            return RestorePlan == null || string.Compare(RestorePlanner.DatabaseName, this.RestoreParams.GetOptionValue<string>(RestoreOptionsHelper.SourceDatabaseName), StringComparison.InvariantCultureIgnoreCase) != 0 ||
```
kburtram: >GetOptionValue [](start = 105, length = 14)

it would be nice if RestoreParams had wrapper methods to get at common option values to make the code a little more readable.  It looks like there are some already, like for BackupFilePaths below.

llali: will make it more readable in the next PR 

---
In reply to: [141914116](https://github.com/Microsoft/sqltoolsservice/pull/470#discussion_r141914116) [](ancestors = 141914116)

PR #472, opened 2017-09-30
Fix backup file validation error message

hyoshis:
Fix backup file validation error message


kburtram:
>         public static bool ValidatePaths(FileBrowserValidateEventArgs args, out string errorMessage)

function is a bit long with a lot of nesting..I count at least 7 levels of nested logic..it may clean up the code to split into multiple functions for the different pieces of logic.

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryFileValidator.cs:28 in ef75e67. [](commit_id = ef75e67d3d360c9660691a46f9e7dbe5bbbb3f6c, deletion_comment = False)

PR #473, opened 2017-10-03
Fixing the bug with connections on database make restore fail

llali:


src/Microsoft.SqlTools.ServiceLayer/Connection/FeatrueWithFullDbAccess.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.Connection
+{
+    /// <summary>
+    /// Any operation that needs full access to databas should implement this interface.
+    /// Make sure to call GainAccessToDatabase before the operation and ReleaseAccessToDatabase after
+    /// </summary>
+    public interface IFeatureWithFullDbAccess
```
kburtram: >IFeatureWithFullDbAccess [](start = 21, length = 24)

file name looks like it has typos

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -296,7 +301,8 @@ public void InitializeService(IProtocolEndpoint serviceHost)
                 if (supported && connInfo != null)
                 {
                     DatabaseTaskHelper helper = AdminService.CreateDatabaseTaskHelper(connInfo, databaseExists: true);
-                    SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo);
+                    SqlConnection sqlConn = ConnectionServiceInstance.OpenSqlConnection(connInfo);
+                    // Connection gets dscounnect when backup is done
```
kburtram: >dscounnect [](start = 39, length = 10)

disconnected

kburtram:
overall looks good.  I'd like to consider adding a connection queue to connnectionservice, similar to what I did in my pending scripting PR.  To do that we may want to consolidate some of the other connections used in the various services.  we should sync-up to go over all the places we open connections to determine which connections can use shared queue vs. require dedicated, persistent connections.

abist:
You can merge master to fix these failing tests.

PR #474, opened 2017-10-03
Fix some issues with Script As Select

kburtram:
This PR fixes issues where SELECT 1000 is failing on Azure DB and servers using non-standard ports.  

It also adds a connection queue to Connection Service, though that chance could be separated from bug fixes for SELECT 1000 failing.  @llali let's chat tomorrow about whether it makes sense to add a connection queue there and if we should move the current connection map over to a queue.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
@@ -740,12 +740,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Scripting
 
         internal string SelectFromTableOrView(Server server, Urn urn, bool isDw)
         {
-            string script = string.Empty;
             DataTable dt = GetColumnNames(server, urn, isDw);
             StringBuilder selectQuery = new StringBuilder();
 
             // build the first line
-            if ((dt != null) && (dt.Rows.Count > 0))
+            if (dt != null && dt.Rows.Count > 0)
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -794,16 +795,23 @@ internal string SelectFromTableOrView(Server server, Urn urn, bool isDw)
             // In Hekaton M5, if it's a memory optimized table, we need to provide SNAPSHOT hint for SELECT.
             if (urn.Type.Equals("Table") && ScriptingUtils.IsXTPSupportedOnServer(server))
             {
-                Table table = (Table)server.GetSmoObject(urn);
-                table.Refresh();
-                if (table.IsMemoryOptimized)
+                try
                 {
-                    selectQuery.Append(" WITH (SNAPSHOT)");
+                    Table table = (Table)server.GetSmoObject(urn);
```
kburtram: this is throwing exceptions for both Azure DB and servers with non-standard ports.  Parsing out the port avoids the exception for on-prem servers.  I'm not sure what to about Azure DB.  It looks like the URN parser what's the server to not include ".database.window.net", but I didn't investigate.  Swallowing the exception is probably fine in most cases, as InMemory tables in Azure DB are currently rare and the user impact is the SELECT won't include WITH (SNAPSHOT) in those rare cases.

PR #477, opened 2017-10-04
Port Azure code from SSDT to the tools service

kevcunnane:
Porting of the vast majority of Azure-related code from SSDT. This is very large, so I want to put this out as one large "lift and shift" PR before I do the tools-service specific JSON-RPC service handlers, connect a new account handler (as the code to get necessary info from accounts and subscriptions isn't fully complete) and add tests over these

**What's in this PR**:
- Created 3 new projects:
  - Microsoft.SqlTools.ResourceProvider will host the executable that accepts requests for Azure-related actions over the JSON-RPC protocol. This must be separate from other DLLs since a direct dependency on the Azure SDK DLLs fails (they're NetStandard 1.4 and you can't reference them if you have RuntimeIdentifiers in your .csproj file)
  - Microsoft.SqlTools.ResourceProvider.Core is where all the main business logic is, including definitions and logic on how to navigate over resources and create firewall rules, etc.
  - Microsoft.SqlTools.ResourceProvider.DefaultImpl is the actual Azure implementation of the resource provider APIs. The reason for separating this is to support eventual integration back into other tools (since their Azure and Identity services will be different).
- Implemented the `AzureResourceManager` that connects to Azure via ARM APIs and handles creating firewall rule and querying databases. The dependent DLLs have had major breaking changes, so will need additional verification to ensure this works as expected
- Ported the unit tests for all code that was not a viewmodel. Viewmodel test code will be ported in a future update as we plumb through a service-equivalent to these. Also, the DependencyManager code which has overlap with our service provider code is commented out. Will work to uncomment in a future update as it has value to test some scenarios

**What's not in this PR**:
- Identity Services. We currently just have a stub for the interface, and even that will likely change a little
- anything JSON-RPC or registered service related. These will be adapted from the viewmodels and added in a separate PR


src/Microsoft.SqlTools.ResourceProvider/Microsoft.SqlTools.ResourceProvider.csproj
```diff
@@ -0,0 +1,28 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.SqlTools.Hosting\Microsoft.SqlTools.Hosting.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="Localization\sr.resx" />
+    <None Include="Localization\sr.strings" />
+  </ItemGroup>
+</Project>
```
llali: you need to add the project to codecoverage.bat if you want to see code coverage on the classes in this project

kevcunnane: Thanks, I didn't realize this. Added this. If we add a lot more DLLs in the future, we'll want to streamline this (e.g. having a different build configuration and setting these via that so we just change the property and it all works).

src/Microsoft.SqlTools.ResourceProvider/Microsoft.SqlTools.ResourceProvider.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <Description>Provides Resource Provider and control plane support.</Description>
+    <Copyright>� Microsoft Corporation. All rights reserved.</Copyright>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <PreserveCompilationContext>true</PreserveCompilationContext>
+    <DebugType>portable</DebugType>
+	  <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
+    <RuntimeIdentifiers>win7-x64;win7-x86;ubuntu.14.04-x64;ubuntu.16.04-x64;centos.7-x64;rhel.7.2-x64;debian.8-x64;fedora.23-x64;opensuse.13.2-x64;osx.10.11-x64</RuntimeIdentifiers>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DefineConstants>TRACE;DEBUG;NETCOREAPP1_0;NETCOREAPP2_0</DefineConstants>
```
kburtram: >;NETCOREAPP1_0;NETCOREAPP2_0 [](start = 32, length = 28)

do we need to define these constants?

src/Microsoft.SqlTools.ResourceProvider/ResourceProviderHostLoader.cs
```diff
@@ -0,0 +1,82 @@
...
+        private static void InitializeRequestHandlersAndServices(UtilityServiceHost serviceHost, SqlToolsContext sqlToolsContext)
+        {
+            // Load extension provider, which currently finds all exports in current DLL. Can be changed to find based
+            // on directory or assembly list quite easily in the future
+            ExtensionServiceProvider serviceProvider = ExtensionServiceProvider.CreateDefaultServiceProvider();
+            serviceProvider.RegisterSingleService(sqlToolsContext);
+            serviceProvider.RegisterSingleService(serviceHost);
+
+            // CredentialService.Instance.InitializeService(serviceHost);
+            // serviceProvider.RegisterSingleService(CredentialService.Instance);
```
kburtram: dead code

src/Microsoft.SqlTools.ResourceProvider/Microsoft.SqlTools.ResourceProvider.csproj
kevcunnane: It appears these are built in - I'm a little unclear on how this works but I think that when I edited things in VS2017 these got added

src/Microsoft.SqlTools.ResourceProvider.Core/Authentication/IAccountManager.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// Returns the current logged in user
+        /// </summary>
+        Task<IUserAccount> GetCurrentAccountAsync();
+
+        /// <summary>
+        /// Returns true if the API supports a login control
+        /// </summary>
+        bool HasLoginDialog
```
kburtram: >HasLoginDialog [](start = 13, length = 14)

I assume a lot of the code is coming over as-is, so I won't comment further on code that may not make sense in the tools service context.

src/Microsoft.SqlTools.ResourceProvider/ResourceProviderHostLoader.cs
kevcunnane: Thanks - fixed in the separate branch I'm working on that actually implements the `ResourceProviderService`. I may just fix in that separate PR since it'll have the real service checked in.

src/Microsoft.SqlTools.ResourceProvider.Core/Common/AutoLock.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// A wrapper around the ReaderWriterLock to make sure the locks are released even if the action fails
+    /// </summary>
+    public class AutoLock
+    {
+        private readonly ReaderWriterLock _lock;
+        private readonly bool _isWriteLocked;
+
+        /// <summary>
+        /// Creates new lock given type of lock and timeput
```
kburtram: >timeput [](start = 52, length = 7)

timeout

src/Microsoft.SqlTools.ResourceProvider.Core/Common/ConcurrentCache.cs
```diff
@@ -0,0 +1,82 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.SqlTools.ResourceProvider.Core
+{
+    internal class ConcurrentCache<T>
```
kburtram: >ConcurrentCache [](start = 19, length = 15)

some of the "utility" classes look like they may be interesting in other use cases.  it may be worthwhile to stick them in hosting or another shared library so they've readily available.

src/Microsoft.SqlTools.ResourceProvider.Core/Discovery/AzureDatabaseDiscoveryProvider.cs
```diff
@@ -0,0 +1,327 @@
...
+        //public async Task<ServiceResponse<DatabaseInstanceInfo>> GetDatabaseInstancesAsync(UIConnectionInfo uiConnectionInfo, CancellationToken cancellationToken)
+        //{
+        //    ServiceResponse<DatabaseInstanceInfo> result = null;
+        //    if (DefaultDatabaseDiscoveryProvider != null && DefaultDatabaseDiscoveryProvider != this)
+        //    {
+        //        result = await DefaultDatabaseDiscoveryProvider.GetDatabaseInstancesAsync(uiConnectionInfo, cancellationToken);
+        //    }
+        //    else
+        //    {
+        //        result = new ServiceResponse<DatabaseInstanceInfo>(); //TODO: add error that we couldn't find any default database provider
```
kburtram: dead code

src/Microsoft.SqlTools.ResourceProvider.Core/Discovery/IDatabaseDiscoveryProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Returns the databases for given server name. 
+        /// </summary> 
+        Task<ServiceResponse<DatabaseInstanceInfo>> GetDatabaseInstancesAsync(string serverName, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Returns the databases for given connection info. 
+        /// The connection info should be used to make the connection for getting databases not the account manager
+        /// </summary> 
+        //Task<ServiceResponse<DatabaseInstanceInfo>> GetDatabaseInstancesAsync(UIConnectionInfo uiConnectionInfo, CancellationToken cancellationToken);
```
kburtram: dead code

src/Microsoft.SqlTools.ResourceProvider.Core/Discovery/ServerTypes.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Microsoft.SqlTools.ResourceProvider.Core
+{
+    /// <summary>
+    /// List of built-in server types used in <see cref="ExportableAttribute" />. 
+    /// Defines a server grouping based on the type of server connection supported (SQL Server, Reporting Server, Analysis Server)
+    /// Additional server types can be defined as needed.
+    /// Note that the Connection Dialog UI may require server type to be set for some resource types such as<see cref="IServerDiscoveryProvider" />. 
+    /// In addition a UI section matching that category may be required, or else the provider will not be used by any UI part and never be called.
+    /// </summary>
+    public static class ServerTypes
```
kburtram: >ServerTypes [](start = 24, length = 11)

at some point it may be worth seeing if we can merge some of the database\server object model code with the similar code we have in the toolsservice project to avoid duplicate, but different, object models.

src/Microsoft.SqlTools.ResourceProvider.Core/Localization/sr.strings
```diff
@@ -0,0 +1,44 @@
...
+FirewallRuleCreationFailed = An error occurred while creating a new firewall rule.
+InvalidIpAddress = Invalid IP address
+InvalidServerTypeErrorMessage = Server Type is invalid.
+LoadingExportableFailedGeneralErrorMessage = A required dll cannot be loaded. Please repair your application.
+NoInternetConnectionError = No internet connection found
+ServerDiscoveryFailedErrorMessage = An error occurred while getting server of type {0} from {1}
+SqlConnectionProperties_CannotTestNonExistentDatabase = This connection cannot be tested because the specified database does not exist or is not visible to the specified user.
+SqlConnectionProperties_MustSpecifyDataSource = This connection cannot be tested because no server name has been specified.
+SqlConnectionProperties_UnsupportedSqlVersion = This server version is not supported.  You must have Microsoft SQL Server 2005 or later.
+SqlFileConnectionProperties_CannotTestNonExistentMdf = This connection cannot be tested because the specified database file does not exist.
```
kburtram: >SqlFileConnectionProperties_CannotTestNonExistentMdf [](start = 0, length = 52)

not sure if all the strings are needed.  if not we may want to scrub at some point to make localization process similar

src/Microsoft.SqlTools.ResourceProvider.Core/Authentication/IAccountManager.cs
kevcunnane: Yes. Keeping most of this since it'll make cross-porting a bit easier in the long term which (hopefully) we want to keep as an option.

src/Microsoft.SqlTools.ResourceProvider.Core/Common/ConcurrentCache.cs
kevcunnane: Yeah, will do

src/Microsoft.SqlTools.ResourceProvider.Core/Discovery/AzureDatabaseDiscoveryProvider.cs
kevcunnane: I'd like to keep this code in here, so I'll add a TODO comment with what I hope to achieve. Essentially I want to move this away from UIConnectionInfo and towards our ConnectionDetails class (which will require a bit more refactoring than I can cover in this PR). With this, we can add back in support for database resource browsing so that long term we enabled Azure browse scenarios just like in SSDT.

src/Microsoft.SqlTools.ResourceProvider.Core/Extensibility/ExportableAttribute.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace Microsoft.SqlTools.ResourceProvider.Core.Extensibility
+{
+    /// <summary>
+    /// Attribute defining a service export, and the metadata about that service. Implements IServiceMetadata,
+    /// which should be used on the importer side to ensure type consistency. Services and providers have to add this property 
+    /// in order to be found by the extension manager
+    /// </summary>
+    [MetadataAttribute]
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class ExportableAttribute : ExportAttribute, IExportableMetadata
```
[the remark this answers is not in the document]
kevcunnane: Agree on this longer term, but for now this is somewhat specialized to the ported code. Can we consider it as we add in new scenarios? I'm wary of putting it all down too deep and bloating the core components with this.

src/Microsoft.SqlTools.ResourceProvider.Core/Localization/sr.strings
kevcunnane: You're right - given some scenarios were not ported over, not all are in use. I will do a quick scrub.

PR #480, opened 2017-10-06
return null response for peek definition if call originates from non-mssql client

praveenbarli:
This PR contains a change to send a null response on handle definition call originated from non-mssql editors


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -381,7 +381,7 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
                     {
                         if (definitionResult.IsErrorResult)
                         {
-                            await requestContext.SendError(definitionResult.Message);
+                            await requestContext.SendError(definitionResult.Message);
```
MattIrv: looks like some whitespace got added here

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -391,12 +391,17 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
                     }
                     else
                     {
-                        // Send an empty result so that processing does not hang
+                        // Send an empty result so that processing does not hang when peek def service called from non-mssql clients
```
MattIrv: Technically I think the non-mssql client comment belongs on the else condition you added below. Since `ShouldSkipIntellisense` will return true for non-mssql editors

PR #481, opened 2017-10-06
ResourceProvider implementation

kevcunnane:
**Note**: This is currently out for reference only - needs tests before being usable for checking.
Provides definition for a ResourceProvider and Authentication service. The ResourceProvider supports firewall rules for now, and since authentication is routed through that method it will call into the auth service to set up the current account to be used.


src/Microsoft.SqlTools.ResourceProvider.Core/AccountOptionsHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+    internal class AccountOptionsHelper
+    {
+        /// <summary>
+        /// The name of the IsMsAccount option
+        /// </summary>
+        internal const string IsMsAccount = "IsMsAccount";
+        /// <summary>
+        /// The name of the Tenants option
+        /// </summary>
+        internal const string Tenants = "Tentants";
```
benrr101: Tenants is spelled wrong

src/Microsoft.SqlTools.ResourceProvider.Core/Contracts/Account.cs
```diff
@@ -0,0 +1,119 @@
...
+        internal bool IsMsAccount
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Tenants for each object
+        /// </summary>
+        public IEnumerable<Tenant> Tenants
```
benrr101: It might be useful to define this as an IList or something with indexing capabilities. The Azure provider implementation will strive to make Tenants[0] the "home tenant" of the user. Not sure if that has benefits to you but it could be useful later.

src/Microsoft.SqlTools.ResourceProvider.Core/Contracts/Account.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// </summary>
+    public class AccountDisplayInfo
+    {
+        /// <summary>
+        /// A display name that offers context for the account, such as "Contoso".
+        /// </summary>
+            
+        public string ContextualDisplayName { get; set; }
+
+        // Note: ignoring ContextualLogo as it's not needed
```
benrr101: Thanks for these comments here 👍

src/Microsoft.SqlTools.ResourceProvider.Core/Contracts/FirewallRule.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// A FirewallRule object, usable in <see cref="CreateFirewallRuleRequest"/>s and other messages
+    /// </summary>
+    public class CreateFirewallRuleParams
+    {
+        /// <summary>
+        /// Account information to use in connecting to Azure
+        /// </summary>
+        public Account Account { get; set; }
+        /// <summary>
+        /// Per-tenant oken mappings. Ideally would be set independently of this call, but for
```
benrr101: "oken mappings" reminds me of https://en.wikipedia.org/wiki/Oakenfold_Anthems

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/AzureAuthenticationManager.cs
```diff
@@ -0,0 +1,314 @@
...
+        Categories.Azure,
+        typeof(IAzureAuthenticationManager),
+        "Microsoft.SqlTools.ResourceProvider.DefaultImpl.AzureAuthenticationManager",
+        1)
+    ]
+    class AzureAuthenticationManager : ExportableBase, IAzureAuthenticationManager
+    {
+        private Dictionary<string, AzureUserAccount> accountsMap;
+        private string currentAccountId = null;
+        private IEnumerable<IAzureUserAccountSubscriptionContext> _selectedSubscriptions = null;
```
benrr101: Can this be implemented as a Concurrent something to avoid the manual locking?

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/AzureAuthenticationManager.cs
```diff
@@ -0,0 +1,314 @@
...
+        private AzureUserAccountDisplayInfo ToDisplayInfo(Account account)
+        {
+            return new AzureUserAccountDisplayInfo()
+            {
+                AccountDisplayName = account.DisplayInfo.DisplayName,
+                ProviderDisplayName = account.Key.ProviderId
+            };
+        }
+
+        private void OnCurrentAccountChanged()
```
benrr101: Overall, I'm still not quite sure how this "current account" stuff is intended to work. It's probably a higher level architecture question, so I won't have this block to review. But, we should think about it a bit more, since I have a suspicion that it's a holdover from SSDT implementation.

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/AzureTenant.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        public string AccountDisplayableId
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Access token for use in login scenarios. Note that we could consider implementing this better in the
```
benrr101: Can we
1) finish this sentence
2) make it a TODO?

src/Microsoft.SqlTools.ResourceProvider.Core/Contracts/FirewallRule.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        /// <summary>
+        /// Account information to use in connecting to Azure
+        /// </summary>
+        public Account Account { get; set; }
+        /// <summary>
+        /// Per-tenant oken mappings. Ideally would be set independently of this call, but for 
+        /// now this allows us to handle this
+        /// </summary>
+        public Dictionary<string,string> SecurityTokenMappings { get; set; }
```
benrr101: This will actually be a (copy/paste definitions from typescript)
```
/**
 * Azure account security token maps a tenant ID to the information returned from a request to get
 * an access token. The list of tenants correspond to the tenants in the account properties.
 */
export type AzureAccountSecurityTokenCollection = {[tenantId: string]: AzureAccountSecurityToken};

/**
 * Token returned from a request for an access token
 */
export interface AzureAccountSecurityToken {
	/**
	 * Access token, itself
	 */
	token: string;

	/**
	 * Date that the token expires on
	 */
	expiresOn: Date;

	/**
	 * Name of the resource the token is good for (ie, management.core.windows.net)
	 */
	resource: string;

	/**
	 * Type of the token (pretty much always 'Bearer')
	 */
	tokenType: string;
}
```

src/Microsoft.SqlTools.ResourceProvider.Core/AccountOptionsHelper.cs
kevcunnane: Hmmm... somehow this file wasn't used. Will delete.

src/Microsoft.SqlTools.ResourceProvider.Core/Contracts/Account.cs
kevcunnane: We should consider removing unnecessary things from the protocol passthrough...

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/AzureAuthenticationManager.cs
kevcunnane: I've ported this as-is from Leila's other codebase and am loathe to change right now. Let's look at this in the future, for now it's highly unlikely to actually lock.

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/AzureAuthenticationManager.cs
kevcunnane: This is all ported over. There's basically 1 active account for most Azure operations and this is an intended behavior. While you can have knowledge of a lot of accounts, you choose one (via the UI) that is the current primary and then all operations are simple.

PR #484, opened 2017-10-09
peek and go to definition less verbose now

abist:
Peek and Go To Definition now have a less verbose script output, with just the create syntax and without comments or other dependent objects.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
@@ -477,7 +477,21 @@ namespace Microsoft.SqlTools.ServiceLayer.Scripting
 			    TargetDatabaseEngineEdition = GetTargetDatabaseEngineEdition(),
 			    TargetDatabaseEngineType = GetTargetDatabaseEngineType(),
 			    ScriptCompatibilityOption = GetScriptCompatibilityOption(),
-                IncludeIfNotExists = true
+                ScriptExtendedProperties = false,
+                ScriptUseDatabase = false,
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -477,7 +477,21 @@ internal ScriptingScriptOperation InitScriptOperation(string objectName, string
 			    TargetDatabaseEngineEdition = GetTargetDatabaseEngineEdition(),
 			    TargetDatabaseEngineType = GetTargetDatabaseEngineType(),
 			    ScriptCompatibilityOption = GetScriptCompatibilityOption(),
-                IncludeIfNotExists = true
+                ScriptExtendedProperties = false,
```
kburtram: if you untabify this file then that would probably fix the misalignment issue here.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -477,7 +477,21 @@ internal ScriptingScriptOperation InitScriptOperation(string objectName, string
...
+                ScriptUseDatabase = false,
+                IncludeIfNotExists = false,
```
kburtram: will these changes also apply to the Script As Create command?  Ideally Script As Create would generate the verbose CREATE syntax and Goto Definition would generate the concise CREATE statement.

abist: This is just for peak definition, the options are intact for script as create

PR #486, opened 2017-10-10
Increase parsing thread stack size to avoid stack overflow

kburtram:
This is to avoid StackOverflow exceptions we see in the parser such as this https://github.com/Microsoft/vscode-mssql/issues/340.  The parser has a recursive error backtracking algorithm that can create stacks 10000+ levels deep, but that eventually unwind.  This typically occurs when a large block of non-SQL content is provided to the parser (such as large XML files, or similar).


mtolentino:
You are a crack, but think as a user, it is best to include an option if they are too ignored to ignore the analyzer?

PR #489, opened 2017-10-11
Improve performance of Create Firewall Rule code

kevcunnane:
- Parallelized the find server code path so all subscriptions are queried in parallel
- Parallelized the list subscriptions code path so this call is in parallel if you have > 1 tenant

This improves perf overall, but we're still struggling with very slow total query perf against Azure. For example a single list subscriptions call can take > 4 seconds, so the process just isn't going fast. It's not clear how to improve this but may warrant some kind of perf tracing to see where the greatest cost is in the future.

### DIFF
--- a/src/Microsoft.SqlTools.ResourceProvider.Core/Discovery/AzureDatabaseDiscoveryProvider.cs
+++ b/src/Microsoft.SqlTools.ResourceProvider.Core/Discovery/AzureDatabaseDiscoveryProvider.cs
@@ -121,7 +121,7 @@ namespace Microsoft.SqlTools.ResourceProvider.Core
                     IEnumerable<IAzureUserAccountSubscriptionContext> subscriptions = await GetSubscriptionsAsync(string.IsNullOrEmpty(serverName));
                     if (!cancellationToken.IsCancellationRequested)
                     {                       
-                        result = await AzureUtil.ExecuteGetAzureResourceAsParallel(null, subscriptions, serverName, cancellationToken,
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ResourceProvider.Core/Discovery/AzureDatabaseDiscoveryProvider.cs
```diff
@@ -121,7 +121,7 @@ public AzureDatabaseDiscoveryProvider()
                     IEnumerable<IAzureUserAccountSubscriptionContext> subscriptions = await GetSubscriptionsAsync(string.IsNullOrEmpty(serverName));
                     if (!cancellationToken.IsCancellationRequested)
                     {                       
-                        result = await AzureUtil.ExecuteGetAzureResourceAsParallel(null, subscriptions, serverName, cancellationToken,
+                        result = await AzureUtil.ExecuteGetAzureResourceAsParallel((object)null, subscriptions, serverName, cancellationToken,
```
kburtram: I thought null would implicitly convert to object?

kburtram: nevermind, it looks like you need the cast for the template to detect the param type.

kevcunnane: Yep, that's the reason here. I could add in 2 separate util methods in the future, but for now this works OK.

PR #498, opened 2017-10-14
Query Execution: Adding better cancellation support

benrr101:
As per https://github.com/Microsoft/vscode-mssql/issues/879#issuecomment-315201612 adding direct cancellation of the active batch's command. This should massively improve the query cancellation experience.


pensivebrian:
This approach works. Another approach would be to wire up the cancel operation with the cancellation token, which keeps the cancellation token as the single mechanism for cancellation:

`cancellationToken.Register(() => command.Cancel());
`

PR #499, opened 2017-10-16
Fix filebrowser to handle cancellation and multiple open/expand requests

hyoshis:
New iteration: 
- added a method to BindingQueue to restart ProcessQueue after cancellation if needed
- added separate request for changing file filter
- tested creating backup and restoring after using file browser 
- tested the file browser connections are closed 
----------------------------------------------------------------
- Created connectedbinding queue for expand requests
- Changed to create new connection each time opening file browser 
Previously the file browser service retrieved existing connection. It can cause a problem if a user opens a dialog, cancels before the file tree is created and then opens the dialog again while cancellation is still in progress.


src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
```diff
@@ -20,6 +21,9 @@ internal class FileBrowserOperation : FileBrowserBase
         private FileTree fileTree;
         private string expandPath;
         private string[] fileFilters;
+        private CancellationTokenSource cancelSource;
```
llali: >cancelSource [](start = 40, length = 12)

token source is disposable, make sure it gets disposed when the class is disposed

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -158,14 +159,26 @@ public void InitializeService(ServiceHost serviceHost)
...
+            try
             {
-                this.ownerToFileBrowserMap.Remove(fileBrowserParams.OwnerUri);
-                response.Succeeded = true;
+                if (this.ownerToFileBrowserMap.ContainsKey(fileBrowserParams.OwnerUri))
+                {
+                    FileBrowserOperation browserOperation = this.ownerToFileBrowserMap[fileBrowserParams.OwnerUri];
+                    if (!browserOperation.FileTreeCreated)
+                    {
+                        browserOperation.Cancel();
```
llali: so cancel doesn't close the connection?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -158,14 +159,26 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    }
+                    else
+                    {
+                        browserOperation.SqlConnection.Close();
```
llali: check if connection is not null and is open before closing

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -158,14 +159,26 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    }
+                    this.ownerToFileBrowserMap.Remove(fileBrowserParams.OwnerUri);
```
llali: shouldn't you use concurrent dictionary? what happens if two request to close comes at the same time?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -185,41 +199,44 @@ public void InitializeService(ServiceHost serviceHost)
 
                 if (connInfo != null)
                 {
-                    DbConnection dbConn;
-                    connInfo.TryGetConnection(ConnectionType.Default, out dbConn);
-                    if (dbConn != null)
-                    {
-                        conn = ReliableConnectionHelper.GetAsSqlConnection((IDbConnection)dbConn);
-                    }
+                    conn = ConnectionService.OpenSqlConnection(connInfo, "RemoteFileBrowser");
```
llali: why do you need a new connection here?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -158,14 +159,26 @@ public void InitializeService(ServiceHost serviceHost)
...
         {
             FileBrowserCloseResponse response = new FileBrowserCloseResponse();
-            if (this.ownerToFileBrowserMap.ContainsKey(fileBrowserParams.OwnerUri))
+            try
             {
-                this.ownerToFileBrowserMap.Remove(fileBrowserParams.OwnerUri);
-                response.Succeeded = true;
+                if (this.ownerToFileBrowserMap.ContainsKey(fileBrowserParams.OwnerUri))
+                {
+                    FileBrowserOperation browserOperation = this.ownerToFileBrowserMap[fileBrowserParams.OwnerUri];
```
kburtram: >FileBrowserOperation [](start = 20, length = 20)

this should probably be disposable and close the connection on dispose.  For example, show closes the connection when an exception is thrown?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: Before it was handling connection close in open task if the file tree was not created yet, but I simplified it and close the connection in Close request task all the time..

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -132,7 +133,12 @@ public void InitializeService(ServiceHost serviceHost)
             FileBrowserCloseResponse response = new FileBrowserCloseResponse();
             FileBrowserOperation removedOperation;
             response.Succeeded = ownerToFileBrowserMap.TryRemove(fileBrowserParams.OwnerUri, out removedOperation);
-
+            if (removedOperation != null)
+            {
+                removedOperation.Cancel();
+                removedOperation.Dispose();
```
llali: so how do you know at this point if the operation canceled and it's safe to dispose?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -132,7 +133,12 @@ public void InitializeService(ServiceHost serviceHost)
...
+                expandNodeQueue.Dispose();
```
hyoshis: i'm looking into changing it to call StopQueueProcessor instead of dispose.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -189,12 +189,29 @@ public void InitializeService(ServiceHost serviceHost)
...
                 result.Succeeded = ownerToFileBrowserMap.TryGetValue(fileBrowserParams.OwnerUri, out browser);
-                if (result.Succeeded && browser != null)
+                this.ConnectionServiceInstance.TryFindConnection(fileBrowserParams.OwnerUri, out connInfo);
+
+                if (result.Succeeded && browser != null && connInfo != null)
                 {
-                    result.Children = browser.GetChildren(fileBrowserParams.ExpandPath).ToArray();
-                    result.ExpandPath = fileBrowserParams.ExpandPath;
-                    result.OwnerUri = fileBrowserParams.OwnerUri;
+                    QueueItem queueItem = expandNodeQueue.QueueBindingOperation(
```
llali: this will block the restore operation. If you open the file browser once the connection stays open in the queue and restore cannot work after that. If you want to the binding queue please register it in the connection service the same way oe did so the connection get closed temporary when restore is working

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -29,6 +28,8 @@ public sealed class FileBrowserService
         private readonly ConcurrentDictionary<string, FileBrowserOperation> ownerToFileBrowserMap = new ConcurrentDictionary<string, FileBrowserOperation>();
         private readonly ConcurrentDictionary<string, ValidatePathsCallback> validatePathsCallbackMap = new ConcurrentDictionary<string, ValidatePathsCallback>();
         private ConnectionService connectionService;
+        private ConnectedBindingQueue expandNodeQueue = new ConnectedBindingQueue(needsMetadata: false);
```
llali: >expandNodeQueue [](start = 38, length = 15)

please test opening file browser for a database. closing and then restoring that database

kburtram:
@hyoshis let's take this change now so I can include in backup restore\testing this weekend.  We can take any follow-up fixes as bugs next week.

PR #500, opened 2017-10-16
Update filebrowser expand contract and methods

hyoshis:
- updated the expand contract
- changed to call GetChildren instead of ExpandSelectedNode since the latter should be only called the first time opening the browser.

(I know this change is included in another PR but i'd like to merge this in separately)


hyoshis:
Merging in since this is a simple fix. Please let me know if you have any feedback.

PR #501, opened 2017-10-17
Improve database properties view:

kevcunnane:
- Compatibility Level should be a numerical value. In using `enum.ToString()` the result was `Version140` instead of `140`
- Backup date should be sensible if never backed up. Added a simple check that states it was never backed up if the backup date is before 1900.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabaseTaskHelper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabaseTaskHelper.cs
@@ -18,6 +18,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Admin
 {
     public class DatabaseTaskHelper: IDisposable
     {
+        private static DateTime minBackupDate = new DateTime(1900, 1, 1);
+
         private DatabasePrototype prototype;
 
         private XmlDocument document;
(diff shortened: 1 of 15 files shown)


src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabaseTaskHelper.cs
```diff
@@ -151,6 +153,16 @@ internal static DatabaseInfo DatabasePrototypeToDatabaseInfo(DatabasePrototype p
...
         }
 
+        private static string GetBackupDate(DateTime backupDate)
+        {
+            if (backupDate == null
+                || backupDate < minBackupDate)
+            {
+                return SR.NeverBackedUp;
+            }
+            return backupDate.ToString();
```
llali: this will return the long datetime format which should be localized. Not a problem for now maybe ...Convert.ToString(dateTime, CultureInfo.CurrentCulture)

kevcunnane: Hmmm, I wonder if we should do a standard format, for example `YYYY-MM-DD`. @erickangMSFT can you comment? The current way is likely to format by default locale which will annoy / confuse any non-US users.

kevcunnane: For now I'll leave this as-is and create a tracking task to verify whether it works as expected on machines with non-US preferences.

PR #502, opened 2017-10-17
Query Execution: Better exception handling on unawaited async tasks

benrr101:
This change adds:
* A new extension method to Task that adds a ContinueWith to the task that will log any and all exceptions. It will optionally call a provided action, as well.
* Utilization of the new ContinueWithOnFaulted in query execution (not simple)
* Utilization of the new ContinueWithOnFaulted in query save as
* Unit tests for the ContinueWithOnFaulted code
* Some light code cleanup in Query class and UnitTest.csproj

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
@@ -1,6 +1,7 @@
 ﻿// 
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
 
 using System;
 using System.Data.Common;
@@ -274,10 +275,11 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
         /// </summary>
         public void Execute()
         {
-            ExecutionTask = Task.Run(ExecuteInternal).ContinueWithExceptionHandling(t =>
-            {
-                QueryFailed?.Invoke(this, t.Exception).Wait();
-            });
+            ExecutionTask = Task.Run(ExecuteInternal)
+                .ContinueWithOnFaulted(t =>
+                {
+                    QueryFailed?.Invoke(this, t.Exception).Wait();
+                });
         }
 
         /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
@@ -17,6 +17,7 @@ using Microsoft.SqlTools.ServiceLayer.SqlContext;
 using Microsoft.SqlTools.Utility;
 using Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode;
 using System.Collections.Generic;
+using Microsoft.SqlTools.ServiceLayer.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -269,7 +276,11 @@ public void Cancel()
         /// </summary>
         public void Execute()
         {
-            ExecutionTask = Task.Run(ExecuteInternal);
+            ExecutionTask = Task.Run(ExecuteInternal)
+                .ContinueWithOnFaulted(t =>
+                {
+                    QueryFailed?.Invoke(this, t.Exception).Wait();
```
kevcunnane: Minor: should we trace too?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -405,16 +418,18 @@ public Task<ExecutionPlan> GetExecutionPlan(int batchIndex, int resultSetIndex)
             }
             finally
             {
+                // Remove the message hadler from the connection
```
kevcunnane: typo: `hadler` -> `handler`

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -504,9 +504,15 @@ public void RemoveRow(long internalId)
                     }
                 }
             });
+            
+            // Add exception handling to the save task
+            Task taskWithHandling = saveAsTask.ContinueWithOnFaulted(t =>
+            {
+                failureHandler?.Invoke(saveParams, t.Exception.Message);
```
kevcunnane: There's no Wait() needed on this one?

kevcunnane: You're awaiting it elsewhere.

src/Microsoft.SqlTools.ServiceLayer/Utility/TaskExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </param>
+        /// <returns>Task with exception handling on continuation</returns>
+        public static Task ContinueWithOnFaulted(this Task task, Action<Task> continuationAction)
+        {
+            return task.ContinueWith(t =>
+            {
+                // If the task hasn't faulted or has an exception, skip processing
+                if (!t.IsFaulted || t.Exception == null)
+                {
+                    return;
```
kevcunnane: Minor: we don't currently have a case where we need to support this on a task with a return type?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
kevcunnane: Ah - I see you're logging in the extension method. Never mind.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/TaskExtensionTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.UnitTests.Utility
+{
+    public class TaskExtensionTests
+    {
+        [Fact]
+        public async Task ContinueWithOnFaultedNullContinuation()
+        {
+            // Setup: Create a task that will definitely fault
+            Task failuretask = new Task(() => throw new Exception("It fail!"));
```
kevcunnane: Minor: `failureTask` (camelcase)

src/Microsoft.SqlTools.ServiceLayer/Utility/TaskExtensions.cs
benrr101: I don't think so. If the point of launching a task asynchronously we probably wouldn't ever need it to return something, so just adding onto Task should be fine.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
benrr101: I think so...?

PR #504, opened 2017-10-18
File Browser: Adding async task exception handling

benrr101:
Changes:
* Adding a new version of the ContinueWithOnFaulted to take async continuations
* Adding unit tests for the above
* Reworking the FileBrowser to use a concurrent dictionary instead of a regular dictionary (since there is the potential for race conditions
* Changing the query execution async tasks to use the async version of ContinueWithOnFaulted

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
@@ -88,7 +88,7 @@ namespace Microsoft.SqlTools.ServiceLayer.FileBrowser
             try
             {
                 var task = Task.Run(() => RunFileBrowserOpenTask(fileBrowserParams, requestContext))
-                    .ContinueWithOnFaulted(t => requestContext.SendResult(false));
+                    .ContinueWithOnFaulted(null);
                 await requestContext.SendResult(true);
             }
             catch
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -53,46 +53,21 @@ internal ConnectionService ConnectionServiceInstance
...
 
         /// <summary>
         /// Initializes the service instance
         /// </summary>
-        /// <param name="serviceHost"></param>
-        /// <param name="context"></param>
+        /// <param name="serviceHost">Service host to register handlers with</param>
         public void InitializeService(ServiceHost serviceHost)
         {
-            this.ServiceHost = serviceHost;
```
llali: ServiceHost is used in other places so you still need to set it here

src/Microsoft.SqlTools.ServiceLayer/Utility/TaskExtensions.cs
```diff
@@ -18,33 +18,83 @@ public static class TaskExtensions
...
+                    Logger.Write(LogLevel.Error, e.StackTrace);
+                }
             });
         }
+
+        private static void LogTaskExceptions(AggregateException exception)
+        {
+            // Construct an error message for an aggregate exception and log it
+            StringBuilder sb = new StringBuilder("Unhandled exception(s) in async task:");
+            foreach (Exception e in exception.InnerExceptions)
```
kevcunnane: You should flatten the exceptions first. This removes nested aggregates which is quite common

PR #506, opened 2017-10-18
Fix tools service crash + improve server edition naming

kevcunnane:
Fix https://github.com/Microsoft/vscode-mssql/issues/986 and add better server edition naming

- Fixed a number of issues related to the binding queue, specifically the fact that it didn't capture exceptions that occurred on the binding thread. This caused the thread to crash the service.
  - The root cause of the error was when you get a connection error during init of the SmoMetadataProvider which threw an exception. Because of this no Binder was created, and the code would null ref later during processing
  - Added logic to handle null ref issue and other related code
  - Added a unit test for the new error handling path, and supported still returning some value in this case
- Separately, have a fix for an issue where the edition is shown as "SQL Azure" for all Azure connections. Fixing to be "Azure SQL DB" for this case and handle when the database reports as DW or Stretch instead. This maps better to users concept of what the edition should be and avoids returning what is an outdated term.
- Also deleted a .resx file that was in the wrong directory.


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -856,8 +859,7 @@ public ParseResult ParseAndBind(ScriptFile scriptFile, ConnectionInfo connInfo)
                     try
                     {
                         scriptInfo.ConnectionKey = this.BindingQueue.AddConnectionContext(info, "languageService");
-                        scriptInfo.IsConnected = true;
-
+                        scriptInfo.IsConnected = this.BindingQueue.IsBindingContextConnected(scriptInfo.ConnectionKey);
```
kburtram: we should only be queuing tasks that are connected, since the whole point is to share connections.  Is this flag to track the current connection state for when connections drops offline?

kevcunnane: This is to set the correct value - the `BindingQueue.AddConnectionContext` does not throw on failure to connect and internally marks the context as not connected. However the scriptInfo assumed that the connection occurred. With this change, the intellisense code should be able to tell that the script isn't connected and return the default completions without ever interacting with the binding queue.

PR #507, opened 2017-10-19
Adding more continuewithonfaulted handlers

benrr101:
* Adding more continue on faulted handlers (these are naïve handlers just to make sure we don't crash the tools service)
* Removing extra nesting for sending scripting task results


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -959,7 +960,7 @@ private void RunConnectRequestHandlerTask(ConnectParams connectParams)
                     };
                     await ServiceHost.SendEvent(ConnectionCompleteNotification.Type, result);
                 }
-            });
+            }).ContinueWithOnFaulted(null);
```
kburtram: should we write these faults to the diag log?

benrr101: Yep, that's part of the method.

PR #510, opened 2017-10-19
Switch back to Tasks in binding queue

kburtram:
This is a partial revert of https://github.com/Microsoft/sqltoolsservice/pull/486.  For binding queue I prefer to use Tasks for now to avoid overhead of creating so many threads.  We can revisit with a more extensive refactor later.

I left the new thread around the disconnected diagnostics parse since that is where most users opening non-SQL content in a SQL file would hit the stack overflow (since presumably they won't connect that SQL editor to a server in most cases, since the file doesn't contain SQL content and would fail to execute if submitted to a server)

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
@@ -288,7 +288,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                             CancellationTokenSource cancelToken = new CancellationTokenSource();
                      
                             // run the operation in a separate thread
-                            var bindThread = new Thread(() =>
+                            var bindTask = Task.Run(() =>
                             {
                                 try
                                 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -288,7 +288,7 @@ private void ProcessQueue()
                             CancellationTokenSource cancelToken = new CancellationTokenSource();
                      
                             // run the operation in a separate thread
-                            var bindThread = new Thread(() =>
+                            var bindTask = Task.Run(() =>
```
kevcunnane: Just pointing out that given the generic try/catch handling here, we don't need to use ContinueWithOnFaulted as it shouldn't do anything so this looks fine to me

kburtram: @kevcunnane that's what I thought.  I left the ContinueWithOnFaulted on the timeout continuation in case the Set() call threw..presumably from a nullref on the bindingcontext

PR #511, opened 2017-10-20
Fix numerous Azure property issues

kevcunnane:
- Avoid crashing when azure edition is System, which it will be for master DB connections. This broke all property queries against Azure.
- Added in a number of properties to send back to the calling app, notably the edition & service objective so we can show the pricing tier. 
- Fixed `IsCloud` handling so this includes Azure SQL DW and Stretch. Intentionally skipping Managed Instance since this should be treated just like on-prem right now.
- Backup/Restore and Scripting will now treat DW like Azure. This means that backup/restore isn't available on DW (correct), and I verified with Ron that the scripter already has logic that treats DW as Azure in the settings.


kburtram:
@kevcunnane I'll merge for you to pick up in tomorrow's build.  Please make any follow-up changes in a separate PR.

PR #512, opened 2017-10-20
fixed failing integration tests

abist:
Fixed integration tests that have been failing for a long while

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
@@ -2,7 +2,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
-
+using System.Data;
 using System.Data.Common;
 using Microsoft.SqlTools.ServiceLayer.Connection;
 using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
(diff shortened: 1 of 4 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ReliableConnectionTests.cs
```diff
@@ -734,6 +734,7 @@ public void InitReliableSqlConnectionTest()
             var retryPolicy = connection.CommandRetryPolicy;
             connection.CommandRetryPolicy = retryPolicy;
             Assert.True(connection.CommandRetryPolicy == retryPolicy);
+            connection.Open();
```
llali: verify it's closed before opening

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
```diff
@@ -83,7 +83,10 @@ public void DatabaseChangesAffectAllConnections()
             // All open DbConnections (Query and Default) should have initialDatabaseName as their database
             foreach (DbConnection connection in connectionInfo.AllConnections)
             {
-                Assert.Equal(connection.Database, initialDatabaseName);
+                if (connection != null && connection.State == ConnectionState.Open)
```
llali: why the database is closed here? can you please check? I know I  made  achange to the connection service so the urr start with "connection" and "dashboard" will be closed. I want to make sure it's not related

abist: Here, connectionInfo.AllConnections has all the connections it has made. So, it actually has 2 connections, one when created by `InitLiveConnectionInfo()` and one created when we execute the query. Only the connection that was used to execute the query remains open while the other one's closed. Also, that connection doesn't have a database name since we don't give any parameter to `InitLiveConnectionInfo()`, which is why it was failing.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ReliableConnectionTests.cs
abist: Fixed

PR #514, opened 2017-10-21
Edit Data: Better errors for possible truncation

benrr101:
* If a cell update is not going to fit in the column, let the user know before they commit (string columns only)
* Tweaks to add row to result set logic to attempt to read before validating the row count. Allows SqlClient exceptions to get priority
* Unit testssssssssssssssssssssssssss (and tweaks to allow for throw on read in mock readers)


kburtram:
@benrr101 I'll merge now to include in testing this weekend and to "get in" pending changes before we start ask-mode process next week.

PR #515, opened 2017-10-23
Fix/restore options issues

llali:
this should fix the issue with isTailLogBackupPossible  was not sent correctly to the client


AbbiePetcht:
>             bool isTailLogBackupWithNoRecoveryPossibleValue = isTailLogBackupPossibleValue;

Delete these lines

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs:924 in 64bc6b4. [](commit_id = 64bc6b4088b3e0ee571343ac90e3a04e3792ff64, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreUtil.cs
```diff
@@ -57,16 +58,13 @@ internal DateTime GetServerCurrentDateTime()
...
-            //TODO: the code is moved from ssms and used for restore differential backups
-            //Uncomment when restore operation for differential backups is supported
-            /*
             string query = "SELECT GETDATE()";
-            DataSet dataset = this.server.ExecutionManager.ExecuteWithResults(query);
+            DataSet dataset = this.server.ExecutionManager.ConnectionContext.ExecuteWithResults(query);
             if (dataset != null && dataset.Tables.Count > 0 && dataset.Tables[0].Rows.Count > 0)
             {
-                dt = Convert.ToDateTime(dataset.Tables[0].Rows[0][0], SmoApplication.DefaultCulture);
+                dt = Convert.ToDateTime(dataset.Tables[0].Rows[0][0], CultureInfo.CurrentCulture);
```
kburtram: >t = Convert.ToDateTime [](start = 17, length = 22)

I think this may fix a different issue than the fix in the other file.  Going forward through ask-mode let's do 1 fix per PR please...disregard if this is related to the same fix.

llali: while I was fixing the issue with tail-gate backup I notice it fails every time if the restore has tail-gate backup and it was because the name of the file was always the something so without this fix the tail-gate backup would be still broken   

---
In reply to: [146411547](https://github.com/Microsoft/sqltoolsservice/pull/515#discussion_r146411547) [](ancestors = 146411547)

PR #516, opened 2017-10-23
change return type for change database from always true

### LINKED ISSUE #517: Change Database returns true even if the operation fails
This impacts the client ability to correctly handle problems that occur.

anthonydresser:
fixes #517


kburtram:
@anthonydresser could you add a link **from** the bug you're fixing to this PR?

PR #519, opened 2017-10-24
Add force change database logic

### LINKED ISSUE #518: Change database fails against azure

anthonydresser:
ChangeDatabase fails against azure, which means the changeDatabase handler fails, meaning the user can't change the connection in an editor. This adds force logic as an option that will close connections to azure servers and reopen them with the correct database.

This doesn't affect query execution behavior because by default it won't try to force and just throw an error like it did before.
fixes #518

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
@@ -111,17 +111,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             }
         }
 
-        /// <summary>
-        /// Get enumerator for types and connections
-        /// </summary>
-        public IEnumerator<KeyValuePair<string, DbConnection>> AllConnectionWithTypes
-        {
-            get
-            {
-                return ConnectionTypeToConnectionMap.GetEnumerator();
-            }
-        }
-
         public bool HasConnectionType(string connectionType)
         {
             connectionType = connectionType ?? ConnectionType.Default;
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
@@ -57,6 +57,13 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         internal readonly ConcurrentDictionary<string, DbConnection> ConnectionTypeToConnectionMap =
             new ConcurrentDictionary<string, DbConnection>();
 
+        public ConcurrentDictionary<string, DbConnection> ConnectionMap
+        {
+            get {
+                return ConnectionTypeToConnectionMap;
+            }
+        }
+
         /// <summary>
         /// Intellisense Metrics
         /// </summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
@@ -104,6 +104,17 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             }
         }
 
+        /// <summary>
+        /// Get enumerator for types and connections
+        /// </summary>
+        public IEnumerator<KeyValuePair<string, DbConnection>> AllConnectionWithTypes
+        {
+            get
+            {
+                return ConnectionTypeToConnectionMap.GetEnumerator();
+            }
+        }
+
         public bool HasConnectionType(string connectionType)
         {
             connectionType = connectionType ?? ConnectionType.Default;
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
@@ -57,13 +57,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         internal readonly ConcurrentDictionary<string, DbConnection> ConnectionTypeToConnectionMap =
             new ConcurrentDictionary<string, DbConnection>();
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1210,23 +1210,43 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
-                        }
+                        if (pair.Value.State == ConnectionState.Open)
+                            {
+                                try
+                                {
+                                    pair.Value.ChangeDatabase(newDatabaseName);
+                                }
+                                catch (SqlException)
+                                {
+                                    if (force == true) {
```
kburtram: nit: `if (force)` also works here.
also, please put the `{` character on a separate line in C#

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1210,23 +1210,43 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
+                                        pair.Value.Dispose();
+                                        info.RemoveConnection(pair.Key);
+                                        
+                                        string connectionString = BuildConnectionString(info.ConnectionDetails);
+
+                                        // create a sql connection instance
+                                        DbConnection connection = info.Factory.CreateSqlConnection(connectionString);
+                                        connection.Open();
+                                        info.AddConnection(pair.Key, connection);
+                                    } else {
```
kburtram: please put `{` and '}` on separate lines in C# per coding guidelines

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -57,6 +57,13 @@ public ConnectionInfo(ISqlConnectionFactory factory, string ownerUri, Connection
         internal readonly ConcurrentDictionary<string, DbConnection> ConnectionTypeToConnectionMap =
             new ConcurrentDictionary<string, DbConnection>();
 
+        public ConcurrentDictionary<string, DbConnection> ConnectionMap
+        {
+            get {
```
kburtram: `{` on separate line in C#...I won't comment on any more of these, but please fix them all

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1210,23 +1210,43 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
+                    info.ConnectionDetails.DatabaseName = newDatabaseName;
+
+                    foreach (KeyValuePair<string, DbConnection> pair in info.ConnectionMap)
                     {
-                        if (connection.State == ConnectionState.Open)
-                        {
-                            connection.ChangeDatabase(newDatabaseName);
-                        }
+                        if (pair.Value.State == ConnectionState.Open)
+                            {
```
kburtram: it looks like this `if()` code block has an extra level of indention

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1201,31 +1201,50 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
+                                string connectionString = BuildConnectionString(info.ConnectionDetails);
+
+                                // create a sql connection instance
+                                DbConnection connection = info.Factory.CreateSqlConnection(connectionString);
+                                connection.Open();
+                                info.AddConnection(key, connection);
+                            }
+                            else
+                            {
+                                conn.ChangeDatabase(newDatabaseName);
```
kevcunnane: Is there ever a case where ChangeDatabase will actually work for IsCloud?
- I think we should probably always use this codepath for cloud, since the ChangeDatabase is failing
- I don't see this having a net impact on the Query execution codepath so we might not need to have a `force` option. The reason is that `USE DATABASE` will not work, so it will never be executed

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1201,31 +1201,50 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
                     {
-                        if (connection.State == ConnectionState.Open)
+                        DbConnection conn;
+                        info.TryGetConnection(key, out conn);
+                        if (conn != null && conn.State == ConnectionState.Open)
                         {
-                            connection.ChangeDatabase(newDatabaseName);
+                            if (info.IsCloud && force)
+                            {
+                                conn.Close();
```
kevcunnane: Can you check whether the connection is already targeting the new database? The reason is that in the scenario where Query Execution changes DB, one connection will already have been changed (via `USE Database`) to target the new DB but the default and any other connections won't. 

Ideally, we only close & reopen connections that don't already target the new database. This will make the code more robust and able to handle other callers who want to make use of it.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
anthonydresser: The current behavior is if you use `use database` in a query again cloud you get an error saying you can't do that. I wanted to keep that behavior for query execution (since it is what ssms does) so thats why I added the force method.

kevcunnane: What I'm saying is that the code path won't get called - what really happens is that the ConnectionType.Query connection is already changed to be the new DB, and this code path just changes the others. In any case, not a blocker.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
anthonydresser: fixed

PR #520, opened 2017-10-24
Scripting should await result of requestcontext.sendresult in case it errors

kevcunnane:
This is a low-pri fix that we may be better off taking at a time where we're more tolerant to risk. I'm pushing it out now as I don't want to forget about it. The summary is that we weren't awaiting the `requestContext.SendResult` message which could throw unexpectedly and theoretically cause a crash / expected error. It also seems a bad practice to hog the binding queue with this as it should be let free up once the main processing is done.

- Refactored to use pattern where we wait on the binding queue action to complete, then send the result / error outside of that.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
@@ -23,6 +23,7 @@ using Microsoft.SqlServer.Management.Smo;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlServer.Management.Sdk.Sfc;
 using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices;
 
 namespace Microsoft.SqlTools.ServiceLayer.Scripting
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -227,55 +228,86 @@ public void InitializeService(ServiceHost serviceHost)
         /// Runs the async task that performs the scripting operation.
         /// </summary>
         private void RunSelectTask(ConnectionInfo connInfo, ScriptingParams parameters, RequestContext<ScriptingResult> requestContext)
-        {            
-            ConnectionServiceInstance.ConnectionQueue.QueueBindingOperation(
-                key: ConnectionServiceInstance.ConnectionQueue.AddConnectionContext(connInfo, "Scripting"),
-                bindingTimeout: ScriptingOperationTimeout,
-                bindOperation: (bindingContext, cancelToken) =>
+        {
+            Task.Run(() =>
```
kburtram: >Task.Run [](start = 12, length = 8)

queuing an operation already creates a task with ContinueWithOnFaulted and it looks like this function is already in a try\catch.

Where do you anticipate faults occurring that require this task\continuewithonfaulted wrapper?

kburtram: ...on second thought I guess you're just trying to make the call actually be async, since we currently do a `queueItem.ItemProcessed.WaitOne();` which blocks, if that's the case then disregard, if the task is for error handling the comment still applies :).

PR #523, opened 2017-10-24
Clear bindingqueue items after cancel

hyoshis:
Clear binding queue items and context map when cancel is requested, so that next time when the process is restarted it does not execute old items.


hyoshis:
I need to make changes and will include it as part of other PR

PR #524, opened 2017-10-24
Add unknown status for DB Nodes whose status cannot be found.

saurabh500:
When SMO's fix for Server.Databases is taken, we will get an exception while querying the Database.Status 
This is the case for paused Datawarehouse nodes. 
Since there is no other way to get this information, catching the exception with Status retrieval is the way to capture this information. 

I tested with updated SMO that we can populate OE with the change in this PR and with changes in SMO


kburtram:
@saurabh500 the appveyor failure doesn't look related to your change, but please make sure to run the unit tests locally to verify it isn't a consistent failure.

saurabh500:
The unit tests have a 100% pass rate on my local environment with the change.

PR #527, opened 2017-10-25
Scripting should await result of requestcontext.sendresult in case it errors

kburtram:
@kevcunnane this is a revert of my revert of you change.  Sorry for merging, I should have checked with you first.


kevcunnane:
I'll close this for now and reopen post-current work.

PR #528, opened 2017-10-25
Fix file browser cancel/connetion issues

hyoshis:
- Changed to queue all file browser operations so that it'll be handled in order. 
- Close operation is queued so that disconnection and resource dispose will happen when all operations are completed.
- When close is requested, it clears all the queued items.
- Re-initialize file tree when file filter is changed

This change fixes the following issues:
https://github.com/Microsoft/carbon/issues/2559
https://github.com/Microsoft/carbon/issues/2557
https://github.com/Microsoft/carbon/issues/2629

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
@@ -28,8 +28,8 @@ namespace Microsoft.SqlTools.ServiceLayer.FileBrowser
         private readonly ConcurrentDictionary<string, FileBrowserOperation> ownerToFileBrowserMap = new ConcurrentDictionary<string, FileBrowserOperation>();
         private readonly ConcurrentDictionary<string, ValidatePathsCallback> validatePathsCallbackMap = new ConcurrentDictionary<string, ValidatePathsCallback>();
         private ConnectionService connectionService;
-        private ConnectedBindingQueue expandNodeQueue = new ConnectedBindingQueue(needsMetadata: false);
-        private static int DefaultExpandTimeout = 120000;
+        private ConnectedBindingQueue fileBrowserQueue = new ConnectedBindingQueue(needsMetadata: false);
+        private static int DefaultTimeout = 120000;
         private string serviceName = "FileBrowser";
 
         /// <summary>
@@ -45,7 +45,7 @@ namespace Microsoft.SqlTools.ServiceLayer.FileBrowser
                 if (connectionService == null)
                 {
                     connectionService = ConnectionService.Instance;
-                    connectionService.RegisterConnectedQueue(this.serviceName, this.expandNodeQueue);
+                    connectionService.RegisterConnectedQueue(this.serviceName, this.fileBrowserQueue);
                 }
                 return connectionService;
             }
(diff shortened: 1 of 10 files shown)


src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -132,90 +132,137 @@ public void InitializeService(ServiceHost serviceHost)
...
-                }
-                else if (!removedOperation.FileTreeCreated)
-                {
-                    removedOperation.Cancel();
-                }
-                else if (hasPendingQueueItems)
-                { 
-                    this.expandNodeQueue.StopQueueProcessor(DefaultExpandTimeout);
-                }
+                var task = Task.Run(() => RunFileBrowserCloseTask(fileBrowserParams, requestContext))
```
kburtram: >ar task = [](start = 17, length = 9)

why are we saving into task variable?  it doesn't look like we use the variable.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -132,90 +132,137 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    .ContinueWithOnFaulted(null);
+                await requestContext.SendResult(new FileBrowserCloseResponse() { Succeeded = true });
+            }
+            catch
+            {
+                await requestContext.SendResult(new FileBrowserCloseResponse());
```
kburtram: >requestContext [](start = 22, length = 14)

should we log the failure or include it in the return response?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
```diff
@@ -132,90 +132,137 @@ public void InitializeService(ServiceHost serviceHost)
...
+                        operation.Cancel();
                     }
-                }
-                else
-                {
-                    ownerToFileBrowserMap.TryGetValue(fileBrowserParams.OwnerUri, out browser);
-                    if (browser != null)
+
+                    // Clear queued items
+                    this.fileBrowserQueue.ClearQueuedItems();
```
kburtram: >ClearQueuedItems [](start = 42, length = 16)

so the design is we can only have 1 file browsing operation running at a time?  this is fine for Carbon use case, but in general we try to design features to work correctly in other tools.  For example would this work in a multi-user app where multiple file browses are going on at a time?

I think this design is probably fine for now, though please be sure to think through scenarios beyond carbon UI when designing features.

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: updated to log the failure

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserService.cs
hyoshis: Yes, for now since we can have only one file browser dialog at a time, I added all the operations in a queue. But this will need to be changed like in the scenario you mentioned so that multiple file browser operations can work. I'll create an issue to track this.

PR #529, opened 2017-10-26
removing the oe binding context on closing session

llali:


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
```diff
@@ -118,6 +118,15 @@ public void OpenConnections(string serverName, string databaseName, int millisec
             }
         }
 
+        public void RemoveBindigContext(ConnectionInfo connInfo)
```
kburtram: fix typo

PR #531, opened 2017-10-26
Add group id and db display name to OE connection queue key

kburtram:


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
```diff
@@ -72,12 +72,24 @@ public ConnectedBindingQueue(bool needsMetadata)
...
             ConnectionDetails details = connInfo.ConnectionDetails;
-            return string.Format("{0}_{1}_{2}_{3}",
+            string key = string.Format("{0}_{1}_{2}_{3}",
                 details.ServerName ?? "NULL",
                 details.DatabaseName ?? "NULL",
                 details.UserName ?? "NULL",
                 details.AuthenticationType ?? "NULL"
             );
+
+            if (!string.IsNullOrEmpty(details.DatabaseDisplayName))
```
llali: in what scenario we need to add the database display name which is different than database name?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
```diff
@@ -72,12 +72,24 @@ public ConnectedBindingQueue(bool needsMetadata)
...
+            {
+                key += "_" + details.DatabaseDisplayName;
```
llali: in the connection context we replace the text with "Null" if the value is null. do we need to do the same thing here? Also may be we should check for null and empty string? is it possible the group id is null in one request and empty string in another?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
kburtram: this is to differentiate connections to "default" (no db name provided) and the default DB (when 'master') has been provided.  Currently if <default> and 'master' connections are in the same group they will have the same connection key, and bad things will happen.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
kburtram: @llali the idea with this change is the key won't change except for OE, where these values are provided.  In the case the display name is null the argument will be omitted from the key, so the key would be different than any connection that has a display name (really we're comparing <default> and master with display name in key), so we don't need to add a "Null" value to the key to differentiate.

PR #533, opened 2017-10-26
Edit Data: Fixing Defaults Behavior

benrr101:
Although this change was originally started to resolve exceptions when creating rows that have invalid defaults, it now features:
* Default values are added to a new list of default values in CreateRow instead of creating CellUpdate objects, this bypasses validation that caused the failed promise in the first place
* Generating the EditRow from the RowCreate uses the default value if a CellUpdate for the column hasn’t been created, this allows calculated column display values to be displayed
* Query generation has been reworked to allow for
  * Excluding default values
  * Special syntax when all columns are calculated or default values

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
@@ -89,7 +89,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
             Validate.IsNotNull(nameof(connection), connection);
 
             // Build the script and generate a command
-            ScriptBuildResult result = BuildInsertScript(true);
+            ScriptBuildResult result = BuildInsertScript(forCommand: true);
             
             DbCommand command = connection.CreateCommand();
             command.CommandText = result.ScriptText;
@@ -123,7 +123,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
         /// <returns>INSERT INTO statement</returns>
         public override string GetScript()
         {
-            return BuildInsertScript(false).ScriptText;
+            return BuildInsertScript(forCommand: false).ScriptText;
         }
 
         /// <summary>
@@ -172,6 +172,21 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
 
         #endregion
 
+        /// <summary>
+        /// Generates an INSERT script that will insert this row
+        /// </summary>
+        /// <param name="forCommand">
+        /// If <c>true</c> the script will be generated with an OUTPUT clause for returning all
+        /// values in the inserted row (including computed values). The script will also generate
+        /// parameters for inserting the values.
+        /// If <c>false</c> the script will not have an OUTPUT clause and will have the values
+        /// directly inserted into the script (with proper escaping, of course). 
+        /// </param>
+        /// <returns>A script build result object with the script text and any parameters</returns>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown if there are columns that are not readonly, do not have default values, and were
+        /// not assigned values.
+        /// </exception>
         private ScriptBuildResult BuildInsertScript(bool forCommand)
         {           
             // Process all the columns in this table
@@ -213,7 +228,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
                 if (forCommand)
                 {
                     // Since this script is for command use, add parameter for the input value to the list
-                    string paramName = $"@Value{RowId}{i}";
+                    string paramName = $"@Value{RowId}_{i}";
                     inValues.Add(paramName);
 
                     SqlParameter param = new SqlParameter(paramName, cell.Column.SqlDbType) {Value = cell.Value};
@@ -284,7 +299,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
                 // Cell has been provided by user, so use that
                 dbCell = cell.AsDbCellValue;
             }
-            return new EditCell(dbCell, true);
+            return new EditCell(dbCell, isDirty: true);
         }
 
         private class ScriptBuildResult
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
@@ -9,6 +9,7 @@ using System.Data;
 using System.Data.Common;
 using System.Data.SqlClient;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -74,50 +88,13 @@ public override DbCommand GetCommand(DbConnection connection)
...
-                SqlParameter param = new SqlParameter(paramName, cell.Column.SqlDbType)
-                {
-                    Value = cell.Value
-                };
-                command.Parameters.Add(param);
-            }
-            string joinedInColumns = string.Join(", ", inColumns);
-            string joinedOutColumns = string.Join(", ", outColumns);
+            // Build the script and generate a command
+            ScriptBuildResult result = BuildInsertScript(true);
```
kevcunnane: Minor: it's nice to name the random bool parameter so I know what `true` actually indicates.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -152,35 +123,7 @@ public override EditRow GetEditRow(DbCellValue[] cachedRow)
...
-                values.Add(SqlScriptFormatter.FormatValue(cell.Value, column));
-            }
-            string joinedValues = string.Join(", ", values);
-
-            // Get the start clause
-            string start = GetTableClause();
-
-            // Put the whole #! together
-            return string.Format(InsertCompleteScript, start, joinedValues);
+            return BuildInsertScript(false).ScriptText;
```
kevcunnane: Again, what does the `false` do? Param name like `includeXyz: false` would be great.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -227,16 +172,125 @@ public override EditUpdateCellResult SetCell(int columnId, string newValue)
 
         #endregion
 
-        private string GetTableClause()
+        private ScriptBuildResult BuildInsertScript(bool forCommand)
```
kevcunnane: And what does `forCommand` mean?

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -227,16 +172,125 @@ public override EditUpdateCellResult SetCell(int columnId, string newValue)
...
+                    
+                    // There is a default value, so trust the db will apply it
+                    continue;
+                }
+
+                // Add the input values
+                if (forCommand)
+                {
+                    // Since this script is for command use, add parameter for the input value to the list
+                    string paramName = $"@Value{RowId}{i}";
```
kevcunnane: This syntax is making me wonder what is string value and what part is looked up. What'll this be? Will it end up being `@Value12` or something? This is a case where simple string.Format might make it more readable.

kevcunnane: Also if it is `@Value12` like I'm guessing, could you ever get a conflict between different params? E.g. Row 11 col 1 is `@Value111`, and Row 1 col 11 is also `@Value111`? Shouldn't there be a `_` in the middle to avoid this? Again, maybe i'm missing something...

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -227,16 +172,125 @@ public override EditUpdateCellResult SetCell(int columnId, string newValue)
...
+            }
+            
+            // Begin the script (ie, INSERT INTO blah)
+            StringBuilder queryBuilder = new StringBuilder();
+            queryBuilder.AppendFormat(InsertScriptStart, AssociatedObjectMetadata.EscapedMultipartName);
+            
+            // Add the input columns (if there are any)
+            if (inColumns.Count > 0)
+            {
+                string joinedInColumns = string.Join(", ", inColumns);
```
kevcunnane: Me being dumb: for a single column this won't have a `, ` correct?

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -227,16 +172,125 @@ public override EditUpdateCellResult SetCell(int columnId, string newValue)
...
+                queryBuilder.AppendFormat(InsertScriptColumns, joinedInColumns);
+            }
+            
+            // Add the output columns (this will be empty if we are not building for command)
+            if (outColumns.Count > 0)
```
kevcunnane: Minor: this is repeated logic. A `joinAndAppend` method could check count > 0 and append comma-delimited values to the builder.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -227,16 +172,125 @@ public override EditUpdateCellResult SetCell(int columnId, string newValue)
...
+                    RawObject = null,
+                    RowId = RowId
+                };
+            }
+            else
+            {
+                // Cell has been provided by user, so use that
+                dbCell = cell.AsDbCellValue;
+            }
+            return new EditCell(dbCell, true);
```
kevcunnane: Again what does `true` mean?

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
benrr101: Good flipping catch! I hadn't thought of a scenario like that. At the moment it's not an issue due to only being able to edit one row at a time, but I don't want to be limited to that.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
benrr101: Correct, `string.Join()` only puts the "glue" (`", "`) in between the values. A single value wouldn't get any glue.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
benrr101: Since this fix still doesn't address triggers, I'm in the process of working on another change that will fix some of the repeated logic. As such, I'll leave this as-is for now.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
benrr101: I thought I didn't need block comments for private methods?

benrr101:
@kevcunnane please take a look at the updates and see if they match your comments

PR #535, opened 2017-10-29
Fix empty result set error handling and metadata service returning correct type name

kevcunnane:
- HandleSimpleExecuteRequest now handles the case where no rows are in a result by cleanly returning a success message but with no rows included. This is handled in the front-end instead and goes through the standard path (with a clean explanation message) instead of showing a `error: no results to return`
- MetadataService was always meant to include the type name in the return result, as otherwise the front end has to guess. In order to fix a bug where this resulted in scripting based on the metadata failing (as front-end used `Procedure` instead of `StoredProcedure`), I'm returning the data here. I'll have a matching front end fix but this is overall a good solution to have.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
@@ -230,8 +230,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                         {
                             // check to make sure any results were recieved
                             if (query.Batches.Length == 0 
-                                || query.Batches[0].ResultSets.Count == 0
-                                || query.Batches[0].ResultSets[0].RowCount == 0) 
+                                || query.Batches[0].ResultSets.Count == 0) 
                             {
                                 await requestContext.SendError(SR.QueryServiceResultSetHasNoResults);
                                 return;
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -245,22 +244,27 @@ public void InitializeService(ServiceHost serviceHost)
...
-                            };
-                            // get the data to send back
-                            ResultSetSubset subset = await InterServiceResultSubset(subsetRequestParams);
                             SimpleExecuteResult result = new SimpleExecuteResult
                             {
-                                RowCount = query.Batches[0].ResultSets[0].RowCount,
+                                RowCount = rowCount,
                                 ColumnInfo = query.Batches[0].ResultSets[0].Columns,
-                                Rows = subset.Rows
+                                Rows = new DbCellValue[0][]
```
anthonydresser: Why is this set to a empty object? if there are no rows, wouldn't we want this value to be null?

kevcunnane: It minimizes the risk that the caller will handle things incorrectly since this way you have a length for the rows no matter what. 

It's done this way in some other calls - basically it's easier to return an empty object since that's what this really is. It's a result set with no rows, not `null` rows.

PR #537, opened 2017-10-31
making database info request handler async

llali:
Create a class which can be used for handling any request async. Using the class in the admin service to handler database info request async. Didn't have to send notification for this. as long as the request handler running a task async, then it won't block other requests

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
@@ -12,6 +12,8 @@ using System.Threading.Tasks;
 using System.Xml;
 using Microsoft.SqlServer.Management.Smo;
 using System.Collections.Concurrent;
+using Microsoft.SqlTools.ServiceLayer.AsyncRequest;
+using System.Threading;
 
 namespace Microsoft.SqlTools.ServiceLayer.Admin
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
```diff
@@ -148,29 +150,37 @@ await requestContext.SendResult(new CreateDatabaseResponse()
...
+            };
+        }
+
         /// <summary>
         /// Return database info for a specific database
         /// </summary>
         /// <param name="connInfo"></param>
         /// <returns></returns>
-        internal static DatabaseInfo GetDatabaseInfo(ConnectionInfo connInfo)
+            internal static DatabaseInfo GetDatabaseInfo(ConnectionInfo connInfo)
```
llali: will remove the extra spaces

src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
```diff
@@ -12,6 +12,8 @@
 using System.Xml;
 using Microsoft.SqlServer.Management.Smo;
 using System.Collections.Concurrent;
+using Microsoft.SqlTools.ServiceLayer.AsyncRequest;
```
kburtram: sort please.

PR #539, opened 2017-10-31
Fix/admin database info simple fix

llali:
This is a simple fix for the admin service blocking other requests. Simply running the code that gets database info in a separate tasks so that other messages don't get blocked by it


src/Microsoft.SqlTools.ServiceLayer/Admin/AdminService.cs
```diff
@@ -163,7 +163,7 @@ await requestContext.SendResult(new GetDatabaseInfoResponse()
                     });
                 };
 
-                Task.Run(() => requestHandler());
+                Task task = Task.Run(async () => await requestHandler());
```
kevcunnane: You need to add the `.ContinueOnFaulted` extension method to any task kicked off by `Task.Run`, or it will crash the process on an unexpected failure. You should add the code to do a `SendError` as part of this method - currently your try/catch doesn't really do anything since it won't actually catch errors inside the `requestHandler = async ...` function.

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -66,24 +66,29 @@ public void InitializeService(ServiceHost serviceHost)
...
-                {
-                    Metadata = metadata.ToArray()
-                });
+                    await requestContext.SendResult(new MetadataQueryResult
+                    {
+                        Metadata = metadata.ToArray()
+                    });
+                };
+
+                Task task = Task.Run(async () => await requestHandler());
```
kevcunnane: Again, this must have `ContinueOnFaulted` and `SendError` inside this.

PR #540, opened 2017-11-01
fix edge case for single batch multi line

anthonydresser:
Fixes an edge case that wasn't being handled


abist:
Are you sure this doesn't break any other scenario? This place is very delicate to touch

anthonydresser:
@abist luckily this code change only affects queries with 1 batch, since otherwise it is reassigned otherwise. I tested what I could think of with 1 batch.

kevcunnane:
@anthonydresser are you planning on committing this in? It's been approved.

PR #542, opened 2017-11-01
Fixing a bug with restore sends wrong exception to client 

llali:
Fixing an issue with connection failure error was sent to client after restore was done. 

Restore operation closes the binding queue connections before restore starts and opens them after it's done, sometimes opening the connection fails after restore is done because the database stays in "Restoring" status. So the we should only try to open then connection and not send error to client if it fails.

Also added restore exceptions to log for troubleshooting

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
@@ -125,7 +125,14 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             {
                 if (bindingContext.BindingLock.WaitOne(millisecondsTimeout))
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingQueue.cs
```diff
@@ -125,7 +125,14 @@ public void OpenConnections(string serverName, string databaseName, int millisec
             {
                 if (bindingContext.BindingLock.WaitOne(millisecondsTimeout))
                 {
-                    bindingContext.ServerConnection.Connect();
+                    try
+                    {
+                        bindingContext.ServerConnection.Connect();
```
kburtram: is it really safe to ignore failures trying to reconnect?  i.e. will this cause random failures in other feature areas?

PR #543, opened 2017-11-02
Fix/scripting server name

llali:
Scripting operation should get the server name using SMO the same way server class does. otherwise they might get different values which fails the scripting.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
@@ -9,11 +9,11 @@ using System.Data.SqlClient;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using Microsoft.SqlServer.Management.Sdk.Sfc;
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingScriptOperation.cs
```diff
@@ -246,27 +246,20 @@ private string GetServerNameFromLiveInstance(string connectionString)
...
                 {
-                    //
-                    // Azure SQL Data Warehouse does not support @@servername, so fallback to SERVERPROPERTY.
-                    //
-
                     Logger.Write(
                         LogLevel.Verbose, 
-                        string.Format("Exception running query 'SELECT @@servername' {0}, fallback to SERVERPROPERTY query", e));
-
-                    cmd.CommandText = "select SERVERPROPERTY('ServerName') AS ServerName";
```
llali: ServerConnection is using SERVERPROPERTY('ServerName') so not need to try again here. If it fails here it means it already failed in the server connection

PR #544, opened 2017-11-03
dashboard now doesnt show temp tables

abist:
Fixes https://github.com/Microsoft/carbon/issues/2824

The OE is 'smart' enough to not show temporary tables however the dashboard showed temp tables in the explorer widget. Made it consistent.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
@@ -11,7 +11,6 @@ using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.Connection;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
 using Microsoft.SqlTools.ServiceLayer.Metadata.Contracts;
-using Microsoft.SqlTools.ServiceLayer.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.Metadata
 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -193,30 +184,27 @@ internal static void ReadMetadata(SqlConnection sqlConn, List<ObjectMetadata> me
...
-                            metadataTypeName = "StoredProcedure";
                         }
                         else
                         {
                             metadataType = MetadataType.Table;
-                            metadataTypeName = "Table";
                         }
-
-                        metadata.Add(new ObjectMetadata
+                        if (metadataType == MetadataType.Table && objectName[0] != '#')
```
kburtram: does this skip the adding non-table type objects?

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -193,30 +184,32 @@ internal static void ReadMetadata(SqlConnection sqlConn, List<ObjectMetadata> me
...
                         else
                         {
                             metadataType = MetadataType.Table;
-                            metadataTypeName = "Table";
                         }
-
-                        metadata.Add(new ObjectMetadata
+                        // remove temporary tables and stored procedures
+                        // since every temp object has a '#' in front of it
+                        if (objectName[0] == '#')
```
kburtram: >if (objectName[0] == '#')  [](start = 24, length = 26)

not sure if I'm missing something.  Where do we add metadata for objects that don't start with '#'?

abist: We don't, that's the change. The OE doesn't show temporary objects (objects starting with '#'), and thus this change to make the dashboard consistent with OE. Otherwise, dashboard shows temporary objects if they exist, while OE doesn't.

kburtram: @abist I'm still not following.  Currently we do add metadata objects that **don't** start with a '#' to the dashboard.  With this change I can't see where we're adding these objects.

abist: This is a PR to **NOT** to add those objects. Those objects start with "#" are temporary. As of now, they are being shown in the dashboard. If you see the bug, we want similar behavior in OE and the dashboard. The OE **doesn't** show these objects with "#".

If it's still not clear, we can sync offline tomorrow.

PR #546, opened 2017-11-14
fixing a restore bug - relocating db files of a windows server db in mac

llali:
when connecting to a db in windows server in mac, the file paths has "/" as separator so getting the directory name fails in mac or linux. Had to convert the separators to be able to work with Path library,


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -680,6 +680,27 @@ private string CombineToServerConnectionPath(string root, string filePath)
             return result;
         }
 
+        private string GetDirectoryName(string filePath)
```
kburtram: should this be a helpwe in the PathWrapper class so we can share logic elsewhere?

PR #549, opened 2017-11-20
Fix Ops Studio issue 97 where query connection is dropped

kevcunnane:
- Fixes https://github.com/Microsoft/sqlopsstudio/issues/97.
- This should fix the bulk of the issues reported by users since it's the 1st use of this connection in the query code path.
- Implemented the fix in the connection service. This will always open a connection when calling `GetOrOpenConnection`. I cannot see a reason why the connection returned from this should ever be closed.
- resolve issues in both this code path and the edit data code path since both use this method.

UnitTest: added a step to an existing unit test.


kevcunnane:
FYI Travis failed due to the flaky `Queue100BindingOperationTest` test. I think we should just disable it at this point, it's our most frequent cause of failure

PR #550, opened 2017-11-21
fixed sproc failing peek definition

abist:
This fixes this issue - https://github.com/Microsoft/sqlopsstudio/issues/120 and https://github.com/Microsoft/sqlopsstudio/issues/110

The problem was it was looking at a wrong value for stored procedures in the dictionary. 
The reason we were getting the create script for `dbo` instead of the stored procedure was because peek definition shifts the cursor to the next token (parent) if a definition wasn't available for the object itself.


abist:
@kevcunnane I agree completely. Since Leila is taking over the scripting issues in general, I'll take over the testing part of this area. I'll add a couple of tests with this PR, but I plan to add some more after this one.

PR #552, opened 2017-11-22
Adding new handler for script as requests

llali:
Fixing the performance problem with scripting one object in the database using the scripting service. The current handler uses smo script publish model for scripting which doesn't perform well with scripting only one object. So Created new handler only for script as requests using smo Scripter class. The performance seems to be much better using this new handler. 

Also changed the peek definition to call scriptingAs operation. This will improve the peek definition performance also

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
@@ -1,0 +1,186 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.Data.SqlClient;
+using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
+using Microsoft.SqlTools.Utility;
+using Microsoft.SqlServer.Management.Common;
+using Microsoft.SqlServer.Management.Smo;
+using System.Collections.Specialized;
+using System.Text;
+using System.Linq;
+using System.Globalization;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    public class ScriptAsScriptingOperation : ScriptingScriptOperation
+    {
+        public ScriptAsScriptingOperation(ScriptingParams parameters): base(parameters)
+        {
+            Validate.IsNotNull("parameters", parameters);
+
+            this.Parameters = parameters;
+        }
+
+        private string BatchTerminator = "GO";
+
+        private string GetScript(ScriptingOptions options, StringCollection stringCollection)
+        {
+            StringBuilder sb = new StringBuilder();
+
+            foreach (var item in stringCollection)
+            {
+                sb.Append(item);
+                if (options != null && !options.NoCommandTerminator)
+                {
+                    //Ensure the batch separator is always on a new line (to avoid syntax errors)
+                    //but don't write an extra if we already have one as this can affect definitions
+                    //of objects such as Stored Procedures (see TFS#9125366)
+                    sb.AppendFormat(CultureInfo.InvariantCulture, "{0}{1}{2}",
+                        item.EndsWith(Environment.NewLine) ? string.Empty : Environment.NewLine,
+                        this.BatchTerminator,
+                        Environment.NewLine);
+                }
+                else
+                {
+                    sb.AppendFormat(CultureInfo.InvariantCulture, Environment.NewLine);
+                }
+            }
+
+            return sb.ToString();
+        }
+
+        private UrnCollection CreateUrns()
+        {
+            IEnumerable<ScriptingObject> selectedObjects = new List<ScriptingObject>(this.Parameters.ScriptingObjects);
+
+            string server = GetServerNameFromLiveInstance(this.Parameters.ConnectionString);
+            string database = new SqlConnectionStringBuilder(this.Parameters.ConnectionString).InitialCatalog;
+            UrnCollection urnCollection = new UrnCollection();
+            foreach (var scriptingObject in selectedObjects)
+            {
+                urnCollection.Add(scriptingObject.ToUrn(server, database));
+            }
+            return urnCollection;
+        }
+
+        private void SetScriptBehavior(ScriptingOptions options)
+        {
+            switch(this.Parameters.ScriptOptions.ScriptCreateDrop)
+            {
+                case "ScriptCreate":
+                    options.ScriptDrops = false;
+                    break;
+                case "ScriptDrop":
+                    options.ScriptDrops = true;
+                    break;
+                default:
+                    options.ScriptDrops = false;
+                    break;
+
+            }
+        }
+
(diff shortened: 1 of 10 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -0,0 +1,322 @@
...
+                        scriptingOptions.TargetDatabaseEngineType = DatabaseEngineType.Standalone;
+                        break;
+                    case SqlScriptOptions.ScriptDatabaseEngineType.SqlAzure:
+                        scriptingOptions.TargetDatabaseEngineType = DatabaseEngineType.SqlAzureDatabase;
+                        break;
+                }
+            }
+
+            object targetDatabaseEngineEdition;
+            if (Enum.TryParse(typeof(SqlScriptOptions.ScriptDatabaseEngineEdition), this.Parameters.ScriptOptions.TargetDatabaseEngineEdition, out targetDatabaseEngineEdition))
```
kevcunnane: Are we missing settings in Ops Studio for controlling this? We should create tracking tasks for all this kind of work. There are a whole bunch of script options and users are starting to request them one by one. We should probably do the work in one batch

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -0,0 +1,322 @@
...
+    /// <summary>
+    /// Class to generate script as for one smo object
+    /// </summary>
+    public class ScriptAsScriptingOperation : SmoScriptingOperation
+    {
+        public ScriptAsScriptingOperation(ScriptingParams parameters): base(parameters)
+        {
+        }
+
+        private string BatchTerminator = "GO";
```
kevcunnane: Can we make this a constant in a utils file? We reuse this in a bunch of places. I hope we'll never have to change it (e.g. support non-GO operators).

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -0,0 +1,322 @@
...
+                using (SqlConnection sqlConnection = new SqlConnection(this.Parameters.ConnectionString))
+                {
+                    sqlConnection.Open();
+                    ServerConnection serverConnection = new ServerConnection(sqlConnection);
+                    Server server = new Server(serverConnection);
+                    scripter = new SqlServer.Management.Smo.Scripter(server);
+                    ScriptingOptions options = new ScriptingOptions();
+                    SetScriptBehavior(options);
+                    PopulateAdvancedScriptOptions(this.Parameters.ScriptOptions, options);
+                    options.WithDependencies = false;
```
kevcunnane: I wonder should scripting dependencies be a config option too? Not vital, just something to think about

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -0,0 +1,322 @@
...
+        {
+            scriptingOptions.AllowSystemObjects = true;
+
+            // setting this forces SMO to correctly script objects that have been renamed
+            scriptingOptions.EnforceScriptingOptions = true;
+
+            //We always want role memberships for users and database roles to be scripted
+            scriptingOptions.IncludeDatabaseRoleMemberships = true;
+
+            if (this.Parameters.ScriptOptions.ScriptCompatibilityOption == SqlScriptOptions.ScriptCompatabilityOptions.Script140Compat.ToString())
```
kevcunnane: Minor: this kind of large if/else can be more efficiently defined in a dictionary mapping.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -61,7 +63,26 @@ internal static ConnectionService ConnectionServiceInstance
             {
                 connectionService = value;
             }
-        }     
+        }
+
+        /// <summary>
+        /// Internal for testing purposes only
+        /// </summary>
+        internal static LanguageService LanguageServiceInstance
```
kevcunnane: Minor: it'd be nice if we took opportunities like this to move the scripting service over to the MEF extension-based service definition. That way you never need this static instance reference - instead the test just passes in a service provider that has the mock services as needed. It's pretty quick to change this over and helps get us on 1 standard going forward.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -108,6 +130,52 @@ public void InitializeService(ServiceHost serviceHost)
...
+                {
+                    ScriptingService.ConnectionServiceInstance.TryFindConnection(parameters.OwnerUri, out connInfo);
+                    if (connInfo != null)
+                    {
+                        connInfo.ConnectionDetails.PersistSecurityInfo = true;
+                        parameters.ConnectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
+                    }
+                    else
+                    {
+                        throw new Exception("Could not find ConnectionInfo");
```
kevcunnane: This should be localized and added to SR.Strings

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Scripting/ScriptingServiceTests.cs
```diff
@@ -13,7 +13,9 @@
 using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
 using Moq;
 using Xunit;
-
+using Microsoft.SqlTools.ServiceLayer.Test.Common;
```
kevcunnane: please sort usings

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScripterCore.cs
```diff
@@ -286,7 +286,7 @@ internal DefinitionResult GetDefinitionUsingDeclarationType(DeclarationType type
             string createSyntax = null;
             if (objectScriptMap.ContainsKey(objectType.ToLower()))
             {
-                createSyntax = string.Format("CREATE {0}", objectScriptMap[objectType.ToLower()]);
+                createSyntax = string.Format("CREATE");
```
abist: This will create a problem because the scripts contain comments on top which have the word "create". I believe @kevcunnane talked about this bug last time, and that was the reason I added the objectType in the search string after "create".

llali: yes but keeping the code causes returning empty string. sometimes there are more than one space between create and type so at least this way the script is returned but the cursor is in the wrong place

abist: Perhaps we can have a check to have the word "create" as well as the objectType in the same line, rather than look for the string ("Create {0}", objectType)

PR #553, opened 2017-11-22
Edit Data: Create Row with Nullable Columns

benrr101:
This change looks big -- it is mostly changes to improve the test matrix for edit data.

The key changes here are to allow nullable columns that don't have default values defined (eg `[col] CHAR(1) NULL`) to not have their values defined. This is a corollary to  https://github.com/Microsoft/sqltoolsservice/pull/533 to resolve all the issues with default handling in Edit Data.

Another useful change: Error message when a required cell is not assigned a value will now tell the user what cell is missing.


test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/Common.cs
```diff
@@ -133,6 +91,29 @@ public static DbDataReader GetNewRowDataReader(DbColumn[] columns, bool includeI
             return new TestDbDataReader(new [] {testResultSet}, false);
         }
 
+        public static EditTableMetadata GetCustomEditTableMetadata(DbColumn[] columns)
+        {
+            // Create column metadata providers            
+            var columnMetas = columns.Select((c, i) => new EditColumnMetadata
```
kevcunnane: Minor: given you're iterating over columns twice, you could just create 2 lists and in a foreach loop assign to each one. Given the num columns is usually small, it's OK if you choose to ignore (just we should always be aware of these kind of optimizations)

kevcunnane: Just realized this is test code... ignore my previous comment

benrr101: You're still right, I'll make the change

saurabh500:
The changes look good to me. 
Can't we preserve the old formatting of the resx files to avoid such noise in future code reviews ?

PR #555, opened 2017-11-28
improving OE tables expand

llali:
The changes are for improving the performance of Object explorer when expanding tables in server which IsSystemVersioned and TemporalType are not supported (SQL DW).

The bug was getting those unsupported properties for tables which caused SMO to try to get the properties and failed. To fix it, now checking if a property is supported and then try to get it from SMO object. By default properties are all supported unless it's specified in TreeNodeDefinition.Xml

Also fixed a bug with sorting smo objects which not needed because the objects already sorted when getting from database


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodeObservableCollection.cs
```diff
@@ -29,6 +30,16 @@ public bool IsPopulating
             get { return numInits.HasValue && numInits != 0; }
         }
 
+        public bool IsSorted
+        {
+            get
+            {
+                // SMO objects are already sorted so no need to sort them again
+                TreeNode firstITem = this.FirstOrDefault();
+                return firstITem != null && firstITem is SmoTreeNode;
```
kevcunnane: Minor: I think you can just do `return this.FirstOrDefault() is SmoTreeNode` since it'll return false if casting `null` to `SmoTreeNode`

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -270,6 +272,14 @@ public override IEnumerable<NodeSmoProperty> SmoProperties
             }
         }
 
+        internal IEnumerable<NodeSmoProperty> CachedSmoProperties
+        {
+            get
+            {
+                return smoPropertis == null ? SmoProperties : smoPropertis;
```
kevcunnane: typo:  `smoPropertis` -> `smoProperties`

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -282,16 +292,31 @@ public virtual bool PassesFinalFilters(TreeNode parent, object context)
...
 
-        public override string GetNodeStatus(object context)
+        public override string GetNodeStatus(object smoObject, SmoQueryContext smoContext)
         {
             return string.Empty;
         }
 
+        public static bool IsPropertySupported(string propertyName, SmoQueryContext context, NamedSmoObject smoObj, IEnumerable<NodeSmoProperty> supportedProperties)
+        {
+            var property = supportedProperties.FirstOrDefault(x => x.Name == propertyName);
```
kevcunnane: Just checking: property name doesn't need ignore case comparison?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -282,16 +292,31 @@ public virtual bool PassesFinalFilters(TreeNode parent, object context)
...
+            if (property != null)
+            {
+                return ServerVersionHelper.IsValidFor(context.ValidFor, property.ValidFor);
+            }
+            else
+            {
+                // Return true if cannot find the proeprty, SMO still tries to get that property but adding the proeprty to supported list can make loading the nodes faster
```
kevcunnane: typo: `proeprty` -> `property`

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -282,16 +292,31 @@ public virtual bool PassesFinalFilters(TreeNode parent, object context)
...
+                Logger.Write(LogLevel.Verbose, $"Smo property name {propertyName} for Smo type {smoObj.GetType()} is not added as supported properties. This can cause the performance of loading the OE nodes");
```
kevcunnane: Nice. Any way we could catch errors like this during automatic testing?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -230,6 +231,7 @@ protected virtual void InitializeChild(TreeNode parent, TreeNode child, object c
             }
             else
             {
+                smoPropertis = SmoProperties;
```
kburtram: typo: smoProperti`e`s

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
llali: o I don't have a good way to do this. I'll think about it 

---
In reply to: [154442107](https://github.com/Microsoft/sqltoolsservice/pull/555#discussion_r154442107) [](ancestors = 154442107)

PR #556, opened 2017-11-28
using existing connection for scripting as operation 

llali:
- Getting connection from binding queue to be able to use existing connection fr script as operations
- Getting existing connection from language service when getting object definition
-Removed the duplicate options from ScriptingOption public class, Created a wrapper for ScriptingOption to be able to add the duplicate option names for Script as operation.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
@@ -9,7 +9,7 @@ using Microsoft.SqlTools.Utility;
 namespace Microsoft.SqlTools.ServiceLayer.Scripting
 {
     /// <summary>
-    /// The options for Script as operation
+    /// A wrpaper of ScriptOptions to map the option name with the oen in SMO.ScriptingOptions
     /// </summary>
     public class ScriptAsOptions : ScriptOptions
     {
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
@@ -1,0 +1,712 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    /// <summary>
+    /// The options for Script as operation
+    /// </summary>
+    public class ScriptAsOptions : ScriptOptions
+    {
+        public ScriptAsOptions(ScriptOptions scriptOptions)
+        {
+            Validate.IsNotNull(nameof(scriptOptions), scriptOptions);
+            ScriptOptions = scriptOptions;
+        }
+
+        public ScriptOptions ScriptOptions { get; private set; }
+
+        /// <summary>
+        /// Generate ANSI padding statements
+        /// </summary>
+        public override bool? ScriptAnsiPadding
+        {
+            get
+            {
+                return ScriptOptions.ScriptAnsiPadding;
+            }
+            set
+            {
+                ScriptOptions.ScriptAnsiPadding = value;
+            }
+        }
+
+
+        /// <summary>
+        /// Append the generated script to a file
+        /// </summary>
+        public override bool? AppendToFile
+        {
+            get
+            {
+                return ScriptOptions.AppendToFile;
+            }
+            set
+            {
+                ScriptOptions.AppendToFile = value;
+            }
+        }
+
+        /// <summary>
+        /// Continue to script if an error occurs. Otherwise, stop.
+        /// </summary>
+        /// <remarks>
+        /// The default is true.
+        /// </remarks>
+        public override bool? ContinueScriptingOnError
+        {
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
```diff
@@ -0,0 +1,712 @@
...
+//
+
+using Microsoft.SqlTools.ServiceLayer.Scripting.Contracts;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    /// <summary>
+    /// The options for Script as operation
+    /// </summary>
```
pensivebrian: Can you add a comment that this class is mainly an adapter to map property names?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
```diff
@@ -0,0 +1,712 @@
...
+
+        /// <summary>
+        /// Script only features compatible with the specified version of SQL Server.  Possible values:
+        ///   Script90Compat
+        ///   Script100Compat
+        ///   Script105Compat
+        ///   Script110Compat
+        ///   Script120Compat
+        ///   Script130Compat
+        ///   Script140Compat
```
kevcunnane: Script150Compat should be present too... right? Or is that not available yet...?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
```diff
@@ -0,0 +1,712 @@
...
+        /// <summary>
+        /// Script only features compatible with the specified SQL Server database engine edition.
+        /// Possible Values:
+        ///   SqlServerPersonalEdition
+        ///   SqlServerStandardEdition 
+        ///   SqlServerEnterpriseEdition 
+        ///   SqlServerExpressEdition
+        ///   SqlAzureDatabaseEdition
+        ///   SqlDatawarehouseEdition
+        ///   SqlServerStretchEdition
```
kevcunnane: SqlManagedInstanceEdition also?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -41,28 +48,26 @@ public override void Execute()
...
-                    options.IncludeHeaders = false;
-                    scripter.Options = options;
-                    scripter.Options.ScriptData = false;
-                    scripter.ScriptingError += ScripterScriptingError;
-                    UrnCollection urns = CreateUrns(serverConnection);
-                    var result = scripter.Script(urns);
-                    resultScript = GetScript(options, result);
-                }
+
+                Server server = new Server(ServerConnection);
```
kevcunnane: What happens if the connection is closed? Will this handle auto-reopening it or do we need to cover that ourselves? We have cases where the connection is closed and in query execution we now auto reopen before doing the operation

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -155,10 +151,34 @@ public void InitializeService(ServiceHost serviceHost)
...
-        public async Task HandleScriptExecuteRequest(ScriptingParams parameters, RequestContext<ScriptingResult> requestContext)
+        private void RunScriptAsTask(ConnectionInfo connInfo, ScriptingParams parameters, RequestContext<ScriptingResult> requestContext)
+        {
+            ScriptAsScriptingOperation operation = new ScriptAsScriptingOperation(parameters);
+            ConnectionServiceInstance.ConnectionQueue.QueueBindingOperation(
+                key: ConnectionServiceInstance.ConnectionQueue.AddConnectionContext(connInfo, "Scripting"),
+                bindingTimeout: ScriptingOperationTimeout,
+                bindOperation: (bindingContext, cancelToken) =>
+                {
+                    string script = string.Empty;
```
kevcunnane: this seems unused? Can you remove?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsOptions.cs
llali: not sure if it's available, I don't see it as a valid option in SM. I'll verify ylater

---
In reply to: [154440666](https://github.com/Microsoft/sqltoolsservice/pull/556#discussion_r154440666) [](ancestors = 154440666)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
llali: I'm changing this code in another PR 

---
In reply to: [154441355](https://github.com/Microsoft/sqltoolsservice/pull/556#discussion_r154441355) [](ancestors = 154441355)

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -155,10 +151,34 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    operation.ServerConnection = bindingContext.ServerConnection;
```
kburtram: this is a bit of a problem in that we are taking values from bindingContext and saving them to a variable that survives outside the locked action section.  this isn't really a good design practice as this context should only be used within the locked callback functions.

PR #557, opened 2017-11-28
Adding Indexes of all types to OE indexes folder

llali:
- Putting primary and Unique keys under indexes folder in OE to be consistence with ssms  
- Added a filter for Keys folder in OE to not show up for SQL dw because it's not supported

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
@@ -194,7 +194,7 @@
   </Node>
 
   <Node Name="Columns" LocLabel="SR.SchemaHierarchy_Columns" BaseClass="ModelBased" Strategy="PopulateParentDetails" NodeType="Column" ChildQuerierTypes="SqlColumn" DisableSort=""/>
-  <Node Name="Keys" LocLabel="SR.SchemaHierarchy_Keys" BaseClass="ModelBased" Strategy="ElementsInRelationship" NodeType="Key" ChildQuerierTypes="SqlIndex;SqlForeignKeyConstraint" >
+  <Node Name="Keys" LocLabel="SR.SchemaHierarchy_Keys" BaseClass="ModelBased" Strategy="ElementsInRelationship" NodeType="Key" ChildQuerierTypes="SqlIndex;SqlForeignKeyConstraint" ValidFor="NotSqlDw">
     <Filters>
       <Filter TypeToReverse="SqlIndex" Property="IndexKeyType" Type="Enum" ValidFor="Sql2016|Sql2017|AzureV12">
         <Value>IndexKeyType.DriPrimaryKey</Value>
@@ -209,6 +209,8 @@
     <Filters>
       <Filter TypeToReverse="SqlIndex" Property="IndexKeyType" Type="Enum" ValidFor="Sql2016|Sql2017|AzureV12">
         <Value>IndexKeyType.None</Value>
+        <Value>IndexKeyType.DriPrimaryKey</Value>
+        <Value>IndexKeyType.DriUniqueKey</Value>
       </Filter>
     </Filters>
   </Node>
--- a/test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestData/ObjectExplorer/Baselines/AllSqlObjects.txt
+++ b/test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestData/ObjectExplorer/Baselines/AllSqlObjects.txt
@@ -42,6 +42,7 @@ NodeType: Constraint Label: CK_Employee_MaritalStatus SubType: Status:
 NodeType: Constraint Label: CK_Employee_SickLeaveHours SubType: Status:
 NodeType: Constraint Label: CK_Employee_VacationHours SubType: Status:
 NodeType: Index Label: NonClusteredIndex-Login (Non-Unique, Non-Clustered) SubType: Status:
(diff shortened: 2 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
```diff
@@ -194,7 +194,7 @@
   </Node>
 
   <Node Name="Columns" LocLabel="SR.SchemaHierarchy_Columns" BaseClass="ModelBased" Strategy="PopulateParentDetails" NodeType="Column" ChildQuerierTypes="SqlColumn" DisableSort=""/>
-  <Node Name="Keys" LocLabel="SR.SchemaHierarchy_Keys" BaseClass="ModelBased" Strategy="ElementsInRelationship" NodeType="Key" ChildQuerierTypes="SqlIndex;SqlForeignKeyConstraint" >
+  <Node Name="Keys" LocLabel="SR.SchemaHierarchy_Keys" BaseClass="ModelBased" Strategy="ElementsInRelationship" NodeType="Key" ChildQuerierTypes="SqlIndex;SqlForeignKeyConstraint" ValidFor="NotSqlDw">
```
kevcunnane: DW doesn't have keys? Is this a bug in the SMO implementation or a real thing?

kevcunnane: If it's a bug, we should raise with their team and track internally...

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
```diff
@@ -209,6 +209,8 @@
     <Filters>
       <Filter TypeToReverse="SqlIndex" Property="IndexKeyType" Type="Enum" ValidFor="Sql2016|Sql2017|AzureV12">
         <Value>IndexKeyType.None</Value>
+        <Value>IndexKeyType.DriPrimaryKey</Value>
```
kevcunnane: Just checking - FKs show up with this change too, correct?

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestData/ObjectExplorer/Baselines/AllSqlObjects.txt
```diff
@@ -42,6 +42,7 @@ NodeType: Constraint Label: CK_Employee_MaritalStatus SubType: Status:
 NodeType: Constraint Label: CK_Employee_SickLeaveHours SubType: Status:
 NodeType: Constraint Label: CK_Employee_VacationHours SubType: Status:
 NodeType: Index Label: NonClusteredIndex-Login (Non-Unique, Non-Clustered) SubType: Status:
+NodeType: Index Label: PK_Employee_BusinessEntityID (Unique, Clustered) SubType:PrimaryKey Status:
```
kevcunnane: Would be great if the test could cover an FK too.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
llali: SSMS object explorer doesn't have a keys folder in dw database

---
In reply to: [154439990](https://github.com/Microsoft/sqltoolsservice/pull/557#discussion_r154439990) [](ancestors = 154439990)

kevcunnane: Yeah I was wondering if that's actually by design. Looking at unsupported features in https://docs.microsoft.com/en-us/azure/sql-data-warehouse/sql-data-warehouse-tables-overview, they aren't supported so you're right. We should probably take a look at that and see if there are any other things we should have filtered for DW too.

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestData/ObjectExplorer/Baselines/AllSqlObjects.txt
llali: there's already an example for FK (FK_Employee_Person_BusinessEntityID) in SSMS foreign keys show up under keys only not indexes  

---
In reply to: [154440262](https://github.com/Microsoft/sqltoolsservice/pull/557#discussion_r154440262) [](ancestors = 154440262)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
llali: I did a round of filtering unsupported objects for dw before and you can see Synonyms and Triggers are not available for dw. I'll verify them again

---
In reply to: [154453792](https://github.com/Microsoft/sqltoolsservice/pull/557#discussion_r154453792) [](ancestors = 154453792)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
llali: FK should only show up under key (based on ssms) and this change doesn't affect that 

---
In reply to: [154440157](https://github.com/Microsoft/sqltoolsservice/pull/557#discussion_r154440157) [](ancestors = 154440157)

PR #561, opened 2017-12-01
check in latest localized XLIFF for December update

Monasr:
check in latest localized XLIFF for December update

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/transXliff/sr.de.xlf
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/transXliff/sr.de.xlf
@@ -374,8 +374,8 @@
         <note/>
       </trans-unit>
       <trans-unit id="EE_ExecutionInfo_FinalizingLoop">
-        <source>Execution completed {0} times...</source>
-        <target state="new">Die Ausführung wurde {0} Mal abgeschlossen...</target>
+        <source>Batch execution completed {0} times...</source>
+        <target state="new">{0}-mal ausgeführt...</target>
         
         <note/>
       </trans-unit>
@@ -398,8 +398,8 @@
         <note/>
       </trans-unit>
       <trans-unit id="EE_ExecutionInfo_InitializingLoop">
-        <source>Starting execution loop of {0} times...</source>
-        <target state="new">{0}-malige Batchausführung wurde gestartet.</target>
+        <source>Beginning execution loop</source>
+        <target state="new">Ausführungsschleife wurde {0}-mal gestartet...</target>
         
         <note/>
       </trans-unit>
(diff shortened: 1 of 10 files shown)


src/Microsoft.SqlTools.ServiceLayer/Localization/transXliff/sr.de.xlf
```diff
@@ -398,8 +398,8 @@
         <note/>
       </trans-unit>
       <trans-unit id="EE_ExecutionInfo_InitializingLoop">
-        <source>Starting execution loop of {0} times...</source>
-        <target state="new">{0}-malige Batchausführung wurde gestartet.</target>
+        <source>Beginning execution loop</source>
+        <target state="new">Ausführungsschleife wurde {0}-mal gestartet...</target>
```
kevcunnane: This will cause trouble - we removed the numerical input so the translation needs updating to include this. Is it possible to do `Ausführungsschleife wurde gestartet...` instead?

src/Microsoft.SqlTools.ServiceLayer/Localization/transXliff/sr.de.xlf
```diff
@@ -566,10 +566,11 @@
         <note/>
       </trans-unit>
       <trans-unit id="EditDataCreateScriptMissingValue">
-        <source>A required cell value is missing</source>
+        <source>The column '{0}' is defined as NOT NULL but was not given a value</source>
```
kevcunnane: In this case, we have added in a numerical parameter. We should test that this doesn't impact localized versions. @benrr101 can you help verify this? I'm just concerned if it throws an error. Otherwise it's OK to still have the old translation available.

src/Microsoft.SqlTools.ServiceLayer/Localization/transXliff/sr.es.xlf
```diff
@@ -398,7 +398,7 @@
         <note/>
       </trans-unit>
       <trans-unit id="EE_ExecutionInfo_InitializingLoop">
-        <source>Starting execution loop of {0} times...</source>
+        <source>Beginning execution loop</source>
         <target state="new">Iniciando bucle de ejecución de {0} veces...</target>
```
kevcunnane: Again if we could alter to `Iniciando bucle de ejecución` this would be ideal. We no longer count number of times for this, so the translation would include `{0}` which looks wrong.

src/Microsoft.SqlTools.ServiceLayer/Localization/transXliff/sr.de.xlf
benrr101: Running the same code as is defined in the sr.cs, it doesn't throw if a parameter is provided but not used in the formatting string. However, if the string expects parameters but none are provided, it will throw.

PR #562, opened 2017-12-02
Edit Data: Better Formatting Errors

benrr101:
This change is to resolve https://github.com/Microsoft/carbon/issues/2642 It improves error messages for formatting errors such that you get the type of the column back. It looks like this:

improvedformatmessages

In order to do this, I had to build a utility function that would generate the TSQL version of the column type from the column. This code is based on the same logic from object explorer. There are unit tests for it, too!

I also refactored script helpers to make it more clear which ones were being used for what.


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/CellUpdate.cs
```diff
@@ -34,50 +35,60 @@ public CellUpdate(DbColumnWrapper column, string valueAsString)
...
+            catch (FormatException fe)
             {
-                // Attempt to go straight to the destination type, if we know what it is, otherwise
-                // leave it as a string
-                Value = columnType != null 
-                    ? Convert.ChangeType(valueAsString, columnType, CultureInfo.CurrentCulture) 
-                    : valueAsString;
-                ValueAsString = Value.ToString();
+                // Pretty up the exception so the user can learn a bit from it
+                // NOTE: Other formatting errors raised by helpers are InvalidOperationException to
```
kevcunnane: I like this, good use of exception types to differentiate the handling.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -303,7 +305,11 @@ TestLocalizationConstant = test
 ############################################################################
 # Utilities
 
-SqlScriptFormatterDecimalMissingPrecision = Decimal column is missing numeric precision or numeric scale
+SqlScriptFormatterDecimalMissingPrecision = Exact numeric column is missing numeric precision or numeric scale
```
kevcunnane: Minor: if you can make sure these errors align to SSMS / SSDT errors, that'd be great.

kevcunnane: E.g. if we already have standardized string error messages it's nice to reuse as they were previously reviewed & localized.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatters/Common.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.SqlTools.ServiceLayer.Utility.SqlScriptFormatters
+{
+    public class Common
```
kevcunnane: Can you remove this since it's not implementing anything at all?

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatters/FromSqlScript.cs
```diff
@@ -0,0 +1,148 @@
...
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Utility.SqlScriptFormatters
+{
+    /// <summary>
+    /// Provides utilities for converting from SQL script syntax into POCOs.
+    /// </summary>
+    public static class FromSqlScript
+    {
+        private static readonly Regex StringRegex = new Regex("^N?'(.*)'$", RegexOptions.Compiled);
```
kevcunnane: Minor: always comment on the regex in case we add to it later. In this case it's to `Detect String Literals inside error messages`?

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatters/FromSqlScript.cs
```diff
@@ -0,0 +1,148 @@
...
+            StringBuilder sb = new StringBuilder();
+            List<string> namedParts = new List<string>();
+            bool insideBrackets = false;
+            bool bracketsClosed = false;
+            for (int i = 0; i < multipartIdentifier.Length; i++)
+            {
+                char iChar = multipartIdentifier[i];
+                if (insideBrackets)
+                {
+                    if (iChar == ']')
```
kevcunnane: Are there cases where we'd have quoted identifiers `"dbo"."test"` instead of brackets? Do we need to account for that? Or do we always control the input here? If this is used for arbitrary scripts this is important. If it's just for internally produced ones that always have `[` then it can be ignored

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatters/FromSqlScript.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// parts of the identifier. Implemented as a state machine that iterates over the
+        /// characters of the multipart identifier.
+        /// </summary>
+        /// <param name="multipartIdentifier">Multipart identifier to decode (eg, "[dbo].[test]")</param>
+        /// <returns>The parts of the multipart identifier in an array (eg, "dbo", "test")</returns>
+        /// <exception cref="FormatException">
+        /// Thrown if an invalid state transition is made, indicating that the multipart identifer
+        /// is not valid. 
+        /// </exception>
+        public static string[] DecodeMultipartIdenfitier(string multipartIdentifier)
```
kevcunnane: Typo: should be `Identifier`

kevcunnane: Don't we have utility methods that already do this? We certainly have ones in the DacFx codebase. There are a number of issues here:
- Whether to handle quoted identifiers or not
- Detection of escaped `[` and `]` characters, which isn't handled
- Potentially others. Quoted identifier handling is an area where I've seen so many bugs raised in the past, that the only "safe" way to handle is to reuse existing code or waste lots of time.

I'm going to wait on the PR and sync with you on where to get resusable code from. There's almost certainly a util method in our existing code that we reference / have ported over.

kevcunnane: OK looking further... I realize you already wrote this (and it's already shipped this way). Checking with Chris, we could use `Microsoft.SqlServer.SqlParser/Parser/EscapeSequence.cs#833153` for these. It covers unescaping both escaped characters and escaped literals. If possible, could we look at using this since it's battle hardened?

llali:
> using System;

copyright missing

---
Refers to: test/Microsoft.SqlTools.ServiceLayer.UnitTests/UtilityTests/FromSqlScriptTests.cs:1 in cab400a. [](commit_id = cab400a607581778d37cee7a1d8c8e561fb54ad6, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatters/Common.cs
benrr101: It *was*... I guess it isn't anymore

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatters/FromSqlScript.cs
benrr101: Right, this is only used for input from ourselves. The values that are being processed here are from SMO or SqlClient. I'm not 100% certain on the SMO stuff (only stuff I'm using is default constraint values) because the values aren't escaped already.

src/Microsoft.SqlTools.ServiceLayer/Utility/SqlScriptFormatters/FromSqlScript.cs
benrr101: I suppose that's fine, but didn't we already try using SqlParser and then got rid of it because it was buggy?

PR #563, opened 2017-12-04
Feature/script execute

llali:
- Added operation type to scriptingRequest to be able to set it to "Select" or "Execute" 
  this property is already set from sqlopsstudio and we were not using it in the service. It's better use that for  "select" script also. using ScriptCreateDrop for select script didn't make sense
  
- Merged scripting and script as requests  one

- Implemented script as execute for stored procedure 

- Removed the binding queue. Getting connection from binding queue and using it in different task without locking the connection was not a good practice and could cause issues. So for now creating new connection for each script request and will improve it later

- Moved the logic for generating script as select to ScriptAsOperation


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -134,90 +134,42 @@ public void InitializeService(ServiceHost serviceHost)
...
-        }
-
-            /// <summary>
-            /// Handles request to start the scripting operation
-            /// </summary>
-            public async Task HandleScriptExecuteRequest(ScriptingParams parameters, RequestContext<ScriptingResult> requestContext)
+        private bool ShouldCreateScriptAsOperation(ScriptingParams parameters)
         {
-            try
+            // Scripting as oepration should be used to script one object.
```
abist: nit: small typo in `operation`

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -106,18 +137,323 @@ public override void Execute()
...
+            StringBuilder parameterList = new StringBuilder();
+            if (sp == null || parentObject == null)
+            {
+                throw new InvalidOperationException(SR.ScriptingExecuteNotSupportedError);
+            }
+            WriteUseDatabase(parentObject, executeStatement, options);
+
+            // character string to put in front of each parameter. First one is just carriage return
+            // the rest will have a "," in front as well.
+            string paramListPreChar = "\r\n   ";
```
kburtram: >\r\n [](start = 39, length = 4)

you may use Environment.NewLine instead of \r\n so that the newlines are generated properly for the platform.

PR #565, opened 2017-12-07
fixed the keys filter to support 2012 and 2014 also

llali:
not sure why the index filter didn't include 2012 and 2014. the index type is valid for those version.


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
```diff
@@ -196,7 +196,7 @@
   <Node Name="Columns" LocLabel="SR.SchemaHierarchy_Columns" BaseClass="ModelBased" Strategy="PopulateParentDetails" NodeType="Column" ChildQuerierTypes="SqlColumn" DisableSort=""/>
   <Node Name="Keys" LocLabel="SR.SchemaHierarchy_Keys" BaseClass="ModelBased" Strategy="ElementsInRelationship" NodeType="Key" ChildQuerierTypes="SqlIndex;SqlForeignKeyConstraint" ValidFor="NotSqlDw">
     <Filters>
-      <Filter TypeToReverse="SqlIndex" Property="IndexKeyType" Type="Enum" ValidFor="Sql2016|Sql2017|AzureV12">
+      <Filter TypeToReverse="SqlIndex" Property="IndexKeyType" Type="Enum" ValidFor="Sql2012|Sql2014|Sql2016|Sql2017|AzureV12">
```
kevcunnane: Were these valid for all versions? E.g. is this not supported in 2005 and up since it's just indexes?

PR #568, opened 2017-12-11
Added functions to metadata list result

llali:


src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -204,6 +207,11 @@ internal static void ReadMetadata(SqlConnection sqlConn, List<ObjectMetadata> me
                             metadataType = MetadataType.SProc;
                             metadataTypeName = "StoredProcedure";
                         }
+                        else if (objectType == "AF" || objectType == "FN" || objectType == "IF")
```
kevcunnane: Minor: would be nice to have test verifying each kind.

llali: the test I added covers table valued and scalar valued functions. not sure how to create aggregate functions to test :) 

---
In reply to: [156225353](https://github.com/Microsoft/sqltoolsservice/pull/568#discussion_r156225353) [](ancestors = 156225353)

PR #569, opened 2017-12-11
Mark database OE nodes as leaf nodes if they are unavailable

MattIrv:
This is the tools service work to fix https://github.com/Microsoft/sqlopsstudio/issues/189

Right now SQL Ops Studio's code assumes that any database with a non-blank status is not available, which is not correct. This PR adds code that determines whether a database is available or not based on its status, and if the database is not available then it marks it as a leaf node.

After this is merged, I'll update SQL Ops Studio to use this property instead of making the availability determination itself.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
@@ -245,6 +245,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 
                 childAsMeItem.NodeSubType = GetNodeSubType(context, smoContext);
                 childAsMeItem.NodeStatus = GetNodeStatus(context, smoContext);
+                childAsMeItem.Properties = GetExtraProperties(context, smoContext);
             }
         }
 
@@ -320,6 +321,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
         public override string GetNodeCustomName(object smoObject, SmoQueryContext smoContext)
         {
             return string.Empty;
+        }
+
+        public override Dictionary<string, object> GetExtraProperties(object smoObject, SmoQueryContext smoQueryContext)
+        {
+            return new Dictionary<string, object>();
         }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
@@ -245,7 +245,6 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 
                 childAsMeItem.NodeSubType = GetNodeSubType(context, smoContext);
                 childAsMeItem.NodeStatus = GetNodeStatus(context, smoContext);
-                childAsMeItem.Properties = GetExtraProperties(context, smoContext);
             }
         }
 
@@ -321,11 +320,6 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
         public override string GetNodeCustomName(object smoObject, SmoQueryContext smoContext)
         {
             return string.Empty;
-        }
-
-        public override Dictionary<string, object> GetExtraProperties(object smoObject, SmoQueryContext smoQueryContext)
-        {
-            return new Dictionary<string, object>();
         }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
@@ -14,14 +14,54 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
     /// </summary>
     internal partial class DatabasesChildFactory : SmoChildFactoryBase
     {
+        private const string databaseUnavailableKey = "databaseUnavailable";
+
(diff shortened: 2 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
```diff
@@ -14,14 +14,54 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+
         public override string GetNodeStatus(object smoObject, SmoQueryContext smoContext)
         {
             return DatabasesCustomNodeHelper.GetStatus(smoObject, smoContext, CachedSmoProperties);
         }
+
+        public override Dictionary<string, object> GetExtraProperties(object smoObject, SmoQueryContext smoQueryContext)
+        {
+            var properties = new Dictionary<string, object>();
+            properties.Add(databaseUnavailableKey, DatabasesCustomNodeHelper.GetDatabaseIsUnavailable(smoObject, smoQueryContext, CachedSmoProperties));
```
kevcunnane: Minor: a nice pattern can be allowing an `AddExtraProperties` method that subclasses implement. This has 2 benefits:
- You always guarantee the dictionary is non-null
- If you have a node inheriting from its parent and both add properties it can call `super.AddExtraProperties` and get the combination instead of having to merge dictionaries

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
```diff
@@ -14,14 +14,54 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+            return properties;
+        }
     }
 
     internal static class DatabasesCustomNodeHelper
     {
+        private static readonly DatabaseStatus[] unavailableDatabaseStatuses = { DatabaseStatus.Inaccessible, DatabaseStatus.Offline, DatabaseStatus.Recovering,
```
kevcunnane: Does this give a style warning? I thought static readonly fields were CapitalCase

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
```diff
@@ -14,14 +14,54 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+            DatabaseStatus.RecoveryPending, DatabaseStatus.Restoring, DatabaseStatus.Suspect, DatabaseStatus.Shutdown };
+
+        internal static bool GetDatabaseIsUnavailable(object smoObject, SmoQueryContext smoContext, IEnumerable<NodeSmoProperty> supportedProperties)
+        {
+            Database db = smoObject as Database;
+            if (db != null && SmoChildFactoryBase.IsPropertySupported("Status", smoContext, db, supportedProperties))
```
kevcunnane: Just confirming for others: Status is pre-fetched as a Property, as defined in the TreeDefinition.xml. As this has perf implications, any property checks in OE should have this verified during PRs 😄

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
MattIrv: No style warning, and elsewhere in the code we're not consistent between starting them with an uppercase and lowercase letter (search 'private static readonly' if you're interested).

Is one of them preferred?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
MattIrv: Since this method is part of the factory, not the node itself, I'm not really sure what the best way to implement it is.

I've added a call to `base.GetExtraProperties` in the factory method implementation, but it looks like most node types don't use one of these factories.

Would you suggest changing the node code somehow?

MattIrv:
@anthonydresser since we use a single class to represent OE nodes, instead of having a separate class for each type of node, I don't think it would make sense to have an `isAvailable` property for most non-database nodes.

I agree that this may make it tough since it won't be an explicit part of the contract though, but the only other options I see are to add an `isAvailable` property that exists on all node types but is only relevant for databases, or to start using a new unique class for database nodes, which would be a pretty complicated refactoring here.

Any thoughts on what you'd prefer?

anthonydresser:
@MattIrv you could use the `isAlwaysLeaf` property rather than making a new one. For inaccessible databases, set isAlwaysLeaf to true. This would also mean it requires no changes on the front end side.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -321,5 +322,10 @@ public override string GetNodeCustomName(object smoObject, SmoQueryContext smoCo
         {
             return string.Empty;
         }
+
+        public override Dictionary<string, object> GetExtraProperties(object smoObject, SmoQueryContext smoQueryContext)
+        {
+            return new Dictionary<string, object>();
```
kburtram: >new Dictionary<string, object>() [](start = 19, length = 32)

should this value be cached after first creation so (1) the same object is returned each time this method is called (2) we don't reallocate memory each time the function is called?

kburtram: ..for the use case it looks like the properties are being added to the collection each time the derived method is called, so I guess you do need to reallocate each time with current design.

---
In reply to: [156536525](https://github.com/Microsoft/sqltoolsservice/pull/569#discussion_r156536525) [](ancestors = 156536525)

PR #571, opened 2017-12-13
Auto detect json columns

anthonydresser:
Added a regex eval for the first row of each column to auto detect json columns since the engine does not support it by default.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
@@ -8,6 +8,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Data.Common;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -589,6 +591,28 @@ private void SingleColumnXmlJsonResultSet()
...
         }
 
+        /// <summary>
+        /// Check columns for json type and set isJson if needed
+        /// </summary>
+        private void CheckForIsJson()
+        {
+            if (Columns?.Length > 0 && RowCount != 0)
+            {
+                Regex regex = new Regex(@"({.*?})");
```
kburtram: will this regex have a lot of false positives?  (i.e., non JSON content tagged as JSON?).  Overall I'm fine with this and we can tune the regex is get customer complaints, though if there is a simple check to the make the regex a little more restrictive that may be worth doing now...for example, is there at least on field separated be a ':' or something like that.

kburtram:
@anthonydresser are you planning to do anything with this PR?  Or should we close it?

anthonydresser:
I'll check this in and we'll react to user feedback

PR #576, opened 2018-01-10
Edit Data: Fix to work with tables with triggers

benrr101:
This change rewrites for a third (and hopefully final time) the logic that generates INSERT and UPDATE scripts in edit data.

**The issue:** Editing tables with triggers fails because of some rule in SQL Server that says you can't use OUTPUT clauses when there's a trigger on that table. OUTPUT clauses are used in the edit data service to immediately get the value of a calculated column (id, etc) in one slick command.

**The solution:** As a workaround, if you OUTPUT INTO a temporary table, you can still use the OUTPUT clause. The change here is to change UPDATE/INSERT scripts into DECLARE, UPDATE/INSERT, SELECT scripts. One quirk of this is that the types in the temp table can't always be the types of their original columns (ie, you can't directly insert a TIMESTAMP column value). To work around this, the FormatColumnType method has been updated to return the semantic equivalent on request (for TIMESTAMP this is VARBINARY(8)).


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -88,13 +89,72 @@ public override DbCommand GetCommand(DbConnection connection)
...
+            string declareStatement = string.Format(DeclareStatement, tempTableName, string.Join(", ", declareColumns));
+
+            // Step 2) Build the insert statement
+            string joinedOutClauseNames = string.Join(", ", outClauseColumnNames);
+            string insertStatement = inValues.Count > 0
+                ? string.Format(InsertOutputValuesStatement, 
+                    AssociatedObjectMetadata.EscapedMultipartName,
+                    string.Join(", ", inColumnNames), 
+                    joinedOutClauseNames,
+                    tempTableName,
```
kburtram: >tempTableName [](start = 20, length = 13)

did you happen to investigate how SSMS supports this scenario?  I wonder if using a temp table is the simplest way to fix this bug?

benrr101: SSMS doesn't do update/insert using an OUTPUT clause. As such it sidesteps the issue entirely, but raises the issue of how to handle calculated column values. Using a temp table is the simplest way to address the issue of OUTPUT clauses on tables with triggers according to various posts on the internet.

benrr101:
@kburtram As far as I know, since we're using table variables and not temp tables (which I may have mistaken called them), the variable will be cleaned up as soon as the batch completes (ie, the command). https://odetocode.com/articles/365.aspx

I'm about 85% confident that there won't be any regressions introduced from this, but we should still give it a once over via our pre-release test.

PR #577, opened 2018-01-17
Bugs fixes so XEvent profiler isn't broken

kburtram:
This is an PR for work in progress so Profiler isn't broken while we're waiting to schedule finishing the feature.  This isn't meant to be final iteration of the work.


kburtram:
..this PR is to shelve Profiler work in-progress so it's not left in broken state.  So this isn't for code that's enabled in any client.

PR #579, opened 2018-02-01
Fix test build break

kburtram:
test break in the integration tests from previous profiler check-in.


kburtram:
Fixing test break, so merging to get code coverage back up.

PR #581, opened 2018-02-06
fixed the issue with default values for options not loaded correctly

llali:
fixes https://github.com/Microsoft/sqlopsstudio/issues/491


kburtram:
@llali I'll merge this and update the tools service reference in SQL Ops to pick up this change in the next build.

PR #589, opened 2018-03-13
Support Object Explorer FindNodes request

MattIrv:
These are the SQL Tools Service changes to support Microsoft/sqlopsstudio#572

Object explorer service now handles the new find nodes request, which it does by getting information about the structure of the Object Explorer tree from `TreeNodeDefinition.xml` and generating all the node paths that match the given metadata, then returning any nodes that actually exist at those paths.


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -46,6 +50,7 @@ public class ObjectExplorerService : HostedService<ObjectExplorerService>, IComp
         private IMultiServiceProvider serviceProvider;
         private ConnectedBindingQueue bindingQueue = new ConnectedBindingQueue(needsMetadata: false);
         private string connectionName = "ObjectExplorer";
+        private Dictionary<string, HashSet<TreeFolder>> typeMap = new Dictionary<string, HashSet<TreeFolder>>();
```
llali: make it ConcurrentDictionary or static to be thread safe

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -689,6 +705,204 @@ private static void AddToApplicableChildFactories(Dictionary<string, HashSet<Chi
...
+                {
+                    nodes.Add(treeNode);
+                }
+            }
+            return nodes;
+        }
+
+        private void LoadTreeStructure()
+        {
+            var assembly = typeof(ObjectExplorerService).Assembly;
```
llali: catch exception and log the error, if for some reason the resource cannot be loaded, the exception will cause the service to fail

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -243,6 +243,26 @@ public ObjectExplorerServiceTests()
             connectionServiceMock.Verify(c => c.Disconnect(It.IsAny<DisconnectParams>()));
         }
 
+        [Fact]
```
llali: please add test for finding column. basically verifying all different options in FindNoresRequest

PR #592, opened 2018-03-15
Feature/agent1 adbist

abist:
Some refactoring following last code review.


src/Microsoft.SqlTools.ServiceLayer/Agent/JobUtilities.cs
```diff
@@ -0,0 +1,106 @@
...
+using Microsoft.SqlServer.Management.Sdk.Sfc;
+using Microsoft.SqlServer.Management.Smo.Agent;
+using Microsoft.SqlTools.ServiceLayer.Agent.Contracts;
+using SMO = Microsoft.SqlServer.Management.Smo;
+
+namespace Microsoft.SqlTools.ServiceLayer.Agent 
+{
+    public class JobUtilities
+    {
+        public const string urnJobName = "JobName";
```
kburtram: public const should typically be Pascal cased, so this should be UrnJobName.  I probably make all these consts follow this naming convention for consistency.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -91,22 +91,6 @@ public void InitializeService(ServiceHost serviceHost)
 
 
         //private const string cUrnEnumerateAgentJobs = "Server/JobServer/Job";
```
kburtram: can you delete this line as well.

PR #597, opened 2018-03-28
Agent work (WIP)

abist:
Separated the steps from the job history to view it separately on the job history page. Sql Ops side PR for the same here - https://github.com/Microsoft/sqlopsstudio/pull/1012


src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -147,12 +147,39 @@ public void InitializeService(ServiceHost serviceHost)
                     DataTable dt = tuple.Item2;
                     int count = dt.Rows.Count;
                     var agentJobs = new List<AgentJobHistoryInfo>();
+
+                    var agentStepMap = new Dictionary<DateTime, List<AgentJobStep>>();
```
kburtram: is RunDate the best way to group these?  I assume these nothing like an ExecutonId field or something?

src/Microsoft.SqlTools.ServiceLayer/Agent/JobUtilities.cs
```diff
@@ -56,6 +56,16 @@ public static AgentJobInfo ConvertToAgentJobInfo(JobProperties job)
             };
         }
 
+        public static Boolean IsStep(DataRow row, SqlConnectionInfo sqlConnectionInfo)
```
kburtram: `Boolean` should use C# primitive type `bool`

src/Microsoft.SqlTools.ServiceLayer/Agent/JobUtilities.cs
```diff
@@ -56,6 +56,16 @@ public static AgentJobInfo ConvertToAgentJobInfo(JobProperties job)
...
+        {
+            int stepId = Convert.ToInt32(row[UrnStepID], System.Globalization.CultureInfo.InvariantCulture);
+            if (stepId != 0)
```
kburtram: nit: lines 62-66 can be written as `return stepId != 0;`

src/Microsoft.SqlTools.ServiceLayer/Agent/JobUtilities.cs
```diff
@@ -99,8 +109,21 @@ public static AgentJobHistoryInfo ConvertToAgentJobHistoryInfo(DataRow jobRow, S
...
-
             return jobHistoryInfo;
         }
+
+        public static AgentJobStep ConvertToAgentJobStep(DataRow jobRow, SqlConnectionInfo sqlConnInfo)
+        {
+            int stepId = Convert.ToInt32(jobRow[UrnStepID], System.Globalization.CultureInfo.InvariantCulture);
+            string stepName = Convert.ToString(jobRow[UrnStepName], System.Globalization.CultureInfo.InvariantCulture);
+            string message = Convert.ToString(jobRow[UrnMessage], System.Globalization.CultureInfo.InvariantCulture);
+            DateTime runDate = Convert.ToDateTime(jobRow[UrnRunDate], System.Globalization.CultureInfo.InvariantCulture);
```
kburtram: any concerns that any of these `Convert` calls may throw an exception based on input data?  Or do we know the inputs will always be safe for some reason?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
abist: RunDate was the only thing I could find to tie the steps to the jobs. They all have the same `jobID`'s, and no other `ID` field.

src/Microsoft.SqlTools.ServiceLayer/Agent/JobUtilities.cs
abist: It will always be safe as `JobHistoryItem.cs` handles and populates the fields the same way.

PR #599, opened 2018-04-05
Expand nodes for findNodes instead of refreshing

MattIrv:
Microsoft/sqlopsstudio#1076 happens because Object Explorer's `findNodes` API call refreshes all nodes along its path, which resets all of their children objects. This PR changes the logic to call `Expand` instead of `Refresh` so that objects that are already expanded will not have their children reset when calling `findNodes`.


kburtram:
I want to include this change in an updated SQL Ops build.  So I'll merge it now.

PR #600, opened 2018-04-06
Use object names instead of labels in node paths

MattIrv:
Right now Object Explorer node paths come directly from object labels, which normally works fine but causes problems when the label includes additional information about the object (see https://github.com/Microsoft/sqlopsstudio/issues/1090)

This PR changes the logic to always use the object name in that case, which will cause node paths to look like the expected ones from that issue

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
@@ -241,6 +241,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
                 if (!string.IsNullOrEmpty(customizedName))
                 {
                     childAsMeItem.NodeValue = customizedName;
+                    childAsMeItem.NodePathName = GetNodePathName(context);
                 }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
```diff
@@ -321,5 +322,10 @@ public override string GetNodeCustomName(object smoObject, SmoQueryContext smoCo
         {
             return string.Empty;
         }
+
+        public override string GetNodePathName(object smoObject)
+        {
+            return (smoObject as NamedSmoObject).Name;
```
llali: I'm not sure if all SMO objects are NamedSmoObject, do you have to check for null here?

kburtram:
@MattIrv I'll merge to include in next VSO rebuild

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoChildFactoryBase.cs
alextercete: This doesn't work for tables that need a customized name (like temporal tables), because it will omit the schema. Looks like this method needs to be overridden in `TablesChildFactory` and `TableChildFactory`.

PR #605, opened 2018-04-18
Add support for using generic queries to filter EditData rows.

corivera:


test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/SessionTests.cs
```diff
@@ -194,14 +194,17 @@ public void CreateRowNotInitialized()
...
+                    EscapedName = cols[0].ColumnName
                 },
                 new EditColumnMetadata                   // Has default
                 {
                     DefaultValue = "default",
-                    EscapedName = cols[0].ColumnName,
+                    EscapedName = cols[1].ColumnName
                 },
-                new EditColumnMetadata()
+                new EditColumnMetadata
```
MattIrv: Why this change? Will real columns always have an `EscapedName`?

corivera: I haven't seen any cases where it's empty, and the rest of the code just seems to assume it's always there. This test case also fails in the new filtering code path if there's no column name.

corivera:
Working on adding some unit tests.

@benrr101 Can you take a look at this PR?

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -55,6 +59,39 @@ public EditTableMetadata()
 
         #endregion
 
+        // Filters out metadata that is not present in the result set, and matches metadata ordering to resultset
```
benrr101: Please use the C# XML style docs for all public methods

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -55,6 +59,39 @@ public EditTableMetadata()
...
+        public static EditColumnMetadata[] FilterColumnMetadata(EditColumnMetadata[] metaColumns, DbColumnWrapper[] resultColumns)
+        {
+            if(metaColumns.Length == 0)
+            {
+                return metaColumns;
+            }
+
+            bool escapeColName = metaColumns[0].EscapedName.Contains("[");
```
benrr101: As per discussion, let's try to keep name escaping and whatnot inside the ToSqlScript/FromSqlScript classes

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -55,6 +59,39 @@ public EditTableMetadata()
...
+            for (int i = 0; i < resultColumns.Length; i++)
+            {
+                DbColumnWrapper column = resultColumns[i];
+                string columnName =  column.ColumnName;
+                if (escapeColName)
+                {
+                    columnName = columnName.Contains("[") ? columnName : ToSqlScript.FormatIdentifier(columnName);
+                }
+
+                columnNameOrdinalMap.Add(columnName, column.ColumnOrdinal.HasValue ? column.ColumnOrdinal.Value : i);
```
benrr101: As per discussion, let's see what happens if the user uses aliases or has duplicate columns via aliases

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -3,8 +3,12 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
```
benrr101: Don't forget to clean up using's

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
```diff
@@ -45,8 +45,10 @@ protected RowEditBase(long rowId, ResultSet associatedResultSet, EditTableMetada
             }
 
             RowId = rowId;
-            AssociatedResultSet = associatedResultSet;
             AssociatedObjectMetadata = associatedMetadata;
+            AssociatedResultSet = associatedResultSet;
+
+            AssociatedObjectMetadata.Columns = EditTableMetadata.FilterColumnMetadata(AssociatedObjectMetadata.Columns, AssociatedResultSet.Columns);
```
benrr101: AFAICT, this is already being done when the associated metadata is extended. If we remove this operation, then we can make the FilterColumnMetadata a private helper method.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/SessionTests.cs
benrr101: Yep, IIRC, a column name is always provided by the server. Worst case it's an empty string, but it's never null, IIRC.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/SessionTests.cs
```diff
@@ -194,14 +194,17 @@ public void CreateRowNotInitialized()
...
                 {
                     DefaultValue = null,
-                    EscapedName = cols[0].ColumnName,
+                    EscapedName = cols[0].ColumnName
                 },
                 new EditColumnMetadata                   // Has default
                 {
                     DefaultValue = "default",
-                    EscapedName = cols[0].ColumnName,
+                    EscapedName = cols[1].ColumnName
```
benrr101: Good catch finding this 👍

corivera:
Closing this PR for now, since this approach needs further review

corivera:
Reopened this PR. I replaced the client-side query string parsing with metadata checking in tools service. The main bulk of these new changes are in Batch.cs and in EditSession.cs. 

The new changes query for additional column metadata and extend the metadata in the resultset. This has to be done in a separate query since the KeyInfo command behavior also appends the primary key columns to the result set, even if they weren't queried for. The metadata is then validated in EditSession once we get a result set back. 

Currently these are the scenarios EditData queries fail on:
1. Multiple Tables, as indicated by duplicate source table names, schemas, or databases.
2. Aliased columns.
3. Expression columns.
4. Multiple columns with the same name.
5. Querying a single table that is not the original table the session was initialized with.

@benrr101 Can you take a look at the new changes?

And just generally, is there any way to add tests for this extra metadata? When trying to add unit tests I wasn't able to get the extra metadata from the mock queries.

corivera:
Currently investigating this new behavior when querying views, since the column metadata for those looks the same as the original source table. Theoretically, when querying a view we can correctly fail due to duplicate table names or column names, or actually update the data in the original table for the corresponding view column.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditTableMetadata.cs
```diff
@@ -55,6 +58,40 @@ public EditTableMetadata()
...
+            {
+                return metaColumns;
+            }
+
+            bool escapeColName = FromSqlScript.IsIdentifierBracketed(metaColumns[0].EscapedName);
+            Dictionary<string, int> columnNameOrdinalMap = new Dictionary<string, int>(capacity: resultColumns.Length);
+            for (int i = 0; i < resultColumns.Length; i++)
+            {
+                DbColumnWrapper column = resultColumns[i];
+                string columnName =  column.ColumnName;
```
MattIrv: extra space

PR #606, opened 2018-04-26
Feature/agent finishes

abist:
Changed the way we get histories for jobs. Realized if we made the LogService a little less strict, we can get the histories for very little overhead and with much more details (Duration in actual time with units, for example). With this change, the Jobs History page loads nearly instantly and most histories for jobs are already loaded in the Jobs View page itself. This will be very helpful for the "Status Health Chart", we want to add later on.

The accompanying Ops Studio PR is here - https://github.com/Microsoft/sqlopsstudio/pull/1284


src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -142,47 +141,27 @@ public void InitializeService(ServiceHost serviceHost)
...
-                                agentStepMap[agentJobStep.RunDate] = agentJobSteps;
-                            }
-                        }
-                        else
-                        {
-                            var agentJobHistoryInfo = JobUtilities.ConvertToAgentJobHistoryInfo(job, sqlConnInfo);
-                            agentJobs.Add(agentJobHistoryInfo);
-                        }
+                    List<AgentJobHistoryInfo> jobHistories = new List<AgentJobHistoryInfo>();
+                    if (count > 0) {
```
kburtram: `{` should be on its own line in C#

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -142,47 +141,27 @@ public void InitializeService(ServiceHost serviceHost)
...
-                    {
-                        if (agentStepMap.ContainsKey(agentJobHistoryInfo.RunDate))
-                        { 
-                            var agentStepList = agentStepMap[agentJobHistoryInfo.RunDate].ToList();
-                            agentStepList.Sort(delegate (AgentJobStep s1, AgentJobStep s2) { return s1.StepId.CompareTo(s2.StepId); });
-                            agentJobHistoryInfo.Steps = agentStepList.ToArray();
-                        }
-                    }
-                    result.Jobs = agentJobs.ToArray();
+                    ServerConnection connection = tuple.Item3;
```
kburtram: I'm not a huge fan of tuples, but if you're going to use them perhaps you should move all the values into name variables closer to where the tuple is created.  Basically moving this variable declaration to line 147.

src/Microsoft.SqlTools.ServiceLayer/Agent/JobHistoryItem.cs
```diff
@@ -173,6 +135,13 @@ public string JobName
                 return m_jobName;
             }
         }
+
+        public List<ILogEntry> LogEntries {
```
kburtram: `{` character should be on it's own line

src/Microsoft.SqlTools.ServiceLayer/Agent/JobUtilities.cs
```diff
@@ -123,5 +73,46 @@ public static AgentJobStep ConvertToAgentJobStep(DataRow jobRow, SqlConnectionIn
             step.RunStatus = runStatus;
             return step;
         }
+
+        public static List<AgentJobHistoryInfo> ConvertToAgentJobHistoryInfo(List<ILogEntry> logEntries, DataRow jobRow) {
```
kburtram: `{` should be on its own line.  Please make this change for all `{` characters in the PR.

src/Microsoft.SqlTools.ServiceLayer/Agent/JobUtilities.cs
```diff
@@ -123,5 +73,46 @@ public static AgentJobStep ConvertToAgentJobStep(DataRow jobRow, SqlConnectionIn
...
+            List<AgentJobHistoryInfo> jobs = new List<AgentJobHistoryInfo>();
+            // get all the values for a job history
+            foreach (ILogEntry entry in logEntries) 
+            {
+                // Make a new AgentJobHistoryInfo object
+                var jobHistoryInfo = new AgentJobHistoryInfo();
+                jobHistoryInfo.InstanceId = Convert.ToInt32(jobRow[UrnInstanceID], System.Globalization.CultureInfo.InvariantCulture);
+                jobHistoryInfo.RunStatus = Convert.ToInt32(jobRow[UrnRunStatus], System.Globalization.CultureInfo.InvariantCulture);
+                jobHistoryInfo.JobId = (Guid) jobRow[UrnJobId];
```
kburtram: presumably this cast won't ever throw?

PR #607, opened 2018-04-27
Reduce unit test time by 1 minute

kevcunnane:
ObjectExplorerTests were adding 1 minute to total test time due to
unnecessarily trying and failing to connect to a non-existent server.
Mocking out this call saves 1minute, reducing test time in half


kevcunnane:
Thanks Karl. I'll just check this in as these tests should not run differently in the lab.

PR #608, opened 2018-04-30
Bump System.Data.SqlClient package version used in SqlTools projects.

corivera:
The EditData query filtering changes require a version bump in the referenced SqlClient package version, since earlier versions have a bug where some column schema metadata doesn't get populated properly. The complete metadata is needed when checking if EditData queries are valid (e.g. query is not querying multiple tables in a database, not querying same-name tables in different databases, etc).

I'm pushing up these changes separately so that we can do some testing with just this version bump while the broader EditData changes are pending approval.

I also bumped all the SqlClient references in the repo for the sake of consistency.

### DIFF
--- a/src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
+++ b/src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
@@ -18,7 +18,7 @@
 	<Reference Include="System.Data.SqlClient" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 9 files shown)


corivera:
@kburtram Looks like the CI build is still installing .NET Core SDK 2.0. Any idea where I would update that?

kburtram:
@corivera don't worry about the CI builds.  I need to fix that later this week.

Do we also need to update the version of SqlClient we use to build SMO?  Or is this change sufficient?

corivera:
@kburtram I only need the extra metadata in the result set columns, so this change should be sufficient. We execute the queries for those in tools service.

PR #611, opened 2018-05-01
Only register message handler once for GO N queries

MattIrv:
Fixes Microsoft/sqlopsstudio#1310 by only registering the message handler once when executing a batch. This fixes a problem where messages got sent multiple times after running `GO N` queries


kburtram:
@MattIrv I'm going to queue a new build to integrate to sqlops now so I'll merge this to be included.

PR #613, opened 2018-05-02
fixed the bug with editing more than one row should fail

llali:
fixes https://github.com/Microsoft/sqlopsstudio/issues/1236

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
@@ -32,7 +32,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
         private const string UpdateScript = "UPDATE {0} SET {1} {2}";
         private const string UpdateScriptMemOptimized = "UPDATE {0} WITH (SNAPSHOT) SET {1} {2}";
         private const string SelectStatement = "SELECT {0} FROM {1}";
-        private string ValidateUpdateOnlyOneRow = "DECLARE @numberOfRows int = 0;" + Environment.NewLine +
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
```diff
@@ -30,6 +32,14 @@ public sealed class RowUpdate : RowEditBase
         private const string UpdateScript = "UPDATE {0} SET {1} {2}";
         private const string UpdateScriptMemOptimized = "UPDATE {0} WITH (SNAPSHOT) SET {1} {2}";
         private const string SelectStatement = "SELECT {0} FROM {1}";
+        private string ValidateUpdateOnlyOneRow = "DECLARE @numberOfRows int = 0;" + Environment.NewLine +
```
kburtram: should this be `const`? If not, I think `validateUpdateOnlyOneRow` is the correct name casing.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
```diff
@@ -30,6 +32,14 @@ public sealed class RowUpdate : RowEditBase
...
+                                                          "Select @numberOfRows = count(*) FROM {0} {1} " + Environment.NewLine +
+                                                          "IF (@numberOfRows > 1) " + Environment.NewLine +
+                                                          "Begin" + Environment.NewLine +
+                                                           " DECLARE @error NVARCHAR(100) = N'The row value(s) updated do not make the row unique or they alter multiple rows(' + CAST(@numberOfRows as varchar(10)) + ' rows)';" + Environment.NewLine +
```
kburtram: It may be hard to localize this error message?

llali: I fix the localization later

PR #614, opened 2018-05-02
Fix crash handling messages when selection is null

MattIrv:
See Microsoft/sqlopsstudio#1331. When running query plan the batch that gets executed has `null` as the selection, causing the tools service to crash when handling messages.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -670,7 +670,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                     // Not a user cancellation error, add all 
                     foreach (var error in errors)
                     {
-                        int lineNumber = error.LineNumber + Selection.StartLine;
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -670,7 +670,7 @@ internal void StatementCompletedHandler(object sender, StatementCompletedEventAr
                     // Not a user cancellation error, add all 
                     foreach (var error in errors)
                     {
-                        int lineNumber = error.LineNumber + Selection.StartLine;
+                        int lineNumber = error.LineNumber + (Selection != null ? Selection.StartLine : 0);
```
AbbiePetcht: >0 [](start = 103, length = 1)

Did you verify in sqlops? Just want to make sure the start line starts from 0 not 1

kburtram:
@MattIrv did you determine how this problem was introduced?  i.e. is this a regression from a recent change?

MattIrv:
@kburtram I think it's coming from https://github.com/Microsoft/sqltoolsservice/pull/605 and is the same root issue as https://github.com/Microsoft/sqlopsstudio/issues/1334 but I haven't tracked it down entirely yet

PR #626, opened 2018-06-04
Update TravisCI to use .Net Core 2.1.300 GA release

kburtram:


.travis.yml
```diff
@@ -25,7 +25,7 @@ env:
 before_install:
   - if [ $TRAVIS_OS_NAME == "linux" ]; then
       sudo apt-get update;
-      sudo apt-get install dotnet-sdk-2.1.300-preview2-008533;
```
llali: please also update the url in the wiki

kburtram: I updated a couple links in the wiki.

PR #627, opened 2018-06-05
Dropping sessions when handling stop requests

madelinemacd:
XEvent sessions now get dropped on the server when the stop button is pressed.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Profiler/IProfilerSessionMonitor.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Profiler/IProfilerSessionMonitor.cs
@@ -27,6 +27,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
         /// <summary>
         /// Stops monitoring a profiler session
         /// </summary>
-        bool StopMonitoringSession(string sessionId);
+        bool StopMonitoringSession(string sessionId, out ProfilerSession session);
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/Profiler/XEventSession.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Profiler/XEventSession.cs
@@ -14,6 +14,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
     {
         public Session Session { get; set; }
 
+        public void Drop()
+        {
+            this.Session.Drop();
+        }
+
         public string GetTargetXml()
         {
             if (this.Session == null)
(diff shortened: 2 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/Profiler/XEventSession.cs
```diff
@@ -14,6 +14,11 @@ public class XEventSession : IXEventSession
     {
         public Session Session { get; set; }
 
+        public void Drop()
+        {
+            this.Session.Drop();
```
kburtram: Do we want to drop the session when the user clicks `Stop` or should we just stop the sessions?  I think SSMS only stops the session in this scenario.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -65,5 +65,62 @@ public class ProfilerServiceTests
             Assert.Equal(sessionListener.PreviousSessionId, sessionId);
             Assert.Equal(sessionListener.PreviousEvents.Count, 1);
         }
+
+        /// <summary>
+        /// Test stopping a session and receiving event callback
+        /// </summary>
+        /// <returns></returns>
+        [Fact]
+        public async Task TestStopProfilingRequest()
```
kburtram: It looks like the code coverage number dropped a bit with this PR.  It could be unrelated flakyness, but please verify it isn't related to a test failure from this change.

madelinemacd:
I made a new commit which stops sessions instead of dropping them, and which comments out flaky tests. 

I set my local vscode to trim trailing whitespace, which I thought would just affect my local view, but it seems like it actually made whitespace changes on files in my commit. I don't think it's harming anything, and I'm not sure how to undo that, but I can look into that if it's a problem.

src/Microsoft.SqlTools.ServiceLayer/Profiler/IProfilerSessionMonitor.cs
```diff
@@ -27,6 +27,6 @@ public interface IProfilerSessionMonitor
         /// <summary>
         /// Stops monitoring a profiler session
         /// </summary>
-        bool StopMonitoringSession(string sessionId);
+        bool StopMonitoringSession(string sessionId, out ProfilerSession session);
```
MattIrv: Does any code actually use the `bool` return value? If not then it might be better to just have it return the session.

PR #634, opened 2018-06-12
Adding pausing functionality for the profiler

madelinemacd:


src/Microsoft.SqlTools.ServiceLayer/Profiler/Contracts/PauseProfilingRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class PauseProfilingParams : GeneralRequestDetails
+    {
+        public string OwnerUri { get; set; }
+    }
+
+    public class PauseProfilingResult
+    {
+        public bool Succeeded { get; set; }
+
+        public string ErrorMessage { get; set; }
```
MattIrv: Do we need the `Succeeded` and `ErrorMessage` properties? I don't see anywhere where `ErrorMessage` is used, and it looks like the code only sends this with `Succeeded` set to true right now.

If neither of them are needed then this class can just be empty - it's not the cleanest code but we do it for other things like `QueryExecuteResult` already.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -170,48 +171,57 @@ public void InitializeService(ServiceHost serviceHost)
...
 
         /// <summary>
-        /// Gets an existing XEvent session or creates one if no matching session exists.
+        /// Gets or creates an XEvent session with the given template per the IXEventSessionFactory contract
         /// Also starts the session if it isn't currently running
         /// </summary>
-        private static Session GetOrCreateSession(SqlStoreConnection connection, string sessionName)
+        public IXEventSession GetOrCreateXEventSession(string template, ConnectionInfo connInfo)
         {
+            // TODO: Change this to handle different names based off of templates
```
MattIrv: It's helpful to link a TODO to a specific GitHub issue, if you have one tracking this work. The top-level profiler issue should be fine

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -24,10 +24,7 @@ public class ProfilerSession
         private TimeSpan pollingDelay = DefaultPollingDelay;
         private ProfilerEvent lastSeenEvent = null;
 
-        /// <summary>
-        /// Unique ID for the session
-        /// </summary>
-        public string SessionId { get; set; }
+        // TODO: Add ability to trigger instant polling
```
MattIrv: What does this mean? Is there an issue tracking it or is it something you meant to include in this PR?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -50,35 +72,111 @@ public void AddSessionListener(IProfilerSessionListener listener)
...
+                {
+                    viewers = new List<string>{ viewerID };
+                    sessionViewers.Add(session.ID, viewers);
                 }
             }
 
             return true;
         }
 
+        // TODO: Clean this up
```
MattIrv: leftover todo. looks like you already took care of it though?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -50,35 +72,111 @@ public void AddSessionListener(IProfilerSessionListener listener)
...
-        public bool StopMonitoringSession(string sessionId, out ProfilerSession session)
+        public void PauseViewer(string viewerID)
         {
+
             lock (this.sessionsLock)
             {
-                if (this.monitoredSessions.ContainsKey(sessionId))
+                //update the status in all viewers
+                Viewer v = this.allViewers[viewerID];
+                this.allViewers[viewerID] = new Viewer(viewerID, !v.active, v.xeSessionID);
```
MattIrv: Is this just updating `v`'s `active` property? Why not do it directly instead of creating a new `Viewer`?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -104,23 +107,78 @@ public class ProfilerServiceTests
...
+            Thread.Sleep(1500);
+
+            // check that session wasn't polled, and no events were sent to listener
+            Assert.Equal(sessionListener.PreviousSessionId, "");
+
+            // unpause viewer
+            await profilerService.HandlePauseProfilingRequest(requestParams, requestContext.Object);
+            Assert.True(success);
+
+            Thread.Sleep(1500);
```
MattIrv: sleeping this long in a test, especially multiple times, isn't great. I know a bunch of our tests already call sleep a bit but we should try to avoid it in new code. I don't think there's any risk of flakiness here like there is in a lot of our other tests where we call sleep because it's only waiting for polling to occur (right?) but is there any way to avoid sleeping here? For example maybe some way to poll manually when testing?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -50,35 +72,111 @@ public void AddSessionListener(IProfilerSessionListener listener)
...
             return true;
         }
 
+        // TODO: Clean this up
+        /// <summary>
+        /// Stop monitoring the session watched by viewerID
+        /// </summary>
+        public bool StopMonitoringSession(string viewerID, out ProfilerSession session)
+        {
+            lock (this.sessionsLock)
```
MattIrv: I don't see where this code actually stops the session on the server side. Does that happen somewhere else or is it missing?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -58,11 +58,14 @@ public class ProfilerServiceTests
             await profilerService.HandleStartProfilingRequest(requestParams, requestContext.Object);
 
             // wait a bit for profile sessions to be polled
-            Thread.Sleep(500);
+            Thread.Sleep(1000);
```
MattIrv: See my comment below for more about sleeping in tests... how do we know that 1000 is really enough if 500 sometimes isn't?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -28,7 +28,7 @@ public class ProfilerServiceTests
         /// </summary>
         /// <returns></returns>
         // TODO: Fix flaky test. See https://github.com/Microsoft/sqltoolsservice/issues/459
-        //[Fact]
+        [Fact]
```
MattIrv: Thanks for fixing this test! You can remove the TODO above too

MattIrv:
It looks like one of your tests is failing in AppVeyor so make sure to take a look at that too

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
madelinemacd: This code is purely for stopping the monitoring of the session, because session monitor only manages what's being polled. it's called by the profiler service's HandleStopProfilingRequest method, and that's where the actual session is stopped.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
madelinemacd: Sorry, these were meant to be more personal notes for myself on features that I'd need to add before 6/20 and where they should go, I'll go back and remove the extraneous TODO's. They're all tracked in one profiler issue that lists all my remaining tasks for the profiler.

src/Microsoft.SqlTools.ServiceLayer/Profiler/IProfilerSessionMonitor.cs
```diff
@@ -22,11 +22,16 @@ public interface IProfilerSessionMonitor
         /// <summary>
         /// Starts monitoring a profiler session
         /// </summary>
-        bool StartMonitoringSession(ProfilerSession session);
+        bool StartMonitoringSession(string viewerID, IXEventSession session);
```
kburtram: please name the parameter `viewerId` following .Net framework naming conventions (see here for additional details https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)

src/Microsoft.SqlTools.ServiceLayer/Profiler/IXEventSession.cs
```diff
@@ -11,13 +11,23 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
     public interface IXEventSession
     {
         /// <summary>
-        /// Reads XEvent XML from the default session target
+        /// Gets unique XEvent session ID
         /// </summary>
-        string GetTargetXml();
+        int ID { get; }
```
kburtram: same comment, please use `Id`

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -129,8 +130,8 @@ public void InitializeService(ServiceHost serviceHost)
 
                 if (connInfo != null)
                 {
-                    ProfilerSession session = StartSession(parameters.OwnerUri, connInfo);
-                    result.SessionId = session.SessionId;
+                    int xEventSessionID = StartSession(parameters.OwnerUri, parameters.TemplateName, connInfo);
```
kburtram: same comment regarding `ID` (I won't tag each occurrence, but please fix them all).

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -32,16 +32,38 @@ public class ProfilerSessionMonitor : IProfilerSessionMonitor
 
         private Task processorThread = null;
 
-        private Dictionary<string, ProfilerSession> monitoredSessions = new Dictionary<string, ProfilerSession>();
+        public struct Viewer
```
kburtram: if this class is internal implementation detail you may consider making it `private` or `internal`

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -50,35 +72,111 @@ public void AddSessionListener(IProfilerSessionListener listener)
...
+                    var profilerSession = new ProfilerSession();
+                    profilerSession.XEventSession = session;
+
+                    this.monitoredSessions.Add(session.ID, profilerSession);
                 }
 
-                if (!this.monitoredSessions.ContainsKey(session.SessionId))
+                // create a new viewer, or configure existing viewer
+                Viewer viewer;
+                if(!this.allViewers.TryGetValue(viewerID, out viewer))
```
kburtram: please add a ' ' between the 'f' and '(' characters in the `if` statement per .Net framework coding convention guidelines.  Please apply this change to all `if` statements in the the PR.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -115,12 +217,20 @@ private void ProcessSession(ProfilerSession session)
...
-                Task.Factory.StartNew(() => 
+                Task.Factory.StartNew(() =>
                 {
                     var events = PollSession(session);
                     if (events.Count > 0)
                     {
-                        SendEventsToListeners(session.SessionId, events);
+                        // notify all viewers for the polled session
+                        List<string> viewerIDs = this.sessionViewers[session.XEventSession.ID];
+                        foreach(string vID in viewerIDs)
```
kburtram: please choose a different name for the loop variable, such as `viewerId`.  Please review the MS doc link and associated code convention pages for .Net Framework recommended naming guidelines.   Here's another good starting point https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions.

Also please add a ' ' between the 'h' and '(' in the foreach.

src/Microsoft.SqlTools.ServiceLayer/Profiler/XEventSession.cs
```diff
@@ -10,10 +10,23 @@
 
 namespace Microsoft.SqlTools.ServiceLayer.Profiler
 {
+    /// <summary>
+    ///  This class performs an important function.
```
kburtram: the class comment could likely be improved 😄

PR #640, opened 2018-06-18
Profiler notifications

madelinemacd:
Notifying users when there's a possibility of dropped events, and when the session is unexpectedly closed on the server.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
@@ -127,8 +127,12 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
         /// Filter the event list to not include previously seen events,
         /// and to exclude events that happened before the profiling session began.
         /// </summary>
-        public void FilterOldEvents(List<ProfilerEvent> events)
+        /// <returns> 
+        /// Bool indicating if there is a possibility that events were lost between polling
+        /// </returns>
+        public bool FilterOldEvents(List<ProfilerEvent> events)
         {
+            bool eventsLost = false;
             if (lastSeenEvent != null)
             {
                 // find the last event we've previously seen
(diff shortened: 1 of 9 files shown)


test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -235,5 +235,52 @@ public class ProfilerServiceTests
 
             requestContext.VerifyAll();
         }
+
+         /// <summary>
```
MattIrv: spacing is off here

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -274,9 +275,15 @@ private List<ProfilerEvent> PollSession(ProfilerSession session)
                     }
                 }
             }
+            catch (XEventException)
+            {
+                SendStoppedSessionInfoToListeners(session.XEventSession.Id);
```
MattIrv: What sort of situations cause an `XEventException`? Would it be possible/better to treat it as a transient exception and try again instead of stopping the session? If we do want to stop the session, maybe we should do something to notify the user that it was stopped due to an error? Nothing that needs to be changed in this PR necessarily but something to think about for future improvements.

madelinemacd: An XEventException is thrown when something disrupts the session's ability to get it's target XML, the error message usually says something about the session being configured wrong and checking to make sure it exists. This is definitely a good area to think about improving for future releases.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -148,6 +162,12 @@ public void FilterOldEvents(List<ProfilerEvent> events)
                 {
                     events.RemoveRange(0, idx + 1);
                 }
+                else
+                {
+                    // last seen event was overwritten
+                    // other events may have been lost as well
+                    this.eventsLost = true;
```
kburtram: I think events may have a sequence number that would allow us to accurately detect if events are lost.  This implementation has a corner case that the previous "last event" is out of the ring buffer, but no event are lost yet.

madelinemacd: @kburtram I've updated the method of detecting lost events, feel free to take a look!

PR #642, opened 2018-06-21
Added support for T-SQL parse

abist:
Added support for a button in the `New Step` dialog, where the command box has a `Parse` button to verify T-SQL syntax without actually executing it. Accompanying PR for dataprotocol side - https://github.com/Microsoft/sqlops-dataprotocolclient/pull/15

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -290,29 +290,32 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         /// <returns></returns>
         internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestContext<SyntaxParseResult> requestContext)
         {            
-            try
-            {            
-                ParseResult result = Parser.Parse(param.Query);
-                SyntaxParseResult syntaxResult = new SyntaxParseResult();
-                if (result != null && result.Errors.Count() == 0)
-                {
-                    syntaxResult.Parseable = true;
-                } else 
-                {
-                    syntaxResult.Parseable = false;
-                    string[] errorMessages = {};
-                    for (int i = 0; i < result.Errors.Count(); i++)
-                    {
-                        errorMessages.Append(result.Errors.ElementAt(i).Message);
-                    }
-                    syntaxResult.Errors = errorMessages;
-                }
-                await requestContext.SendResult(syntaxResult);
-            }
-            catch (Exception ex)
+            await Task.Run(async () => 
             {
-                await requestContext.SendError(ex.ToString());
-            }
+                try
+                {   
+                    ParseResult result = Parser.Parse(param.Query);
+                    SyntaxParseResult syntaxResult = new SyntaxParseResult();
+                    if (result != null && result.Errors.Count() == 0)
+                    {
+                        syntaxResult.Parseable = true;
+                    } else 
+                    {
+                        syntaxResult.Parseable = false;
+                        string[] errorMessages = new string[result.Errors.Count()];
+                        for (int i = 0; i < result.Errors.Count(); i++)
+                        {
+                            errorMessages[i] = result.Errors.ElementAt(i).Message;
+                        }
+                        syntaxResult.Errors = errorMessages;
+                    }
+                    await requestContext.SendResult(syntaxResult);
+                }
+                catch (Exception ex)
+                {
+                    await requestContext.SendError(ex.ToString());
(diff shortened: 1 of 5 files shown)


anthonydresser:
Would it make more since to make this a different request type and call it validate or something? If I'm looking to validate a bit of tsql I don't think my first instinct would be to look at one of the query execution calls.

abist:
@anthonydresser I agree and I could do that, but since the functionality already implemented, adding a couple fields makes more sense. @kburtram should I make another request handler for this?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -190,6 +193,21 @@ public void InitializeService(ServiceHost serviceHost)
         {
             try
             {
+                if (executeParams.IsParse) {
```
kburtram: `{` characters should be on their own line in C# files

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -3,20 +3,23 @@
...
-using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
 using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests;
 using Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage;
 using Microsoft.SqlTools.ServiceLayer.SqlContext;
 using Microsoft.SqlTools.ServiceLayer.Workspace;
 using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
-using Microsoft.SqlTools.Utility;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
+using Microsoft.SqlServer.Management.SqlParser.Parser;
```
kburtram: Please sort the using statements..  System includes first, then others sorted alphabetically.

kburtram:
@abist it probably makes sense to have a separate parse method.  In SQL Server there is an option to  execute a query against the engine in "Parse-Only" mode, in which case this implementation makes sense.

Two problems though (1) we'd like the API to be somewhat generic and it's not clear this "Execute as Parse" is really generic in non-SQL Server scenarios (like CosmosDB, etc, though realistically there are several option options like this already like IncludeShowPlan).  and (2) you aren't actually executing the query against the engine to do the parse.  You're using the language service parser.  In this case, probably a parse or validate option in the language service makes more sense (though given the language service is closely coupled with VS Code lanaguageservice-protocol definition, I'm fine this being in the DMP query contracts).

Overall, I'm fine with this change, noting that we'll likely want to cleanup these Query APIs soon.  Or you can add a new method now if you want.

abist:
I can add it to the languageservice instead, it's a small change

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -281,6 +282,39 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        /// T-SQL syntax parse request callback
+        /// </summary>
+        /// <param name="param"></param>
+        /// <param name="requestContext"></param>
+        /// <returns></returns>
+        internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestContext<SyntaxParseResult> requestContext)
+        {            
+            try
+            {            
+                ParseResult result = Parser.Parse(param.Query);
```
kburtram: should the entire request handler be wrapped in a Task to avoid blocking the dispatcher queue?  Or are we sure the `Parse` will be quick.  Since it's disconnected parse it probably will be quick, but may be safe to run on different thread.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -281,6 +282,39 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                SyntaxParseResult syntaxResult = new SyntaxParseResult();
+                if (result != null && result.Errors.Count() == 0)
+                {
+                    syntaxResult.Parseable = true;
+                } else 
+                {
+                    syntaxResult.Parseable = false;
+                    string[] errorMessages = {};
```
kburtram: why not preallocate the array since you know the desired size `Errors.Count()`, which would avoid needing to dynamically grow the array.

PR #643, opened 2018-06-22
Support profiling on Azure instances

madelinemacd:
Fixes one of the two problems raised in https://github.com/Microsoft/sqlopsstudio/issues/1700

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
@@ -211,7 +211,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
             {
                 if (DatabaseUtils.IsSystemDatabaseConnection(connInfo.ConnectionDetails.DatabaseName))
                 {
-                    throw new NotSupportedException("Cannot profile Azure system databases");
+                    throw new NotSupportedException(SR.AzureSystemDbProfilingError);
                 }
                 store = new DatabaseXEStore(connection, connInfo.ConnectionDetails.DatabaseName);
             }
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -202,6 +204,24 @@ internal int StartSession(string ownerUri, string template, ConnectionInfo connI
             return xeSession.Id;
         }
 
+        private static BaseXEStore CreateXEventStore(ConnectionInfo connInfo, SqlStoreConnection connection)
+        {   
+            BaseXEStore store = null;
```
kburtram: not related to these changes, but I would change `BaseXEStore` to `XEStoreBase`,  putting `Base` at the end makes related types sort better alphabetically if there are several different types of "base" classes in a project.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -202,6 +204,24 @@ internal int StartSession(string ownerUri, string template, ConnectionInfo connI
...
+            if (connInfo.IsCloud)
+            {
+                if (DatabaseUtils.IsSystemDatabaseConnection(connInfo.ConnectionDetails.DatabaseName))
+                {
+                    throw new NotSupportedException("Cannot profile Azure system databases");
```
kburtram: should the error message be localized?  I assume this will get displayed to the user.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
madelinemacd: @kburtram I agree, but 'BaseXEStore' comes from 'Microsoft.SqlServer.Management.XEvent.BaseXEStore' so changing this isn't very feasible

PR #647, opened 2018-06-28
Delete unused Agent-related code

kburtram:
This removes several SQL Agent related classes that we're originally ported over, but aren't directly needed.  The relevant code has or will be moved into the other classes to align with the execution framework refactoring I did when porting.


kburtram:
Just deleting code..

PR #649, opened 2018-06-30
Add Name field to AlertInfo

kburtram:


kburtram:
minor bug fix that I want to include in next insiders build.

PR #654, opened 2018-07-11
Fix GUID parse error breaking some on-prem connections

kevcunnane:
- Fixes https://github.com/Microsoft/sqlopsstudio/issues/1896
- Verified this is only used in retry logic on Azure, so the fact it's failing on-prem isn't relevant. It's still unclear why in some cases the session ID is non-null but not a GUID.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableSqlConnection.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableSqlConnection.cs
@@ -434,12 +434,12 @@ SET NUMERIC_ROUNDABORT OFF;";
                         if (DBNull.Value != result)
                         {
                             string sessionId = (string)command.ExecuteScalar();
-                            _azureSessionId = new Guid(sessionId);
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableSqlConnection.cs
```diff
@@ -434,12 +434,12 @@ private void RetrieveSessionId()
                         if (DBNull.Value != result)
                         {
                             string sessionId = (string)command.ExecuteScalar();
-                            _azureSessionId = new Guid(sessionId);
+                            Guid.TryParse(sessionId, out _azureSessionId);
```
llali: should we add an error or warning when the parsing fails here?

kevcunnane: Good point - it previously would fall through to the catch statement & log (for SQLExceptions, not the FormatException causing this issue). I'll add logging

PR #663, opened 2018-07-20
Fix bug where strings that looked like dates were transformed when reading messages

MattIrv:
Fixes Microsoft/sqlopsstudio#1982. This bug was happening because Json.NET automatically converts strings that look like dates to a different format (see https://github.com/JamesNK/Newtonsoft.Json/issues/862 for lots of discussion).


kburtram:
@MattIrv I'd like to test this fix today so I'll merge to get a new tools service into SQL Ops Studio.

PR #664, opened 2018-07-23
Multiple batch execution time

abist:
Fixed https://github.com/Microsoft/sqlopsstudio/issues/1494.  Added a test to verify the same.


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -353,7 +353,7 @@ private bool IsExecutingMultipleTimes()
                 dbCommand.CommandText = BatchText;
                 dbCommand.CommandType = CommandType.Text;
                 dbCommand.CommandTimeout = 0;
-                executionStartTime = DateTime.Now;
+                executionStartTime = this.BatchExecutionCount > 1 ? executionStartTime : DateTime.Now;
```
anthonydresser: I fixed this by just moving the assignment of this property. Take a look at mine. Your's has tests so we should take yours, but I think my fix was cleaner.

abist: Sorry, I didn't realize you already had a PR out for this.

PR #666, opened 2018-07-25
Update Sql Management Objects to latest version

shueybubbles:
The NetStandard SMO binaries have individual packages on NuGet.org now. We can consume them directly instead of having a custom package. As we update SSMS and the public Microsoft.SqlServer.SqlManagementObjects package we will also update these xplat packages.


kburtram:
@shueybubbles if you have problems getting the CI build to pass cleanly, you can go ahead and merge, then queue a build in VSO to validate there.

PR #669, opened 2018-07-31
Added more options to SaveResultsAsCsvRequestParams

SebastianPfliegel:
Fixes Microsoft/sqlopsstudio#203

This PR adds more options for the SaveAsCsv request:
- LineSeperator
- TextIdentifier
- Encoding

I also created these PRs:
Microsoft/vscode-mssql#1128
Microsoft/sqlopsstudio#2099

This commit also fixes a bug introduced by #653, where custom delimiters didn't set the text identifier correctly (function EncodeCsvField).

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsCsvFileStreamWriter.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsCsvFileStreamWriter.cs
@@ -48,23 +48,57 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
         /// </param>
         public override void WriteRow(IList<DbCellValue> row, IList<DbColumnWrapper> columns)
         {
-            string delimiter = ",";
+            char delimiter = ',';
             if(!string.IsNullOrEmpty(saveParams.Delimiter))
             {
-                delimiter = saveParams.Delimiter;
-            }            
+                // first char in string
+                delimiter = saveParams.Delimiter[0];
+            }
+
+            string lineSeperator = Environment.NewLine;
+            if(!string.IsNullOrEmpty(saveParams.LineSeperator))
+            {
+                lineSeperator = saveParams.LineSeperator;
+            }
+
+            char textIdentifier = '"';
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsCsvFileStreamWriter.cs
```diff
@@ -48,23 +48,57 @@ public SaveAsCsvFileStreamWriter(Stream stream, SaveResultsAsCsvRequestParams re
         /// </param>
         public override void WriteRow(IList<DbCellValue> row, IList<DbColumnWrapper> columns)
         {
-            string delimiter = ",";
+            char delimiter = ',';
```
kburtram: would we ever want to support multi-character delimiters?  I don't think we need to, but just asking.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsCsvFileStreamWriter.cs
```diff
@@ -48,23 +48,57 @@ public SaveAsCsvFileStreamWriter(Stream stream, SaveResultsAsCsvRequestParams re
...
+            Encoding encoding;
+            try
+            {
+                if(int.TryParse(saveParams.Encoding, out codepage))
+                {
+                    encoding = Encoding.GetEncoding(codepage);
+                }
+                else
+                {
+                    encoding = Encoding.GetEncoding(saveParams.Encoding);
```
kburtram: will this throw in the most common case where `Encoding` isn't provided?  If so, you may consider rewriting to first null check `Encoding` and set `encoding = Encoding.GetEncoding("utf-8");` in the normal flow control (either in the `else` or as a default value when declaring `encoding`.

Basically, I like to avoid 1st-chance handled exceptions when possible since I often run the debugger to break on all exceptions which works best when exceptions are only used for real error conditions.

abist:
@SebastianPfliegel Lets get master merged to this branch so we can get the conflicts resolved before merging. Looks good otherwise.

PR #672, opened 2018-08-02
Add back missing XEventEnum dependency

kburtram:


kburtram:
This is needed to fix a profiler break.

shueybubbles:
is there a test gap to fill?

kburtram:
It looks like the Profiler integration tests are disabled https://github.com/Microsoft/sqltoolsservice/blob/master/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Profiler/ProfilerServiceTests.cs.

@madelinemacd should we enable these tests?  I think they may have caught the missing dependency issue.

PR #673, opened 2018-08-02
Remove unneeded checked in Nuget package

kburtram:


kburtram:
deleting an unused binary file...

PR #678, opened 2018-08-08
Upgrade SMO packages to fix issue with scripting tables 

shueybubbles:
when Script Dependencies is true on the script options. This change also creates a common.props file to centralize values like package version numbers, with SmoPackageVersion as the initial shared value.


kevcunnane:
Actually - this breaks CI builds. Please fix :)

PR #679, opened 2018-08-08
Fix failing Loc lookups in hosting service

kevcunnane:
It turns out the RootNamespace needed to be set or else .Net Core would decide to omit ".v2", which broke unit testing.
- Enabled CI testing to catch these errors in the future
- Added RootNamespace
- Fixed a host of other issues that added warnings to the tests so we get a cleaner run output.


src/Microsoft.SqlTools.Hosting.v2/Properties/AssemblyInfo.cs
```diff
@@ -42,5 +42,4 @@
 [assembly: AssemblyInformationalVersion("1.0.0.0")]
 
 [assembly: InternalsVisibleTo("Microsoft.SqlTools.Hosting.UnitTests")]
-[assembly: InternalsVisibleTo("Microsoft.SqlTools.CoreServices")]
```
shueybubbles: Is there a team guidance for appropriate use of InternalsVisibleTo? It got way out of hand in previous projects.

PR #680, opened 2018-08-08
Serialize Connection String

anthonydresser:
Adds the handler for serialization of connection string feature.
https://github.com/Microsoft/sqlopsstudio/pull/2192
https://github.com/Microsoft/sqlops-dataprotocolclient/pull/22

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -952,7 +952,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             serviceHost.SetRequestHandler(ListDatabasesRequest.Type, HandleListDatabasesRequest);
             serviceHost.SetRequestHandler(ChangeDatabaseRequest.Type, HandleChangeDatabaseRequest);
             serviceHost.SetRequestHandler(GetConnectionStringRequest.Type, HandleGetConnectionStringRequest);
-            serviceHost.SetRequestHandler(SerializeConnectionStringRequest.Type, HandleSerializeConnectionStringRequest);
+            serviceHost.SetRequestHandler(BuildConnectionInfoRequest.Type, HandleBuildConnectionInfoRequest);
         }
 
         /// <summary> 
@@ -1292,7 +1292,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         /// <summary>
         /// Handles a request to serialize a connection string
         /// </summary>
-        public async Task HandleSerializeConnectionStringRequest(
+        public async Task HandleBuildConnectionInfoRequest(
             string connectionString,
             RequestContext<ConnectionDetails> requestContext)
         {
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SerializeConnectionStringRequest.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
+{
+    /// <summary>
+    /// Serialize Connection String request 
+    /// </summary>
+    public class SerializeConnectionStringRequest
+    {
+        public static readonly
+            RequestType<string, ConnectionDetails> Type =
+            RequestType<string, ConnectionDetails>.Create("connection/serializeconnectionstring");
```
kburtram: I'm going back-and-forth whether `serialize` is really the correct verb here (verses something like parse).  I think it's fine, since we are "serializing" into an object, though I typically think of creating objects from a raw value as "deserialize".  In any case, just food for thought.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1288,6 +1289,63 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
+                    // rather than an error.
+                    await requestContext.SendResult(null);
+                }
+            });
+        }
+
+        public ConnectionDetails ParseConnectionString(string connectionString)
+        {
+            SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connectionString);
+            ConnectionDetails details = new ConnectionDetails(){
```
kburtram: the `{` should be on it's own line in C#

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1288,6 +1289,63 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
+                DatabaseName = builder.InitialCatalog,
+                Encrypt = builder.Encrypt,
+                FailoverPartner = builder.FailoverPartner,
+                LoadBalanceTimeout = builder.LoadBalanceTimeout,
+                MaxPoolSize = builder.MaxPoolSize,
+                MinPoolSize = builder.MinPoolSize,
+                MultipleActiveResultSets = builder.MultipleActiveResultSets,
+                MultiSubnetFailover = builder.MultiSubnetFailover,
+                PacketSize = builder.PacketSize,
+                Password = !builder.IntegratedSecurity ? builder.Password : "",
```
kburtram: nit: `string.Empty` is the preferred way to write `""` in C#

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SerializeConnectionStringRequest.cs
anthonydresser: Maybe `buildConnectionInfo` might be more telling?

kburtram: @anthonydresser that sounds good as well, though the convention has been to use all lower-case for JSON-RPC method names (e.g., `"connection/buildconnectioninfo"`), though I realize that isn't followed consistently across all feature areas.

PR #681, opened 2018-08-14
Fix for https://github.com/Microsoft/sqlopsstudio/issues/1317

alanrenmsft:
When SQL tools service passes the query result back, a ToString operation is used to get the display value for each cell, this works fine if we are only showing the results in the query result window, but can be problematic if we need to use the display value to convert to specific types, for example decimal type. the fix is to introduce a new property to the contract type DbCellType: InvariantCultureDisplayValue, then from the client side we can safely convert the data to specific types.

This fix to the issue needs to be implemented in both SqlToolsService and SqlOpsStudio. there will be a separate PR for sqlopsstudio.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
@@ -151,11 +151,11 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
 
                 // Use the right read function for the type to read the data from the file
                 ReadMethod readFunc;
-                if(!readMethods.TryGetValue(colType, out readFunc))
+                if (!readMethods.TryGetValue(colType, out readFunc))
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -216,6 +218,17 @@ private void AssureBufferLength(int newBufferLength)
                 T resultObject = convertFunc(length.ValueLength);
                 result.RawObject = resultObject;
                 result.DisplayValue = toStringFunc == null ? result.RawObject.ToString() : toStringFunc(resultObject);
+                if(setInvariantCultureDisplayValue)
```
kburtram: nit: in this code base the convention is to put a `' '` character between `'f'` and `'('` in `if` statements.

alanrenmsft: done

PR #684, opened 2018-08-24
Save as XML feature added

sadedil:
Similar approach used like Save as JSON or Save as CSV.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsXmlFileStreamWriter.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsXmlFileStreamWriter.cs
@@ -1,0 +1,142 @@
+﻿// 
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Xml;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
+{
+    /// <summary>
+    /// Writer for writing rows of results to a XML file.
+    /// </summary>
+    /// <remarks>
+    /// This implements its own IDisposable because the cleanup logic closes the element that was
+    /// created when the writer was created. Since this behavior is different than the standard
+    /// file stream cleanup, the extra Dispose method was added.
+    /// </remarks>
+    public class SaveAsXmlFileStreamWriter : SaveAsStreamWriter, IDisposable
+    {
+        // Root element name for the output XML
+        private const string RootElementTag = "data";
+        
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsXmlFileStreamWriter.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// request params.
+        /// </param>
+        public override void WriteRow(IList<DbCellValue> row, IList<DbColumnWrapper> columns)
+        {
+            // Write the header for the object
+            xmlTextWriter.WriteStartElement(ItemElementTag);
+
+            // Write the items out as properties
+            int columnStart = ColumnStartIndex ?? 0;
+            int columnEnd = ColumnEndIndex + 1 ?? columns.Count;
```
kburtram: Can `ColumnEndIndex + 1` ever evaluate to `null`?  I would expect that to throw is ColumnEndIndex is null.

sadedil: Hi, @kburtram. Thanks for the review.

I think `ColumnStartIndex` and `ColumnEndIndex` are optional values, so it's OK to be null. Am I missing something?

And I thought the first expression was simpler way to write second one:
1. `int columnEnd = ColumnEndIndex + 1 ?? columns.Count;`
2. `int columnEnd = (ColumnEndIndex != null) ? ColumnEndIndex.Value + 1 : columns.Count;` (*this expression can be found at `SaveAsJsonFileStreamWriter.cs`*)

benrr101: You are right that ColumnStartIndex and ColumnEndIndex are optional, so supporting null is important here. But I did a quick bit of research and learned something new!

> The predefined unary and binary operators and any user-defined operators that exist for value types may also be used by nullable types. These operators produce a null value if the operands are null; otherwise, the operator uses the contained value to calculate the result.

>...no matter what the operand value is, if another operand is null, then the result is null.

(https://stackoverflow.com/questions/7660449/math-operations-with-null)

So yes, that's a nice way to simplify the code -- though it might not be as easy to understand at first 😄

kburtram: @benrr101 thanks, I learned something new too 😄.  @sadedil the code looks good to me.  For clarity, the below snippet illustrates the confusion.  I this case I wasn't sure if `val1 + 1` would raise a nullref exception.

```
int? val1 = null;
int val2 = val1 + 1 ?? 100;
```

sadedil: @benrr101 thanks for the clarification.

I think you're right, it's not so easy to see how is working at first sight.

inalpolat:
Great feature!

PR #686, opened 2018-09-04
Revert "Switch to the unified SMO NuGet"

kburtram:
The SMO update is causing breaks in Credentials and Resource Providers so we need to either fix this or revert back to previous SMO consumption model until the issue is resolved.


kburtram:
@shueybubbles as we discussed offline, let's revert for now and work on a fix in parallel.  Thanks!

PR #693, opened 2018-09-18
 fixed couple of issues in the perf tests

llali:


kburtram:
@llali thanks for fixing 😄

PR #698, opened 2018-09-26
Logging for ga

ranasaria:
Adding a robustness change to call close on logger before process exit.

Also adding a designer.cs file needed by a script csproj that got missed out in previous checkin due to .gitignore entry.


kburtram:
@ranasaria I guess there isn't a good way to force the close during the finalizer if the log is still open?

ranasaria:
Yeah. I will research this more on the internet later and see if there is a more comprehensive fix corner case scenarios.


-  Arvind.

PR #701, opened 2018-10-02
Upgrade SMO to 150 preview 

shueybubbles:
This enables full compatibility with SQL 2019 CTP. 
While testing the upgrade I found an issue with scripting objects that have a single quote in their name so I'm including that fix here.


src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
```diff
@@ -20,10 +20,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="System.Data.SqlClient" Version="4.5.0-preview2-26406-04" />
-    <PackageReference Include="Microsoft.SqlServer.Management.SmoMetadataProvider" Version="$(SmoPackageVersion)" />
-    <PackageReference Include="Microsoft.SqlServer.Management.SqlScriptPublishModel" Version="$(SmoPackageVersion)" />
-    <PackageReference Include="Microsoft.SqlServer.SqlParser" Version="$(SmoPackageVersion)" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
```
ranasaria: Academic question: This is a little unexpected removal I would think. What I am wondering at is the following. Does the code in this project depend directly on Newtonsoft.json or did it have this dependency due to one of the SmoMetadataProvider, SqlScriptPublishModel, and SqlParser codebases.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -103,10 +103,10 @@ public static Urn ToUrn(this ScriptingObject scriptingObject, string server, str
             string urn = string.Format(
                 "Server[@Name='{0}']/Database[@Name='{1}']/{2}[@Name='{3}' {4}]",
                 server.ToUpper(),
-                database,
+                Urn.EscapeString(database),
```
ranasaria: How did you end up uncovering this bug? Do we have a test coverage hole? If you can spare on some thoughts what missing test we ought to go and write then kindly file an issue for our backlog.

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestLogger.cs
```diff
@@ -97,13 +97,15 @@ public void WriteWithCallstack()
 
         public void Verify(TraceEventType eventType, string message, string callstackMessage, bool shouldVerifyCallstack = false, bool expectLogMessage = true)
         {
+            Logger.Flush();
+            // The Regex uses .* between the severity and the message to allow SMO to vary the content. 140 SMO has nothing there, 150 has a timestamp
             if (expectLogMessage)
             {
-                Assert.True(File.Exists(Logger.LogFileFullPath) && Regex.IsMatch(LogContents, $@"\b{eventType}:\s+\d+\s+:\s+{message}", RegexOptions.Compiled));
+                Assert.True(File.Exists(Logger.LogFileFullPath) && Regex.IsMatch(LogContents, $@"\b{eventType}:.*{message}", RegexOptions.Compiled));
```
ranasaria: The previous version of code was verifying that a trace eventid (like an error code) which is superimposed and emitted by the toolsservice logger is present in the log entry. So before we relax this verification I would like to understand why this relaxation is needed. Maybe there is a bug in my original implementation that you have uncovered. More likely this error code is just not present for the smo traces and we need to make this chagne only for those tests.

ranasaria: eventid comes right after eventType. So for some reason the above check was getting satisfied before while it is not in smo. This does not look related to emission of date in smo log entry or is it?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
ranasaria: I see that you have added some new tests below, so if they were sufficient to reproduce the original bug and verify the fix then we are good here and thank you for the tests :)

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestLogger.cs
shueybubbles: I think it's not practical for these tests to be so tightly coupled to the trace format of external components like SMO. What if we want to put something else in the trace between the message and the event type in the future? Having to also come back and update this test for something that should have no functional impact seems like excessive overhead. I changed this regex simply to allow SMO to have whatever it wants between the type and the message. The event id is still there afaik.

ranasaria: The idea is to have one kind of verification logic for tools service logging and the other for smo logging. That will give you what you need. My concern is that if something breaks emission of the eventid in the toolsservice code then I want that to be caught by the tests.

src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
shueybubbles: looks like linux FileTokenStorage needs it. I think I just selected one too many lines before I hit delete.

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestLogger.cs
ranasaria: so you do not have to do much perhaps
revert the chagne to TestLogger
and change LoggerGetValidTableDefinitionTest 
to not call test.Verify() and instead do your own verify method
later if we have more tests for smo logging then we can create a dedicated test object for it

ranasaria: David and I had an offline discussion and I see his point on over verifying the format, especially since in this case the eventid's presence in toolservice logging is dictated by the api when the log entry is generated. What gets written to the actual log file is purely determined by the listener.

PR #702, opened 2018-10-03
Fix OE and Binding Queue reliability bugs

MattIrv:
This fixes some [Object Explorer reliability problems](https://github.com/Microsoft/azuredatastudio/issues/1503), specifically:
- Fixes a bug where the binding queue kept waiting for a task that had timed out, meaning that no other tasks could be executed in that binding context
- Fixes a bug in OE expand logic where the expansion task kept running even after it was canceled, blocking any other OE expansions since the connection would already have an active query
- The above 2 fixes mean that an OE timeout no longer blocks other operations for that connection
- Fixes a bug where the binding queue would wait on its main thread whenever a new task tried to execute in the same binding context as an existing task, blocking all other binding queue operations even for other connections. This is fixed by using one task to execute binding queue operations per context.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
@@ -39,7 +39,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         /// </summary>
         internal Dictionary<string, IBindingContext> BindingContextMap { get; set; }
 
-        private Dictionary<IBindingContext, Task> bindingContextTasks = new Dictionary<IBindingContext, Task>();
+        internal Dictionary<IBindingContext, Task> BindingContextTasks { get; set; } = new Dictionary<IBindingContext, Task>();
 
         /// <summary>
         /// Constructor for a binding queue instance
@@ -149,7 +149,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                 {
                     var bindingContext = new T();
                     this.BindingContextMap.Add(key, bindingContext);
-                    this.bindingContextTasks.Add(bindingContext, Task.Run(() => null));
+                    this.BindingContextTasks.Add(bindingContext, Task.Run(() => null));
                 }
 
                 return this.BindingContextMap[key];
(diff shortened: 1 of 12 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -286,86 +296,119 @@ private void ProcessQueue()
...
+                            {                                                    
+                                // prefer the queue item binding item, otherwise use the context default timeout
+                                int bindTimeout = queueItem.BindingTimeout ?? bindingContext.BindingTimeout;
+
+                                // handle the case a previous binding operation is still running
+                                if (!bindingContext.BindingLock.WaitOne(queueItem.WaitForLockTimeout ?? 0))
+                                {
+                                    try
+                                    {
+                                        Logger.Write(TraceEventType.Warning, "Binding queue operation timed out waiting for previous operation to finish");
```
ranasaria: Consider defining a new Logger.LogEvent entry and using that as the argument to Logger.Write. This will help in future to quickly search for all log entries with a given LogEvent.

The LogEvents could be specific areas of code that we designate them to be.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -286,86 +296,119 @@ private void ProcessQueue()
...
+                        var bindingContextTask = this.BindingContextTasks[bindingContext];
 
-                            // handle the case a previous binding operation is still running                                                 
-                            if (!bindingContext.BindingLock.WaitOne(queueItem.WaitForLockTimeout ?? 0))
-                            {
-                                queueItem.Result = queueItem.TimeoutOperation != null
-                                    ? queueItem.TimeoutOperation(bindingContext)
-                                    : null;
+                        // Run in the binding context task in case this task has to wait for a previous binding operation
+                        this.BindingContextTasks[bindingContext] = bindingContextTask.ContinueWith((task) =>
```
ranasaria: Do we want this new task to run irrespective of what happened to original task? I suggest deliberately specify TaskContinuationOptions parameter even if default value as appropriate for us might be a good idea.

PR #704, opened 2018-10-04
Reopen connections prior to creating query execution data readers

kburtram:
This is a very limited fix that may help with https://github.com/Microsoft/azuredatastudio/issues/2759.  It adds a couple checks to reopen the query execution DbConnection prior to opening the DataReader object.  

This doesn't address the core problem that connections are being closed nearly immediately after being opened, or if the connection gets dropped after the data reader is created, but it would help if the connect is closed in the short period of time after Open succeeds and creating the DataReaders.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -17,6 +17,7 @@ using Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage;
 using Microsoft.SqlTools.Utility;
 using System.Globalization;
 using System.Collections.ObjectModel;
+using Microsoft.SqlTools.ServiceLayer.Connection;
 
 namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
 {
@@ -364,7 +365,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                     // so add a newline to the end of the query. See https://github.com/Microsoft/sqlopsstudio/issues/1424
                     dbCommand.CommandText += Environment.NewLine;
 
-                    EnsureConnectionIsOpen(conn);
+                    ConnectionService.EnsureConnectionIsOpen(conn);
 
                     // Fetch schema info separately, since CommandBehavior.KeyInfo will include primary
                     // key columns in the result set, even if they weren't part of the select statement.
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -364,6 +364,8 @@ private async Task ExecuteOnce(DbConnection conn, CancellationToken cancellation
                     // so add a newline to the end of the query. See https://github.com/Microsoft/sqlopsstudio/issues/1424
                     dbCommand.CommandText += Environment.NewLine;
 
+                    EnsureConnectionIsOpen(conn);
```
alanrenmsft: Your fix will address the particular symptom, I am not sure whether you are aware, but in ADO.NET, the connection state might appear open even though the underlining connection is broken until you try to use the connection to execute some commands, the state will be changed accordingly only when the command cannot be executed.

I would suggest to introduce a retry helper with retry strategy and use it whenever we need reliable connection.

kburtram: ReliableConnection is only for DbConnection and DbCommand operations as far as I could see.  It doesn't do anything for DataReaders.  

I'll look to see if there's a way to incorporate DataReaders into that component as well, though in general I'd agree that elaborate retry for all data operations seems to be looking at symptoms instead of root problems.  For example, when we create the DbCommand immediately prior to the DataReader the retry logic is already applied, then a few milliseconds later the connection is dropped.  

I haven't seen anywhere in any of our tools code an attempt to add retries for all interactions with SqlClient.  And this creates concerns as to how are 3rd-parties supposed to build applications using SqlClient if the expectation is that every API needs to be retried several times on failure.  I'm not yet convinced we're looking in the correct place here.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -719,6 +723,29 @@ private SpecialAction ProcessResultSetSpecialActions()
...
+            if (conn.State != ConnectionState.Open)
+            {
+                try
+                {
+                    // close it in case it's in some non-Closed state
+                    conn.Close();
+                }
+                catch
+                {
+                    // ignore any exceptions thrown from .Close
```
ranasaria: Should we log a warning/informational message here?

kburtram:
@ranasaria I'll add logging separately since I want to do that as part of providing reconnect notifications and other connection error messages.

PR #706, opened 2018-10-09
Load properties used for OE node status by group instead of individually

MattIrv:
Loading logins, users, and triggers in OE right now is slow when there are thousands of objects because each object's status is determined by making an individual SMO call.

It is much quicker to have SMO load those properties when running the query to list the objects instead of querying them individually.

Now 1000 logins on a local server load in about 1 second instead of before where it would usually time out after 45 seconds

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoColumnCustomNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoColumnCustomNode.cs
@@ -4,9 +4,11 @@
 //
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlTools.ServiceLayer.ObjectExplorer.Nodes;
 using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoColumnCustomNode.cs
```diff
@@ -20,6 +22,81 @@ public override string GetNodeCustomName(object smoObject, SmoQueryContext smoCo
         {
             return SmoColumnCustomNodeHelper.CalculateCustomLabel(smoObject, smoContext);
         }
+
+        public override IEnumerable<NodeSmoProperty> SmoProperties
+        {
+            get
+            {
+                return new List<NodeSmoProperty>
```
ranasaria: This will create a new List every time SmoProperties is called. Can we make it either a lazily initialized object or construction is when ColumnsChildFactory is constructed and keep returning that readonly private variable in this property.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
```diff
@@ -3,9 +3,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System.Collections.Generic;
```
ranasaria: You can revert the redundant change to this file.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoKeyCustomNode.cs
```diff
@@ -23,6 +25,31 @@ public override string GetNodeSubType(object smoObject, SmoQueryContext smoConte
     /// </summary>
     internal partial class IndexesChildFactory : SmoChildFactoryBase
     {
+        public override IEnumerable<NodeSmoProperty> SmoProperties
+        {
+            get
+            {
+                return new List<NodeSmoProperty>
```
ranasaria: Same comment as that for ColumnsChildFactory

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTriggerCustomNode.cs
```diff
@@ -12,10 +14,27 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
     /// </summary>
     internal partial class TriggersChildFactory : SmoChildFactoryBase
     {
+        public static readonly List<NodeSmoProperty> SmoPropertyList = new List<NodeSmoProperty>
+        {
+            new NodeSmoProperty
```
ranasaria: repeat of above.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoLoginCustomNode.cs
```diff
@@ -16,6 +18,21 @@ public override string GetNodeStatus(object smoObject, SmoQueryContext smoContex
         {
             return LoginCustomNodeHelper.GetStatus(smoObject);
         }
+
+        public override IEnumerable<NodeSmoProperty> SmoProperties
+        {
+            get
+            {
+                return new List<NodeSmoProperty>
```
ranasaria: Same issue.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTriggerCustomNode.cs
```diff
@@ -12,10 +14,27 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
     /// </summary>
     internal partial class TriggersChildFactory : SmoChildFactoryBase
     {
+        public static readonly List<NodeSmoProperty> SmoPropertyList = new List<NodeSmoProperty>
```
ranasaria: Minor concern: Slightly better to make this lazily initialized for two reasons. Both are not big concerns for small objects like this one.
1. Static object constructs increase process startup time/load time.
2. If no one ever expands the triggers node then this is not constructed

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoUserCustomNode.cs
```diff
@@ -16,6 +18,21 @@ public override string GetNodeStatus(object smoObject, SmoQueryContext smoContex
         {
             return UserCustomeNodeHelper.GetStatus(smoObject);
         }
+
+        public override IEnumerable<NodeSmoProperty> SmoProperties
+        {
+            get
+            {
+                return new List<NodeSmoProperty>
```
ranasaria: same concern as before.

MattIrv:
I'm going to avoid doing any extra optimization since these are small lists

PR #707, opened 2018-10-10
Make sample work with current code

eashi:
The JSONRPC sample under docs seems to have fallen behind the changes in the main code.

This PR will update the sample code so the program displays the records of the query successfully in the client's console output. There are still things need to be done, e.g. for some reason the ColumnInfo is not properly populated, but will need another PR for that.


MattIrv:
Looks good to me, thanks!

Issue #708, opened 2018-10-10
handling of hierachyid datatype


alanrenmsft:
we are handling it as byte[] data type and as a result, when using Edit Data for a table with this type of columns in Azure Data Studio, we will get an error indicating cannot convert string to byte[].

this issue is found while editing the OrganizationNode column of HumanResources.Employee table in adventureWorks database.

BTW, SSMS handles this data type properly.

benrr101:
For the record, this was a feature that I didn't really understand at the time so I left it to be worked on later. Though I thought the logic for converting 0x* strings to byte[] was working correctly

PR #709, opened 2018-10-12
Handling the HierarchyId for edit data scenario

alanrenmsft:
#708 

**Before:**
image

**After (Same as SSMS):** 
image


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/DbColumnWrapper.cs
```diff
@@ -199,8 +199,8 @@ public DbColumnWrapper(DbColumn column)
                 object assemblyQualifiedName = column.UdtAssemblyQualifiedName;
                 const string hierarchyId = "MICROSOFT.SQLSERVER.TYPES.SQLHIERARCHYID";
 
-                if (assemblyQualifiedName != null &&
-                    string.Equals(assemblyQualifiedName.ToString(), hierarchyId, StringComparison.OrdinalIgnoreCase))
+                if (assemblyQualifiedName != null
+                && assemblyQualifiedName.ToString().StartsWith(hierarchyId, StringComparison.OrdinalIgnoreCase))
```
alanrenmsft: I think this code is from SSMS, I found the original version, and it is using string.compare only to the length of hierarchyId string. bascially the same to StartWith. BTW, it seems StartWith is only available after .net 3.5.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditColumnMetadata.cs
```diff
@@ -36,6 +36,22 @@ public EditColumnMetadata()
...
         public string EscapedName { get; set; }
 
+        /// <summary>
+        /// The column's expression for select statement
+        /// </summary>
+        public string ExpressionForSelectStatement
+        {
+            get
+            {
+                return IsHierarchyId ? string.Format("{0}.ToString() {0}", EscapedName) : EscapedName;
```
benrr101: Using string interpolation here would be nice
I suspect there will be other column types that will use similar logic (Geometry type perhaps?). If so, let's double check that this pattern is scalable

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -496,10 +498,22 @@ public EditUpdateCellResult UpdateCell(long rowId, int columnId, string newValue
             }
         }
 
+        internal void UpdateColumnInformationWithMetadata(DbColumnWrapper[] columns)
```
benrr101: Make sure internal stuff goes in the private helpers section

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -496,10 +498,22 @@ public EditUpdateCellResult UpdateCell(long rowId, int columnId, string newValue
...
+        {
+            if (columns != null && this.objectMetadata != null)
```
benrr101: Inverting this if statement to return would help with cleanliness

src/Microsoft.SqlTools.ServiceLayer/EditData/EditColumnMetadata.cs
benrr101: I know this is implicit, but it's really nice to have the explicit `AS` added between the two

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -96,14 +96,21 @@ public override DbCommand GetCommand(DbConnection connection)
...
+            for (int i = 0; i < AssociatedObjectMetadata.Columns.Length; i++)
             {
                 DbColumnWrapper column = AssociatedResultSet.Columns[i];
                 EditColumnMetadata metadata = AssociatedObjectMetadata.Columns[i];
                 CellUpdate cell = newCells[i];
-                
+
                 // Add the output columns regardless of whether the column is read only
-                outClauseColumnNames.Add($"inserted.{metadata.EscapedName}");
+                if (metadata.IsHierarchyId)
```
benrr101: Again, I'm not sure this pattern is scalable. You may want to consider having this as a new utilities class that can convert the columns to identifiers.
Alternatively, this is pretty much just adding `inserted.` to `metadata.ExpressionForSelectStatement`. Removing the check for `IsHierarchyId` in favor of using that property would simplify this code a bit

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowCreate.cs
```diff
@@ -112,16 +119,25 @@ public override DbCommand GetCommand(DbConnection connection)
...
                 // Add the input column
                 inColumnNames.Add(metadata.EscapedName);
-                
+
                 // Add the input values as parameters
                 string paramName = $"@Value{RowId}_{i}";
-                inValues.Add(paramName);
-                inParameters.Add(new SqlParameter(paramName, column.SqlDbType) {Value = cell.Value});
+
+                if (metadata.IsHierarchyId)
```
benrr101: This part should definitely go into the utilities. I can't exactly remember how it all works right now. If I recall correctly, when we put together a query to add/update there's a utility (ToSqlScript) that turns the value into an expression SQL Server can work with. Take a look at the FormatDecimalLike method in that utility class to see an example of how we take decimals and use a `CAST` on them. There's even a line that calls out how hierarchyid doesn't work right now, so that might make it easier!

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowUpdate.cs
```diff
@@ -97,32 +97,46 @@ public override DbCommand GetCommand(DbConnection connection)
             for (int i = 0; i < AssociatedObjectMetadata.Columns.Length; i++)
             {
                 EditColumnMetadata metadata = AssociatedObjectMetadata.Columns[i];
-                
+
                 // Add the output columns regardless of whether the column is read only
                 declareColumns.Add($"{metadata.EscapedName} {ToSqlScript.FormatColumnType(metadata.DbColumn, useSemanticEquivalent: true)}");
-                outClauseColumns.Add($"inserted.{metadata.EscapedName}");
+                if (metadata.IsHierarchyId)
```
benrr101: Same comments from CreateRow apply here

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/DbColumnWrapper.cs
benrr101: I could have missed that. When you see so many `string.Compare == 0` you start to just ignore the contents and replace with `string.Equals` blindly

kburtram:
@alanrenmsft are you still tracking this PR?  Or should we close it out to clean-up the pending PR list?

alanrenmsft:
@kburtram ah oh! I haven't looked at this repo for a long time and almost forgot about this. one, I actually would still like to finish this since it is almost there. Hopefully next week.

alanrenmsft:
being a really long time, i don't really have the time to look into other comments, i tested the feature and will merge as is.

PR #711, opened 2018-10-17
Bug fix for https://github.com/Microsoft/azuredatastudio/issues/2923 

ranasaria:
Bug fix for https://github.com/Microsoft/azuredatastudio/issues/2923 and misc other fixes
1. Fixes the spurious additional numbers in credentialstore log file name. Microsoft/azuredatastudio#2923
2. Removed old deprecated command line parameters no longer needed:

- --log-dir
- --enable-logging

3. Enables support for --autoflush-log to pave way for a future fix for Microsoft/azuredatastudio#2926
4. By default emit the log file name when the Logger is initialized.
5. Fixed affected tests and added coverage for new autoflush-log functionality. 
6. Did some refactoring SmoModel\Smo*CustomNode.cs files to ensure that we do not create new objects within a property getter but keep sending back singleton objects that are created once.

### DIFF
--- a/src/Microsoft.SqlTools.Credentials/Program.cs
+++ b/src/Microsoft.SqlTools.Credentials/Program.cs
@@ -36,15 +36,12 @@ namespace Microsoft.SqlTools.Credentials
                 string logFilePath = commandOptions.LogFilePath;
                 if (string.IsNullOrWhiteSpace(logFilePath))
                 {
-                    logFilePath = "credentials";
-                }
-                if (!string.IsNullOrWhiteSpace(commandOptions.LoggingDirectory))
-                {
-                    logFilePath = Logger.GenerateLogFilePath(Path.Combine(commandOptions.LoggingDirectory, logFilePath));
+                    logFilePath = Logger.GenerateLogFilePath("credentials");
                 }
 
+                Logger.AutoFlush = commandOptions.AutoFlushLog;
+
                 Logger.Initialize(tracingLevel: commandOptions.TracingLevel, logFilePath: logFilePath, traceSource: "credentials");
-                Logger.Write(TraceEventType.Information, "Starting SqlTools Credentials Provider");
 
                 // set up the host details and profile paths 
                 var hostDetails = new HostDetails(
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoColumnCustomNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoColumnCustomNode.cs
@@ -23,80 +23,76 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             return SmoColumnCustomNodeHelper.CalculateCustomLabel(smoObject, smoContext);
         }
 
-        public override IEnumerable<NodeSmoProperty> SmoProperties
+        private readonly Lazy<List<NodeSmoProperty>> smoPropertiesLazy = new Lazy<List<NodeSmoProperty>>(() => new List<NodeSmoProperty>
         {
-            get
+            new NodeSmoProperty
             {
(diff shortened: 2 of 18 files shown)


src/Microsoft.SqlTools.Credentials/Program.cs
```diff
@@ -36,15 +36,12 @@ internal static void Main(string[] args)
                 string logFilePath = commandOptions.LogFilePath;
```
ranasaria: This code is common to the 3 backend services: resourceprovider, credential service and tools service layer. We should probably refactor it out into a programbase class, so we do not leave behind unintended code in one when we fix others. Filed issue #712 for this.

src/Microsoft.SqlTools.Credentials/Program.cs
```diff
@@ -36,15 +36,12 @@ internal static void Main(string[] args)
...
                 {
-                    logFilePath = "credentials";
-                }
-                if (!string.IsNullOrWhiteSpace(commandOptions.LoggingDirectory))
-                {
-                    logFilePath = Logger.GenerateLogFilePath(Path.Combine(commandOptions.LoggingDirectory, logFilePath));
+                    logFilePath = Logger.GenerateLogFilePath("credentials");
                 }
 
+                Logger.AutoFlush = commandOptions.AutoFlushLog;
```
MattIrv: is there any time we wouldn't want auto flush now that it integrates with the output window?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoColumnCustomNode.cs
```diff
@@ -23,80 +23,76 @@ public override string GetNodeCustomName(object smoObject, SmoQueryContext smoCo
             return SmoColumnCustomNodeHelper.CalculateCustomLabel(smoObject, smoContext);
         }
 
-        public override IEnumerable<NodeSmoProperty> SmoProperties
+        private readonly Lazy<List<NodeSmoProperty>> smoPropertiesLazy = new Lazy<List<NodeSmoProperty>>(() => new List<NodeSmoProperty>
```
MattIrv: Thanks for fixing this and the other ones

test/ScriptGenerator/Properties/Resources.Designer.cs
```diff
@@ -66,10 +66,19 @@ internal class Resources {
...
+        ///IF (1 = FULLTEXTSERVICEPROPERTY(&apos;IsFullTextInstalled&apos;))
+        ///begin
+        ///EXEC [AdventureWorks].[dbo].[sp_fulltext_database] @action = &apos;enable&apos;
+        ///end
+        ///GO
+        ///ALTER DATABASE [AdventureWorks] SET ANSI_NULL_DEFAULT OFF 
+        ///GO
+        ///ALTER DATABASE [AdventureWorks] SET ANSI_NULLS ON 
+        ///GO
+        ///ALTER DATABASE  [rest of string was truncated]&quot;;.
```
MattIrv: what is this change?

ranasaria: This is just a generated file. Hopefully it won't keep getting generated every time. Else I will figure something out for it.

src/Microsoft.SqlTools.Credentials/Program.cs
ranasaria: This will get exposed as User option with default of false. There is penalty to autoflushing so we will advice user as such. Only where user if ok for some time to play perf penalty and want to see log emitted real time in out output window, will they turn this on.

src/Microsoft.SqlTools.Credentials/Program.cs
ranasaria: Since an issue has been filed we will deal with this in a future PR.

PR #719, opened 2018-10-26
Update SqlClient patch to set KeepAlive retry interval to 1 second

kburtram:
Update binary picks-up this change https://github.com/kburtram/corefx/commit/94686c4e44a5a9472f77417ee622d1fde612fa52.


kburtram:
I'm still validating if this is the correct fix.  I think the retry interval is incorrect, but this doesn't seem to behave as I'd expect either.

kburtram:
It looks like the SqlClient sockets are still sending a consistent 30 second keep alive with this change.  Unfortunately, I don't have an environment to trigger keepalive retries, so I can't trace that network traffic.  Based on the documentation the change looks correct so I'll go ahead with it.

screen shot 2018-10-26 at 3 31 13 pm

PR #722, opened 2018-10-30
Rollback SMO to latest 14.0 version

kburtram:
The SMO update appears to be related to https://github.com/Microsoft/azuredatastudio/issues/3012.  Rolling-back to 14.0 SMO fixed the IntelliSense issue in my repro scenario.  Since this is such a high-impact bug we should probably rollback until we have a tested fix.

### DIFF
--- a/Common.props
+++ b/Common.props
@@ -1,8 +1,7 @@
 <Project>
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
-    <SmoPackageVersion>150.18040.0-preview</SmoPackageVersion>
+    <SmoPackageVersion>140.17282.0-xplat</SmoPackageVersion>
     <HighEntropyVA>true</HighEntropyVA>
-	<TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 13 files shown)


kburtram:
@shueybubbles I'll leave this PR active while doing more investigations to make sure that this actually is related to SMO rather than "random" timing variations between test scenarios.

shueybubbles:
Did you just try changing the package version to the non-xplat 140 version in common.props instead of changing all the csproj files?

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

Common.props
```diff
@@ -1,8 +1,7 @@
 <Project>
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
-    <SmoPackageVersion>150.18040.0-preview</SmoPackageVersion>
+    <SmoPackageVersion>140.17282.0-xplat</SmoPackageVersion>
```
shueybubbles: i think you can just set this to 140.17283.0 and not change the csproj files, so we keep using the unified package. If that still fails but using the separate xplat packages works, then we'll know there's something wrong with the distribution itself.

kburtram: @shueybubbles I'll try this.  I'm seeing an inconsistent repro now where sometimes the 15.0 binaries are able to get the completion suggestions and other times not.  

Though overall we have seen a big spike in IntelliSense issues opened in the past month, so I still think this is the most likely related change, but it doesn't look as clear cut as 14 works & 15 doesn't.  It needs more investigation, but if I can't root cause today I'll probably rollback to a "known good" pending further investigation.

kburtram:
The repro for the intellisense issue is intermittent and I haven't nailed down exactly what's going on yet.  That said, from my testing this PR seems to work the best.  I'll go with this for November release and we can work on getting the 15 package tested and validated for December.  Worst case, the SMO bump is unrelated to the problem and we ship 15 preview a month later, so it seems worth it to see if this resolves the spike in intellisense issues reported with the October builds.

PR #723, opened 2018-10-30
Revert "Rollback SMO to latest 14.0 version"

kburtram:
Based on @shueybubbles's analysis this looks more likely to be related to BindingQueue changes for OE concurrency.  I'll leave the 14.0 revert in to produce a private test build while we root cause the intellisense issue.  I am not aware of anything in the 15.0 bump that we "need" for Nov (though of course we don't want to rollback for no good reason).


kburtram:
Based on investigation it doesn't look related to SMO.  I'll restore the 15.0 binaries and try the fix @MattIrv suggested.

PR #727, opened 2018-11-05
Add support for Azure Active Directory connections

MattIrv:
.NET Core supports Azure authentication now so we can use it to enable AAD authentication.

There are 2 ways of making connections that are used throughout our code. The main path goes through `ConnectionService`'s `Connect` method, but other parts of the code open and manage their own connections using `ConnectionService`'s static `OpenSqlConnection` method. I've updated both of these paths to support Azure access tokens that we will receive from the front end as part of the connection details. I've also added a new static `OpenServerConnection` method since many of the places that called `OpenSqlConnection` immediately wrapped it in a `ServerConnection` object.

This is for Microsoft/azuredatastudio#13


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
```diff
@@ -48,8 +48,9 @@ internal static ConnectionProviderOptions BuildConnectionProviderOptions()
                         ValueType = ConnectionOption.ValueTypeCategory,
                         SpecialValueType = ConnectionOption.SpecialValueAuthType,
                         CategoryValues = new CategoryValue[]
-                        { new CategoryValue {DisplayName = "SQL Login", Name = "SqlLogin" },
-                          new CategoryValue {DisplayName =  "Windows Authentication", Name= "Integrated" }
+                        { new CategoryValue { DisplayName = "SQL Login", Name = "SqlLogin" },
+                          new CategoryValue { DisplayName = "Windows Authentication", Name = "Integrated" },
+                          new CategoryValue { DisplayName = "Azure Active Directory (MFA)", Name = "AzureMFA" }
```
kburtram: Should we use the same text as SSMS?  I think this corresponds to the `Active Directory - Universal with MFA support`.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1552,4 +1584,23 @@ public static void EnsureConnectionIsOpen(DbConnection conn, bool forceReopen =
...
+        public string UserId { get; set; }
+
+        private string accessToken;
+
+        public AzureAccessToken(string accessToken)
+        {
+            this.accessToken = accessToken;
+        }
+
+        public string GetAccessToken() {
```
kburtram: `{` should be on its own line.

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -21,9 +21,9 @@
 </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Data.SqlClient" Version="4.5.0" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.6.0-preview3-27014-02" />
```
kburtram: this is going to cause problems with our private SqlClient patch for the macOS KeepAlive.  If we really need to move to a new SqlClient version let's sync-up on how to build a new private build with that fix.

MattIrv: we need the new SqlClient version because 4.5.0 doesn't have `AccessToken` on connections

kburtram: @MattIrv we'll need to build a private SqlClient for macOS that matches the `4.6.0-preview3-27014-02` sync point.  I'll track down where this is building from, but you'll need to include that updated binary with this PR to avoid breaking the KeepAlive scenario.  And the build scripts currently apply the patch, so whatever you reference will be overwritten in the packaged archives.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
MattIrv: In our design meeting we had agreed it should include the word "Azure" at the start but I can update it to be "Azure Active Directory - Universal with MFA support"

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
MattIrv: I haven't figured out how to track down what code corresponds to that release yet but when I tested with the patched binaries everything worked

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
ranasaria: +1

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1529,6 +1540,27 @@ internal static SqlConnection OpenSqlConnection(ConnectionInfo connInfo, string
             return null;
         }
 
+        /// <summary>
+        /// Create and open a new ServerConnection from a ConnectionInfo object.
+        /// This calls ConnectionService.OpenSqlConnection and then creates a
+        /// ServerConnection from it.
+        /// </summary>
+        internal static ServerConnection OpenServerConnection(ConnectionInfo connInfo, string featureName = null)
```
ranasaria: it would be good to add some description in doc comments about the 'featureName' parameter. It is not obvious how that parameter is used or what it does.

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
MattIrv: The [nuget gallery for that package](https://www.nuget.org/packages/System.Data.SqlClient/4.6.0-preview3-27014-02) shows that it comes from commit a8b44a7 in the corefx release/2.2 branch. In your repo that's https://github.com/kburtram/corefx/commit/a8b44a739a4ffef665a02bc83ba1355e0ee945df. So if you haven't updated your release/2.2 branch since building the custom binaries then they should work 4.6.0-preview3-27014-02.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -46,18 +46,18 @@ internal static class ReliableConnectionHelper
         /// Opens the connection and sets the lock/command timeout and pooling=false.
         /// </summary>
         /// <returns>The opened connection</returns>
-        public static IDbConnection OpenConnection(SqlConnectionStringBuilder csb, bool useRetry)
+        public static IDbConnection OpenConnection(SqlConnectionStringBuilder csb, bool useRetry, string azureAccountToken)
```
ranasaria: Should this be an optional parameter?

MattIrv: I made it required because everywhere that calls this method needs to take into account the possibility of an azure account token. If it were not required then it would be too easy for us to write code that doesn't use that argument at all and then fails when the user actually has an AAD connection

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
MattIrv: done

PR #728, opened 2018-11-07
DacFx import/export

kisantia:
Adding DacFx import,export, extract, and deploy handling to SqlToolsService so that it can be used in ADS.


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
+{
+    /// <summary>
+    /// Parameters for a DacFx deploy request.
+    /// </summary>
+    public class DacFxDeployParams : IScriptableRequestParams
```
pensivebrian: I don't think DacFx operations are scriptable, so we shouldn't have to implement this interface.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
```
kburtram: Do you think it makes sense to have so many references in file, folder, namespace, and class names to `DacFx`?  Instead of having more generic names like `DeploymentService` or `ExportService` etc.  For example, we don't call out SMO for all the classes that depend on that library.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>
+    /// Parameters for a DacFx deploy request.
+    /// </summary>
+    public class DacFxDeployParams : IScriptableRequestParams
+    {
+        /// <summary>
+        /// Gets or sets connection string of the target database the deploy operation will run against.
+        /// </summary>
+        public string ConnectionString { get; set; }
```
kburtram: this application is generally using connection URIs that maintain connection details with the ConnectionService.  Is there a reason not to follow that pattern with these new classes?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// Gets or sets name for deployed database
+        /// </summary>
+        public string TargetDatabaseName { get; set; }
+
+        /// <summary>
+        /// Connection uri
+        /// </summary>
+        public string OwnerUri { get; set; }
```
kburtram: seems redundant with the ConnectionString?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxDeployOperation.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using System.Data.SqlClient;
+using System.Diagnostics;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx
+{
+    /// <summary>
+    /// Class to represent an in-progress deploy operation
+    /// </summary>
+    class DacFxDeployOperation : DacFxOperation
```
kburtram: what's the thinking of how this relates to the existing Tasks support?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxExportOperation.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Microsoft.SqlServer.Dac;
```
kburtram: all the files should include the MIT copyright header

kevcunnane:
One question/comment: this seems like a good point to start moving things into their own projects. Tools like mssql-cli won't need this stuff and shouldn't long-term pull in so many dependencies. Most do need the existing core things (SMO, Intellisense, Connect, Query) but stuff like this should really start being put in their own DLL, even if it's pulled into the exe for now.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
kisantia: No I will remove ConnectionString and switch to using SqlConnection. I initially didn't see that I could get the connectionstring (which is needed to instantiate DacServices) from it and was just trying to get something working.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
kisantia: Will remove ConnectionString

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxDeployOperation.cs
kisantia: Do you mean the TaskServices api?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
pensivebrian: My opinion - I like have a DacFxService, which will wrap all of DacFx operations: import, export, schema compare, extract, deploy, etc.. However I don't think we need a DacFx prefix to all request/response type names in the RPC layer.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxExtractOperation.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using System.Data.SqlClient;
+using System.Diagnostics;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx
+{
+    /// <summary>
+    /// Class to represent an in-progress extract operation
+    /// </summary>
+    class DacFxExtractOperation : DacFxOperation
```
pensivebrian: Seems like we're repeating a log of operation code - should we centralize this using a helper or a generic method?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxDeployOperation.cs
```diff
@@ -0,0 +1,47 @@
...
+        public override void Execute(TaskExecutionMode mode)
+        {
+            if (this.CancellationToken.IsCancellationRequested)
+            {
+                throw new OperationCanceledException(this.CancellationToken);
+            }
+
+            try
+            {
+                DacServices ds = new DacServices(this.sqlConnection.ConnectionString);
```
pensivebrian: What about progress and message events? We want clients to see these as dacfx operations will be long-running.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -0,0 +1,258 @@
...
+using System.Collections.Concurrent;
+using System.Data.SqlClient;
+using System.Threading.Tasks;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx
+{
+    /// <summary>
+    /// Main class for DacFx service
+    /// </summary>
+    class DacFxService
```
pensivebrian: What about cancellation? Or should cancel be done from the TaskServices API?

pensivebrian: Also, think about progress and message events - should they go through the TaskServices API as well?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
kisantia: I removed the DacFx prefix from everything except DacFxService and the DacFxOperation base class that is used by import, export, extract, and deploy operations

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExportOperation.cs
```diff
@@ -0,0 +1,51 @@
...
+        public override void Execute(TaskExecutionMode mode)
+        {
+            if (this.CancellationToken.IsCancellationRequested)
+            {
+                throw new OperationCanceledException(this.CancellationToken);
+            }
+
+            try
+            {
+                DacServices ds = new DacServices(this.sqlConnection.ConnectionString);
```
pensivebrian: Can we use the constructor that takes a SecureString?

pensivebrian: Applies to entire PR where we instantiate DacServices.

kisantia: Currently, the SqlCredential with the SecureString password doesn't get populated when the ConnectionService opens the connection.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxExtractOperation.cs
kisantia: Moved the common code to base DacFxOperation class

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxDeployOperation.cs
kisantia: I added a handler for ProgressChanged, but it doesn't do anything since there isn't a way for the progress to be shown in ADS yet.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxDeployRequest.cs
kisantia: This is necessary for the TaskService, since it uses IScriptableRequestParams. Can be fixed in another code review

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
```diff
@@ -0,0 +1,122 @@
...
+            catch (Exception e)
+            {
+                Logger.Write(TraceEventType.Error, string.Format("DacFx import operation {0} failed with exception {1}", this.OperationId, e));
+                throw;
+            }
+        }
+
+        public abstract void Execute();
+
+        protected void DacProgressUpdates(object obj, DacProgressEventArgs args)
```
pensivebrian: We're unbinding event handlers - I generally don't expect in an event handler to unbind themselves, but to do something with the event arg data. What is the method attempting to accomplish?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -0,0 +1,257 @@
...
+            {
+                await requestContext.SendError(e);
+            }
+        }
+
+        /// <summary>
+        /// Handles request to deploy a dacpac
+        /// </summary>
+        /// <returns></returns>
+        public async Task HandleDeployRequest(DeployParams parameters, RequestContext<DeployResult> requestContext)
```
pensivebrian: Still seeing a lot of duplication in this HandleXXXXRequest methods. Can each handler instantiate the operation to run (ExtractOperation, DeployOperation, etc. ) and then pass the operation to a common method which handles the execution details of the operation?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
```diff
@@ -0,0 +1,122 @@
...
+        protected CancellationToken CancellationToken { get { return this.cancellation.Token; } }
+
+        /// <summary>
+        /// The error occurred during operation
+        /// </summary>
+        public string ErrorMessage
+        {
+            get
+            {
+                return string.Empty;
```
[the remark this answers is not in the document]
kisantia: This is what's in BackupOperation, but I'll remove it since it isn't doing anything.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
kisantia: Removed for now since the event handler isn't doing anything

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
kisantia: Sure I added an ExecuteOperation method that gets called by the handlers

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DacFxRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
+{
+    /// <summary>
+    /// Parameters for a DacFx request.
+    /// </summary>
+    public interface IDacFxParams : IScriptableRequestParams
```
pensivebrian: Seems like we can just make this a class and do away with all of the IDacFxParams implementations. But, this abstraction will break down in the future since, each operations will have operation specific parameters. This works for now, but think about how we can address this in the future.

kisantia: Changed it to be a base class instead

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
```diff
@@ -0,0 +1,93 @@
...
+        private bool disposed = false;
+
+        /// <summary>
+        /// Gets the unique id associated with this instance.
+        /// </summary>
+        public string OperationId { get; private set; }
+
+        public SqlTask SqlTask { get; set; }
+
+        protected SqlConnection SqlConnection { get; set; }
```
pensivebrian: Can this be "private set"?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        protected DacServices DacServices { get; set; }
```
pensivebrian: Can this be "private set"?

src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
```diff
@@ -22,9 +22,11 @@
     <PackageReference Include="System.Data.SqlClient" Version="4.6.0-preview3-27014-02" />
     <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" Version="$(SmoPackageVersion)" /><PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.0.0" />
+    <PackageReference Include="System.IO.Packaging" Version="4.5.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
     <PackageReference Include="System.Composition" Version="1.1.0" />
 	<PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
+	<PackageReference Include="System.Security.Permissions" Version="4.5.0" />
```
alanrenmsft: please fix the indentions here.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// </summary>
+        internal ConcurrentDictionary<string, DacFxOperation> ActiveOperations => operations.Value;
+
+        /// <summary>
+        /// Handles request to export a bacpac
+        /// </summary>
+        /// <returns></returns>
+        public async Task HandleExportRequest(ExportParams parameters, RequestContext<DacFxResult> requestContext)
+        {
+            try
```
alanrenmsft: I would recommend to use Task.Run to wrap the whole thing, take a look at the other providers for reference.

alanrenmsft: for example, profiler.

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -23,6 +23,7 @@
   <ItemGroup>
     <PackageReference Include="System.Data.SqlClient" Version="4.6.0-preview3-27014-02" />
     <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" Version="$(SmoPackageVersion)" />
+	<PackageReference Include="Microsoft.SqlServer.DacFx" Version="150.4240.1-preview" />
```
alanrenmsft: indention

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
kisantia: I will look into changing this in another PR

PR #729, opened 2018-11-08
fix for issue 3123

alanrenmsft:
SqlToolsService side of fix for issue: https://github.com/Microsoft/azuredatastudio/issues/3123

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
@@ -124,41 +124,44 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
         /// </summary>
         internal async Task HandleCreateXEventSessionRequest(CreateXEventSessionParams parameters, RequestContext<CreateXEventSessionResult> requestContext)
         {
-            try
+            await Task.Run(async () =>
             {
-                ConnectionInfo connInfo;
-                ConnectionServiceInstance.TryFindConnection(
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -205,8 +205,27 @@ internal async Task HandleStopProfilingRequest(StopProfilingParams parameters, R
 
                 if (session != null)
                 {
-                    session.XEventSession.Stop();
-                    await requestContext.SendResult(new StopProfilingResult{});
+                    // Occasionally we might see the InvalidOperationException due to a read is 
+                    // in progress, add the following retry logic will solve the problem.
+                    int remainingAttampts = 3;
```
kburtram: typo `remainingAttampts` => `remainingAttempts`

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -205,8 +205,27 @@ internal async Task HandleStopProfilingRequest(StopProfilingParams parameters, R
...
+                            break;
+                        }
+                        catch(InvalidOperationException)
+                        {
+                            remainingAttampts--;
+                            if (remainingAttampts == 0)
+                            {
+                                throw;
+                            }
+                            Thread.Sleep(500);
```
kburtram: this request handler should probably be moved into a different thread.  I think as the dispatcher is currently written this will block other events while the retries are going on.

PR #739, opened 2018-11-15
Send a Object Explorer session disconnect message on socket exceptions

kburtram:


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -353,9 +360,19 @@ private void ProcessQueue()
                                         {
                                             result = queueItem.ErrorHandler(ex);
                                         }
+
+                                        if (IsExceptionOfType(ex, typeof(SqlException)) || IsExceptionOfType(ex, typeof(SocketException)))
```
MattIrv: we might want to wrap the call to `queueItem.ErrorHandler` above in a try/catch, otherwise error handlers that fail will cause this code not to execute

kburtram: thanks, I've added a try\catch, though hopefully the error handler isn't also throwing an exception 😄

PR #740, opened 2018-11-16
Use consistent version of SqlClient

MattIrv:
My original commit to bump SqlClient to the version we need for AAD missed a few projects which didn't cause problems on local builds but causes problems on the Windows packaged build. This fixes it


kburtram:
@MattIrv I'm going to queue a new build and update ADS, so I'll merge to pick this up.

PR #743, opened 2018-11-20
Treat databases used by Polybase as system databases when listing databases in a server.

corivera:
When Polybase is enabled in SQL Server, it creates the databases DWConfiguration, DWDiagnostics, and DWQueue. These databases should not be altered or dropped by the user, so they are effectively treated like the built in system DBs. This change excludes those databases when retrieving the list of user databases in a server, so that the user can't select them by accident in something like the Create External Table wizard, for example.

Fixes https://github.com/Microsoft/azuredatastudio/issues/3229


corivera:
@kevcunnane Looks like dotnet sdk 2.2 isn't being found by the CI builds.

corivera:
Seems like the build failure is because dotnet sdk 2.2 isn't in the Microsoft package respository yet. I had to manually install 2.2 on my machine.

PR #752, opened 2018-11-30
Fix CodePages version issue

MattIrv:
The version of SqlClient we ended up using doesn't actually depend on a newer version of CodePages so we can use the stable 4.5.0 version

Once Azure Data Studio is updated it should fix the "save as csv" exception from Microsoft/azuredatastudio#3329


kburtram:
@MattIrv I'll merge and queue a new build.

PR #756, opened 2018-12-05
Batch parser project

NiranjanVirtuosity:
- This SQL Tools Service have BatchParser module in it. I am working on another project which have dependency on this BatchParser. So I am moving this batch 	  parser part to a separate library so that multiple projects can reference/utilize this library.

- To avoid the circular dependency issue between Microsoft.SqlTools.BatchParser and Microsoft.SqlTools.ServiceLayer, we have added another project[Microsoft.SqlTools.Common]. This is only used to resolve dependency issue in Microsoft.SqlTools.BatchParser and Microsoft.SqlTools.ServiceLayer.

- Changed few access modifiers from internal to public in "Microsoft.SqlTools.BatchParser" and "Microsoft.SqlTools.Common" 

To achieve this we have restructured the project by moving things around. Could someone review this and let us know your feedback.

+ @Matteo-T


Matteo-T:
Is the commit from November an unwanted one?

Matteo-T:
Does anybody know a good way to actually *see* what changes are being made here or maybe that's just the way it is when files are moved around (as opposed to being edited)?

NiranjanVirtuosity:
Hi, 
I have made some framework oriented changes and Created a new PR in branch named "BatchParser_NetStandard".

PR #759, opened 2018-12-11
Add DacFx Generate Deploy Script operation

kisantia:
Adds Generate Script functionality for upgrade database scenario that was requested (https://github.com/Microsoft/azuredatastudio/issues/3171#issuecomment-444470787)

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployScriptRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployScriptRequest.cs
@@ -1,0 +1,30 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
+{
+    /// <summary>
+    /// Parameters for a DacFx generate deploy script request.
+    /// </summary>
+    public class GenerateDeployScriptParams : DacFxParams
+    {
+        /// <summary>
+        /// Gets or sets the filepath where to save the generated script
+        /// </summary>
+        public string ScriptFilePath { get; set; }
+    }
+
+    /// <summary>
+    /// Defines the DacFx generate deploy script request type
+    /// </summary>
+    class GenerateDeployScriptRequest
+    {
+        public static readonly RequestType<GenerateDeployScriptParams, DacFxResult> Type =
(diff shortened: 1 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployScriptRequest.cs
```diff
@@ -0,0 +1,30 @@
...
+        public string ScriptFilePath { get; set; }
+    }
+
+    /// <summary>
+    /// Defines the DacFx generate deploy script request type
+    /// </summary>
+    class GenerateDeployScriptRequest
+    {
+        public static readonly RequestType<GenerateDeployScriptParams, DacFxResult> Type =
+            RequestType<GenerateDeployScriptParams, DacFxResult>.Create("dacfx/generateUgradeScript");
```
pensivebrian: spelling: generateUgradeScript. Also, it's not always an upgrade script. Maybe 'generateDeploymentScript' would be more clear?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployScriptOperation.cs
```diff
@@ -0,0 +1,36 @@
...
+        public GenerateDeployScriptOperation(GenerateDeployScriptParams parameters, SqlConnection sqlConnection) : base(sqlConnection)
+        {
+            Validate.IsNotNull("parameters", parameters);
+            this.Parameters = parameters;
+        }
+
+        public override void Execute()
+        {
+            DacPackage dacpac = DacPackage.Load(this.Parameters.PackageFilePath);
+            string incrementalScript = this.DacServices.GenerateDeployScript(dacpac, this.Parameters.DatabaseName, null, this.CancellationToken);
```
pensivebrian: Would the method DacServices.Script() work better since it generates the master script and database script for Azure?

pensivebrian: It also includes the deployment report.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -150,6 +151,45 @@ public async Task HandleDeployRequest(DeployParams parameters, RequestContext<Da
...
+        {
+            try
+            {
+                ConnectionInfo connInfo;
+                ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                if (connInfo != null)
+                {
+                    SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo, "GenerateScript");
```
alanrenmsft: I looked into the implementation of the Dac operations, it seems to me we don't really need a sqlconnection object, we only need the connection string to be passed to the DacServices object. Can we remove all the OpenSqlConnection call in this file?

kisantia: Ok I removed all the OpenSqlConnection calls and build the connection string instead

PR #760, opened 2018-12-12
BatchParser project in .NetStandard

NiranjanVirtuosity:
--Moved BatchParser folder from Microsoft.SqlTools.ServiceLayer to Microsoft.SqlTools.ManagerBatchParser
--Moved some dependency files from Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection to Microsoft.SqlTools.Common
--Added new project Microsoft.SqlTools.ManagerBatchParser
--Added new project Microsoft.SqlTools.Common
--Changing access modifiers from internal to public
These projects build in .Net Standard 2.0 
@Matteo-T


Matteo-T:
1) what's the error above about conflicting files?
2) Did you move/split the strings in resx files? How are those changes affecting the other loc files (I forgot their extension)
3) Are the tests passing 100%?
4) Should we spend time adding XML comments to the methods/types/etc that are now public? And, more importantly, add tests for them? Check with Karl and the other owners of this project. IMHO, this could be done in a different PR, as part of a different work item.

src/Microsoft.SqlTools.ManagedBatchParser/Localization/sr.cs
```diff
@@ -0,0 +1,6663 @@
...
+
+        public static string StoredProcedureScriptParameterComment
+        {
+            get
+            {
+                return Keys.GetString(Keys.StoredProcedureScriptParameterComment);
+            }
+        }
+
+        public
```
kburtram: it looks like the resource files include strings not related to batch parser.  you should set up srgen to build localization resources only for the strings needed for this project, then remove those resources from the other project to avoid unneeded duplication.

kburtram:
@Matteo-T for your questions.

1. The merge conflicts mean things have changed in master since this PR branch was created.  Please remerge or rebase master with the PR branch to resolve.  
2. The LOC files don't appear to be done correctly.  Please move only the needed resources to the new project, delete those resources from the original project, and setup the resource files to be autogen'ed using srgen by the build.bat command similar to what we're doing for other projects.
3. The AppVeyor build appears to be broken, so it's unlikely the unit tests are passing.
4. We don't need XML comments for all methods, but generally it's good to add these for key public methods of the API.  These can be added later if needed.  Please do make sure any new code is adequately tested, the code coverage doesn't drop, and the test projects are correctly organized prior to merging.

PR #763, opened 2019-01-03
This change ensures that we send back only on 'isComplete=true' message for a given result set

ranasaria:
Fixes:
1. The streaming protocol is not tightened to ensure that only the last message for a result set contain isCompleted=true.
2. Tightened logic in sending messages to make sure no duplicate messages get sent out due to concurrent processing.
3. Made a fix to a null reference exception when processing special action which was a pre-existing benign bug.

Testing: 1 more tests and improved verifications to existing tests


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SpecialAction.cs
```diff
@@ -72,7 +72,7 @@ public bool ExpectYukonXMLShowPlan
         /// </summary>
         public void CombineSpecialAction(SpecialAction action)
         {
-            flags |= action.flags;
+            flags |= ((action?.flags) ?? ActionFlags.None);
```
ranasaria: This just fixes an independent pre-existing null reference exception as we were derefencing action which was allowed to be passed in a null value.

anthonydresser:
@ranasaria whats the plan on getting this into ADS?

ranasaria:
I will wait for the official build to drop tonight and then on Monday, I will bump up the tools version.


-  Arvind.

PR #765, opened 2019-01-17
Unified connection support

YurongHe:
- Add the query to get big data server property Note: current the server property is not implemented yet. Right now using query of sys.assembilities to unblock
- Add the query to get BDC endpoints if it is BDC cluster. This is the temp solution provided so far. Platform will change this to use stored proc to get BDC endpoints.
- Add option property bag to ServerInfo to hold the new properties. So that any new metadata doesn't require contract change in the further
  Will access if any existing properties should move to properties bag. Will do the refactor later.
- Fixed the null reference bug when the node is found. Return empty nodes in this case. For example: the request node path is server/Data Services/HDFS.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
@@ -5,7 +5,6 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
 {
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
@@ -3,6 +3,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
+using System.Collections.Generic;
+using static Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection.ReliableConnectionHelper;
+
 namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
 {
     /// <summary>
@@ -64,5 +68,20 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
         /// The Operating System version string of the machine running the SQL Server instance.
         /// </summary>
         public string MachineName { get; set; }
+
+        public List<BigDataClusterEndpoint> BigDataClusterEndpoints { get; set; }
+    }
+}
+
+public class BigDataClusterEndpoint
+{
+    public string ServiceName;
+    public string IpAddress;
+    public int Port;
+    public BigDataClusterEndpoint(string serviceName, string ipAddress, int port)
+    {
+        ServiceName = serviceName;
+        IpAddress = ipAddress;
+        Port = port;
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
@@ -69,7 +69,10 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
         /// </summary>
         public string MachineName { get; set; }
 
-        public List<BigDataClusterEndpoint> BigDataClusterEndpoints { get; set; }
+        /// <summary>
+        /// Server options
+        /// </summary>
+        public Dictionary<string, object> Options { get; set; }
     }
 }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
@@ -5,7 +5,7 @@
 
 using System;
 using System.Collections.Generic;
-using static Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection.ReliableConnectionHelper;
+using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
 {
@@ -76,15 +76,9 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
     }
 }
 
-public class BigDataClusterEndpoint
+public class ClusterEndpoint
 {
     public string ServiceName;
     public string IpAddress;
     public int Port;
-    public BigDataClusterEndpoint(string serviceName, string ipAddress, int port)
-    {
-        ServiceName = serviceName;
-        IpAddress = ipAddress;
-        Port = port;
-    }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
@@ -817,6 +817,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
                     serverInfo.OsVersion = reader[0].ToString();
                 });
 
+                // Get BDC endpoints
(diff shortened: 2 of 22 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/CreateSessionRequest.cs
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System.Collections.Generic;
```
MattIrv: unused import

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
```diff
@@ -3,6 +3,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
+using System.Collections.Generic;
+using static Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection.ReliableConnectionHelper;
```
kevcunnane: Please remove this using as you're not making any use of it, thanks.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
```diff
@@ -64,5 +68,23 @@ public class ServerInfo
...
+        /// <summary>
+        /// Server options
+        /// </summary>
+        public Dictionary<string, object> Options { get; set; }
+    }
+}
+
+public class BigDataClusterEndpoint
+{
+    public string ServiceName;
```
kevcunnane: The pattern is to use getters / setters for all fields in JSON-serializable contract classes. I'd also suggest following standard practice to not have a constructor, and just do
```c#
new BigDataClusterEndpoint() { ServiceName = 'xyz', ... }
```
to assign to setters during construction instead

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -806,6 +817,23 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
                     serverInfo.OsVersion = reader[0].ToString();
                 });
 
+                int? isBigDataCluster = serverInfo.Options[ServerInfo.OptionIsBigDataCluster] as int?;
```
kevcunnane: What happens in C# if the key doesn't exist, can you remind me? I had thought it might throw.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -806,6 +817,23 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
...
+                if (isBigDataCluster != null && isBigDataCluster.Value == 1)
```
kevcunnane: I like this implementation, where we add the 1 extra call if needed but in most cases we're not affected. This seems like a good tradeoff.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -806,6 +817,23 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
...
+                {
+                    ExecuteReader(
+                    connection,
+                    SqlConnectionHelperScripts.GetBigDataClusterEndpoints,
+                    delegate (IDataReader reader)
+                    {
+                        List<BigDataClusterEndpoint> bigDataClusterEndpoints = new List<BigDataClusterEndpoint>();
```
kevcunnane: Minor: I think we can just use `clusterEndpoints` everywhere. It's a lot shorter and has the same basic meaning.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/SqlConnectionHelperScripts.cs
```diff
@@ -47,5 +47,10 @@ @rc 			int
 ";
 
     public const string GetOsVersion = @"SELECT OSVersion = RIGHT(@@version, LEN(@@version)- 3 -charindex (' ON ', @@version))";
+    public const string GetBigDataClusterEndpoints = @"
+SELECT [service_name]
```
kevcunnane: Minor: seems like convention is to put the queries on 1 line if possible. This is small enough that this seems easy to do.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -699,6 +701,8 @@ public class ServerInfo
             public string OsVersion;
 
             public string MachineName;
+
+            public Dictionary<string, object> Options { get; set; }
```
kburtram: I would recommend deriving from `GeneralRequestDetails` which contains an Options member a several nice related helper methods.  Then implementing getter\setters that read\write to the property bag.  This makes the internal code nice using strongly typed properties, but keeps the wire-format generic.  Take a look at `ConnectionDetails` for this pattern.

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
ranasaria: Dictionary will throw KeyNotFoundException: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.item?view=netframework-4.7.2

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -754,6 +758,8 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
                 delegate (IDataReader reader)
                 {
                     reader.Read();
+                    serverInfo.Options = new Dictionary<string, object>();
```
ranasaria: Do we need to construct serverInfo.Options even when reader.FieldCount <= 6?, could serverInfo.Options stay null in those scenarios? or does it need to be an empty Dictionary?

src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -806,6 +817,27 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
...
+                {
+                    int? isBigDataCluster = serverInfo.Options[ServerInfo.OptionIsBigDataCluster] as int?;
+                    if (isBigDataCluster != null && isBigDataCluster.Value == 1)
+                    {
+                        ExecuteReader(
+                        connection,
+                        SqlConnectionHelperScripts.GetClusterEndpoints,
+                        delegate (IDataReader reader)
+                        {
+                            List<ClusterEndpoint> clusterEndpoints = new List<ClusterEndpoint>();
```
kevcunnane: Minor: what happens if we don't have permissions to this table / can't find it? In the future, this is a possibility. We should consider this.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -406,7 +405,18 @@ internal ExpandResponse QueueExpandNodeRequest(ObjectExplorerSession session, st
         {
             NodeInfo[] nodes = null;
             TreeNode node = session.Root.FindNodeByPath(nodePath);
-            ExpandResponse response = new ExpandResponse { Nodes = new NodeInfo[] { }, ErrorMessage = node.ErrorMessage, SessionId = session.Uri, NodePath = nodePath };
+            ExpandResponse response = null;
+
+            if (node == null)
+            {
+                response = new ExpandResponse { Nodes = new NodeInfo[] { }, ErrorMessage = string.Empty, SessionId = session.Uri, NodePath = nodePath };
```
kevcunnane: Please add a comment explaining this scenario, basically `// This node was likely returned from a different node provider. Ignore expansion and return an empty array since we do not need to add any nodes under this section of the tree`

PR #768, opened 2019-01-29
Add DacFx Generate Deploy Plan Operation

kisantia:
This adds the generate deploy plan operation which returns the report with the operations that will be done when the dacpac is deployed to the database.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployPlanRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployPlanRequest.cs
@@ -29,6 +29,9 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     /// </summary>
     public class GenerateDeployPlanRequestResult : DacFxResult
     {
+        /// <summary>
+        /// An xml string that details the alerts and the operations for deploying the specified dacpac to the database
+        /// </summary>
         public string Report { get; set; }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployPlanRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployPlanRequest.cs
@@ -9,25 +9,25 @@ using Microsoft.SqlTools.ServiceLayer.Utility;
 namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 {
     /// <summary>
-    /// Parameters for a DacFx upgrade plan request.
+    /// Parameters for a DacFx generate deploy plan request.
     /// </summary>
-    public class UpgradePlanParams : DacFxParams
+    public class GenerateDeployPlanParams : DacFxParams
     {
     }
 
     /// <summary>
-    /// Defines the DacFx upgrade plan request type
+    /// Defines the DacFx generate deploy plan request type
     /// </summary>
-    class UpgradePlanRequest
+    class GenerateDeployPlanRequest
     {
-        public static readonly RequestType<UpgradePlanParams, UpgradePlanRequestResult> Type =
-            RequestType<UpgradePlanParams, UpgradePlanRequestResult>.Create("dacfx/upgradePlan");
+        public static readonly RequestType<GenerateDeployPlanParams, GenerateDeployPlanRequestResult> Type =
+            RequestType<GenerateDeployPlanParams, GenerateDeployPlanRequestResult>.Create("dacfx/generateDeployPlan");
     }
 
     /// <summary>
     /// Parameters returned from a generate deploy script request.
     /// </summary>
-    public class UpgradePlanRequestResult : DacFxResult
+    public class GenerateDeployPlanRequestResult : DacFxResult
     {
         public string Report { get; set; }
     }
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -168,7 +168,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
 
                     // want to show filepath in task history instead of server and database
                     metadata.ServerName = parameters.ScriptFilePath;
-                    metadata.DatabaseName = "";
+                    metadata.DatabaseName = string.Empty;
 
                     sqlTask = SqlTaskManagerInstance.CreateAndRun<SqlTask>(metadata);
 
@@ -176,7 +176,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
                     {
                         OperationId = operation.OperationId,
                         Success = true,
-                        ErrorMessage = ""
+                        ErrorMessage = string.Empty
                     });
                 }
             }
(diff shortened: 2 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        public GenerateDeployPlanParams Parameters { get; }
+
+        public GenerateDeployPlanOperation(GenerateDeployPlanParams parameters, ConnectionInfo connInfo): base(connInfo)
+        {
+            Validate.IsNotNull("parameters", parameters);
+            this.Parameters = parameters;
+        }
+
+        public override void Execute()
```
pensivebrian: Shouldn't execute do something?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
```diff
@@ -0,0 +1,125 @@
...
+            DacDeployOptions options = GetDefaultDeployOptions();
+            // this is so that operations with possible data loss will be in the report
+            options.BlockOnPossibleDataLoss = false;
+
+            return ds.GenerateDeployReport(dacpac, this.Parameters.DatabaseName, options, this.CancellationToken);
+        }
+
+        private DacDeployOptions GetDefaultDeployOptions()
+        {
+            DacDeployOptions options = new DacDeployOptions
```
pensivebrian: I'd had to have defaults hard coded since they can change. Do we have to specify them here? Are the default not already set when instantiating a new instance?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateDeployPlanRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+        public static readonly RequestType<GenerateDeployPlanParams, GenerateDeployPlanRequestResult> Type =
+            RequestType<GenerateDeployPlanParams, GenerateDeployPlanRequestResult>.Create("dacfx/generateDeployPlan");
+    }
+
+    /// <summary>
+    /// Parameters returned from a generate deploy script request.
+    /// </summary>
+    public class GenerateDeployPlanRequestResult : DacFxResult
+    {
+        public string Report { get; set; }
```
pensivebrian: A comment would be nice? Is the report an xml string? Or is this a file path?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
kisantia: These are the defaults used in SSMS, which are different from the default DacDeployOptions when instantiating a new instance. For example, the default for DropObjectsNotInSource is false, but in SSMS the default is true.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
kisantia: The report string needs to be returned to be sent with the result, so the inherited Execute() can't be used.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
pensivebrian: That's unfortunate. sqlpackage, DacFx API, and now SSMS have different defaults. Are you proposing SSMS should be defaults going forward?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
pensivebrian: There are probably more elegant ways to deal with this. If the abstraction no longer make sense, let's fix the abstraction. Not hack around it.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
kisantia: It would be good to have the same defaults for all three, but I don't know if the SSMS values should be the default since it could result in unintentionally dropping objects. In this case, using the SSMS values makes sense so that the same deploy report is generated in both SSMS and ADS.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
kisantia: I changed the abstract method in the base class to be virtual instead. Is this a better way to deal with this?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
pensivebrian: Have any proposals to resolve this?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
pensivebrian: Well, it look like the intent of the Execute() method was to be some sort of command pattern. Given that, I see three other approaches:
1. Do we use the DacFxOperation.Execute() method class directly? If not, and we only deal with concrete classes, then maybe we can delete the Execute() method from the base class altogether? 
2. Execute can set a ReportResults property when complete
3. Enhance Execute() to provide a result, e.g. Execute<T> where the concrete class provides the return type.

kisantia: Ok I added a DeployReport property to the class to store the generated report. Execute() was initially added in the base class so that the code in Execute(TaskExecutionMode mode) can be reused easily.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -246,5 +279,15 @@ internal void PerformOperation(DacFxOperation operation)
         {
             operation.Execute(TaskExecutionMode.Execute);
         }
+
+        /// <summary>
+        /// For testing purpose only
+        /// </summary>
+        /// <param name="operation"></param>
+        internal string PerformGenerateDeployPlanOperation(GenerateDeployPlanOperation operation)
```
pensivebrian: Do we need this method? It seems so similiar to PerformOperation() above. Can PerformOperation just return a string?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -185,6 +186,38 @@ await requestContext.SendResult(new DacFxResult()
...
+                if (connInfo != null)
+                {
+                    GenerateDeployPlanOperation operation = new GenerateDeployPlanOperation(parameters, connInfo);
+                    operation.Execute();
+
+                    await requestContext.SendResult(new GenerateDeployPlanRequestResult()
+                    {
+                        OperationId = operation.OperationId,
+                        Success = true,
+                        ErrorMessage = "",
```
pensivebrian: nit: use string.Empty

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
kisantia: it isn't needed anymore after adding the deploy report property to the operation

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateDeployPlanOperation.cs
kisantia: Most of the SSMS default options are the same as the DacFx ones, so here this can be changed to only set the options that are different rather than setting all of them. It looks like only 7 of them are different from the default value set. It would be good to also change this in SSMS to simplify.

pensivebrian: This works for now. Let's make sure that we use the same options when running the deployment in the deploy operation. Long term, I think we should align on the DacFx defaults, and make that change in SSMS. This was, we specify the defaults in one place.

PR #770, opened 2019-01-29
Fixed #3839 model.sys.assemblies doesn't exist issue

YurongHe:
Cloud doesn't have model.sys.assemblies. Also can't use three part naming to access. The workaround is changing the query to check if the cluster_endpoint_info exist or not


src/Microsoft.SqlTools.ServiceLayer/Connection/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -818,30 +818,29 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
                 });
 
                 // Get BDC endpoints
-                if (serverInfo != null && serverInfo.Options != null && serverInfo.Options.ContainsKey(ServerInfo.OptionIsBigDataCluster))
+                if (!serverInfo.IsCloud)
```
kburtram: this check is no longer needed `serverInfo != null`?

PR #775, opened 2019-02-16
Added more options to job changes and elaborated error messages

abist:
Minor changes as part of committer work.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
@@ -298,8 +298,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Agent
                 catch (Exception e)
                 {
                     result.Success = false;
-                    result.ErrorMessage = e.Message + "\r\n\t" + e.InnerException.Message
-                        + "\r\n\t" + e.InnerException.InnerException.Message;
+                    result.ErrorMessage = e.Message + Environment.NewLine + "\t" + e.InnerException.Message
+                        + Environment.NewLine + "\t" + e.InnerException.InnerException.Message;
                     await requestContext.SendResult(result);
                 }
             });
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/JobStepsActions.cs
```diff
@@ -4,6 +4,7 @@
 //
 
 using System;
+using Microsoft.SqlTools.ServiceLayer.Agent;
```
kburtram: we should try to keep using statements sorted (though it is common & fine to sort `System` namespaces ahead of others).

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -298,7 +298,8 @@ internal async Task HandleJobActionRequest(AgentJobActionParams parameters, Requ
                 catch (Exception e)
                 {
                     result.Success = false;
-                    result.ErrorMessage = e.Message;
+                    result.ErrorMessage = e.Message + Environment.NewLine + "\t" + e.InnerException.Message
```
kburtram: shouldn't we check that `InnerException` isn't null prior to accessing it?  Or for some reason we know it will always be non-null here?

abist: Usually there's always an inner exception (which is the indented exception details shown in SSMS). But I'll add the check to be safe.

kburtram: When accessing a member without a null check we need to be 100% sure it's not null, so if it's "usually" not null we still need a null check.  FYI, in general most exceptions have InnerException null, though SqlExeptions often do have an InnerException.

Unless you have a block of code that sets the innerexception that you know runs prior to your catch block, you should always null check prior to access.

PR #776, opened 2019-02-22
Feature/cms backend

udeeshagautam:
Cms SQL Tools Service first set of Apis
- Connect to CMS Server
- Add/Remove Registered Servers to CMS
- Add remove Server Groups to CMS


src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,364 @@
...
+            serviceHost.SetRequestHandler(ListRegisteredServerRequest.Type, this.HandleListRegisteredServersRequest);
+            serviceHost.SetRequestHandler(AddRegisteredServerRequest.Type, this.HandleAddRegisteredServerRequest);
+            serviceHost.SetRequestHandler(RemoveRegisteredServerRequest.Type, this.HandleRemoveRegisteredServerRequest);
+            serviceHost.SetRequestHandler(AddServerGroupRequest.Type, this.HandleAddServerGroupRequest);
+            serviceHost.SetRequestHandler(RemoveServerGroupRequest.Type, this.HandleRemoveServerGroupRequest);
+        }
+
+        public async Task HandleCreateCentralManagementServerRequest(CreateCentralManagementServerParams createCmsParams, RequestContext<ListRegisteredServersResult> requestContext)
+        {
+            Logger.Write(TraceEventType.Verbose, "HandleCreateCentralManagementServerRequest");
```
anthonydresser: I noticed none of the handlers are async. If there are any long running operations we should make sure they are async so we don't block the ipc handler.

udeeshagautam: Sure will do for new connections. I noticed we do async notification for new connection but not for list databases like scenarios - so will follow the same.

anthonydresser: I wouldn't recommend following the connection service example. The notification system it uses is outdated and complicated.

anthonydresser: Just run async tasks like https://github.com/Microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs#L230

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisterServerParams.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace Microsoft.SqlTools.ServiceLayer.Cms.Contracts
+{
+    /// <summary>
+    /// Paramaters to create Top level Central Management Server
+    /// </summary>
+    public class CreateCentralManagementServerParams
+    {
+        public string RegisteredServerName { get; set; }
+
+        public string RegisterdServerDescription { get; set; }
```
abist: small typo:  RegisteredServerDescription

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                CmsTask = Task.Run(async () =>
+                {
+                    try
+                    {
+                        //Validate params and connect
+                        ServerConnection conn = await ValidateAndCreateConnection(createCmsParams.ConnectParams);
+
+                        // Get Current Reg Servers on CMS
+                        RegisteredServersStore store = new RegisteredServersStore(conn);
```
udeeshagautam: TODO: Using the alias/display name passed for CMS

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,434 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.SqlTools.ServiceLayer.Cms
+{
+    /// <summary>
+    /// Main class for DacFx service
```
kburtram: `DacFx` should be `CmsService`

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,434 @@
...
+                CmsTask = Task.Run(async () =>
+                {
+                    try
+                    {
+                        //Validate params and connect
+                        ServerConnection conn = await ValidateAndCreateConnection(createCmsParams.ConnectParams);
+
+                        // Get Current Reg Servers on CMS
+                        RegisteredServersStore store = new RegisteredServersStore(conn);
+                        ServerGroup parentGroup = store.DatabaseEngineServerGroup; //TODO Do we need other types like Integrated Service and Analysis Service
```
kburtram: we're not supporting SSIS or AS in this codebase, so probably don't need the `TODO`.  If we want that in the future there will need to be large scale refactoring in a bunch of places.

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,434 @@
...
+                    catch (Exception ex)
+                    {
+                        // Exception related to connection/creation will only be caught here. Note that the outer catch will not catch them
+                        await requestContext.SendError(ex);
+                    }
+                });
+            }
+            catch (Exception e)
+            {
+                // Exception related to run tak will be captured here
```
kburtram: nit: typo `tak` -> `task`

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,434 @@
...
+                {
+                    registerdServerStore = value;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Internal variable for testability
+        /// </summary>
+        internal Task CmsTask { get; set; }
```
kburtram: what are you using this property for?  I don't see it being accessed outside this class, and then only to assign to it.

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,434 @@
...
+        /// <summary>
+        /// Internal for testing purposes only
+        /// </summary>
+        internal static RegisteredServersStore ServerStore
+        {
+            get
+            {
+                if (registerdServerStore == null)
+                {
+                    registerdServerStore = RegisteredServersStore.LocalFileStore;
```
kburtram: should you assign using the setter (i.e. `this.ServerStore = …`) to go through the lock, or take the lock explicitly here?

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -0,0 +1,434 @@
...
+            ListRegisteredServersResult result = new ListRegisteredServersResult() { RegisteredServersList = serverResults, RegisteredServerGroups = groupsResults };
+            return result;
+        }
+
+        #endregion
+
+        /// <summary>
+        /// Internal for testing purposes only
+        /// </summary>
+        internal static ConnectionService ConnectionServiceInstance
```
kburtram: I think we've been putting properties above methods inside class definitions, at least that's what the .Net Framework coding guidelines recommend we do.

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisterServerParams.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <summary>
+    /// Paramaters to create Top level Central Management Server
+    /// </summary>
+    public class CreateCentralManagementServerParams
+    {
+        public string RegisteredServerName { get; set; }
+
+        public string RegisteredServerDescription { get; set; }
+
+        public ConnectParams ConnectParams { get; set; }
```
kburtram: should this class derive from ConnectParams instead?

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisterServerParams.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /// <summary>
+    /// Parmaters to Add Registered Server to top level CMS
```
kburtram: nit: `Parmaters` -> `Parameters`

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisterServerParams.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// </summary>
+    public class AddRegisteredServerParams
+    {
+        public string RegisteredServerName { get; set; }
+
+        public string RegisteredServerDescription { get; set; }
+        
+        public ConnectionDetails RegServerConnectionDetails { get; set; }
```
kburtram: we should be consistent whether to use `RegisteredServer` or `RegServer`, personally I would always use `RegisteredServer` event though it's too verbose.

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisterServerParams.cs
kburtram: ..note this same typo appears to have been copy\pasted a few times

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisteredServerRequest.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Microsoft.SqlTools.ServiceLayer.Cms.Contracts
+{
+
+    public class CreateCentralManagementServerRequest
+    {
+        public static readonly RequestType<CreateCentralManagementServerParams, ListRegisteredServersResult> Type =
+            RequestType<CreateCentralManagementServerParams, ListRegisteredServersResult>.Create("cms/createCms");
+    }
+
+    public class ListRegisteredServerRequest
```
kburtram: should this be `ListRegisteredServersRequest` (plural)?

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisteredServerRequest.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        public static readonly RequestType<ListRegisteredServerParams, ListRegisteredServersResult> Type =
+            RequestType<ListRegisteredServerParams, ListRegisteredServersResult>.Create("cms/listRegServers");
```
kburtram: I'd write out RegisteredServers completely in the JSON-RPC method names.

src/Microsoft.SqlTools.ServiceLayer/Cms/Contracts/RegisteredServersResult.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public string Name { get; set; }
+
+        public string ServerName { get; set; }
+
+        public string Description { get; set; }
+
+        public string RelativePath { get; set; }
+
+        public ConnectionDetails connectionDetails { get; set; }
```
kburtram: `connectionDetails` should be `ConnectionDetails`

src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
udeeshagautam: This is only used in tests. Since the call to Handle functions in Service is async, we need a way to wait for task to complete to validate requestContext result. This was one pattern used for testing across multiple services so used the same.

kburtram: makes sense.  For some reason when I search for usages yesterday I didn't see any, but now I see this being used through out the tests.  Maybe I had a typo in search box 😄

PR #780, opened 2019-03-08
Fix for : 4045: Cannot cancel a Query. Query runs too long. (And SMO update)

udeeshagautam:
**Fix for : 4045: Cannot cancel a Query. Query runs too long**
HandleExecuteRequest was returning a task to awaiter - which was getting waited on. changed it to async function to start task in new thread and return nothing to awaiter . Also the cancellation token set by cancel request was getting checked only after making the connection to execute batches. Added an additional check for cancellation token before the connection has been made.

**Fix for 4319: Error showing dbs when using AAD in... 1.5.0-alpha.74**
David has already created a new version of SMO nuget with the fix. - incorporating the same (150.18096.0-preview).

### DIFF
--- a/Common.props
+++ b/Common.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
-    <SmoPackageVersion>150.18085.0-preview</SmoPackageVersion>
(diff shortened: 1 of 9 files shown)


anthonydresser:
Looks like some tests broke; be sure to fix them

udeeshagautam:
@anthonydresser, @abist, @kburtram - just wanted to let you know that that I am completing this - so that it can be part of next update to ADS whenever that happens.

PR #782, opened 2019-03-14
fix CMS Service issue

abist:
Fixed issue with add registered servers call


src/Microsoft.SqlTools.ServiceLayer/Cms/CmsService.cs
```diff
@@ -105,15 +105,15 @@ public async Task HandleAddRegisteredServerRequest(AddRegisteredServerParams cms
                             RegisteredServerCollection servers = parentGroup.RegisteredServers;
 
                             // Add the new server (intentionally not cheching existence to reuse the exception message)
-                            RegisteredServer registeredServer = new RegisteredServer(parentGroup, cmsCreateParams.RegisteredServerName);
```
udeeshagautam: The value used here should be alias... i.e. it can be xyz and not usable in connection string- which may be different from connectiondetails.ServerName .. which should be real name that can be used in a connection string

PR #787, opened 2019-03-22
Add schema compare generate script operation

kisantia:
This adds the schema compare generate script operation that generates the script for deploying the source to the target. It generates the script with the default options and does not yet support include/exclude differences. The public schema compare API only supports generating the script when the target is a database.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
@@ -9,9 +9,10 @@ using Microsoft.SqlTools.ServiceLayer.Hosting;
 using Microsoft.SqlTools.ServiceLayer.TaskServices;
 using System;
 using System.Collections.Concurrent;
-using System.Data.SqlClient;
 using System.Threading.Tasks;
 using Microsoft.SqlTools.ServiceLayer.SchemaCompare;
+using Microsoft.SqlServer.Dac.Compare;
+using Microsoft.SqlTools.ServiceLayer.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.SchemaCopmare
 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
```diff
@@ -85,6 +100,59 @@ await requestContext.SendResult(new SchemaCompareResult()
...
+                sqlTask = SqlTaskManagerInstance.CreateAndRun<SqlTask>(metadata);
+
+                await requestContext.SendResult(new ResultStatus()
+                {
+                    Success = true,
+                    ErrorMessage = operation.ErrorMessage
+                });
+            }
+            catch (Exception e)
+            {
```
udeeshagautam: We are not sending back e.Message. Might miss on some failure message..

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
```diff
@@ -73,9 +83,14 @@ await requestContext.SendResult(new SchemaCompareResult()
                             Differences = operation.Differences
```
udeeshagautam: Will operation.diff be available here in all cases (considering async/failure without exception if any)?

udeeshagautam:
Looks like the cc decrease is in schema comapar area. Am I seeing it wrong? I do see tests so not sure if we need to do something there.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
kisantia: operation.ErrorMessage gets set to e.Message before the exception gets thrown to this catch

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
kisantia: It should be ok if operation.Differences is null as long as it's handled properly in the ADS extension

PR #788, opened 2019-03-25
Make "--enable-logging" an ignored option for mssql back-compat

kburtram:
Fixes https://github.com/Microsoft/vscode-mssql/issues/1200


kburtram:
I'm going to merge and get an updated build published.  We can keep a bug open to remove this "no-op" argument and change the mssql settings to no longer send this argument separately.

PR #789, opened 2019-03-26
Fix invalid dacpac version crashing sqltoolsservice

kisantia:
SqlToolsService when trying to convert an invalid version string to a Version object. This fixes that by trying to do the conversion later and throwing an error.

image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
@@ -31,10 +31,9 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
         {
             // try to get version
             Version version;
-            Version.TryParse(this.Parameters.ApplicationVersion, out version);
-            if (version == null)
+            if (!Version.TryParse(this.Parameters.ApplicationVersion, out version))
             {
-                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Invalid version {0} passed. Version must be in the format x.x.x.x where x is a number", this.Parameters.ApplicationVersion));
+                throw new ArgumentException(string.Format(SR.ExtractInvalidVersion, this.Parameters.ApplicationVersion));
             }
             this.DacServices.Extract(this.Parameters.PackageFilePath, this.Parameters.DatabaseName, this.Parameters.ApplicationName, version, null, null, null, this.CancellationToken);
         }
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
@@ -37,7 +37,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
             }
             catch
             {
-                Logger.Write(TraceEventType.Warning, string.Format(CultureInfo.InvariantCulture, "Invalid version {0} passed", this.Parameters.ApplicationVersion));
+                throw new Exception(string.Format("Invalid version {0} passed. Version must be in the format x.x.x.x where x is a number", this.Parameters.ApplicationVersion));
             }
 
             this.DacServices.Extract(this.Parameters.PackageFilePath, this.Parameters.DatabaseName, this.Parameters.ApplicationName, version, null, null, null, this.CancellationToken);
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
@@ -29,13 +29,19 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
 
         public override void Execute()
         {
-            // try to get version
-            Version version;
-            if (!Version.TryParse(this.Parameters.ApplicationVersion, out version))
-            {
-                throw new ArgumentException(string.Format(SR.ExtractInvalidVersion, this.Parameters.ApplicationVersion));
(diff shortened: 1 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
```diff
@@ -37,7 +37,7 @@ public override void Execute()
             }
             catch
             {
-                Logger.Write(TraceEventType.Warning, string.Format(CultureInfo.InvariantCulture, "Invalid version {0} passed", this.Parameters.ApplicationVersion));
+                throw new Exception(string.Format("Invalid version {0} passed. Version must be in the format x.x.x.x where x is a number", this.Parameters.ApplicationVersion));
```
pensivebrian: Shouldn't this error message be localized?

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
```diff
@@ -37,7 +37,7 @@ public override void Execute()
             }
             catch
             {
```
pensivebrian: Using Version.TryParse would be more concise and readable since it avoid the try{} catch{}.

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
pensivebrian: I throwing a generic exception right here? Consider an ArgumentException of an InvalidOperationException.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -253,10 +253,16 @@ private async Task<Mock<RequestContext<DacFxResult>>> SendAndValidateExtractInva
 
             DacFxService service = new DacFxService();
             ExtractOperation operation = new ExtractOperation(extractParams, result.ConnectionInfo);
```
pensivebrian: Do you have to have an integration test to validate a version? Could the operation be tested without the DacFxService?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -253,10 +253,16 @@ private async Task<Mock<RequestContext<DacFxResult>>> SendAndValidateExtractInva
...
-            service.PerformOperation(operation);
+            try
+            {
+                service.PerformOperation(operation);
+            }
+            catch (Exception e)
+            {
+                Assert.Contains("Invalid version", e.Message);
```
pensivebrian: This only works for english. How can we have this run for other locales?

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
kisantia: changed it to be an ArgumentException

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
```diff
@@ -28,7 +29,14 @@ public ExtractOperation(ExtractParams parameters, ConnectionInfo connInfo) : bas
 
         public override void Execute()
         {
-            this.DacServices.Extract(this.Parameters.PackageFilePath, this.Parameters.DatabaseName, this.Parameters.ApplicationName, this.Parameters.ApplicationVersion, null, null, null, this.CancellationToken);
+            // try to get version
+            Version version;
+            Version.TryParse(this.Parameters.ApplicationVersion, out version);
```
pensivebrian: Version is value type, so won't comparing to null always be false? Version.TryParse() returns a boolean, so you can put that method invocation as if statement conditional expression: `if (!TryParse(...)) throw`

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
```diff
@@ -28,7 +29,14 @@ public ExtractOperation(ExtractParams parameters, ConnectionInfo connInfo) : bas
...
+            if (version == null)
+            {
+                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Invalid version {0} passed. Version must be in the format x.x.x.x where x is a number", this.Parameters.ApplicationVersion));
```
pensivebrian: Don't we localize error message from the API?

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
kisantia: The comparison was true when the version was invalid, but I'll change it since it's more clear to have the Version.TryParse() in the if statement.

pensivebrian: Turns out version is not a value type  :)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
kisantia: A test to validate a valid version?

kisantia: I removed the integration test and added a unit test to validate the version parsing.

PR #790, opened 2019-03-27
Fix "Save to CSV" CodePages assembly version mismatch

kburtram:
Resolve https://github.com/Microsoft/azuredatastudio/issues/4392 and https://github.com/Microsoft/vscode-mssql/issues/1205.  It looks like ResourceProvider & Credentials are consuming one set of dependencies and ServiceLayer has another.  When the self-contained apps get combined into a single folder they're stomping on each other.

This change adds the same dependency to all projects and changes the app merge order so that ServiceLayer wins the conflicts.

### DIFF
--- a/build.json
+++ b/build.json
@@ -16,9 +16,9 @@
     "netcoreapp2.2"
   ],
   "MainProjects": [     
-    "Microsoft.SqlTools.ServiceLayer",
     "Microsoft.SqlTools.Credentials",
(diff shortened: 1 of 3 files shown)


kburtram:
I've verified this fixes the issues mentioned above.  This change only moves references and build scripts around, so I'm merging to get a function vscode-mssql build out.

PR #795, opened 2019-04-05
Add schema compare publish changes operation

kisantia:
This adds the schema compare publish changes operation that publishes the changes to the target and also localizes a few strings used for the DacFx operations.

This does not yet support include/exclude, so all the differences will be published. The public schema compare API only supports generating the script when the target is a database.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
@@ -84,13 +84,13 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCopmare
                             Differences = operation.Differences
                         });
                     }
-                    catch
+                    catch(Exception e)
                     {
                         await requestContext.SendResult(new SchemaCompareResult()
                         {
                             OperationId = operation != null ? operation.OperationId : null,
                             Success = false,
-                            ErrorMessage = operation.ErrorMessage,
+                            ErrorMessage = operation == null ? e.Message : operation.ErrorMessage,
                         });
                     }
                 });
@@ -128,12 +128,12 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCopmare
                     ErrorMessage = operation.ErrorMessage
                 });
             }
-            catch
+            catch (Exception e)
             {
                 await requestContext.SendResult(new ResultStatus()
                 {
                     Success = false,
-                    ErrorMessage = operation.ErrorMessage
+                    ErrorMessage = operation == null ? e.Message : operation.ErrorMessage,
                 });
             }
         }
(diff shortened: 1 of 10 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
```diff
@@ -117,7 +118,43 @@ public async Task HandleSchemaCompareGenerateScriptRequest(SchemaCompareGenerate
...
+                    Success = true,
+                    ErrorMessage = operation.ErrorMessage
+                });
+            }
+            catch
+            {
+                await requestContext.SendResult(new ResultStatus()
+                {
+                    Success = false,
+                    ErrorMessage = operation.ErrorMessage
```
udeeshagautam: Check if operation is null, if yes use exception message as error message?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -301,6 +301,117 @@ private async Task<Mock<RequestContext<SchemaCompareResult>>> SendAndValidateSch
...
+                    SourceEndpointInfo = sourceInfo,
+                    TargetEndpointInfo = targetInfo
+                };
+
+                SchemaCompareOperation schemaCompareOperation = new SchemaCompareOperation(schemaCompareParams, result.ConnectionInfo, result.ConnectionInfo);
+                schemaCompareOperation.Execute(TaskExecutionMode.Execute);
+
+                Assert.True(schemaCompareOperation.ComparisonResult.IsValid);
+                Assert.False(schemaCompareOperation.ComparisonResult.IsEqual);
+                Assert.NotNull(schemaCompareOperation.ComparisonResult.Differences);
```
udeeshagautam: Is it possible to add some more validation than differences not being null? for example just some table/column being present in diff or so...

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -301,6 +301,117 @@ private async Task<Mock<RequestContext<SchemaCompareResult>>> SendAndValidateSch
...
+                // update target
+                var publishChangesParams = new SchemaComparePublishChangesParams
+                {
+                    TargetDatabaseName = targetDb.DatabaseName,
+                    OperationId = schemaCompareOperation.OperationId,
+                };
+
+                SchemaComparePublishChangesOperation publishChangesOperation = new SchemaComparePublishChangesOperation(publishChangesParams, schemaCompareOperation.ComparisonResult);
+                publishChangesOperation.Execute(TaskExecutionMode.Execute);
+                Assert.True(publishChangesOperation.PublishResult.Success);
```
udeeshagautam: same here - is it possible to check that the changes really happened?

kisantia: added a second schema compare comparison after the publish to make sure they are equal and no differences.

PR #797, opened 2019-04-11
Proposed fix for : 3653 : Cancel not working even after disconnecting

udeeshagautam:
The issue in bug (https://github.com/Microsoft/azuredatastudio/issues/3653): Cancel not working even after disconnecting seems to be that if any internal error happens while before all batches are executed sqltoolsservice is not able to send back batch result summary or any error messages and hence the query seems to be executing forever. Here are the proposed changes to stop query execution if an exception has occurred before/during batch execution. 

**Before :** 
image

**After:** (with a fake error)
image

Please note that I havent added test yet - will add in the same PR once we have agreed on the fix.


udeeshagautam:
@anthonydresser - this is the PR regarding query executing after disconnect that we talked about - if you get a chance please take a look. Do you this this seems like the right (or at least partial) fix?

kburtram:
@udeeshagautam do we still need this PR?  Is it blocked on anything?

udeeshagautam:
@kburtram No its not blocked. I missed to complete. Merging from master for one final check and will complete.

PR #798, opened 2019-04-16
Feature/schemacompare options

udeeshagautam:
SQL tools service side of changes for Adding Options (and object types to exclude) for Schema compare. This is using same options as DacDeployOptions with same defaults.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptions.cs
@@ -1,0 +1,239 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlServer.Dac;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Text;
+
+namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
+{
+    /// <summary>
+    /// Class to define schema compare and publish options. 
+    /// Keeping the same defaults as DacFx
+    /// The default values here should also match the default values in ADS UX
+    /// </summary>
+    public class SchemaCompareOptions
+    {
+        public bool IgnoreTableOptions { get; set; }
+
+        [DefaultValue(true)]
+        public bool IgnoreSemicolonBetweenStatements { get; set; } = true;
+
+        [DefaultValue(true)]
+        public bool IgnoreRouteLifetime { get; set; } = true;
+
+        public bool IgnoreRoleMembership { get; set; } 
+
+        [DefaultValue(true)]
+        public bool IgnoreQuotedIdentifiers { get; set; } = true;
+
+        public bool IgnorePermissions { get; set; }
+
+        public bool IgnorePartitionSchemes { get; set; }
+
+        [DefaultValue(true)]
+        public bool IgnoreObjectPlacementOnPartitionScheme { get; set; } = true;
+
+        public bool IgnoreNotForReplication { get; set; }
+
+        [DefaultValue(true)]
+        public bool IgnoreLoginSids { get; set; } = true;
+
+        public bool IgnoreLockHintsOnIndexes { get; set; }
+
+        [DefaultValue(true)]
+        public bool IgnoreKeywordCasing { get; set; } = true;
+
+        [DefaultValue(true)]
+        public bool IgnoreIndexPadding { get; set; } = true;
+
+        public bool IgnoreIndexOptions { get; set; }
+
+        public bool IgnoreIncrement { get; set; }
+
+        public bool IgnoreIdentitySeed { get; set; }
+
+        public bool IgnoreUserSettingsObjects { get; set; }
+
+        [DefaultValue(true)]
+        public bool IgnoreFullTextCatalogFilePath { get; set; } = true;
+
+        [DefaultValue(true)]
+        public bool IgnoreWhitespace { get; set; } = true;
+
+        public bool IgnoreWithNocheckOnForeignKeys { get; set; }
+
+        [DefaultValue(true)]
+        public bool VerifyCollationCompatibility { get; set; } = true;
+
+        [DefaultValue(true)]
+        public bool UnmodifiableObjectWarnings { get; set; } = true;
+
+        public bool TreatVerificationErrorsAsWarnings { get; set; }
+
+        [DefaultValue(true)]
+        public bool ScriptRefreshModule { get; set; } = true;
+
+        [DefaultValue(true)]
+        public bool ScriptNewConstraintValidation { get; set; } = true;
+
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -84,8 +87,14 @@ public void Execute(TaskExecutionMode mode)
             {
                 SchemaCompareEndpoint sourceEndpoint = CreateSchemaCompareEndpoint(this.Parameters.SourceEndpointInfo, this.SourceConnectionString);
                 SchemaCompareEndpoint targetEndpoint = CreateSchemaCompareEndpoint(this.Parameters.TargetEndpointInfo, this.TargetConnectionString);
-
+                
                 SchemaComparison comparison = new SchemaComparison(sourceEndpoint, targetEndpoint);
+
+                if(this.Parameters.SchemaCompareOptions != null)
```
kisantia: nit: space after if

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -488,49 +482,6 @@ public async void SchemaCompareGenerateScriptDacpacToDatabase()
...
         {
             Assert.NotNull(await SendAndValidateSchemaComparePublishChangesRequestDacpacToDatabase());
-        }
-
-
-        /// <summary>
-        /// Verify the schema compare publish changes request comparing a database to a database
-        /// </summary>
-        [Fact]
-        public async void SchemaComparePublishChangesDatabaseToDatabase()
```
kisantia: Was this deleted on accident?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
```diff
@@ -0,0 +1,376 @@
...
+END
+";
+        private SchemaCompareOptions GetIgnoreColumnOptions()
+        {
+            var options = new SchemaCompareOptions();
+            options.IgnoreColumnOrder = true;
+            return options;
+        }
+
+        private SchemaCompareOptions GetExcludeTableValuedFucntionOptions()
```
kisantia: nit: typo

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptions.cs
```diff
@@ -0,0 +1,207 @@
...
+using System.Text;
+
+namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
+{
+    /// <summary>
+    /// Class to define schema compare and publish options. 
+    /// Keeping the order and defaults same as DacFx
+    /// The default values here should also match the default values in ADS UX
+    /// </summary>
+    public class SchemaCompareOptions
```
pensivebrian: Could we make the default constructor initialize the defaults using reflection from the DacDeployOptions? Currently, we are already in a situation where defaults are different between the API, sqlpackage, and SSMS, and I don't want to make the situation get worse. I think the API should be the source of truth.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptions.cs
```diff
@@ -0,0 +1,207 @@
...
+    {
```
pensivebrian: Can we add a test the compare the properties from DacDeployOptions and SchemaCompareOptions, and fails when a DacDeployOptions is not present in SchemaCompareOptions? This we, can catch when properties are added when we pull in an updated nuget package.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptions.cs
pensivebrian: Do we need the SchemaCompareOptions type? Could we reuse the existing options type defined for the publish/script tools service operations?

udeeshagautam: Yes that sounds good.. will try

udeeshagautam: @pensivebrian not sure which ones you are referring to. We did not pass any options so far to Schema Compare operations, adding this inside the existing Schema compare params only. Please let me know if I am missing something here...

pensivebrian: We had talked about exposing the deployment options in the dac deployment wizards, so it's probably not there - my bad. But if we do in the future, I think it makes sense to use the same type, so perhaps we should rename to this type to DeploymentOptions ?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
udeeshagautam: Sorry my mistake in merge.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptions.cs
udeeshagautam: Done

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptions.cs
udeeshagautam: Sure I can do that (just wanted to double check to confirm - it wouldn't be confusing right?)

pensivebrian: Long term, are we going to have to deal with other schema compare options? For example, when we expose a checkbox to ignore a change - those options wouldn't make sense on a DeploymentOptions type. If that's the case, maybe it make more sense to make DeploymentOptions is own type which is a property of the SchemaCompareOptions type?

udeeshagautam: Currently - SchemaComapreOption or DacDeployOptions both - have two types of things 1) the options to ignore/verify and so on and 2) excluded object types. But the individual diff or change excludes though checkboxes are not set through options - Schema Comparison Result has a separate API for those. 

So given the parity with DacDeployOptions - I am changing the name for now to DeploymentOptions as you suggested. We can revisit the name if it doesn't fit once we share the extension side of changes :)

PR #801, opened 2019-04-25
update smo to 150 GA

shueybubbles:
this has a fix for SqlParser to reduce memory and cpu usage when it detects invalid TSQL syntax


shueybubbles:
are the code coverage deltas anything to be concerned with?

PR #802, opened 2019-04-26
Fixed - Invoke-SqlCmd -Variable does not match the behavior of sqlcmd.exe -v

NiranjanVirtuosity:
"-Variable" functionality is working unexpected, and here's the fix. 

TFS: http://sqlbuvsts01:8080/Main/SQL%20Server/_workitems#id=12817543&triage=true&_a=edit

### DIFF
--- a/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
@@ -103,7 +103,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
 
         private void CheckVariableReferences(Token token)
         {
-            // AddVariableReferences will raise error if reference is not constructed correctly 
+            // AddVariableReferences will raise error if reference is not constructed correctly
             AddVariableReferences(token, 0, null);
         }
 
@@ -126,7 +126,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                 if (state == 0 && ch == '$')
                 {
                     state = 1;
-                } 
+                }
                 else if (state == 1)
                 {
                     if (ch == '(')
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
```diff
@@ -573,15 +582,10 @@ internal string ResolveVariables(Token inputToken, int offset, List<VariableRefe
                     offset: reference.Start - offset,
                     filename: inputToken.Filename);
                 string value = variableResolver.GetVariable(variablePos.Value, reference.VariableName);
-                if (value == null)
+                if (value == null || ThrowOnUnresolvedVariable == true /* Undefined variable */)
```
kburtram: the previous code was an `&&` conditional test.  is there a reason this should be an `||` now?

NiranjanVirtuosity: The issue is "-variable" parameter functionality is not working as expected.

"ThrowOnUnresolvedVariable" is set to true only on ConfigureBatchParser() in ExecutionEngine.cs

When I pass powershell command "invoke-sqlcmd -query 'select $(a)' -Variable 'b=10'", ConfigureBatchParser() method is not called. So the "ThrowOnUnresolvedVariable" is not set and the '&&' condition fails.

Instead of getting error I am getting a result with empty data.

kburtram: it seems like the `ThrowOnUnresolvedVariable` flag no longer makes sense if we write the condition this way.  do we know why the code was originally written to only throw if both conditions are true?  i.e. in what conditions does value equal null and we don't want to throw.  clearly that was a case the original code author anticipated, inferred by the existence of `ThrowOnUnresolvedVariable`.  Basically I want to make sure we're not now throwing in situations that we "by design" didn't previously.

NiranjanVirtuosity: This is no longer needed, we made the changes in Poweshell itself. So am closing this PR.

NiranjanVirtuosity:
This is no longer needed, we made the changes in Powershell itself. So am closing this PR.

PR #804, opened 2019-05-01
Fix Dashboard OS version script

abist:
Fixes https://github.com/Microsoft/azuredatastudio/issues/2793

Before -

image

Now -

afterJPG

after2

### DIFF
--- a/src/Microsoft.SqlTools.CoreServices/Connection/ReliableConnection/SqlConnectionHelperScripts.cs
+++ b/src/Microsoft.SqlTools.CoreServices/Connection/ReliableConnection/SqlConnectionHelperScripts.cs
@@ -46,6 +46,6 @@ IF (@filepath IS NULL)
 SELECT @filepath AS FilePath
 ";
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.CoreServices/Connection/ReliableConnection/SqlConnectionHelperScripts.cs
```diff
@@ -46,6 +46,6 @@ @rc 			int
 SELECT @filepath AS FilePath
 ";
 
-    public const string GetOsVersion = @"SELECT OSVersion = RIGHT(@@version, LEN(@@version)- 3 -charindex (' ON ', @@version))";
+    public const string GetOsVersion = @"SELECT OSVersion = RIGHT(@@version, LEN(@@version)- 3 -charindex (' on ', @@version))";
```
kburtram: is there a way to write this so it's case-insensitive?  Perhaps adding a `tolower` call somewhere so we can be sure this condition won't start failing in the future?

abist: done

PR #806, opened 2019-05-07
 Invoke-SqlCmd handles ":!!if" in inconsistent way (PS5 vs PS6)

NiranjanVirtuosity:
Fix: 

- Set wordBounday to false for "!!" in Lexer.cs

### DIFF
--- a/test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
+++ b/test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
@@ -3,17 +3,17 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-using System;
-using System.Data.SqlClient;
-using System.Globalization;
-using System.IO;
-using System.Text;
 using Microsoft.SqlTools.ManagedBatchParser.IntegrationTests.TSQLExecutionEngine;
 using Microsoft.SqlTools.ManagedBatchParser.IntegrationTests.Utility;
 using Microsoft.SqlTools.ServiceLayer.BatchParser;
 using Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode;
 using Microsoft.SqlTools.ServiceLayer.Test.Common;
 using Microsoft.SqlTools.ServiceLayer.Test.Common.Baselined;
+using System;
+using System.Data.SqlClient;
+using System.Globalization;
+using System.IO;
+using System.Text;
 using Xunit;
 
 namespace Microsoft.SqlTools.ManagedBatchParser.UnitTests.BatchParser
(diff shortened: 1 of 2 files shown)


Matteo-T:
Is the only change to parse "!!" with "wordboundary = false"? If so, the change makes sense to me.

Please add tests!

kburtram:
@NiranjanVirtuosity please ping me once you've addressed @Matteo-T feedback and I can merge this for you.

NiranjanVirtuosity:
> @NiranjanVirtuosity please ping me once you've addressed @Matteo-T feedback and I can merge this for you.

yes sure, Sorry for the dealy.

NiranjanVirtuosity:
> Please, add tests!

I have added tests.

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
```diff
@@ -232,15 +230,52 @@ public void VerifyLexerSetState()
...
-
                 executionResult = ScriptExecutionResult.Failure;
-
             }
             //  we doesn't expect any exception or testCase failures
             Assert.Equal<ScriptExecutionResult>(ScriptExecutionResult.Success, executionResult);
-             
+        }
+
+        // This test case is to verify that, Powershell's Invoke-SqlCmd handles ":!!if" in an inconsistent way
```
kburtram: Seems like the comment should say `in a consistent way` instead of `inconsistent way`.  But maybe I'm reading the text wrong.

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
```diff
@@ -232,15 +230,52 @@ public void VerifyLexerSetState()
...
+        // Inconsistent way means, instead of throwing an exception as "Command Execute is not supported." it was throwing "Incorrect syntax near ':'."
+        [Fact]
+        public void VerifySqlCmdExecute()
```
Matteo-T: This is a really terrible name: it is not descriptive at all of that the test does.

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
```diff
@@ -232,15 +230,52 @@ public void VerifyLexerSetState()
...
                 executionResult = ScriptExecutionResult.Failure;
-
             }
             //  we doesn't expect any exception or testCase failures
             Assert.Equal<ScriptExecutionResult>(ScriptExecutionResult.Success, executionResult);
-             
+        }
+
+        // This test case is to verify that, Powershell's Invoke-SqlCmd handles ":!!if" in an inconsistent way
+        // Inconsistent way means, instead of throwing an exception as "Command Execute is not supported." it was throwing "Incorrect syntax near ':'."
```
Matteo-T: I would not try to explain what "inconsistent" mean: just don't use that word - 'cuz if you don't provide the context is  not easy to understand. Just say something like: "Verify that lexing/parsing !!if is consistent with the behavior of sqlcmd.exe, i.e. a space is not needed after an execute token ("!!").

This also shows that you are missing 1 test: they case where the is a space between "!!" and "if".

PR #811, opened 2019-05-14
Fix for Full/Log/Differential backup #5370

udeeshagautam:
Fix for https://github.com/microsoft/azuredatastudio/issues/5370 
The properties for backup type were not set correctly and hence not propagating to actual backup operation. Added a new test for the same and enabled few old tests which seemed consistent. Not re-enabling file browser ones since they will fail on Linux.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
@@ -75,57 +75,73 @@ CREATE CERTIFICATE {1} WITH SUBJECT = 'Backup Encryption Certificate'; ";
         public void CreateBackupTest()
         {
             DisasterRecoveryService service = new DisasterRecoveryService();
-            string databaseName = "testbackup_" + new Random().Next(10000000, 99999999);
-            SqlTestDb testDb = SqlTestDb.CreateNew(TestServerType.OnPrem, false, databaseName);
-            var liveConnection = LiveConnectionHelper.InitLiveConnectionInfo(databaseName);
-            DatabaseTaskHelper helper = AdminService.CreateDatabaseTaskHelper(liveConnection.ConnectionInfo, databaseExists: true);
-            SqlConnection sqlConn = ConnectionService.OpenSqlConnection(liveConnection.ConnectionInfo);
+            string databaseName = "SqlToolsService_TestBackup_" + new Random().Next(10000000, 99999999);
+            SqlTestDb testDb = SqlTestDb.CreateNew(TestServerType.OnPrem, false, databaseName);            
 
-            string backupPath = GetDefaultBackupFullPath(service, databaseName, helper.DataContainer, sqlConn);
+            try
+            {
+                var liveConnection = LiveConnectionHelper.InitLiveConnectionInfo(databaseName);
+                DatabaseTaskHelper helper = AdminService.CreateDatabaseTaskHelper(liveConnection.ConnectionInfo, databaseExists: true);
+                SqlConnection sqlConn = ConnectionService.OpenSqlConnection(liveConnection.ConnectionInfo);
 
-            BackupInfo backupInfo = CreateDefaultBackupInfo(databaseName,
-                BackupType.Full,
-                new List<string>() { backupPath },
-                new Dictionary<string, int>() { { backupPath, (int)DeviceType.File } });
-            BackupOperation backupOperation = CreateBackupOperation(service, liveConnection.ConnectionInfo.OwnerUri, backupInfo, helper.DataContainer, sqlConn);
+                string backupPath = GetDefaultBackupFullPath(service, databaseName, helper.DataContainer, sqlConn);
 
-            // Backup the database
-            service.PerformBackup(backupOperation);
+                BackupInfo backupInfo = CreateDefaultBackupInfo(databaseName,
+                    BackupType.Full,
+                    new List<string>() { backupPath },
+                    new Dictionary<string, int>() { { backupPath, (int)DeviceType.File } });
+                BackupOperation backupOperation = CreateBackupOperation(service, liveConnection.ConnectionInfo.OwnerUri, backupInfo, helper.DataContainer, sqlConn);
 
-            VerifyAndCleanBackup(backupPath);
-            testDb.Cleanup();
+                // Backup the database
+                service.PerformBackup(backupOperation);
+
+                VerifyAndCleanBackup(backupPath);
+                sqlConn.Close();
+            }
+            finally
+            {
+                testDb.Cleanup();
+            }
         }
 
         [Fact]
         public void ScriptBackupTest()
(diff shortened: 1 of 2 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
```diff
@@ -393,6 +431,36 @@ private bool ContainsFileInTheFolder(FileTreeNode folderNode, string filePath)
             }
             return false;
         }
+
+        private string GenerateScriptForBackupType(BackupType backupType, string databaseName)
+        {
+            DisasterRecoveryService service = new DisasterRecoveryService();
+            var liveConnection = LiveConnectionHelper.InitLiveConnectionInfo(databaseName);
+            DatabaseTaskHelper helper = AdminService.CreateDatabaseTaskHelper(liveConnection.ConnectionInfo, databaseExists: true);
+            SqlConnection sqlConn = ConnectionService.OpenSqlConnection(liveConnection.ConnectionInfo);
```
abist: We should put this in a `using` or close the connection later

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
```diff
@@ -232,6 +232,44 @@ public void ScriptBackupWithAdvancedOptionsTest()
...
+            script = GenerateScriptForBackupType(BackupType.Differential, databaseName);
+
+            // Validate differential backup script 
+            Assert.Contains("BACKUP DATABASE", script, StringComparison.OrdinalIgnoreCase);
+            Assert.DoesNotContain("BACKUP LOG", script, StringComparison.OrdinalIgnoreCase);
+            Assert.Contains("WITH  DIFFERENTIAL", script, StringComparison.OrdinalIgnoreCase);
+            
+            // Clean up the database
+            Console.WriteLine("Clean up database..");
+            testDb.Cleanup();
```
Charles-Gagnon: This isn't going to execute if there's an assertion failure - make sure we always clean up regardless

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
```diff
@@ -393,6 +431,36 @@ private bool ContainsFileInTheFolder(FileTreeNode folderNode, string filePath)
...
+            backupInfoLog.MediaDescription = "backup test";
+            BackupOperation backupOperation = CreateBackupOperation(service, liveConnection.ConnectionInfo.OwnerUri, backupInfoLog, helper.DataContainer, sqlConn);
+
+            // Generate Script
+            Console.WriteLine("Generate script for backup operation..");
+            service.ScriptBackup(backupOperation);
+            string script = backupOperation.ScriptContent;
+
+            // There shouldnt be any backup file created
+            Assert.True(!File.Exists(backupPath));
```
Charles-Gagnon: Give this a descriptive message if it does exist saying what the problem is

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
```diff
@@ -232,6 +232,44 @@ public void ScriptBackupWithAdvancedOptionsTest()
             testDb.Cleanup();
         }
 
+        /// <summary>
+        /// Test the correct script generation for different backup action types
+        /// </summary>
+        [Fact]
+        public void ScriptBackupWithDifferentActionTypesTest()
+        {
+            string databaseName = "testbackup_" + new Random().Next(10000000, 99999999);
```
Charles-Gagnon: Suggest giving this a more descriptive name (SqlToolsService_TestBackup_**** or something) so it's really easy to identify where it's coming from

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
```diff
@@ -232,6 +232,44 @@ public void ScriptBackupWithAdvancedOptionsTest()
...
+            // Validate Full backup script 
+            Assert.Contains("BACKUP DATABASE", script, StringComparison.OrdinalIgnoreCase);
+            Assert.DoesNotContain("BACKUP LOG", script, StringComparison.OrdinalIgnoreCase);
+            Assert.DoesNotContain("DIFFERENTIAL", script, StringComparison.OrdinalIgnoreCase);
+            
+            // Create log backup script
+            script = GenerateScriptForBackupType(BackupType.TransactionLog, databaseName);
+            
+            // Validate Log backup script 
+            Assert.Contains("BACKUP LOG", script, StringComparison.OrdinalIgnoreCase);
```
Charles-Gagnon: If this using the VS Assert? If so I'd suggest adding in a descriptive message saying what the problem is - the VS default messages are quite unhelpful most of the time

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
udeeshagautam: Good point,copied from existing tests and missed this.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
udeeshagautam: Don't want to elongate name unless required. there aren't any other backup tests which do this - so seems like backup should do.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
udeeshagautam: Doesnt have a message parameter and gives good enough message so leaving as is.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
Charles-Gagnon: We really shouldn't be concerned about length here. And my issue is that often I'll go on a test server and see a bunch of databases and have no idea where they're from. TestBackup is not descriptive enough to know that it's coming from the SqlToolsService tests. Being able to tell immediately where they're from will help save time if something goes wrong (like failing to clean up DBs)

udeeshagautam: sure, will change

PR #812, opened 2019-05-15
Query execution settings support

kburtram:


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
```diff
@@ -0,0 +1,231 @@
...
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage;
+using Microsoft.SqlTools.ServiceLayer.SqlContext;
+using Microsoft.SqlTools.ServiceLayer.Workspace;
+using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Hosting;
+using Microsoft.SqlTools.Utility;
+using System.Diagnostics;
+using System.Xml.Serialization;
```
Charles-Gagnon: Do you really need all these?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        private QueryExecutionSettings settings;
+
+        public QuerySettingsHelper(QueryExecutionSettings settings)
+        {
+            this.settings = settings;
+        }
+
+        [XmlIgnore]
+        public string SetNoCountString
```
Charles-Gagnon: Why these attributes? Are you planning on having this be serialized at all anyways?

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -108,8 +235,389 @@ public ExecutionPlanOptions ExecutionPlanOptions
...
+        }
+
+        /// <summary>
+        /// Set transaction isolation level
+        /// </summary>
+        public string TransactionIsolationLevel
+        {
+            get
+            {
+                return GetOptionValue<string>("transactionIsolationLevel");
```
Charles-Gagnon: Missing default?

Charles-Gagnon:
>         public void Update(QueryExecutionSettings newSettings)

Add new options to this as well?

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs:631 in 0faff77. [](commit_id = 0faff7751540a0a952592c56ccca07848e34e315, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/Workspace/WorkspaceService.cs
```diff
@@ -263,7 +265,7 @@ public void RegisterTextDocOpenCallback(TextDocOpenCallback task)
                 }
```
Charles-Gagnon: Do we need to do this anymore if we're passing the URI in directly? When would openedFile be null anyways?

Charles-Gagnon:
>             Assert.Equal(openedFile, closedFile);

Add assert for URI being the same as well?

---
Refers to: test/Microsoft.SqlTools.ServiceLayer.UnitTests/Workspace/WorkspaceTests.cs:57 in 0faff77. [](commit_id = 0faff7751540a0a952592c56ccca07848e34e315, deletion_comment = False)

Charles-Gagnon:
>         /// <param name="scriptFile"></param>

nit: add xml doc for new param

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs:548 in 0faff77. [](commit_id = 0faff7751540a0a952592c56ccca07848e34e315, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
```diff
@@ -0,0 +1,231 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
+{
+    /// <summary>
+    /// Service for executing queries
+    /// </summary>
+    public class QuerySettingsHelper
+    {
+        //strings for various "SET <option> ON/OFF" statements
+        private static string s_On = "ON";
```
Charles-Gagnon: nit: private static readonly

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
kburtram: this class is copy\paste from SSMS so I left it basically as implemented there.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
kburtram: Done.  Copy\paste issue.  Unfortunately VS Code doesn't have "Remove unneeded using" like VS, so you have to manually figure out which ones you need.

Charles-Gagnon: I poked around a bit and turns out there is - it's just hidden under a random keybinding. ctrl + . will give you that option :

image

PR #813, opened 2019-05-17
Open schema compare generated script

kisantia:
This changes the schema compare generate script operation to open the generated script in ADS as a temp file instead of saving it to the file path passed as a parameter.

A separate PR for ADS(https://github.com/microsoft/azuredatastudio/pull/5515) removes the file save as dialog and passes the target server and database names to show in the task viewlet instead of the file name.

openScript2

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareGenerateScriptOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareGenerateScriptOperation.cs
@@ -38,16 +38,23 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
 
         public SchemaComparisonResult ComparisonResult { get; set; }
 
+        public SchemaCompareScriptGenerationResult ScriptGenerationResult { get; set; }
+
         public SchemaCompareGenerateScriptOperation(SchemaCompareGenerateScriptParams parameters, SchemaComparisonResult comparisonResult)
         {
             Validate.IsNotNull("parameters", parameters);
-            Validate.IsNotNull("scriptFilePath", parameters.ScriptFilePath);
             this.Parameters = parameters;
             Validate.IsNotNull("comparisonResult", comparisonResult);
             this.ComparisonResult = comparisonResult;
         }
 
         public void Execute(TaskExecutionMode mode)
+        {
+            Execute();
+            AddScriptToTask();
+        }
+
+        public void Execute()
         {
             if (this.CancellationToken.IsCancellationRequested)
             {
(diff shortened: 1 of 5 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
```diff
@@ -292,19 +290,15 @@ private async Task<Mock<RequestContext<SchemaCompareResult>>> SendAndValidateSch
...
 
                 SchemaCompareGenerateScriptOperation generateScriptOperation2 = new SchemaCompareGenerateScriptOperation(generateScriptParams2, schemaCompareOperation2.ComparisonResult);
-                generateScriptOperation2.Execute(TaskExecutionMode.Execute);
+                generateScriptOperation2.Execute();
 
-                // validate script
-                var filePath2 = Path.Combine(folderPath, string.Concat(sourceDb.DatabaseName, "_", "Update.publish2.sql"));
-                Assert.True(File.Exists(filePath2) && !string.IsNullOrEmpty(File.ReadAllText(filePath2)), "Should have differences differences");
+                // validate script generation succeeded
+                Assert.True(generateScriptOperation2.ScriptGenerationResult.Success);
```
udeeshagautam: can we add little validation for content... may be that 1) content isn't empty 2) has a basic substring that we expect..?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareGenerateScriptOperation.cs
```diff
@@ -56,15 +63,7 @@ public void Execute(TaskExecutionMode mode)
 
             try
             {
-                SchemaCompareScriptGenerationResult result = this.ComparisonResult.GenerateScript(this.Parameters.TargetDatabaseName);
-                File.WriteAllText(this.Parameters.ScriptFilePath, result.Script);
-
-                if (!string.IsNullOrEmpty(result.MasterScript))
-                {
-                    // master script is only used if the target is Azure SQL db and the script contains all operations that must be done against the master database
-                    string masterScriptPath = Path.Combine(Path.GetDirectoryName(this.Parameters.ScriptFilePath), string.Concat("master_", Path.GetFileName(this.Parameters.ScriptFilePath)));
```
udeeshagautam: we don't need this behavior anymore?

udeeshagautam:
@kisantia just noticed that the coverage decrease is significant (-15%).. can you please double check if its because some tests started failing in SC after the change

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
kisantia: added back check if script is null or empty. I didn't add a check for a substring since the two tests that use this changed different objects.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareGenerateScriptOperation.cs
```diff
@@ -38,16 +38,23 @@ class SchemaCompareGenerateScriptOperation : ITaskOperation
...
             this.ComparisonResult = comparisonResult;
         }
 
         public void Execute(TaskExecutionMode mode)
+        {
+            Execute();
+            AddScriptToTask();
+        }
+
+        public void Execute()
```
pensivebrian: It's strange to see two executes methods that are public. Any reason we can't have one public?

kisantia: This Execute() needs to be public because it's what the tests use. The other one Execute(TaskExecutionMode mode) needs to be public because that's what is being used by the service.

pensivebrian: Why can't the tests use Execute(TaskExecutionMode mode)?

kisantia: The AddScriptToTask() had to separated because the tests don't create a SqlTask, which is why Execute(TaskExecutionMode mode) calls Execute() then AddScriptToTask().

pensivebrian: Please write comments for issues like this.

kisantia: I put a comment about this above the AddScriptToTask() method

pensivebrian: I missed that comment. I think it's in the wrong spot to explain two public execute methods.

In the future, when you have to introduce workarounds like this, I'd try taking a step back and reconsider the design. For example, instead of taking on another service as a dependency, in this case SqlTask service, you could define and event ScriptCompleted. Then the service host could subscribe to the event and wire up the services. These keeps things decoupled.

PR #816, opened 2019-05-22
Change to scope file tree to the known path if root folders dont return correctly

udeeshagautam:
Aims to fix https://github.com/microsoft/azuredatastudio/issues/4767 : 
Issue : File browser for back up result does not load when the root folders return are not valid and hence the expand path i.e. the default backup path can not be found in those. 
Fix: workaround to scope the file browser tree to the default backup path

Following is how it looks when root not found. (Please note that if there is nothing present in the default folder it will show as empty)

backup folder

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
@@ -160,7 +160,19 @@ namespace Microsoft.SqlTools.ServiceLayer.FileBrowser
                         {
                             this.fileTree.SelectedNode = lastNode;
                         }
-                        throw new FileBrowserException(string.Format(SR.InvalidPathError, this.expandPath));
+
+                        // If root folders returned from sys.dm_os_enumerate_filesystem don't match the folders present, expandPath might not be found
+                        // Bug: https://github.com/microsoft/azuredatastudio/issues/4767
+                        // Workaround : scope file browser tree to expandPath explicitly 
+                        try
+                        {
+                            this.ScopeFileTreeToPath(expandPath);
+                            return;
+                        }
+                        catch
+                        {
+                            throw new FileBrowserException(string.Format(SR.InvalidPathError, this.expandPath));
+                        }
                     }
                 }
 
@@ -204,7 +216,7 @@ namespace Microsoft.SqlTools.ServiceLayer.FileBrowser
         }
 
         public List<FileTreeNode> GetChildren(string filePath)
-        { 
+        {
             List<FileTreeNode> children = new List<FileTreeNode>();
             foreach (var file in EnumerateFilesInFolder(Enumerator, connection, filePath))
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
```diff
@@ -248,6 +260,26 @@ internal bool FilterFile(string fileName, string[] masks)
...
+            this.fileTree = new FileTree();
+            FileTreeNode node = new FileTreeNode()
+            {
+                Name = expandPath,
+                FullPath = expandPath,
+                IsExpanded = true
+            };
+
+            this.fileTree.RootNode.AddChildNode(node);
+            node.Children = node.Children = this.GetChildren(node.FullPath);
```
shueybubbles: duplicate assignment of node.Children

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
```diff
@@ -248,6 +260,26 @@ internal bool FilterFile(string fileName, string[] masks)
...
+        {
+            this.fileTree = new FileTree();
+            FileTreeNode node = new FileTreeNode()
+            {
+                Name = expandPath,
+                FullPath = expandPath,
+                IsExpanded = true
+            };
+
+            this.fileTree.RootNode.AddChildNode(node);
```
shueybubbles: is this adding '/var/opt/mssql/data' as a child node of the root in addition to any other nodes that came back from enumerating the root?

shueybubbles:
Can you add a screenshot?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
udeeshagautam: /var/opt/mssql/data is the this.fileTree.RootNode. This is adding the bak files and any child folders to it.

kburtram:
@udeeshagautam do you know what happened with the code coverage?  perhaps some tests are failing?

udeeshagautam:
@kburtram yes, the last run had -1% decline.. so I re-ran and it went -4.5%.
I see quite a few "Login failed for user 'sa'." - not related to this area as such but will take a look before merging...

udeeshagautam:
@shueybubbles @chlafreniere would you like to take a look or I should close?

src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
```diff
@@ -160,7 +160,19 @@ public void ExpandSelectedNode(string expandPath)
...
+
+                        // If root folders returned from sys.dm_os_enumerate_filesystem don't match the folders present, expandPath might not be found
+                        // Bug: https://github.com/microsoft/azuredatastudio/issues/4767
+                        // Workaround : scope file browser tree to expandPath explicitly 
+                        try
+                        {
+                            this.ScopeFileTreeToPath(expandPath);
+                            return;
+                        }
+                        catch
```
ranasaria: What is the type of Exception that we are replacing with the FileBrowserException? Should we wrap the exception caught as an inner exception of log the details of the exception that we are swallowing?

udeeshagautam: Yes that would be good idea, however in file browser service the result only seems to pick the message from exception as result.Message = ex.Message; so it wouldn't matter. I can take the change of enabling passing on the whole exception as a separate change overall.

PR #818, opened 2019-05-23
Adding some default values to match SSDT and other tools

udeeshagautam:
Adding defaults close to what is respected by schema compare Compare().

These values help defaults match SSDT and ADS DacFx extension. And help with issue : https://github.com/microsoft/azuredatastudio/issues/5599 where the schema compare result shows different value but doesn't apply those changes.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
@@ -208,7 +208,7 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
         {
             DacDeployOptions options = new DacDeployOptions();
 
-            // Adding these defaults to ensure behavior similarity with other tools.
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -207,6 +207,18 @@ public class DeploymentOptions
         public DeploymentOptions()
         {
             DacDeployOptions options = new DacDeployOptions();
+
+            // Adding these defaults to ensure behavior similarity with other tools.
```
kisantia: add that the dacfx wizard and ssms import/export wizards use these defaults

udeeshagautam: Hey @kisantia, you mean in comment? haha I had that and then deleted :P 
Will add comment.

kisantia: yup I meant in the comment :)

PR #819, opened 2019-05-31
Remove cancellable task which does not cancel anything

udeeshagautam:
Issue : https://github.com/microsoft/azuredatastudio/issues/5804
DacFx Api's does not allow cancel on generate script and apply operation.
But the default task created is in Task view has a context menu for cancel which is clickable and changes the UX to cancelled. This might create lot of confusion.

image

Reason : TaskInfo IsCancellable is set to true as long as task is has a "taskToCancel" object. This object is automatatically created based on Operation in current Create Task Call.

Fix: Setting the Task to cancel object to null explicitly since so that Iscancellable is set to false. 

**This is a temporary fix till DacFx Api changes are not in place for real cancellation.**


udeeshagautam:
holding it off since we are working on cancel operation.

PR #822, opened 2019-06-07
Open dacfx deploy script instead of saving to file path

kisantia:
This changes the DacFx generate deploy script operation to open the script after it's generated instead of saving it to the specified file path. 

Also changing the Schema Compare generate script operation to remove the two public Execute methods.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
@@ -57,19 +57,24 @@ CREATE TABLE [dbo].[table3]
             string folderPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "DacFxTest");
             Directory.CreateDirectory(folderPath);
 
-            var exportParams = new ExportParams
+            try
             {
-                DatabaseName = testdb.DatabaseName,
-                PackageFilePath = Path.Combine(folderPath, string.Format("{0}.bacpac", testdb.DatabaseName))
-            };
+                var exportParams = new ExportParams
+                {
+                    DatabaseName = testdb.DatabaseName,
+                    PackageFilePath = Path.Combine(folderPath, string.Format("{0}.bacpac", testdb.DatabaseName))
+                };
 
-            DacFxService service = new DacFxService();
-            ExportOperation operation = new ExportOperation(exportParams, result.ConnectionInfo);
-            service.PerformOperation(operation);
+                DacFxService service = new DacFxService();
+                ExportOperation operation = new ExportOperation(exportParams, result.ConnectionInfo);
+                service.PerformOperation(operation, TaskExecutionMode.Execute);
 
-            // cleanup
-            VerifyAndCleanup(exportParams.PackageFilePath);
-            testdb.Cleanup();
+                VerifyAndCleanup(exportParams.PackageFilePath);
+            }
+            finally
+            {
+                testdb.Cleanup();
+            }
 
             return requestContext;
         }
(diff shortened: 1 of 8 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -266,20 +268,21 @@ private async Task<Mock<RequestContext<DacFxResult>>> SendAndValidateGenerateDep
...
+
+            // Verify script was generated
+            Assert.NotEmpty(generateScriptOperation.Result.DatabaseScript);
+            Assert.Contains("CREATE TABLE", generateScriptOperation.Result.DatabaseScript);
 
             // cleanup
-            VerifyAndCleanup(generateScriptParams.ScriptFilePath);
             VerifyAndCleanup(extractParams.PackageFilePath);
             sourceDb.Cleanup();
             targetDb.Cleanup();
```
udeeshagautam: Not specifically for this change but please add source db and target db cleanup in a finally block so that its cleaned up even if something goes wrong...

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
```diff
@@ -266,7 +266,7 @@ private async Task<Mock<RequestContext<SchemaCompareResult>>> SendAndValidateSch
                 };
 
                 SchemaCompareGenerateScriptOperation generateScriptOperation1 = new SchemaCompareGenerateScriptOperation(generateScriptParams1, schemaCompareOperation1.ComparisonResult);
-                generateScriptOperation1.Execute();
+                generateScriptOperation1.Execute(TaskExecutionMode.Script);
```
udeeshagautam: for my understanding - Does the TaskExecutionMode.Script do anything different (I see it setting the task name - but anything else?)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -266,20 +268,21 @@ private async Task<Mock<RequestContext<DacFxResult>>> SendAndValidateGenerateDep
...
             {
                 PackageFilePath = extractParams.PackageFilePath,
-                DatabaseName = string.Concat(sourceDb.DatabaseName, "-deployed"),
-                ScriptFilePath = Path.Combine(folderPath, string.Concat(sourceDb.DatabaseName, "_", "UpgradeDACScript.sql"))
+                DatabaseName = targetDb.DatabaseName
             };
 
+            // Generate script for deploying source dacpac to target db
             GenerateDeployScriptOperation generateScriptOperation = new GenerateDeployScriptOperation(generateScriptParams, result.ConnectionInfo);
             service.PerformOperation(generateScriptOperation);
```
udeeshagautam: why no TaskExecutionMode.Script here?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
kisantia: This is how ADS knows to open the script after the task is finished

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
kisantia: It doesn't matter since the tests don't create s SqlTask, but I'll change this since it's unclear.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
kisantia: added try{} finally{} to all the dacfx tests

PR #825, opened 2019-06-10
Schema Compare open SCMP file

kisantia:
This takes a request to open an SCMP file and returns the source and target endpoint info, deployment options, and excluded elements. Some xml parsing still had to be done to get the included/excluded elements and because the source and target info of SchemaComparison are internal.

Will add deployment options checks to tests after Udeesha's changes get checked in since it has a helper to compare options.


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
```diff
@@ -0,0 +1,171 @@
...
+                endpointInfo.PackageFilePath = dacpacEndpoint.FilePath;
+            }
+            else
+            {
+                // need to parse xml to get connection string of database
+                XmlNodeList connectionBasedModelProviderNodes = this.scmpInfo.DocumentElement.SelectNodes("descendant::ConnectionBasedModelProvider");
+                string searchingFor = source ? "Source" : "Target";
+
+                foreach (XmlNode node in connectionBasedModelProviderNodes)
+                {
```
udeeshagautam: Please check for null connectionBasedModelProviderNodes before enumeration

I would also suggest a try catch so that in case parsing goes wrong we can send back endpoint info with at least database name.... and in ADS while doing the comparison user will have to reconnect

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
```diff
@@ -0,0 +1,171 @@
...
+                this.scmpInfo = new XmlDocument();
+                this.scmpInfo.Load(this.Parameters.filePath);
+
+                this.Result = new SchemaCompareOpenScmpResult()
+                {
+                    DeploymentOptions = new DeploymentOptions(compare.Options),
+                    Success = true,
+                    SourceEndpointInfo = this.GetEndpointInfo(true, compare.Source),
+                    TargetEndpointInfo = this.GetEndpointInfo(false, compare.Target),
+                    OriginalTargetName = this.GetOriginalTargetName(),
```
udeeshagautam: when we are parsing Target from scmp file - won't the target and original target return the same endpoint?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
```diff
@@ -227,6 +229,41 @@ await requestContext.SendResult(new SchemaCompareOptionsResult()
...
 
+        /// <summary>
+        /// Handles schema compare open SCMP request
+        /// </summary>
+        /// <returns></returns>
+        public async Task HandleSchemaCompareOpenScmpRequest(SchemaCompareOpenScmpParams parameters, RequestContext<SchemaCompareOpenScmpResult> requestContext)
+        {
+            try
+            {
+                Task schemaCompareTask = Task.Run(async () =>
```
udeeshagautam: Can you please use the currentTask object in class - that way when we test service call.. we can wait on that to ensure this completes before test exits

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -495,5 +496,140 @@ private void ValidateSchemaCompareScriptGenerationWithExcludeIncludeResults(Sche
...
+                SchemaCompareDatabaseEndpoint targetEndpoint = new SchemaCompareDatabaseEndpoint(targetDb.ConnectionString);
+                SchemaCompareOpenScmpResult result = this.CreateAndOpenScmp(sourceEndpoint, targetEndpoint, out filePath);
+
+                Assert.Equal(targetDb.DatabaseName, result.OriginalTargetName);
+                Assert.Equal(sourceDb.DatabaseName, result.SourceEndpointInfo.DatabaseName);
+                Assert.Equal(targetDb.DatabaseName, result.TargetEndpointInfo.DatabaseName);
+                Assert.Equal(SchemaCompareEndpointType.Database, result.SourceEndpointInfo.EndpointType);
+                Assert.Equal(SchemaCompareEndpointType.Database, result.TargetEndpointInfo.EndpointType);
+                Assert.Empty(result.ExcludedSourceElements);
+                Assert.Empty(result.ExcludedTargetElements);
```
udeeshagautam: Nit : can we combine some of the code to common functions in the three tests - they seems to be almost very similar other than dacpac/db distinction...

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
kisantia: If the scmp was saved after the source and target are switched, then original target will be the same as the source. The original target is needed since the ExcludedSourceElements and ExcludedTargetElements are based on the original source and target.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
kisantia: Sure, made this change in the latest changes

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -495,5 +496,117 @@ private void ValidateSchemaCompareScriptGenerationWithExcludeIncludeResults(Sche
...
             Assert.True(initialScript.Length == afterIncludeScript.Length, $"Changes should be same as inital since we included what we excluded, before {initialScript}, now {afterIncludeScript}");
         }
+
+        /// <summary>
+        /// Verify opening an scmp comparing two databases
+        /// </summary>
+        [Fact]
+        public async void SchemaCompareOpenScmpDatabaseToDatabaseRequest()
+        {
+            await CreateAndOpenScmp(SchemaCompareEndpointType.Database, SchemaCompareEndpointType.Database);
```
udeeshagautam: Doesn't have to be in this PR - We should have a validation for service call too i.e. the Handle* function (I have added a test for all service calls - we can add there).

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
```diff
@@ -0,0 +1,176 @@
...
+            // if the endpoint is a dacpac we don't need to parse the xml
+            SchemaCompareDacpacEndpoint dacpacEndpoint = endpoint as SchemaCompareDacpacEndpoint;
+            if (dacpacEndpoint != null)
+            {
+                endpointInfo.EndpointType = SchemaCompareEndpointType.Dacpac;
+                endpointInfo.PackageFilePath = dacpacEndpoint.FilePath;
+            }
+            else
+            {
+                // need to parse xml to get connection string of database
```
udeeshagautam: @pensivebrian should we make the connection string getter to be public in DacFx SchemaCompareDatabaseEndpoint - so that we don't have to do parsing here? What do you think? I am worried that parsing might need constant maintenance if things around this change in DacFx.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
kisantia: yes will add after your PR adding the test gets merged.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
```diff
@@ -0,0 +1,176 @@
...
+                    OriginalTargetName = this.GetOriginalTargetName(),
+                    OriginalTargetConnectionString = this.GetOriginalTargetConnectionString(),
+                    ExcludedSourceElements = this.GetExcludedElements(compare.ExcludedSourceObjects),
+                    ExcludedTargetElements = this.GetExcludedElements(compare.ExcludedTargetObjects)
+                };
+            }
+            catch (Exception e)
+            {
+                ErrorMessage = e.Message;
+                Logger.Write(TraceEventType.Error, string.Format("Schema compare open scmp operation failed with exception {0}", e.Message));
```
pensivebrian: We should log the exception object so that the stack trace and inner exceptions are included in addition to the message.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
```diff
@@ -0,0 +1,176 @@
...
+            SchemaCompareDacpacEndpoint dacpacEndpoint = endpoint as SchemaCompareDacpacEndpoint;
+            if (dacpacEndpoint != null)
+            {
+                endpointInfo.EndpointType = SchemaCompareEndpointType.Dacpac;
+                endpointInfo.PackageFilePath = dacpacEndpoint.FilePath;
+            }
+            else
+            {
+                // need to parse xml to get connection string of database
+                XmlNodeList connectionBasedModelProviderNodes = this.scmpInfo.DocumentElement.SelectNodes("descendant::ConnectionBasedModelProvider");
```
pensivebrian: Consider using the XML Linq APIs instead of the old XPath API for querying xml.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
pensivebrian: I don't have enough context to have an opinion. I'll leave it up to you two.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -232,7 +232,7 @@ public static string FormatScript(string script)
             return script;
         }
 
-        private static string GetName(string name)
+        public static string GetName(string name)
```
pensivebrian: Is the robust enough to handle databases with brackets in the name? I made the same comment in another PR.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
```diff
@@ -227,6 +232,41 @@ await requestContext.SendResult(new SchemaCompareOptionsResult()
...
+            try
+            {
+                CurrentSchemaCompareTask = Task.Run(async () =>
+                {
+                    SchemaCompareOpenScmpOperation operation = null;
+
+                    try
+                    {
+                        operation = new SchemaCompareOpenScmpOperation(parameters);
+                        operation.Execute(TaskExecutionMode.Execute);
```
pensivebrian: Why make async methods with the operation is sync? Shouldn't this be async?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        private List<SchemaCompareObjectId> GetExcludedElements(IList<SchemaComparisonExcludedObjectId> excludedObjects)
+        {
+            List<SchemaCompareObjectId> excludedElements = new List<SchemaCompareObjectId>();
+
+            foreach (SchemaComparisonExcludedObjectId entry in excludedObjects)
+            {
+                excludedElements.Add(new SchemaCompareObjectId()
+                {
+                    Name = SchemaCompareOperation.GetName(entry.Identifier.ToString()),
```
udeeshagautam: Changes to string.Join("." identifier.parts)
same reason as our discussion..

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOpenScmpRequest.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
+{
+    public class SchemaCompareObjectId
+    {
+        /// <summary>
+        /// Name to create object identifier
+        /// </summary>
+        public string Name;
```
udeeshagautam: string[] NameParts
to preserve the "." and "[" in name and avoid any string operations in sqltools service as we talked
(don't have to be in this PR)

kisantia: changed to NameParts

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
kisantia: not using this anymore

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOpenScmpOperation.cs
kisantia: switched to use the XML Linq APIs

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
pensivebrian: Hmmm. Blocking the json/rpc channel seems like a bad idea as no other request can be serviced, right? I guess the question is why makes this method async in the first place? But if this is existing pattern that's used elsewhere, then I guess it makes sense to be consistent.

PR #827, opened 2019-06-14
Agent - Fix last step action 

abist:
Fixes https://github.com/microsoft/azuredatastudio/issues/5896

Also turned strings to enums

Along with PR in ADS here - https://github.com/microsoft/azuredatastudio/pull/6034

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/JobStepData.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/JobStepData.cs
@@ -365,6 +365,12 @@ namespace Microsoft.SqlTools.ServiceLayer.Agent
                 }
                 return this.failureAction;
             }
+            set
+            {
+                if (this.failureAction != value) {
+                    this.failureAction = value;
+                }
+            }
         }
 
         public JobStepData FailStep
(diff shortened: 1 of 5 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Agent/AgentTestUtils.cs
```diff
@@ -30,7 +30,7 @@ public static class AgentTestUtils
                 Id = 1,
                 JobName = job.Name,
                 StepName = stepName,
-                SubSystem = "T-SQL",
+                SubSystem = SqlServer.Management.Smo.Agent.AgentSubSystem.TransactSql,
```
Charles-Gagnon: Would suggest adding a unit test that has a hardcoded list of all the currently known SMO enum types and verifying that the enum only contains those types so that we can catch newly added types ASAP (for both enums used here)

src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/JobStepData.cs
```diff
@@ -365,6 +365,12 @@ public StepCompletionAction FailureAction
                 }
                 return this.failureAction;
             }
+            set
+            {
+                if (this.failureAction != value) {
```
Charles-Gagnon: Seems a little silly to have this check but I guess if that's the pattern I don't have a huge issue with it...

PR #828, opened 2019-06-18
Change to pass Schema compare Operation id from ADS 

udeeshagautam:
This PR contains 

Changes for Schema compare operation id to be passed from ADS so that 
1) ADS knows the operation id to cancel if needed without having to wait for operation to return
2) We can have only one id per SC page so that only the open page's result is stored in memory not the previous compares. 
Test fix to add Cancel call in service call test

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
@@ -528,8 +528,9 @@ CREATE TABLE [dbo].[table3]
         [Fact]
         public async Task VerifySchemaCompareServiceCalls()
         {
-            string operationId = null;
+            string operationId = Guid.NewGuid().ToString();
             DiffEntry diffEntry = null;
+            bool cancelled = false;
             var connectionObject = SchemaCompareTestUtils.GetLiveAutoCompleteTestObjects();
 
             SqlTestDb sourceDb = await SqlTestDb.CreateNewAsync(TestServerType.OnPrem, false, null, SourceScript, "SchemaCompareSource");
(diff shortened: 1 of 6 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -970,18 +1003,33 @@ private void CreateAndValidateScmpFile(SchemaCompareEndpointInfo sourceInfo, Sch
...
-                return false;
+                Assert.True(diffResult.Differences == null, "Differences should be null after cancel");
+                Assert.True(diffResult.Success == false, "Result success forschema compare should be false after cancel");
+                diffEntry = null;
+                return true;
             }
+
+            diffEntry = diffResult.Differences.ElementAt(0);
+            Assert.True(diffResult.Success == true, "Result success is false for schema compare");
+            Assert.True(diffResult.Differences != null, "Schema comapre Differences should not be null");
```
kisantia: nit typo

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -970,18 +1003,33 @@ private void CreateAndValidateScmpFile(SchemaCompareEndpointInfo sourceInfo, Sch
...
-                operationId = diffResult.OperationId;
-                diffEntry = diffResult.Differences.ElementAt(0);
-                return (diffResult.Success == true && diffResult.Differences != null && diffResult.Differences.Count > 0);
-            }
-            catch
+            if (cancelled)
             {
-                return false;
+                Assert.True(diffResult.Differences == null, "Differences should be null after cancel");
+                Assert.True(diffResult.Success == false, "Result success forschema compare should be false after cancel");
```
kisantia: nit space between "forschema"

PR #829, opened 2019-06-18
Default to Server level context for all OE nodes

Charles-Gagnon:
Before : 

image

After : 

image


kburtram:
@Charles-Gagnon it may be worth noting this in the changelog so it's clear this change in behavior is intended.

kburtram:
@Charles-Gagnon it's probably worth determining the change in code coverage is unrelated.  I'd assume it is unless this is breaking a test or something.

Charles-Gagnon:
Yeah - the code coverage numbers are really sketchy. It's on my backlog to investigate - but I've seen the numbers fluctuate wildly for even simple changes (I've seen changes of up to 17%) so it's likely that same issue here since my changes shouldn't cause this kind of drop in numbers.

PR #831, opened 2019-06-26
FIx intellisense cache not refreshing

Charles-Gagnon:
Fixes https://github.com/microsoft/azuredatastudio/issues/4825. When refreshing we weren't clearing the cached ScriptParseInfo so it basically wasn't doing anything since we'd use that binding info next time an Intellisense query request was submitted.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -290,17 +290,17 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         /// <param name="requestContext"></param>
         /// <returns></returns>
         internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestContext<SyntaxParseResult> requestContext)
-        {            
-            await Task.Run(async () => 
+        {
+            await Task.Run(async () =>
             {
                 try
-                {   
+                {
                     ParseResult result = Parser.Parse(param.Query);
                     SyntaxParseResult syntaxResult = new SyntaxParseResult();
                     if (result != null && result.Errors.Count() == 0)
                     {
                         syntaxResult.Parseable = true;
-                    } else 
+                    } else
                     {
                         syntaxResult.Parseable = false;
                         string[] errorMessages = new string[result.Errors.Count()];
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -628,13 +628,16 @@ public async Task HandleDidChangeTextDocumentNotification(ScriptFile[] changedFi
...
                         ScriptParseInfo scriptInfo = GetScriptParseInfo(connInfo.OwnerUri, createIfNotExists: false);
-                        if (scriptInfo != null && scriptInfo.IsConnected && 
+                        if (scriptInfo != null && scriptInfo.IsConnected &&
                             Monitor.TryEnter(scriptInfo.BuildingMetadataLock, LanguageService.OnConnectionWaitTimeout))
                         {
                             try
                             {
                                 this.BindingQueue.AddConnectionContext(connInfo, featureName: "LanguageService", overwrite: true);
+                                RemoveScriptParseInfo(rebuildParams.OwnerUri);
+                                UpdateLanguageServiceOnConnection(connInfo).Wait();
```
Charles-Gagnon: Here's the only actual changes. I call UpdateLanguageServiceOnConnection to immediately rebuild the cache (it would do it the next time a request came in but this frontloads it so we can start that process early for slow connections)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -290,17 +290,17 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
         /// <param name="requestContext"></param>
         /// <returns></returns>
         internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestContext<SyntaxParseResult> requestContext)
-        {            
-            await Task.Run(async () => 
+        {
```
kburtram: it often nice to separate non-function formatting changes from functional changes into different PRs.  I see you called out the relevant changes, but it's hard to spot typos, etc.

Charles-Gagnon: Yeah sorry about that - I have an extension that deletes whitespace on save automatically so I don't notice the changes until I'm committing them in usually and undoing them would be a pain. If there were more actual changes I'd do that to make it clearer but since this was just 2 lines I figured it'd be fine to just call it out like I did.

PR #832, opened 2019-07-02
Materialize notebook in powershell

aasimkhan30:
The script reads a string JSON notebook template, executes it and returns a string materialized JSON notebook. To be later used to execute notebook jobs

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
+++ b/src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
@@ -1,0 +1,211 @@
+param(
+    [System.String]
+    $TemplateNotebook = $(throw New-Object System.Exception "TemplateNotebook cannot be null"),
+
+    [System.String]
+    $DatabaseInstance,
+
+    [System.String]
+    $DatabaseServer,
+
+    [System.String]
+    $DatabaseUsername,
+
+    [System.String]
+    $DatabasePassword
+)
+<#
+    Parses a DataTable returned by Invoke-SQL Command into a Jupyter Notebook output cell.
+#>
+function ParseTableToNotebookOutput {
+    param (
+        [System.Data.DataTable]
+        $DataTable,
+
+        [int]
+        $CellExecutionCount
+    )
+    $TableHTMLText = "<table>"
+    $TableSchemaFeilds = @()
+    $TableHTMLText += "<tr>"
+    foreach ($ColumnName in $DataTable.Columns) {
+        $TableSchemaFeilds += @(@{name = $ColumnName.toString() })
+        $TableHTMLText += "<th>" + $ColumnName.toString() + "</th>"
+    }
+    $TableHTMLText += "</tr>"
+    $TableSchema = @{ }
+    $TableSchema["fields"] = $TableSchemaFeilds
+
+    $TableDataRows = @()
+    foreach ($Row in $DataTable) {
+        $TableDataRow = [ordered]@{ }
+        $TableHTMLText += "<tr>"
+        $i = 0
+        foreach ($Cell in $Row.ItemArray) {
+            $TableDataRow[$i.ToString()] = $Cell.toString()
+            $TableHTMLText += "<td>" + $Cell.toString() + "</td>"
+            $i++
+        }
+        $TableHTMLText += "</tr>"
+        $TableDataRows += $TableDataRow
+    }
+
+    $TableDataResource = @{ }
+    $TableDataResource["schema"] = $TableSchema
+    $TableDataResource["data"] = $TableDataRows
+    $TableData = @{ }
+    $TableData["application/vnd.dataresource+json"] = $TableDataResource
+    $TableData["text/html"] = $TableHTMLText
+    $TableOutput = @{ }
+    $TableOutput["output_type"] = "execute_result"
+    $TableOutput["data"] = $TableData
+    $TableOutput["metadata"] = @{ }
+    $TableOutput["execution_count"] = $CellExecutionCount
+    return $TableOutput
+}
+
+
+function ParseQueryErrorToNotebookOutput {
+    param (
+        $QueryError
+    )
+    <#
+    Following the current syntax of errors in T-SQL notebooks from ADS
+    #>
+    $ErrorString = "Msg " + $QueryError.Exception.InnerException.Number +
+    ", Level " + $QueryError.Exception.InnerException.Class +
+    ", State " + $QueryError.Exception.InnerException.State +
+    ", Line " + $QueryError.Exception.InnerException.LineNumber +
+    "`r`n" + $QueryError.Exception.Message
+    
+    $ErrorOutput = @{ }
+    $ErrorOutput["output_type"] = "error"
+    $ErrorOutput["traceback"] = @()
+    $ErrorOutput["evalue"] = $ErrorString
+    return $ErrorOutput
+}
+
+function ParseStringToNotebookOutput {
+    param (
+        [System.String]
+        $InputString
+    )
+    <#
+    Parsing the string to notebook cell output. 
+    It's the standard Jupyter Syntax
+    #>
+    $StringOutputData = @{ }
+    $StringOutputData["text/html"] = $InputString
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
+param(
+    [System.String]
+    $TemplateNotebook = $(throw New-Object System.Exception "TemplateNotebook cannot be null"),
```
Charles-Gagnon: You should be able to just use [Parameter(Mandatory=$true)] to ensure that it has a value

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
...
+function ParseTableToNotebookOutput {
+    param (
+        [System.Data.DataTable]
+        $DataTable,
+
+        [int]
+        $CellExecutionCount
+    )
+    $TableHTMLText = "<table>"
+    $TableSchemaFeilds = @()
```
Charles-Gagnon: spelling - TableSchemaFields

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
...
+}
+
+<#
+Start of script
+#>
+try {
+    $TemplateNotebookJsonObject = ConvertFrom-Json -InputObject $TemplateNotebook
+}
+catch {
+    Throw $_.Exception
```
Charles-Gagnon: If you're just rethrowing it you don't need the try/catch

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
...
+Setting params for Invoke-Sqlcmd
+#>
+$DatabaseQueryHashTable = @{ }
+if ($DatabaseInstance) {
+    $DatabaseQueryHashTable["ServerInstance"] = $DatabaseInstance
+}
+if ($DatabaseServer) {
+    $DatabaseQueryHashTable["HostName"] = $DatabaseServer
+}
+if ($DatabasePassword) {
```
Charles-Gagnon: Are you only supporting sql logins? What about Integrated/AAD?

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
...
+    <#
+    Adding the data tables
+    #>
+    $NotebookCellOutputs += $NotebookCellTableOutputs
+    $NotebookCell.outputs = $NotebookCellOutputs
+    
+}
+
+
+return ($TemplateNotebookJsonObject | ConvertTo-Json -Depth 100)
```
Charles-Gagnon: Why the limit? This should be documented at the very least if you think it's necessary.

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
...
+        $DataTable,
+
+        [int]
+        $CellExecutionCount
+    )
+    $TableHTMLText = "<table>"
+    $TableSchemaFeilds = @()
+    $TableHTMLText += "<tr>"
+    foreach ($ColumnName in $DataTable.Columns) {
+        $TableSchemaFeilds += @(@{name = $ColumnName.toString() })
```
kevcunnane: Typo: Feilds -> Fields

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
...
+    Following the current syntax of errors in T-SQL notebooks from ADS
+    #>
+    $ErrorString = "Msg " + $QueryError.Exception.InnerException.Number +
+    ", Level " + $QueryError.Exception.InnerException.Class +
+    ", State " + $QueryError.Exception.InnerException.State +
+    ", Line " + $QueryError.Exception.InnerException.LineNumber +
+    "`r`n" + $QueryError.Exception.Message
+    
+    $ErrorOutput = @{ }
+    $ErrorOutput["output_type"] = "error"
```
kevcunnane: Is this how we send errors from the notebook or does it come out as `'application/vnd.jupyter.stderr'`?

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
kevcunnane: I'm confused too - why do you only set username if there's a password? Isn't it (theoretically) possible someone has a no-password login?

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
```diff
@@ -0,0 +1,211 @@
...
+    #>
+    $DatabaseQueryHashTable["Query"] = $NotebookCell.source
+    <#
+    Executing the T-SQL Query and storing the result and the time taken to execute
+    #>
+    $SqlQueryExecutionTime = Measure-Command { $SqlQueryResult = @(Invoke-Sqlcmd @DatabaseQueryHashTable  4>&1) }
+    <#
+    Setting the the Notebook Cell Execution Count
+    #>
+    $NotebookCell.execution_count = $CellExcecutionCount++
```
kevcunnane: typo: ` $CellExecutionCount`

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
kevcunnane: Actually - checked the SQLKernel code and you're matching it. Please ignore.

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
kevcunnane: I'm hoping it's fairly short since most of the outputs should be strings, correct?

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
aasimkhan30: The error thrown wasn't terminating and I needed to terminate the script after JSON conversion error.

src/Microsoft.SqlTools.ServiceLayer/Agent/Notebooks/Exec-Notebook.ps1
aasimkhan30: Yeah, I think the maximum depth of  nested JSON objects in jupyter notebooks is 5 or 6. I'll check that and make changes accordingly

Charles-Gagnon: I'd just remove the limit if it's only going to be 5 or 6, that way we don't have to update that if it ever changes later. And with the depth being that small it doesn't seem like enough of a concern to warrant explicitly limiting.

aasimkhan30: The default limit of ConvertTo-Json cmdlet is 2, which is why I had to overwrite it.

PR #833, opened 2019-07-03
 Adding completion extension loading and execution logic

shengyfu:


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +287,61 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        /// <summary>
+        /// Completion extension load request callback
+        /// </summary>
+        /// <param name="param"></param>
+        /// <param name="requestContext"></param>
+        /// <returns></returns>
+        internal async Task HandleCompletionExtLoadRequest(CompletionExtensionParams param, RequestContext<bool> requestContext)
+        {
+            var errorMsg = "Failed to create ICompletionExtensionProvider";
+
```
Charles-Gagnon: This is never used, you always overwrite it. Just put the strings in the SendError call - no reason to separate them out like this

src/Microsoft.SqlTools.Hosting.v2/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -83,8 +86,23 @@ public static ExtensionServiceProvider CreateFromAssembliesInDirectory(string di
...
             return new ExtensionServiceProvider(conventions => new ContainerConfiguration().WithAssemblies(assemblies, conventions));
         }
+        
+        public static ExtensionServiceProvider AddAssembly(string path)
+        {            
+            Assembly assembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
+            assemblyCache.AddOrUpdate(assembly.FullName, assembly, (_, previous) => {
+                return assembly;
+            });
+            return new ExtensionServiceProvider(conventions => new ContainerConfiguration().WithAssemblies(assemblyCache.Values, conventions));
```
Charles-Gagnon: Just call Create(IEnumerable<Assembly>) from here

src/Microsoft.SqlTools.Hosting.v2/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -83,8 +86,23 @@ public static ExtensionServiceProvider CreateFromAssembliesInDirectory(string di
...
+            {
+                assemblyCache.AddOrUpdate(assembly.FullName, assembly, (_, previous) => {
+                    return assembly;
+                });
+            }
             return new ExtensionServiceProvider(conventions => new ContainerConfiguration().WithAssemblies(assemblies, conventions));
         }
+        
+        public static ExtensionServiceProvider AddAssembly(string path)
+        {
```
Charles-Gagnon: Would prefer to see this be

        public static ExtensionServiceProvider CreateFromAssemblies(IEnumerable<string> assemblyPaths) {...} if you're going to be adding a new method. You could also just convert the assembly path to an Assembly in the calling code and then just call the existing Create

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +287,61 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                    {
+                        extName = ext.Name;
+                        await ext.Initialize(cancellationToken);
+                    }
+                    catch (NotImplementedException)
+                    {
+                    }
+                    catch (NotSupportedException)
+                    {
+                    }
```
Charles-Gagnon: Are these expected or something? Seems like you shouldn't be silently ignoring these

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -23,14 +23,15 @@
...
-using Microsoft.SqlTools.ServiceLayer.QueryExecution;
 using Microsoft.SqlTools.ServiceLayer.Scripting;
 using Microsoft.SqlTools.ServiceLayer.SqlContext;
 using Microsoft.SqlTools.ServiceLayer.Utility;
 using Microsoft.SqlTools.ServiceLayer.Workspace;
 using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
 using Microsoft.SqlTools.Utility;
 using Location = Microsoft.SqlTools.ServiceLayer.Workspace.Contracts.Location;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion.Extension;
+using Microsoft.SqlTools.Extensibility;
```
Charles-Gagnon: Please keep these in alphabetical order

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -90,7 +97,53 @@ public void AutoCompleteFindCompletions()
...
+                string codeBase = Assembly.GetExecutingAssembly().CodeBase;
+                UriBuilder uri = new UriBuilder(codeBase);
+                string path = Uri.UnescapeDataString(uri.Path);
+                return Path.GetDirectoryName(path);
+            }
+        }
+
+        // This test currently requires a live database connection to initialize 
+        // SMO connected metadata provider.  Since we don't want a live DB dependency
+        // in the CI unit tests this scenario is currently disabled.
```
Charles-Gagnon: Update this comment to be accurate. And please use XML doc (<summary>)

sqltoolsservice.sln
```diff
@@ -95,6 +95,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlTools.ManagedB
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlTools.ManagedBatchParser.IntegrationTests", "test\Microsoft.SqlTools.ManagedBatchParser.IntegrationTests\Microsoft.SqlTools.ManagedBatchParser.IntegrationTests.csproj", "{D3696EFA-FB1E-4848-A726-FF7B168AFB96}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CompletionExtSample", "test\CompletionExtSample\CompletionExtSample.csproj", "{0EC2B30C-0652-49AE-9594-85B3C3E9CA21}"
```
Charles-Gagnon: >CompletionExtSample [](start = 53, length = 19)

I would prefer this named like the others - something like Microsoft.SqlTools.Test.CompletionExtension

test/CompletionExtSample/CompletionExt1.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public CompletionExt1(IReadOnlyDictionary<string, object> properties)
+        {
+            _modelPath = (string)properties["modelPath"];
+        }
+
+        void IDisposable.Dispose()
+        {
+            throw new NotImplementedException();
+        }
```
Charles-Gagnon: Do you really need to throw here? Just have it no-op IMO

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1506,9 +1565,15 @@ internal SignatureHelp GetSignatureHelp(TextDocumentPosition textDocumentPositio
             // if there are no completions then provide the default list
             if (resultCompletionItems == null)
             {
-                resultCompletionItems = AutoCompleteHelper.GetDefaultCompletionItems(scriptDocumentInfo, useLowerCaseSuggestions);
+                resultCompletionItems = AutoCompleteHelper.GetDefaultCompletionItems(scriptDocumentInfo, useLowerCaseSuggestions);
```
Charles-Gagnon: >                 [](start = 130, length = 16)

nit: Please clean up these trailing whitespaces (there's lots of extensions out there that will visualize this for you and/or auto-cleanup)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1506,9 +1565,15 @@ internal SignatureHelp GetSignatureHelp(TextDocumentPosition textDocumentPositio
...
             if (resultCompletionItems == null)
             {
-                resultCompletionItems = AutoCompleteHelper.GetDefaultCompletionItems(scriptDocumentInfo, useLowerCaseSuggestions);
+                resultCompletionItems = AutoCompleteHelper.GetDefaultCompletionItems(scriptDocumentInfo, useLowerCaseSuggestions);                
             }
 
+            //invoke the completion extensions            
+            foreach(var completionExt in _completionExtensions.Values)
+            {
+                resultCompletionItems = await completionExt.HandleCompletionAsync(connInfo, scriptDocumentInfo, resultCompletionItems, new CancellationTokenSource(200).Token).ConfigureAwait(false);
```
Charles-Gagnon: >200 [](start = 163, length = 3)

Suggest using a const value for this and put an explanation of why you chose this value

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
```diff
@@ -74,5 +95,11 @@ public class CompletionItem
         /// resolve request.
         /// </summary>
         public object Data { get; set; }
+
+        //Exposing a command field for a completion item for passing telemetry 
+        public Command Command { get; set; }
```
Charles-Gagnon: Please use XML doc tags for all documentation

test/CompletionExtSample/CompletionExt1.cs
```diff
@@ -0,0 +1,85 @@
...
+            Console.WriteLine("Enter ExecuteAsync");
+
+            var sortedItems = completions.OrderBy(item => item.SortText);
+            sortedItems.First().PreSelect = true;
+            foreach(var item in sortedItems)
+            {
+                item.Command = new Command
+                {
+                    command = "vsintellicode.completionItemSelected",
+                    arguments = new object[] { new Dictionary<string, string> { { "IsCommit", "True" } } }
```
Charles-Gagnon: So what is this command doing? Why do you need to call additional commands on the ADS side?

Charles-Gagnon:
Note that the CI build is failing - looks like it can't build your CompletionExtSample project. Make sure you investigate and get that fixed before checkin.

test/CompletionExtSample/CompletionExt1.cs
shengyfu: This is the command exposed on the IntelliCode side for logging telemetry when one completion item is selected.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: The intention was the Initialize method doesn't have to be implemented.

Charles-Gagnon: Is this standard practice for you? I would generally think that if someone didn't need it they would just have it be a no-op - exceptions should be reserved for actual errors.

ranasaria: +1. It would be very counter-intuitive to have the NotImplemented and NotSupported exception caught and eaten up like it. It is better to make the Initialize method be required to be implemented which could be an no-op implementation.

shengyfu: This is debatable, but I'll remove the try/catch here.

shengyfu:
@kevcunnane , did you get a chance to review the latest code? @Charles-Gagnon , let me know when I can merge the change to master?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -90,11 +97,59 @@ public void AutoCompleteFindCompletions()
...
+                UriBuilder uri = new UriBuilder(codeBase);
+                string path = Uri.UnescapeDataString(uri.Path);
+                return Path.GetDirectoryName(path);
+            }
+        }
+
+        /// <summary>
+        /// This test currently requires a live database connection to initialize
+        /// SMO connected metadata provider.  Since we don't want a live DB dependency
+        /// in the CI unit tests this scenario is currently disabled.
```
Charles-Gagnon: This comment is still inaccurate as you do use a live connection

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +293,46 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        internal async Task HandleCompletionExtLoadRequest(CompletionExtensionParams param, RequestContext<bool> requestContext)
+        {
+            try
+            {
+                var serviceProvider = ExtensionServiceProvider.Create(new Assembly[] { AssemblyLoadContext.Default.LoadFromAssemblyPath(param.Assembly) });
+                foreach (var provider in serviceProvider.GetServices<ICompletionExtensionProvider>())
+                {
+                    var cancellationToken = new CancellationTokenSource(ExtensionLoadingTimeout).Token;
+                    var ext = await provider.CreateAsync(param.Properties ?? new Dictionary<string, object>(), cancellationToken);
+                    if (ext == null)
```
Charles-Gagnon: Hm, we shouldn't just be relying on just the extension to properly cancel the request. There's nothing preventing it from just ignoring the cancellation and taking a huge amount of time. Would prefer we wrap these in our own logic to handle the timeout instead. Same for below with the completion request

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
```diff
@@ -74,5 +95,15 @@ public class CompletionItem
...
+
+        /// <summary>
+        /// Exposing a command field for a completion item for passing telemetry
+        /// </summary>
+        public Command Command { get; set; }
+
+        /// <summary>
+        /// Whether this completion item is preselected or not
+        /// </summary>
+        public bool? PreSelect { get; set; }
```
kevcunnane: Minor: is this casing what's sent? E.g. it's really `preSelect` from the vscode side?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +293,46 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                    {
+                        await requestContext.SendError("Failed to create ICompletionExtension");
```
kevcunnane: Do we need to localize these?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +293,46 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        /// Completion extension load request callback
+        /// </summary>
+        /// <param name="param"></param>
+        /// <param name="requestContext"></param>
+        /// <returns></returns>
+        internal async Task HandleCompletionExtLoadRequest(CompletionExtensionParams param, RequestContext<bool> requestContext)
+        {
+            try
+            {
+                var serviceProvider = ExtensionServiceProvider.Create(new Assembly[] { AssemblyLoadContext.Default.LoadFromAssemblyPath(param.Assembly) });
```
kevcunnane: Is there a reason you don't use the existing `ServiceHostInstance.ServiceProvider` and update it with the new assembly? The reason I am asking is that this should ensure that it can share common services to the extension if needed. Otherwise this won't happen

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +293,46 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                            previous?.Dispose();
+                            return ext;
+                        });
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex.Message);
+            }
```
kevcunnane: Don't we need to send a `requestContext.SendResult(true)` here, or `requestContext.SendResult(false)` if no provider was found? The VSCode side may be stuck waiting forever on completion of the promise.

test/CompletionExtSample/CompletionExt1.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
```
kevcunnane: Copyright (even for samples) needed

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
shengyfu: In VSCode, the property is "preselect", so I should rename it to "Preselect" as other property here.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: I'm not sure, what's the normal practice on error messages?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Actually I would like to use ServiceHostInstance.ServiceProvider, what's the API to add a new assembly? Create is a static API.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
kevcunnane: If it's possible to show up as an error / warning popup to the user, needs localization. if it's mostly diagnostic but _might_ show in the Output area I think it can skip localization

shengyfu: This should be diagnostic message to output and telemetry, shouldn't show up as a error/warning popup to the user. So we don't need localization here.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Fixed as you suggested

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Fixed as you suggested

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Added timeout in the calling code.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
shengyfu: Fixed.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +293,51 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                    cancellationTokenSource.Dispose();
+                    if (!string.IsNullOrEmpty(extName))
+                    {
+                        _completionExtensions.AddOrUpdate(extName, ext, (_, previous) =>
+                        {
+                            previous?.Dispose();
+                            return ext;
+                        });
+                    }
+                    await requestContext.SendResult(true);
```
kevcunnane: This needs to be outside the for loop. Otherwise if you do have 2 or more providers you'll send twice.

shengyfu: Good catch, fixed now.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -90,11 +97,57 @@ public void AutoCompleteFindCompletions()
...
+            {
+                string codeBase = Assembly.GetExecutingAssembly().CodeBase;
+                UriBuilder uri = new UriBuilder(codeBase);
+                string path = Uri.UnescapeDataString(uri.Path);
+                return Path.GetDirectoryName(path);
+            }
+        }
+
+        /// <summary>
+        /// This test tests completion extension interface
```
Charles-Gagnon: >This test tests completion extension interface [](start = 12, length = 46)

Generally I'd prefer these be more descriptive - this tells us nothing useful. What part of the interface is it testing? What is it doing to verify that? Given the name and description I have no idea what this is doing without reading the code itself. Adding comments to the code itself to help this (so the summary doesn't get too verbose) is useful too.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -90,11 +97,57 @@ public void AutoCompleteFindCompletions()
...
+            ScriptParseInfo scriptInfo = new ScriptParseInfo { IsConnected = true };
+            autoCompleteService.ParseAndBind(result.ScriptFile, result.ConnectionInfo);
+            scriptInfo.ConnectionKey = autoCompleteService.BindingQueue.AddConnectionContext(result.ConnectionInfo);
+
+            var completions = autoCompleteService.GetCompletionItems(
+                result.TextDocumentPosition,
+                result.ScriptFile,
+                result.ConnectionInfo).Result;
+
+            Assert.True(completions.Length > 0 && completions[0].Preselect.HasValue && completions[0].Preselect.Value);
```
Charles-Gagnon: >True [](start = 19, length = 4)

Add a message here explaining the issue - right now if this fails it's just going to say "expected true got false" which is not very helpful in figuring out the issue. Make sure you include the contents of the completions list as well since that often times helps with figuring out an issue.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +293,57 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        internal async Task HandleCompletionExtLoadRequest(CompletionExtensionParams param, RequestContext<bool> requestContext)
+        {
+            bool loaded = false;
+            try
+            {
+                var serviceProvider = (ExtensionServiceProvider)ServiceHostInstance.ServiceProvider;
+                serviceProvider.AddAssembliesToConfiguration(new Assembly[] { AssemblyLoadContext.Default.LoadFromAssemblyPath(param.Assembly) });
+                int extCount = 0;
+                foreach (var provider in serviceProvider.GetServices<ICompletionExtensionProvider>())
+                {
```
Charles-Gagnon: This doesn't seem right - every time someone calls load extension from ADS it's going to re-initialize all of the loaded extensions. I would think that you'd just want to call Initialize on the single assembly from AssemblyLoadContext.Default.LoadFromAssemblyPath(param.Assembly) and then add it to the completionExtensions list. 

A big issue you have right now is this situation :

1. Extension assembly requested to be loaded. Let's say there's already 2 other ICompletionExtensionProvider extensions loaded as well
2. Calling Initialize on the new extension fails and it throws an exception - but the other two passed fine
3. We'll now call SendError with the error message as expected
4. But then we'll continue on and SendResult(true) since the other extensions loaded properly

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
shengyfu: Added more comments to the test.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Make sense, I fixed it by calling Create to get a new ExtensionProvider to find the completion extensions in the passed in assembly.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
shengyfu: Added more details to the description.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
Charles-Gagnon: Please add messages to the actual asserts as well - that way when they fail the log will have useful information instead of just saying "expected true got false"

---
In reply to: [304579406](https://github.com/microsoft/sqltoolsservice/pull/833#discussion_r304579406) [](ancestors = 304579406)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
Charles-Gagnon: Hm, but now you're calling CreateAsync on this new provider. We still want the ServiceHostInstance.ServiceProvider to have all the assemblies initialized.

What I think you'll need to do is loop through and just skip any that are already initialized. 

---
In reply to: [304580100](https://github.com/microsoft/sqltoolsservice/pull/833#discussion_r304580100) [](ancestors = 304580100)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
shengyfu: I see, added.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +293,57 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                            return ext;
+                        });
+                        extCount++;
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex.Message);
+            }
```
Charles-Gagnon: Should return here otherwise we could end up sending both an error and a result - but only one of those should be expected

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
```diff
@@ -45,6 +53,19 @@ public enum CompletionItemKind
             Reference = 18
     }
 
+    public class Command
+    {
+        /// <summary>
+        /// The identifier of the actual command handler, like `vsintellicode.completionItemSelected`.
+        /// </summary>
+        public string command;
```
kburtram: public variables should Uppercase, e.g. Command.  Ideally we'd use properties instead of fields if there is not reason not too.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1509,6 +1570,24 @@ internal SignatureHelp GetSignatureHelp(TextDocumentPosition textDocumentPositio
                 resultCompletionItems = AutoCompleteHelper.GetDefaultCompletionItems(scriptDocumentInfo, useLowerCaseSuggestions);
             }
 
+            //invoke the completion extensions
+            foreach(var completionExt in _completionExtensions.Values)
```
kburtram: nit: missing ' ' before '('

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1509,6 +1570,24 @@ internal SignatureHelp GetSignatureHelp(TextDocumentPosition textDocumentPositio
...
+            {
```
kburtram: I suspect this code should only run in the code paths that call `GetDefaultCompletionItems`.  In the other cases we are providing schema-based suggestions and my understanding is the model is only providing keyword suggestions.  This implies that these suggestions will most likely not be useful in the other code paths and we probably shouldn't pay the cost to make these calls.

I guess this is a more generic completion provider extensibility mechanism, so ultimately it could make sense to run this all the time.  But until we have a use-case that will provide schema completions from an extension I think the above comment makes sense.

test/CompletionExtSample/CompletionExt1.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Linq;
+using Microsoft.SqlTools.ServiceLayer.Connection;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion.Extension;
+using Microsoft.SqlTools.ServiceLayer.LanguageServices.Contracts;
+
+namespace Microsoft.SqlTools.Test.CompletionExtension
+{
+    [Export(typeof(ICompletionExtensionProvider))]
+    public class CompletionExtProvider1 : ICompletionExtensionProvider
```
kburtram: what is the point of the `1` suffix.  are we going to be adding a `2` set of classes?

shengyfu: We could add more extensions for testing purpose.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: This is a generic completion extension point, I don't see the reason to restrict it. And also we are looking at schema-based recommendations as well, we can discuss more offline.

shengyfu: If you are concerned about the performance, I can lower the time budget to 50ms.

kburtram: The underlying concern is performance.  The high-level point is if we don't run the `GetDefaultCompletionItems` code path we have already determined from the AST that we are in a location that needs schema-based suggestions.

If we don't have any extensions that provide schema-based suggestions, and don't have any medium-term plans to have such an extension, I would prefer not to pay any cost in this perf-critical code path.  It doesn't really matter if it's 50ms or 200ms if we have very high confidence the suggestions won't be useful.  If we do have a scenario where an extension can provide schema-based suggestions it seems easy enough to relocate this code to run all the time at that time.  

Do see any value in the existing completion extensions running at a place in the code that requires schema-based completions (e.g. in a columns or table list)?

test/CompletionExtSample/CompletionExt1.cs
kburtram: hmm, ok, seems unusual to number classes when there is only one of them.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
kburtram: ...just noted your previous comment "schema-based recommendations as well, we can discuss more offline."  Please ping me to discuss if you plan to support this scenario.  In that case the current integration point would make sense.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
Charles-Gagnon: @kevcunnane To verify - but my understanding is that an error is a response that has additional error metadata (which IIRC is actually turned into an exception on the ADS side). We should only ever send one response for a single request so sending both an error response and a normal response isn't going to work correctly and will likely cause another error.

kevcunnane: Yes, we should return here and only send result on success. Also re-reviewing, there's a line that sends error if 1 provider is null and then continues, that's a bug too. 

@shengyfu - please verify that max 1 `SendError` or `SendResult` is ever sent. Otherwise we'll get unexpected behavior.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Fixed, please review again.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Yes, fixed now.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
shengyfu: Changed to properties.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Based on our offline discussion, we'll limit to default completion list for now.

test/CompletionExtSample/CompletionExt1.cs
shengyfu: Removed the numbering.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
kburtram: thanks!  looks good to me.  we can switch it back once we have that use-case a little further along.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
```diff
@@ -23,6 +24,13 @@ public static readonly
             RequestType<CompletionItem, CompletionItem>.Create("completionItem/resolve");
     }
 
+    public class CompletionExtLoadRequest
```
Charles-Gagnon: >CompletionExtLoadRequest [](start = 17, length = 24)

Also please update [https://github.com/microsoft/sqltoolsservice/blob/master/docs/guide/jsonrpc_protocol.md](https://github.com/microsoft/sqltoolsservice/blob/master/docs/guide/jsonrpc_protocol.md) with the new request/response items. I'd suggest making a new section called "Language Service Protocol Extensions" since this isn't part of the official language service protocol.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +294,62 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        internal async Task HandleCompletionExtLoadRequest(CompletionExtensionParams param, RequestContext<bool> requestContext)
+        {
+            int extCount = 0;
+            try
+            {
+                //register the new assembly
+                var serviceProvider = (ExtensionServiceProvider)ServiceHostInstance.ServiceProvider;
+                var assemblies = new Assembly[] { AssemblyLoadContext.Default.LoadFromAssemblyPath(param.Assembly) };
+                serviceProvider.AddAssembliesToConfiguration(assemblies);
+                foreach (var provider in serviceProvider.GetServices<ICompletionExtensionProvider>())
```
Charles-Gagnon: >provider [](start = 29, length = 8)

Shouldn't this be var service instead? You're getting the services from the provider - not other providers.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +294,62 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                    var ext = await provider.CreateAsync(param.Properties ?? new Dictionary<string, object>(), cancellationToken).WithTimeout(ExtensionLoadingTimeout);
+                    if (ext == null)
+                    {
+                        await requestContext.SendError("Failed to create ICompletionExtension from " + provider.GetType().FullName);
+                        return;
+                    }
+                    string extName = ext.Name;
+                    await ext.Initialize(cancellationToken).WithTimeout(ExtensionLoadingTimeout);
+                    cancellationTokenSource.Dispose();
+                    if (!string.IsNullOrEmpty(extName))
```
Charles-Gagnon: >if (!string.IsNullOrEmpty(extName)) [](start = 20, length = 35)

If this is true then we should throw an error since we won't actually be loading the extension

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +294,62 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                var serviceProvider = (ExtensionServiceProvider)ServiceHostInstance.ServiceProvider;
+                var assemblies = new Assembly[] { AssemblyLoadContext.Default.LoadFromAssemblyPath(param.Assembly) };
+                serviceProvider.AddAssembliesToConfiguration(assemblies);
+                foreach (var provider in serviceProvider.GetServices<ICompletionExtensionProvider>())
+                {
+                    if (completionExtensionProviders.Contains(provider))
+                    {
+                        //skipping completion providers that already loaded
+                        continue;
+                    }
```
Charles-Gagnon: I don't think is necessary - just do the check below in the completionExtensions list to see if the assembly is already added there.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +294,62 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                    string extName = ext.Name;
+                    await ext.Initialize(cancellationToken).WithTimeout(ExtensionLoadingTimeout);
+                    cancellationTokenSource.Dispose();
+                    if (!string.IsNullOrEmpty(extName))
+                    {
+                        completionExtensions.AddOrUpdate(extName, ext, (_, previous) =>
+                        {
+                            previous?.Dispose();
+                            return ext;
+                        });
```
Charles-Gagnon: I don't think we should be doing this like this - the dispose should be unnecessary and could possible cause issues with stuff being disposed in the middle of execution if we somehow did get to a point where this was called for already loaded extensions. The way I'd do it is check if the extension in the params is already loaded (in the map) and if it is send an error since that's unexpected behavior. If it isn't then go ahead and initialize it and add it to the map at that point.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +294,62 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                        extCount++;
```
Charles-Gagnon: >extCount++; [](start = 24, length = 11)

Instead of doing the count like this I'd just do a SendResult(true) here and return. Then below where you're currently sending a result do a SendError instead since we should never get to that point.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: Since we are using MEF for service discovery, I think I should drop ICompletionExtensionProvider, what do you think?

shengyfu: I think I can drop ICompletionExtensionProvider interface, what do you think?

Charles-Gagnon: Oh yeah good point - definitely doesn't seem useful anymore. Sounds good to me!

---
In reply to: [304700466](https://github.com/microsoft/sqltoolsservice/pull/833#discussion_r304700466) [](ancestors = 304700466)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
shengyfu: There is real user scenario here, in vscode if the user upgrades intellicode extension with the updated bits for the sql completion extension, it will ask mssql extension to reload the completion extension, in this case, we should reload and update the current completion extension instance.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/Extension/ICompletionExtension.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public interface ICompletionExtension : IDisposable
+    {
+        /// <summary>
+        /// Unique name for the extension
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Completion extension initialization
```
Charles-Gagnon: >Completion extension initialization [](start = 12, length = 35)

I'd suggest something more useful like "Method for initializing the extension, this is called once when the extension is loaded"

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/Extension/ICompletionExtension.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        /// <param name="properties">Parameters needed by the extension</param>
+        /// <param name="cancelToken">Cancellation token for cancel the initialization</param>
```
Charles-Gagnon: >cancel the initialization [](start = 61, length = 25)

"Cancellation token used to indicate that the initialization should be cancelled"

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/Extension/ICompletionExtension.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <returns></returns>
+        Task Initialize(IReadOnlyDictionary<string, object> properties, CancellationToken token);
+
+        /// <summary>
+        /// Implement the completion extension logic
```
Charles-Gagnon: >Implement the completion extension logic [](start = 12, length = 40)

Handles the completion request, returning the modified CompletionItemList if used

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/Extension/ICompletionExtension.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        /// <param name="connInfo">Connection information for the completion session</param>
+        /// <param name="scriptDocumentInfo">Script parsing information</param>
+        /// <param name="completions">Current completion list</param>
+        /// <param name="cancelToken">Cancellation token for cancel the completion extension logic</param>
```
Charles-Gagnon: >Cancellation token for cancel the completion extension logic [](start = 38, length = 60)

"Token used to indicate that the completion request should be cancelled"

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/Extension/CompletionExtensionParams.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion.Extension
+{
+
+    [Serializable]
+    public class CompletionExtensionParams
+    {
+        public string AssemblyPath { get; set; }
```
Charles-Gagnon: >AssemblyPath [](start = 22, length = 12)

Please put XML doc summaries on these

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +295,86 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                }
+                //register the new assembly
+                var serviceProvider = (ExtensionServiceProvider)ServiceHostInstance.ServiceProvider;
+                var assembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(param.AssemblyPath);
+                var assemblies = new Assembly[] { assembly };
+                serviceProvider.AddAssembliesToConfiguration(assemblies);
+                foreach (var ext in serviceProvider.GetServices<ICompletionExtension>())
+                {
+                    var cancellationTokenSource = new CancellationTokenSource(ExtensionLoadingTimeout);
+                    var cancellationToken = cancellationTokenSource.Token;
```
Charles-Gagnon: Move this after the check for the TypeName

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +295,86 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        /// <param name="requestContext"></param>
+        /// <returns></returns>
+        internal async Task HandleCompletionExtLoadRequest(CompletionExtensionParams param, RequestContext<bool> requestContext)
+        {
+            try
+            {
+                if (!CheckExtAssemblyToBeLoaded(param.AssemblyPath))
+                {
+                    await requestContext.SendError(string.Format("Skip loading {0} because it's already loaded once or doesn't exist", param.AssemblyPath));
+                    return;
```
Charles-Gagnon: I'd rather we be specific and send a separate error message for each scenario

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +295,86 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        }
+
+        /// <summary>
+        /// Check a particular assembly needs to be loaded or not
+        /// </summary>
+        /// <param name="assemblyPath"></param>
+        /// <returns></returns>
+        private bool CheckExtAssemblyToBeLoaded(string assemblyPath)
+        {
+            if (File.Exists(assemblyPath))
```
Charles-Gagnon: >if (File.Exists(assemblyPath)) [](start = 12, length = 30)

This isn't really necessary - File.GetLastWriteTime will throw if it doesn't exist which will be caught by the catch in the calling method anyways

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +295,86 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex.Message);
+                return;
+            }
+        }
+
+        /// <summary>
+        /// Check a particular assembly needs to be loaded or not
```
Charles-Gagnon: >Check a particular assembly needs to be loaded or not [](start = 12, length = 53)

Check whether a particular assembly should be reloaded based on whether it's been updated since it was last loaded.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +295,86 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+        /// </summary>
+        /// <param name="assemblyPath"></param>
+        /// <returns></returns>
+        private bool CheckExtAssemblyToBeLoaded(string assemblyPath)
```
Charles-Gagnon: >CheckExtAssemblyToBeLoaded [](start = 21, length = 26)

Suggest that this is renamed "CheckIfAssemblyShouldBeLoaded" - it's worded a little better

Charles-Gagnon: >CheckExtAssemblyToBeLoaded [](start = 21, length = 26)

Please add tests for this new logic too to verify that the behavior is correct when called multiple times when the assembly has/hasn't been updated

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -283,6 +295,86 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
...
+                        return;
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex.Message);
+                return;
+            }
+        }
```
Charles-Gagnon: You need to do a SendError if we get past the foreach loop since that means that we didn't find an assembly from the ServiceProvider that matches the one we just added.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
shengyfu: Updated the doc.

shengyfu:
@kburtram, I added a list of TSQL functions into the default completion list, please review.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
```diff
@@ -29,305 +29,394 @@ public static class AutoCompleteHelper
...
-            "where",
-            "while",
-            "with",
-            "within",
-            "without",
-            "writetext",
-            "xact_abort",
-            "xml",
+        {
+            "abs",
```
kburtram: I think adding built-in SQL functions is fine.  Though I'm not sure how effective it will be.  The way the IntelliSense works is that we first try to use the T-SQL parser to get suggestions.  The parser will only provide schema-based suggestions (unfortunately it doesn't support keyword suggestions).  Only if there are no schema-based suggestions then we return this default keyword list.

I suspect that most the code locations where a user would want to call a function they will likely have schema suggestions available (e.g. in column lists, where clauses, having clauses, etc).  So the default completion list would not be provided.

There are couple exceptions, such as when the editor is not connected or when no tables have been listed in the from clause yet.  But I suspect in the most common use-cases the functions still won't be suggested.  Overall, I'm still fine adding them though in case they help in some cases.

shengyfu: I see, the TSQL functions should merge with the schema-based suggestions after a space or comma.

PR #834, opened 2019-07-10
Added endpoint for fetching all notebook jobs

aasimkhan30:
Created a simple endpoint that fetches all the notebook jobs accessible by the user. To be used in ADS, AgentViewComponent.


src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                {
+                    var result = new AgentNotebooksResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+
+                    if (connInfo != null)
+                    {
+                        using (SqlConnection connection = new SqlConnection(ConnectionService.BuildConnectionString(connInfo.ConnectionDetails)))
```
kburtram: I think we're generally using the QueryService and ConnectionService to execute SQL rather than directly executing like this.  I think these services provide some abstractions and supporting functionality that's useful.  

Can you look for other examples of executing SQL in the code base and follow those patterns?  Or was this copied from an example in the code?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                        {
+                            connection.Open();
+                            /*
+                            This query fetches all the notebook Jobs across all the databases accessible by the user.
```
kburtram: nit: I'd recommend using single line comment style (i.e. `//`) for single line comments

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                "(SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = N''nb_template'')\n"+
+                                "BEGIN\n" + 
+                                "select dbo.nb_template.*,\n" +
+                                "DB_NAME() as db_name,\n" +
+                                "msdb.dbo.sysjobs.*\n" +
+                                "from [?].dbo.nb_template\n" +
+                                "LEFT JOIN\n" + 
+                                "msdb.dbo.sysjobs \n" +
+                                "ON \n" + 
+                                "[?].dbo.nb_template.job_id = msdb.dbo.sysjobs.job_id\n" +
```
kburtram: should we put `dbo.nb_template` in it's own schema or should we use `dbo`?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                    foreach (var job in jobs.Values)
+                                    {
+                                        allJobsHashTable.Add(job.JobID, job);
+                                    }
+                                }                                
+                                foreach (DataTable templateTable in jobIdsDataSet.Tables)
+                                {
+                                    foreach (DataRow templateRow in templateTable.Rows){
+                                        NotebookJobInfo notebookJob = new NotebookJobInfo();
+                                        notebookJob.Template = (string)templateRow["json"];
```
kburtram: will this throw if the "json" value is `null`?  perhaps using `as` to cast is safer.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                if(jobs != null)
+                                {
+                                    foreach (var job in jobs.Values)
+                                    {
+                                        allJobsHashTable.Add(job.JobID, job);
+                                    }
+                                }                                
+                                foreach (DataTable templateTable in jobIdsDataSet.Tables)
+                                {
+                                    foreach (DataRow templateRow in templateTable.Rows){
```
kburtram: please put `{` on their own line in C# code per the project coding conventions.

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Request definition
+        /// </summary>
+        public static readonly
+        RequestType<AgentNotebooksParams, AgentNotebooksResult> Type = 
+        RequestType<AgentNotebooksParams, AgentNotebooksResult>.Create("agent/notebooks");
+    }
+    /// <summary>
+    /// SQL Agent Notebook Job Info Class
+    /// </summary>
+    public class NotebookJobInfo
```
kburtram: we shouldn't define the Info classes in the contracts..they should be defined in with the info classes.  For this simple class probably sticking next to AgentJobInfo makes sense

kburtram: hmm, looks like AgentJobInfo is in contracts. I guess I'm thinking of the Data classes.  In any case, I'd stick next to AgentJobInfo.

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -0,0 +1,51 @@
...
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Agent.Contracts
+{
+    /// <summary>
```
kburtram: a couple formatting notes.  Please add a blank line between class definitions and in getters\setters please add a space character after `{` and before `}`.  Take a look at some of the other class for examples.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                "DB_NAME() as db_name,\n" +
+                                "msdb.dbo.sysjobs.*\n" +
+                                "from [?].dbo.nb_template\n" +
+                                "LEFT JOIN\n" + 
+                                "msdb.dbo.sysjobs \n" +
+                                "ON \n" + 
+                                "[?].dbo.nb_template.job_id = msdb.dbo.sysjobs.job_id\n" +
+                                "END '\n" +  
+                                "EXEC sp_MSforeachdb @script\n";
+                            var agentnotebooks = new List<NotebookJobInfo>();
```
kburtram: this should be `agentNotebooks` per camel-casing naming convention.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                        using (SqlConnection connection = new SqlConnection(ConnectionService.BuildConnectionString(connInfo.ConnectionDetails)))
+                        {
+                            connection.Open();
+                            /*
+                            This query fetches all the notebook Jobs across all the databases accessible by the user.
+                             */
+                            string getJobIdsFromDatabaseQueryString = 
+                                "DECLARE @script as varchar(MAX)\n" +
+                                "SET @script =\n" + 
+                                "'USE [?];\n" +
```
Charles-Gagnon: >USE [?] [](start = 34, length = 7)

What if the user doesn't have permission to some DBs? You should probably do a try/catch so the entire thing doesn't fail in that case

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                "IF EXISTS\n" + 
+                                "(SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = N''nb_template'')\n"+
```
Charles-Gagnon: >nb_template [](start = 96, length = 11)

Check schema too?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                "BEGIN\n" + 
+                                "select dbo.nb_template.*,\n" +
```
Charles-Gagnon: >select [](start = 33, length = 6)

nit: T-SQL keywords should be all caps (SELECT, AS, FROM, etc)

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                            var agentnotebooks = new List<AgentNotebookInfo>();
+                            using( SqlCommand getJobIdsFromDatabaseQueryCommand = new SqlCommand(getJobIdsFromDatabaseQueryString, connection))
+                            {
+                                SqlDataAdapter jobIdsAdapter = new SqlDataAdapter(getJobIdsFromDatabaseQueryCommand);
+                                DataSet jobIdsDataSet = new DataSet();
+                                jobIdsAdapter.Fill(jobIdsDataSet);
+                                var serverConnection = ConnectionService.OpenServerConnection(connInfo);
+                                var fetcher = new JobFetcher(serverConnection);
+                                var filter = new JobActivityFilter();
+                                // Fetching all the jobs
```
Charles-Gagnon: What is this filter doing?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                    {
+                                        allJobsHashTable.Add(job.JobID, job);
+                                    }
+                                }                                
+                                foreach (DataTable templateTable in jobIdsDataSet.Tables)
+                                {
+                                    foreach (DataRow templateRow in templateTable.Rows){
+                                        AgentNotebookInfo notebookJob = new AgentNotebookInfo();
+                                        notebookJob.Template = (string)templateRow["json"];
+                                        notebookJob.TargetDatabase = (string)templateRow["db_name"];
```
Charles-Gagnon: nit: You can use an initializer to make this a bit cleaner.

var notebookJob = new AgentNotebookInfo() {
    Template = (string)templateRow["json"],
    TargetDatabase = (string)templateRow["db_name"],
    ...
}

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
Charles-Gagnon: >dbo.nb_template.* [](start = 40, length = 17)

Have you thought about how you're going to handle versioning? What happens if the schema for this table changes?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1214,88 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                            string getJobIdsFromDatabaseQueryString = 
+                                "DECLARE @script as varchar(MAX)\n" +
+                                "SET @script =\n" + 
+                                "'USE [?];\n" +
+                                "IF EXISTS\n" + 
+                                "(SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = N''nb_template'')\n"+
+                                "BEGIN\n" + 
+                                "select dbo.nb_template.*,\n" +
+                                "DB_NAME() as db_name,\n" +
+                                "msdb.dbo.sysjobs.*\n" +
```
Charles-Gagnon: >"msdb.dbo.sysjobs.* [](start = 32, length = 19)

Do you really need all the columns? I'd try to limit it to just the ones you need. Also keep in mind that these values could potentially be null if the job doesn't exist (say if it was deleted after running the job)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Agent/AgentNotebookTests.cs
```diff
@@ -0,0 +1,7 @@
+namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.Agent
+{
+    public class AgentNotebookTests
+    {
+
```
Charles-Gagnon: Please add tests =)

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
aasimkhan30: The system procedure "sp_MSforeachdb" runs it on only those dbs which user has permission to access. 
Code Referenced.
`EXEC sp_MSforeachdb @script`

Charles-Gagnon: Where are you getting that information from? I looked around briefly and didn't see anything saying that specifically.

Regardless though - this procedure is specifically undocumented and unsupported so relying on behavior like that seems dangerous. There's no harm in being extra careful here I think.

aasimkhan30:
I've added new stuff can you re-review?
@kburtram @Charles-Gagnon

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
             return new Tuple<SqlConnectionInfo, DataTable, ServerConnection>(sqlConnInfo, dt, serverConnection);
         }
 
+                // Notebook helper function to execute queires to be later replaced by QueryExecutionService method.
+        public DataSet executeQuery(ConnectionInfo connInfo, string SqlQuery)
```
Charles-Gagnon: Use xml doc comments for method headers (/** should auto-populate the header if you're using VS)

Charles-Gagnon: >executeQuery [](start = 23, length = 12)

Style for C# is methods are capitalized

Charles-Gagnon: >SqlQuery [](start = 68, length = 8)

Style for C# is params are non-capitalized

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                {
+                    var result = new AgentNotebooksResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+
+                    if (connInfo != null)
+                    {
+                        string getJobIdsFromDatabaseQueryString =
```
Charles-Gagnon: >getJobIdsFromDatabaseQueryString [](start = 31, length = 32)

nit: You can prefix the string with @ to allow newlines within the string and get rid of all the +'s

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
Charles-Gagnon: This is still fetching all columns - I still would prefer to see these specifically listed out

---
In reply to: [302604528](https://github.com/microsoft/sqltoolsservice/pull/834#discussion_r302604528) [](ancestors = 302604528)

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                            // This query fetches all the notebook Jobs across all the databases accessible by the user.
+                            
+                        var agentNotebooks = new List<AgentNotebookInfo>();
+                        DataSet jobIdsDataSet = executeQuery(connInfo, getJobIdsFromDatabaseQueryString);
+                        var serverConnection = ConnectionService.OpenServerConnection(connInfo);
+                        var fetcher = new JobFetcher(serverConnection);
+                        var filter = new JobActivityFilter();
+                        // Fetching all the jobs
+                        var jobs = fetcher.FetchJobs(filter);
+                        // Creating Dictionary of jobs for effcient access later on.
```
Charles-Gagnon: >effcient [](start = 59, length = 8)

efficient

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                try
+                {
+                    var result = new AgentNotebooksResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+
+                    if (connInfo != null)
+                    {
```
Charles-Gagnon: If connInfo is null you never send a response and thus will time out on the ADS side. You should send an error in that case.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+            {
+                try
+                {
+                    var result = new AgentNotebookHistoryResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                    if (connInfo != null)
+                    {
```
Charles-Gagnon: If connInfo is null you never send a response and thus will time out on the ADS side. You should send an error in that case.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                        foreach (JobSchedule schedule in schedules)
+                        {
+                            jobSchedules.Add(AgentUtilities.ConvertToAgentScheduleInfo(schedule));
+                        }
+                        result.Schedules = jobSchedules.ToArray();
+                        // Add histories
+                        int count = dt.Rows.Count;
+                        List<AgentNotebookHistoryInfo> notebookHistories = new List<AgentNotebookHistoryInfo>();
+                        if (count > 0)
+                        {
```
Charles-Gagnon: If count is <= 0 you never send a response and thus will time out on the ADS side. You should send an error in that case.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                {
+                    var result = new AgentNotebookHistoryResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                    if (connInfo != null)
+                    {
+                        string getNotebookHistoryQueryString = 
+                            "USE "+ parameters.TargetDatabase + ";" +
```
Charles-Gagnon: >parameters.TargetDatabase [](start = 36, length = 25)

Database name needs to be escaped and bracket-quoted.

You also should avoid embedding parameters directly into a query like this and instead use SqlCommand.Parameters. I'd suggest adding an optional parameter to ExecuteQuery that takes in a list of parameters that are then attached to the created SqlCommand

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                            "SELECT * FROM notebooks.nb_materialized " +
+                            "WHERE JOB_ID = '" + parameters.JobId + "'";
+
+                        DataSet NotebookHistoryDataset = executeQuery(connInfo, getNotebookHistoryQueryString);
```
Charles-Gagnon: >NotebookHistoryDataset [](start = 32, length = 22)

var names should be non-capitalized

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                        CDataContainer dataContainer = CDataContainer.CreateDataContainer(connInfo, databaseExists: true);
+                        var jobServer = dataContainer.Server.JobServer;
+                        var jobs = jobServer.Jobs;
+                        Tuple<SqlConnectionInfo, DataTable, ServerConnection> tuple = CreateSqlConnection(connInfo, parameters.JobId);
+                        SqlConnectionInfo sqlConnInfo = tuple.Item1;
+                        DataTable dt = tuple.Item2;
+                        ServerConnection connection = tuple.Item3;
```
Charles-Gagnon: >connection [](start = 41, length = 10)

This doesn't seem to be used?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+
+                        var agentNotebookHistories = new List<AgentNotebookHistoryInfo>();
+                        JobStepCollection steps = jobs[parameters.JobName].JobSteps;
+                        var jobSteps = new List<AgentJobStepInfo>();
+                        foreach (JobStep step in steps)
+                        {
+                            jobSteps.Add(AgentUtilities.ConvertToAgentJobStepInfo(step, parameters.JobId, parameters.JobName));
+                        }
```
Charles-Gagnon: You should be able to just use  jobSteps.AddRange(steps)

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                string jobRuntime = jobHistory.RunDate.ToString("yyyyMMddHHmmss");
+                                AgentNotebookHistoryInfo notebookHistory = jobHistory;
+                                if(allNotebookHistoriesHashTable.ContainsKey(jobRuntime))
+                                    notebookHistory.MaterializedNotebookId = allNotebookHistoriesHashTable[jobRuntime].MaterializedNotebookId;
+                                notebookHistories.Add(notebookHistory);
+                            }
+                            result.Histories = notebookHistories.ToArray();
+                            result.Success = true;
+                            tlog.CloseReader();
+                            await requestContext.SendResult(result);
```
Charles-Gagnon: This should be in a finally block with the try starting right above creating the LogSourceJobHistory (or whever the reader is opened)

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                    var result = new AgentNotebookMaterializedResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                    if (connInfo != null)
+                    {
+                         string MaterializedNotebookQueryString = 
+                            "USE "+ parameters.TargetDatabase + ";" +
+                            "SELECT * FROM notebooks.nb_materialized " +
```
Charles-Gagnon: Same here about bracketing/escaping the db, using parameters and using @ for the newlines

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                            "WHERE materialized_id = '" + parameters.NotebookMaterializedID + "'";
+
+                        DataSet MaterializedNotebookDataSet = executeQuery(connInfo, MaterializedNotebookQueryString);
```
Charles-Gagnon: >MaterializedNotebookDataSet [](start = 32, length = 27)

var names non-capitalized

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                try
+                {
+                    var result = new AgentNotebookMaterializedResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                    if (connInfo != null)
+                    {
+                         string MaterializedNotebookQueryString =
```
Charles-Gagnon: >MaterializedNotebookQueryString [](start = 32, length = 31)

var names non-capitalized

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                        out connInfo);
+                    if (connInfo != null)
+                    {
+                         string MaterializedNotebookQueryString = 
+                            "USE "+ parameters.TargetDatabase + ";" +
+                            "SELECT * FROM notebooks.nb_materialized " +
+                            "WHERE materialized_id = '" + parameters.NotebookMaterializedID + "'";
+
+                        DataSet MaterializedNotebookDataSet = executeQuery(connInfo, MaterializedNotebookQueryString);
+                        if (MaterializedNotebookDataSet != null){
```
Charles-Gagnon: Shouldn't this be an error if you don't get a result set back?

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1219,215 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+            {
+                try
+                {
+                    var result = new AgentNotebookMaterializedResult();
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                    if (connInfo != null)
+                    {
```
Charles-Gagnon: If connInfo is null you never send a response and thus will time out on the ADS side. You should send an error in that case.

src/Microsoft.SqlTools.ServiceLayer/Agent/Common/AgentUtilities.cs
```diff
@@ -66,6 +66,37 @@ public static AgentJobInfo ConvertToAgentJobInfo(JobProperties job)
             };
         }
 
+        public static AgentNotebookInfo convertToAgentNotebookInfo(JobProperties job)
```
Charles-Gagnon: >convertToAgentNotebookInfo [](start = 40, length = 26)

method names capitalized

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -0,0 +1,221 @@
...
+    /// </summary>
+    public class CreateAgentNotebookResult  : ResultStatus
+    {
+        public AgentNotebookInfo Job { get; set; }
+    }
+
+    /// <summary>
+    /// SQL Agent create Notebook request type
+    /// </summary>
+    public class CreateAgentNotebookRequest
```
Charles-Gagnon: >CreateAgentNotebookRequest [](start = 17, length = 26)

I don't see this, updatenotebook or delete notebook handled in AgentService. If you aren't planning on implementing those now you should delete them

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -253,6 +253,7 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
             {
                 Logger.Write(TraceEventType.Verbose, "Shutting down language service");
                 DeletePeekDefinitionScripts();
+                DeleteAgentNotebooks();
                 this.Dispose();
```
Charles-Gagnon: This should go in the InitializeService method of AgentService, not here

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
aasimkhan30: I still haven't implemented the endpoints yet.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
aasimkhan30: Moved

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1232,461 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                var result = new CreateAgentNotebookResult();
+                try
+                {
+                    var assembly = Assembly.GetAssembly(typeof(AgentService));
+                    using (Stream scriptStream = assembly.GetManifestResourceStream("Microsoft.SqlTools.ServiceLayer.Agent.NotebookResources.NotebookJobScript.ps1"))
+                    {
+                        using (StreamReader reader = new StreamReader(scriptStream))
+                        {
+                            string execNotebookScript = "$TargetDatabase = \"" + parameters.Notebook.TargetDatabase + "\"" + Environment.NewLine + reader.ReadToEnd();
+                            AgentJobStepInfo notebookJobStep = new AgentJobStepInfo()
```
kburtram: consider a createDefaultJobStep method somewhere

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1232,461 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                            @"
+                            IF NOT EXISTS (
+                            SELECT  schema_name
+                            FROM    information_schema.schemata
+                            WHERE   schema_name = 'notebooks' ) 
+                            BEGIN
+                            EXEC sp_executesql N'CREATE SCHEMA notebooks'
+                            END
+
+                            IF  NOT EXISTS (SELECT * FROM sys.objects
```
kburtram: `IF  NOT EXISTS (SELECT * ` could be `IF  NOT EXISTS (SELECT TOP 1 id` or something

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1232,461 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                            ";
+                            List<SqlParameter> insertTemplateJsonQueryParams = new List<SqlParameter>();
+                            insertTemplateJsonQueryParams.Add(new SqlParameter("jobId", jobId));
+                            insertTemplateJsonQueryParams.Add(new SqlParameter("templateFileContents", templateFileContents));
+                            insertTemplateJsonQueryParams.Add(new SqlParameter("executeDatabase", parameters.Notebook.ExecuteDatabase));
+                            ExecuteQuery(connInfo, insertTemplateJsonQuery, insertTemplateJsonQueryParams, targetDatabase);
+                        }
+                    }
+                    result.Success = true;
+                    result.Job = parameters.Notebook;
```
kburtram: should the property here be `result.Notebook`

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1232,461 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                    WHERE 
+                    job_id = @jobId;
+                    DELETE FROM notebooks.nb_materialized
+                    WHERE
+                    job_id = @jobId;
+                    IF NOT EXISTS (SELECT * FROM notebooks.nb_template)
+                    BEGIN
+                        DROP TABLE notebooks.nb_template;
+                        DROP TABLE notebooks.nb_materialized;
+                        DROP SCHEMA notebooks;
```
kburtram: this may fail if the user creates new object within here.  is there a way to either (1) check if there's still content here before trying to delete or (2) try\catch this statement or isolate in a different batch such that the drop tables and other cleanup succeeds even when the drop schema fails.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1232,461 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                                                parameters.OwnerUri,
+                                                out connInfo);
+                    string targetDatabase = parameters.Notebook.TargetDatabase;
+                    string jobId = parameters.Notebook.JobId;
+                    if (parameters.TemplateFilePath != null)
+                    {
+
+                        string templateFilePath = parameters.TemplateFilePath;
+                        string templateFileContents = File.ReadAllText(templateFilePath);
+                        templateFileContents = templateFileContents.Replace("'", "''");
```
kburtram: I don't think you need to escape the `'` characters if you are using SqlParameters.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1232,461 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                        UPDATE notebooks.nb_template SET notebook = @templateFileContents WHERE job_id = @jobId
+                        ";
+                        List<SqlParameter> insertTemplateJsonQueryParams = new List<SqlParameter>();
+                        insertTemplateJsonQueryParams.Add(new SqlParameter("templateFileContents", templateFileContents));
+                        insertTemplateJsonQueryParams.Add(new SqlParameter("jobId", jobId));
+                        ExecuteQuery(connInfo, insertTemplateJsonQuery, insertTemplateJsonQueryParams, targetDatabase);
+                    }
+                    string updateExecuteDatabaseQuery =
+                    @"
+                     UPDATE notebooks.nb_template SET execute_database = @executeDatabase WHERE job_id = @jobId
```
kburtram: maybe consider `execution_database` instead `execute_database`

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1211,6 +1232,461 @@ public static XmlDocument CreateJobXmlDocument(string svrName, string jobName)
...
+                    ";
+                    List<SqlParameter> updateExecuteDatabaseQueryParams = new List<SqlParameter>();
+                    updateExecuteDatabaseQueryParams.Add(new SqlParameter("executeDatabase", parameters.Notebook.ExecuteDatabase));
+                    updateExecuteDatabaseQueryParams.Add(new SqlParameter("jobId", jobId));
+                    ExecuteQuery(connInfo, updateExecuteDatabaseQuery, updateExecuteDatabaseQueryParams, targetDatabase);
+                    var tempResult =
```
kburtram: why are we capturing `tempResult`?

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentJobHistoryInfo.cs
```diff
@@ -58,4 +58,14 @@ public  class AgentJobStep
...
         public AgentJobStepInfo stepDetails;
 	}
+
+    /// <summary>
+    /// a class for storing various properties of a agent notebook history
+    /// </summary>
+    public class AgentNotebookHistoryInfo : AgentJobHistoryInfo
+    {
+        public int MaterializedNotebookId { get; set; }
+        public int MaterializedNotebookErrorFlag { get; set; }
```
kburtram: remove or rename to `...ErrorCode`

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentJobHistoryInfo.cs
```diff
@@ -58,4 +58,14 @@ public  class AgentJobStep
...
+        public string MaterializedNotebookErrorInfo { get; set; }
```
kburtram: check other classes for naming conventions, but I think this would be something like `ErrorMessage`

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentJobInfo.cs
```diff
@@ -61,4 +61,16 @@ public class AgentJobInfo
         public AgentScheduleInfo[] JobSchedules { get; set; }
         public AgentAlertInfo[] Alerts { get; set; }
     }
+
+    /// <summary>
+    /// a class for storing variour properties of notebook Jobs
+    /// </summary>
+    public class AgentNotebookInfo : AgentJobInfo
+    {
+        public string TemplateId { get; set; }
```
kburtram: consider adding comments of these properties describing they are..specifically the distinction between Target and Execution databases, etc.

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /// <summary>
+    /// SQL Agent update Notebook params
+    /// </summary>
+    public class UpdateAgentNotebookParams : TaskRequestDetails
+    {
+        public string OwnerUri { get; set; }
+
+        public string OriginalNotebookName { get; set; }
```
kburtram: maybe just `OriginalName` is fine?

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /// <summary>
+    /// SQL Agent Notebook materialized params
+    /// </summary>
+    public class AgentNotebookMaterializedParams : TaskRequestDetails
+    {
+        public string OwnerUri { get; set; }
+        public string TargetDatabase { get; set; }
+        public int NotebookMaterializedID { get; set; }
```
kburtram: in C# the convention is abbreviations should be Pascal-cased.  i.e. so `Id` here.

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /// <summary>
+    /// SQL Agent Notebook materialized result
+    /// </summary>
+    public class AgentNotebookMaterializedResult : ResultStatus
+    {
+        public string NotebookMaterializedJson { get; set; }
```
kburtram: maybe `MaterializedNotebook` is better since it ignores some implementation details (i.e. JSON).

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// <summary>
+    /// SQL Agent Notebook templates request type
+    /// </summary>
+    public class AgentNotebookTemplateRequest
+    {
+        public static readonly
+            RequestType<AgentNotebookTemplateParams, AgentNotebookTemplateResult> Type =
+            RequestType<AgentNotebookTemplateParams, AgentNotebookTemplateResult>.Create("agent/notebooktemplate");
+    }
+
```
kburtram: blank lines

src/Microsoft.SqlTools.ServiceLayer/Utility/FileUtils.cs
```diff
@@ -9,6 +9,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Utility
     internal static class FileUtilities
     {
         internal static string PeekDefinitionTempFolder = Path.GetTempPath() + "mssql_definition"; 
+        internal static string AgentNotebookTempFolder = Path.GetTempPath() + "agent_notebooks";
```
kburtram: maybe `mssql_notebooks` so the folders group together.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Agent/AgentNotebookTests.cs
```diff
@@ -0,0 +1,128 @@
+using System;
```
kburtram: using statements should be sorted.  The convention in this code base is (1) System imports first alphabetically then (2) everything else alphabetically (3) the using static can be last.

PR #836, opened 2019-07-24
Fix DacFx wizard not supporting AAD auth

kisantia:
This fixes AAD auth not previously working by passing in the Azure authentication token when DacServices is created if there is a token.

Fixes https://github.com/microsoft/azuredatastudio/issues/3486

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
@@ -38,6 +38,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
         protected DacFxOperation(ConnectionInfo connInfo)
         {
             Validate.IsNotNull("connectionInfo", connInfo);
+            Validate.IsNotNull("connectionDetails", connInfo.ConnectionDetails);
             this.ConnInfo = connInfo;
             this.ConnectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
             this.OperationId = Guid.NewGuid().ToString();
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
```diff
@@ -79,7 +83,8 @@ public void Execute(TaskExecutionMode mode)
 
             try
             {
-                this.DacServices = new DacServices(this.ConnectionString);
+                // Pass in Azure authentication token if needed
+                this.DacServices = this.ConnInfo.ConnectionDetails.AzureAccountToken != null ? new DacServices(this.ConnectionString, new AccessTokenProvider(this.ConnInfo.ConnectionDetails.AzureAccountToken)) : new DacServices(this.ConnectionString);
```
udeeshagautam: this.ConnInfo.ConnectionDetails is already checked for null in some way? if not please check that one too...

src/Microsoft.SqlTools.ServiceLayer/Utility/AccessTokenProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            if (string.IsNullOrEmpty(accessToken))
+            {
+                throw new ArgumentNullException("accessToken");
+            }
+
+            _accessToken = accessToken;
+        }
+
+        public bool IsTokenExpired() { return false; }
```
udeeshagautam: never expires? or we don't care in current case?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
kisantia: added to constructor of DacFxOperation

src/Microsoft.SqlTools.ServiceLayer/Utility/AccessTokenProvider.cs
kisantia: Yeah I don't think we care since the token is passed from ADS. This is the same implementation that the SqlPackageAccessTokenProvider has

PR #837, opened 2019-07-26
Fix node metadata and fully populate metadata from SMO objects

Charles-Gagnon:
This started off as just wanting to get the MetadataType to be correct - it was being set to the default (0 - Table) for all nodes. Along the way I fixed it up so we'll actually send the correct metadata for the node and cleaned up the logic to be more centralized.

Note that this is a breaking change for the MetadataType enum since I added the Unknown value for value 0. I'd rather do this now since it's better style and is a simple fix for anyone who is actually using it. But regardless we should major vBump the release version to follow versioning guidelines.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/ObjectMetadata.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/ObjectMetadata.cs
@@ -3,6 +3,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using Microsoft.SqlServer.Management.Smo;
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/ObjectMetadata.cs
```diff
@@ -3,34 +3,124 @@
...
     /// </summary>
     public enum MetadataType
     {
-        Table = 0,
-        View = 1,
-        SProc = 2,
-        Function = 3,
-        Schema = 4,
-        Database = 5
+        Unknown = 0,
```
kburtram: I think other extensions, such as the Postgres provider, have a dependency on this.  If we're going to change we should probably update all the providers at the same time.

PR #838, opened 2019-07-26
Revert "Default to Server level context for all OE nodes (#829)"

Charles-Gagnon:
Fixes https://github.com/microsoft/azuredatastudio/issues/6497 https://github.com/microsoft/azuredatastudio/issues/6394

This reverts commit a4b6c300ac91875028ba4ed74f3304afaf45aa96. Two reasons for reverting this :

1. There's been numerous complaints (https://github.com/microsoft/azuredatastudio/issues/6394) about the change in behavior
2. It broke Data Explorer nodes and while the long term "best" fix may be to have a single connection per node and then have child nodes update their context appropriately - but this is going to require more investigation and time to implement correctly so doing this lets us fix the completely broken experience until that can be done correctly


Charles-Gagnon:
@kburtram @kevcunnane I'm planning on merging this in today unless either of you have any issue with this

PR #839, opened 2019-07-31
Feature/sqlcmd : Enable running scripts with SQLCMD variables - Part 1

udeeshagautam:
This contains:
1. Changes to enable running of a script file containing (and defined) SQLCMD variables. This will evolve to enable getting arguments from outside the script in later changes.
2. Changes to switch on/off Intellisense with changing SQLCMD mode. We will make changes to better handle the specific SQLCMD Intellisense later.

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Utility/GeneralRequestDetails.cs
+++ b/src/Microsoft.SqlTools.Hosting/Utility/GeneralRequestDetails.cs
@@ -14,7 +14,7 @@ namespace Microsoft.SqlTools.Utility
     {
         public GeneralRequestDetails()
         {
-            Options = new Dictionary<string, object>(new OptionsKeyComparer());
+            Options = new Dictionary<string, object>(StringComparer.InvariantCultureIgnoreCase);
         }
 
         public T GetOptionValue<T>(string name, T defaultValue = default(T))
@@ -110,22 +110,6 @@ namespace Microsoft.SqlTools.Utility
         /// Gets or Sets the options
         /// </summary>
         public Dictionary<string, object> Options { get; set; }
-    }
-           
-    /// <summary>
-    /// Compare class for Execution Setting options to ensure ignoring case in dictionary keys
-    /// </summary>
-    public class OptionsKeyComparer : IEqualityComparer<string>
-    {
-        public bool Equals(string x, string y)
-        {
-            return string.Equals(x, y, StringComparison.InvariantCultureIgnoreCase);
-        }
-
-        public int GetHashCode(string obj)
-        {
-            return obj.ToLowerInvariant().GetHashCode();
-        }
     }
 }
 
--- a/src/Microsoft.SqlTools.Hosting/Utility/GeneralRequestDetails.cs
+++ b/src/Microsoft.SqlTools.Hosting/Utility/GeneralRequestDetails.cs
@@ -14,7 +14,7 @@ namespace Microsoft.SqlTools.Utility
     {
         public GeneralRequestDetails()
         {
-            Options = new Dictionary<string, object>();
+            Options = new Dictionary<string, object>(new OptionsKeyComparer());
         }
 
         public T GetOptionValue<T>(string name, T defaultValue = default(T))
@@ -111,4 +111,21 @@ namespace Microsoft.SqlTools.Utility
         /// </summary>
         public Dictionary<string, object> Options { get; set; }
     }
+           
+    /// <summary>
+    /// Compare class for Execution Setting options to ensure ignoring case in dictionary keys
+    /// </summary>
+    public class OptionsKeyComparer : IEqualityComparer<string>
+    {
+        public bool Equals(string x, string y)
+        {
+            return string.Equals(x, y, StringComparison.InvariantCultureIgnoreCase);
+        }
+
+        public int GetHashCode(string obj)
+        {
+            return obj.ToLowerInvariant().GetHashCode();
+        }
+    }
 }
+
--- a/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
@@ -51,7 +51,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                     int offset = offsets[0];
                     int startColumn = batchInfos[0].startColumn;
                     int count = batchInfos.Count;
-                    string batchText = content.Substring(offset, batchInfos[0].length);
+                    string batchText = batchInfos[0].batchText; // content.Substring(offset, batchInfos[0].length);
 
                     // if there's only one batch then the line difference is just startLine
                     if (count > 1)
@@ -82,15 +82,15 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                     }
 
                     // Generate the rest batch definitions
-                    for (int index = 1; index < count - 1; index++)
+                    for (int index = 1; index < count - 1 ; index++)
                     {
                         lineDifference = batchInfos[index + 1].startLine - batchInfos[index].startLine;
                         position = ReadLines(reader, lineDifference, endLine);
                         endLine = position.Item1;
                         endColumn = position.Item2;
                         offset = offsets[index];
-                        batchText = content.Substring(offset, batchInfos[index].length);
-                        startLine = batchInfos[index].startLine;
+                        batchText = batchInfos[index].batchText;
+                        startLine = batchInfos[index].startLine + 1; // this is done for first batch but was missing for following batches
                         startColumn = batchInfos[index].startColumn;
 
                         // make a new batch definition for each batch
@@ -109,7 +109,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                     if (count > 1)
                     {
 
-                        batchText = content.Substring(offsets[count-1], batchInfos[count - 1].length);
+                        batchText = batchInfos[count - 1].batchText;
                         BatchDefinition lastBatchDef = GetLastBatchDefinition(reader, batchInfos[count - 1], batchText);
                         batchDefinitionList.Add(lastBatchDef);
                     }
@@ -209,7 +209,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
         private static BatchDefinition GetLastBatchDefinition(StringReader reader,
             BatchInfo batchInfo, string batchText)
         {
-            int startLine = batchInfo.startLine;
+            int startLine = batchInfo.startLine + 1;
             int startColumn = batchInfo.startColumn;
             string prevLine = null;
             string line = reader.ReadLine();
@@ -328,12 +328,12 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
         /// <summary>
         /// Takes in a query string and returns a list of BatchDefinitions
         /// </summary>
-        public List<BatchDefinition> GetBatches(string sqlScript)
+        public List<BatchDefinition> GetBatches(string sqlScript, ExecutionEngineConditions conditions = null)
         {
             batchInfos = new List<BatchInfo>();
 
             // execute the script - all communication / integration after here happen via event handlers
-            executionEngine.ParseScript(sqlScript, notificationHandler);
+            executionEngine.ParseScript(sqlScript, notificationHandler, conditions);
 
             // retrieve a list of BatchDefinitions 
             List<BatchDefinition> batchDefinitionList = ConvertToBatchDefinitionList(sqlScript, batchInfos);
@@ -381,7 +381,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                     }
 
                     // Add the script info
-                    batchInfos.Add(new BatchInfo(args.Batch.TextSpan.iStartLine, args.Batch.TextSpan.iStartIndex, batchTextLength, args.Batch.ExpectedExecutionCount));
+                    batchInfos.Add(new BatchInfo(args.Batch.TextSpan.iStartLine, args.Batch.TextSpan.iStartIndex, batchTextLength, batchText, args.Batch.ExpectedExecutionCount));
                 }
             }
             catch (NotImplementedException)
@@ -474,17 +474,19 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
 
         private class BatchInfo
         {
-            public BatchInfo(int startLine, int startColumn, int length, int repeatCount = 1)
+            public BatchInfo(int startLine, int startColumn, int length, string batchText, int repeatCount = 1)
             {
                 this.startLine = startLine;
                 this.startColumn = startColumn;
                 this.length = length;
                 this.executionCount = repeatCount;
+                this.batchText = batchText;
             }
             public int startLine;
             public int startColumn;
             public int length;
             public int executionCount;
+            public string batchText;
         }
 
     }
--- a/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
@@ -51,7 +51,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                     int offset = offsets[0];
                     int startColumn = batchInfos[0].startColumn;
                     int count = batchInfos.Count;
-                    string batchText = batchInfos[0].batchText; // content.Substring(offset, batchInfos[0].length);
+                    string batchText = batchInfos[0].batchText;
 
                     // if there's only one batch then the line difference is just startLine
                     if (count > 1)
@@ -90,7 +90,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                         endColumn = position.Item2;
                         offset = offsets[index];
                         batchText = batchInfos[index].batchText;
-                        startLine = batchInfos[index].startLine + 1; // this is done for first batch but was missing for following batches
+                        startLine = batchInfos[index].startLine + 1; //positions is 0 index based
                         startColumn = batchInfos[index].startColumn;
 
                         // make a new batch definition for each batch
@@ -381,7 +381,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                     }
 
                     // Add the script info
-                    batchInfos.Add(new BatchInfo(args.Batch.TextSpan.iStartLine, args.Batch.TextSpan.iStartIndex, batchTextLength, batchText, args.Batch.ExpectedExecutionCount));
+                    batchInfos.Add(new BatchInfo(args.Batch.TextSpan.iStartLine, args.Batch.TextSpan.iStartIndex, batchText, args.Batch.ExpectedExecutionCount));
                 }
             }
             catch (NotImplementedException)
@@ -474,17 +474,15 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
 
         private class BatchInfo
         {
-            public BatchInfo(int startLine, int startColumn, int length, string batchText, int repeatCount = 1)
+            public BatchInfo(int startLine, int startColumn, string batchText, int repeatCount = 1)
             {
                 this.startLine = startLine;
                 this.startColumn = startColumn;
-                this.length = length;
                 this.executionCount = repeatCount;
                 this.batchText = batchText;
             }
             public int startLine;
             public int startColumn;
-            public int length;
             public int executionCount;
             public string batchText;
         }
--- a/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
@@ -193,7 +193,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode
             commandParser.SetRecognizeSqlCmdSyntax(sqlCmdMode);
             commandParser.SetBatchDelimiter(BatchSeparator);
             commandParser.ThrowOnUnresolvedVariable = true;
-
+            
             batchParser.Execute = new BatchParser.ExecuteDelegate(ExecuteBatchInternal);
             batchParser.ErrorMessage = new BatchParser.ScriptErrorDelegate(RaiseScriptError);
             batchParser.Message = new BatchParser.ScriptMessageDelegate(RaiseBatchMessage);
@@ -237,7 +237,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode
         {
             try
             {
-                DisconnectSqlCmdInternal();
+                DisconnectSqlCmdInternal();                
 
                 ConfigureBatchEventHandlers(currentBatch, batchEventHandlers, false);
 
@@ -325,7 +325,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode
                 batchEventHandlers.OnBatchMessage(this, args);
             }
         }
-
+        
         /// <summary>
         /// Executes a given batch given the number of times
         /// </summary>
@@ -334,8 +334,8 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode
         /// <param name="lineNumber"></param>
         /// <returns>True if we should continue processing, false otherwise</returns>
         private bool ExecuteBatchInternal(
-            string batchScript,
-            int num,
+            string batchScript, 
+            int num, 
             int lineNumber)
         {
             if (lineNumber == -1)
@@ -392,7 +392,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode
                 currentBatch.TextSpan = textSpan;
                 currentBatch.BatchIndex = currentBatchIndex;
                 currentBatch.ExpectedExecutionCount = numBatchExecutionTimes;
-
+                
                 currentBatchIndex++;
 
                 if (conditions != null)
@@ -1021,7 +1021,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode
         #endregion
 
         #region Public members
-
+        
         /// <summary>
         /// Executes the script
         /// </summary>
@@ -1045,7 +1045,8 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode
         /// Parses the script locally
         /// </summary>
         /// <param name="script">script to parse</param>
-        /// <param name="batchEventsHandler">batch handler</param>        
+        /// <param name="batchEventsHandler">batch handler</param>   
+        /// <param name="conditions">execution engine conditions if specified</param>
         /// <remarks>
         /// The batch parser functionality is used in this case
         /// </remarks>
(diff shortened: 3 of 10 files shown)


src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
```diff
@@ -51,7 +51,7 @@ private static List<BatchDefinition> ConvertToBatchDefinitionList(string content
                     int offset = offsets[0];
                     int startColumn = batchInfos[0].startColumn;
                     int count = batchInfos.Count;
-                    string batchText = content.Substring(offset, batchInfos[0].length);
+                    string batchText = batchInfos[0].batchText; // content.Substring(offset, batchInfos[0].length);
```
Charles-Gagnon: >// content.Substring(offset, batchInfos[0].length); [](start = 64, length = 51)

This seems fine (was it broken before) but get rid of the commented out code

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
```diff
@@ -82,15 +82,15 @@ private static List<BatchDefinition> ConvertToBatchDefinitionList(string content
...
                     {
                         lineDifference = batchInfos[index + 1].startLine - batchInfos[index].startLine;
                         position = ReadLines(reader, lineDifference, endLine);
                         endLine = position.Item1;
                         endColumn = position.Item2;
                         offset = offsets[index];
-                        batchText = content.Substring(offset, batchInfos[index].length);
-                        startLine = batchInfos[index].startLine;
+                        batchText = batchInfos[index].batchText;
+                        startLine = batchInfos[index].startLine + 1; // this is done for first batch but was missing for following batches
```
Charles-Gagnon: >// this is done for first batch but was missing for following batches [](start = 69, length = 69)

I'd say just put the first comment here about the pos being 0 based

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
```diff
@@ -474,17 +474,19 @@ private void Dispose(bool disposing)
 
         private class BatchInfo
         {
-            public BatchInfo(int startLine, int startColumn, int length, int repeatCount = 1)
+            public BatchInfo(int startLine, int startColumn, int length, string batchText, int repeatCount = 1)
```
Charles-Gagnon: >int length [](start = 61, length = 10)

I'd get rid of length since we can get that from batchText

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
```diff
@@ -1045,23 +1045,27 @@ public void ExecuteBatch(ScriptExecutionArgs scriptExecutionArgs)
...
         /// The batch parser functionality is used in this case
         /// </remarks>
-        public void ParseScript(string script, IBatchEventsHandler batchEventsHandler)
+        public void ParseScript(string script, IBatchEventsHandler batchEventsHandler, ExecutionEngineConditions conditions = null)
         {
             Validate.IsNotNull(nameof(script), script);
             Validate.IsNotNull(nameof(batchEventsHandler), batchEventsHandler);
 
-
+            if (conditions != null)
```
Charles-Gagnon: >if (conditions != null) [](start = 12, length = 23)

Should be fine to just set this.conditions = conditions since you'll want to be checking that conditions is not null later on anyways

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -116,7 +116,13 @@ public class Query : IDisposable
 
             // Process the query into batches 
             BatchParserWrapper parser = new BatchParserWrapper();
-            List<BatchDefinition> parserResult = parser.GetBatches(queryText);
+            ExecutionEngineConditions conditions = null;
+            if (settings.IsSqlCmdMode)
+            {
+                conditions = new ExecutionEngineConditions();
+                conditions.IsSqlCmd = settings.IsSqlCmdMode;
```
Charles-Gagnon: >conditions.IsSqlCmd = settings.IsSqlCmdMode; [](start = 16, length = 44)

nit : use initializer here

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -163,6 +163,11 @@ public class QueryExecutionSettings : GeneralRequestDetails
         /// </summary>
         private const int DefaultQueryGovernorCostLimit = 0;
 
+        /// <summary>
+        /// Flag to run query in sqlcmd mode
```
Charles-Gagnon: >Flag to run query in sqlcmd mode [](start = 12, length = 32)

Specifically this is the default for the flag

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
```diff
@@ -338,6 +338,41 @@ public void VerifyIsSqlCmd()
             }
         }
 
+        // Verify whether the batchParser execute SqlCmd successfully
+        [Fact]
```
Charles-Gagnon: Use xml doc comments for this

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
```diff
@@ -338,6 +338,41 @@ public void VerifyIsSqlCmd()
...
+GO
+select $(__var1) + $(__var2) as col
+GO";
+
+                using (SqlConnection con = new SqlConnection(CONNECTION_STRING))
+                {
+                    con.Open();
+                    var condition = new ExecutionEngineConditions();
+                    condition.IsSqlCmd = true;
+                    TestExecutor testExecutor = new TestExecutor(sqlCmdQuery, con, condition);
```
Charles-Gagnon: initializer

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -328,5 +332,71 @@ public async Task RebuildIntellisenseCacheClearsScriptParseInfoCorrectly()
                 testDb.Cleanup();
             }
         }
+
+        [Fact]
+        // This test validates switching off editor intellisesnse for now. 
+        // Will change to better handling once we have specific SQLCMD intellisense in Language Service
```
Charles-Gagnon: xml doc comments

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -328,5 +332,71 @@ public async Task RebuildIntellisenseCacheClearsScriptParseInfoCorrectly()
...
+        public async Task HandleReuestTochangeToSQLCMDFile()
```
Charles-Gagnon: >HandleReuestTochangeToSQLCMDFile [](start = 26, length = 32)

HandleRequestToChangeToSqlcmdFile

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -328,5 +332,71 @@ public async Task RebuildIntellisenseCacheClearsScriptParseInfoCorrectly()
...
+                eventContextSqlCmd.Setup(x => x.SendEvent(PublishDiagnosticsNotification.Type, It.Is<PublishDiagnosticsNotification>((notif) => ValidateNotification(notif, 0, ref countOfValidationCalls)))).Returns(Task.FromResult(new object()));
+                await langService.HandleDidChangeLanguageFlavorNotification(new LanguageFlavorChangeParams
+                {
+                    Uri = scriptFile.ClientFilePath,
+                    Language = LanguageService.SQL_CMD_LANG.ToLower(),
+                    Flavor = "MSSQL"
+                }, eventContextSqlCmd.Object);
+                await langService.DelayedDiagnosticsTask;
+
+                Assert.True(countOfValidationCalls == 2, string.Format("Validation should be call 2 time but is called {0} times", countOfValidationCalls));
```
Charles-Gagnon: >call [](start = 93, length = 4)

should be called

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -328,5 +332,71 @@ public async Task RebuildIntellisenseCacheClearsScriptParseInfoCorrectly()
...
+            finally
+            {
+                if (File.Exists(scriptFile.ClientFilePath))
+                {
+                    File.Delete(scriptFile.ClientFilePath);
+                }
+            }
+        }
+
+        private bool ValidateNotification(PublishDiagnosticsNotification notif, int errors, ref int countofValidationCalls)
```
Charles-Gagnon: >countofValidationCalls [](start = 100, length = 22)

countOfValidationCalls

src/Microsoft.SqlTools.Hosting/Utility/GeneralRequestDetails.cs
```diff
@@ -14,7 +14,7 @@ public class GeneralRequestDetails
     {
         public GeneralRequestDetails()
         {
-            Options = new Dictionary<string, object>();
+            Options = new Dictionary<string, object>(new OptionsKeyComparer());
         }
```
pensivebrian: Could we use StringComparer.InvariantCultureIgnoreCase instead of defining our own?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -328,5 +332,71 @@ public async Task RebuildIntellisenseCacheClearsScriptParseInfoCorrectly()
...
+        {
+            countofValidationCalls++;
+            Assert.True(notif.Diagnostics.Length == errors, string.Format("Notification errors {0} are not as expected {1}", notif.Diagnostics.Length, errors));
+            return true;
+        }
```
Charles-Gagnon: If this is going to throw when it fails anyways might as well just have it not return anything

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
```diff
@@ -109,7 +109,7 @@ private static List<BatchDefinition> ConvertToBatchDefinitionList(string content
                     if (count > 1)
                     {
 
-                        batchText = content.Substring(offsets[count-1], batchInfos[count - 1].length);
+                        batchText = batchInfos[count - 1].batchText;
                         BatchDefinition lastBatchDef = GetLastBatchDefinition(reader, batchInfos[count - 1], batchText);
```
Charles-Gagnon: Are the changes in this file relevant to the sqlcmd stuff or just bugs that you're fixing in this PR?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
abist: Be careful with the batch indices. We've had issues here where the batch starting lines and end lines don't match up. You can verify this by running any query and then clicking on the `Line xx` in the Query Results window. It should highlight the actual run batch.

Charles-Gagnon: Yeah - I'm kind of worried about making changes here. When I was doing the query history stuff I found a lot of weirdness with how the numbers end up and so making changes here could easily break functionality elsewhere.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
udeeshagautam: This was broken in context of substitute text which was not used from anywhere. will remove the comment, thanks for pointing out

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
udeeshagautam: dont want to change code path so will keep this

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
udeeshagautam: good catch :)

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
udeeshagautam: not sure what you mean

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
udeeshagautam: the mock call requires bool return.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
udeeshagautam: - BatchText is required for SQLCMD substituion of sqlcmd variables.
- +1 in start line is just to fix wrong start line in messages

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
pensivebrian: Agreed. Make sure we test out on mac/Linux as I see some code that searches for "\n" in OnBatchParserExecutionFinished and adjusts batch length.

---
In reply to: [309457799](https://github.com/microsoft/sqltoolsservice/pull/839#discussion_r309457799) [](ancestors = 309457799)

udeeshagautam: Sure I will test that way.  

The batch indecies were giving wrong result for sustitution - that is why I added the text to be sure.

Regarding line numbers - I tried comparing with DacFx but couldntfully since its changed much. In there we have no 0 based index so we use the startline as is. Here we have 0 based index and we add +1 to first line but ot to later ones. 

Those are the only two changes here. Let's discuss if the second one looks too risky.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
```diff
@@ -474,17 +474,19 @@ private void Dispose(bool disposing)
...
                 this.length = length;
                 this.executionCount = repeatCount;
+                this.batchText = batchText;
             }
             public int startLine;
             public int startColumn;
             public int length;
             public int executionCount;
+            public string batchText;
         }
```
pensivebrian: Remember we have to work with files GB in size. Be adding a member here, we're allocating a string which is tied to the lifetime of a BatchInfo instance. If BatchInfo instances are created up front, we could be allocating a lot of strings. Perhaps the old code allocated a string on demand for a reason?

src/Microsoft.SqlTools.Hosting/Utility/GeneralRequestDetails.cs
udeeshagautam: yes absolutely... will change

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
udeeshagautam: Sure. Just wanted clarity since I (rightly) expected this line to cause questions ;)

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
udeeshagautam: hmm... thats a good point. I will fight with substitution a little bit more to see if there is an alternative way.

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
Charles-Gagnon: Use an initializer here 😄 

```
var condition = new ExecutionEngineConditions() {
    IsSqlCmd = true
};
```

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
udeeshagautam: pinged you with an rough alternative... let's discuss tomorrow

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
udeeshagautam: @abist update - validated selecting lines from message

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
udeeshagautam: @pensivebrian as discussed, I did the validation with big script of 110MB with multiple batches and big batches both... did not see any significant difference in memory/CPU or time taken. 

(local debug mode shows other weird issue i.e. query not showing complete/cancel even if all is done but that is with or without new sqltoolsservice)

So, do you think its okay that we go on with this one for the first cut and try the in-place execution (as discussed) in later cycles.

pensivebrian: Fine for the first cut.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DisasterRecovery/BackupServiceTests.cs
```diff
@@ -42,7 +42,7 @@ public class BackupServiceTests
         /// Get backup configuration info
         /// </summary>
         /// Test is failing in code coverage runs. Reenable when stable.
-        ///[Fact]
+        [Fact]
```
pensivebrian: >   [Fact] [](start = 5, length = 9)

Changes to this file intentional?

udeeshagautam: no, thanks for the catch. Was trying to re-enable the. not part of the PR.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
ranasaria: A code spell checker extension in vscode had helped a lot with spelling issues. They are not perfect but they have helped me elminate 90% of my silly typos.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -116,7 +116,12 @@ public class Query : IDisposable
 
             // Process the query into batches 
             BatchParserWrapper parser = new BatchParserWrapper();
-            List<BatchDefinition> parserResult = parser.GetBatches(queryText);
+            ExecutionEngineConditions conditions = null;
+            if (settings.IsSqlCmdMode)
+            {
+                conditions = new ExecutionEngineConditions() { IsSqlCmd = settings.IsSqlCmdMode };
+            }
+            List<BatchDefinition> parserResult = parser.GetBatches(queryText, conditions);
```
ranasaria: What happens if settings.IsSqlCmdMode is not true? Are we ok calling parser.GetBatches with parameter conditions as null?

udeeshagautam: The default will continue to be IsSqlCmdMode = false. Yeah, GetBatches handles the conditions being null and follows the simple query creation path.

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
```diff
@@ -338,6 +338,42 @@ public void VerifyIsSqlCmd()
...
 
+        /// <summary>
+        /// Verify whether the batchParser execute SqlCmd successfully
+        /// </summary>
+        [Fact]
+        public void VerifyRunSqlCmd()
+        {
+            using (ExecutionEngine executionEngine = new ExecutionEngine())
+            {
+                string sqlCmdQuery = @"
```
ranasaria: const string sqlCmdQuery, that makes it a compile time constant embeded in generated code.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -328,5 +332,73 @@ public async Task RebuildIntellisenseCacheClearsScriptParseInfoCorrectly()
...
+                    File.Delete(scriptFile.ClientFilePath);
+                }
+            }
+        }
+
+        private bool ValidateNotification(PublishDiagnosticsNotification notif, int errors, ref int countOfValidationCalls)
+        {
+            countOfValidationCalls++;
+            Assert.True(notif.Diagnostics.Length == errors, string.Format("Notification errors {0} are not as expected {1}", notif.Diagnostics.Length, errors));
+            return true;
```
ranasaria: >true [](start = 19, length = 4)

even though the code should never get past to this line when notif.Diagnostics.Length == errors is false, I suggest returning (notif.Diagnostics.Length == errors). Because when this condition is false line 401 is never executed and when it is true both returning true and returning that condition are equivalent. Where this suggested change gains is in scenarios where in future this assert is removed, then the suggested change would be more meaningful return value.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -328,5 +332,73 @@ public async Task RebuildIntellisenseCacheClearsScriptParseInfoCorrectly()
...
+                eventContextSqlCmd.Setup(x => x.SendEvent(PublishDiagnosticsNotification.Type, It.Is<PublishDiagnosticsNotification>((notif) => ValidateNotification(notif, 0, ref countOfValidationCalls)))).Returns(Task.FromResult(new object()));
+                await langService.HandleDidChangeLanguageFlavorNotification(new LanguageFlavorChangeParams
+                {
+                    Uri = scriptFile.ClientFilePath,
+                    Language = LanguageService.SQL_CMD_LANG.ToLower(),
+                    Flavor = "MSSQL"
+                }, eventContextSqlCmd.Object);
+                await langService.DelayedDiagnosticsTask;
+
+                Assert.True(countOfValidationCalls == 2, string.Format("Validation should be called 2 time but is called {0} times", countOfValidationCalls));
```
ranasaria: >Assert.True [](start = 16, length = 11)

Assert.Equal produces more readable output in case of failures. So changing this to soemthing like:
Assert.Equal(countOfValidationCalls, 2, string.Format("Validation should be called 2 time but is called {0} times", countOfValidationCalls))  is more useful.

Also suggest using string interpolation instead of string.Format so the change would become:
Assert.Equal(countOfValidationCalls, 2, $"Validation should be called 2 time but is called {countOfValidationCalls} times")

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
```diff
@@ -1045,23 +1045,27 @@ public void ExecuteBatch(ScriptExecutionArgs scriptExecutionArgs)
...
+            {
+                this.conditions = conditions;
+            }
```
ranasaria: what value of this.conditions do you want when conditions is in fact null? Are you afraid or overwrite a previous value?

Terse for the above if condition(and modern programmers like the brevity) would be:
this.conditions = conditions ?? this.conditions;

ranasaria:
>             return Task.FromResult(true);

even though this is not your change I suggest changing returns of Task.FromResult(true) to Task.CompletedTask.  The previous code was the best option before .Net 4.6. But from .Net 4.6 it is best to use this predefined constant. 

Here is some discussion on this topic:
[https://stackoverflow.com/questions/13127177/if-my-interface-must-return-task-what-is-the-best-way-to-have-a-no-operation-imp](https://stackoverflow.com/questions/13127177/if-my-interface-must-return-task-what-is-the-best-way-to-have-a-no-operation-imp)

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs:1758 in 26eca53. [](commit_id = 26eca53e95d0c211ca64029d9d3fbb72326b0f44, deletion_comment = False)

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
udeeshagautam: hmm.. I would prefer to use the same practice as used for rest of the tests in this file for consistency unless it breaks something

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
```diff
@@ -338,6 +338,42 @@ public void VerifyIsSqlCmd()
...
+:setvar __var1 1
```
ranasaria: That would be your call. There is no downside/risk to adding the keyword const just a very miniscule perf gain. Ideally all our code with literal strings would attributed as 'const' and if they use runtime values be attributed readonly but I do not have the luxury of making that change across the board. What would be nice if we start doing it right new code and change code as we come across it.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
udeeshagautam: So the issue with Assert.Equal is that it doesnt let me give a message in case of failure... which is generally good clarity and for finding where the issue happened.

Sure will chang to string interpolation

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
udeeshagautam: in that case the old code path hits. so my idea here is, since this is a very crutial path I want the new code path to hit in minimal possible cases thats why checking before assigning.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
udeeshagautam: Yeah I agree, I did that before.. but realized while running that false value doesnt fail the test intemittenlty... hence changed to assert model for certainty

test/Microsoft.SqlTools.ManagedBatchParser.IntegrationTests/BatchParser/BatchParserTests.cs
udeeshagautam: will do

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
ranasaria: Adding a comment like the following would help with the maintainability:
            if (settings.IsSqlCmdMode)
            {
                conditions = new ExecutionEngineConditions() { IsSqlCmd = settings.IsSqlCmdMode };
            }
             // else  conditions remains null and is handled correctly by parser.GetBatches to continue to nun in non-SqlCmdMode correctly

PR #840, opened 2019-08-01
Add serialization service support

kevcunnane:
Add serialization service support
- Added a method that handles serialization requests
- Support splitting save over multiple requests to reduce overall message size
- Added unit tests

String changes used a new version of the string tool for generation.
Will publish PR separately for the changes to build & localization
so this can run on Mac without .Net Core 1.0

Not covered yet:
- Excel testing. This is a bit complex to set up compared to the others
- Negative tests (exception cases).


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
```diff
@@ -0,0 +1,256 @@
...
+            }
+            return columnWrappers;
+        }
+
+
+        public SerializeDataResult ProcessRequest(SerializeDataRequestParams serializeParams)
+        {
+            SerializeDataResult result = new SerializeDataResult()
+            {
+            };
```
Charles-Gagnon: remove?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
```diff
@@ -0,0 +1,256 @@
...
+            try
+            {
+                this.WriteData(serializeParams.Rows, serializeParams.IsComplete);
+                if (serializeParams.IsComplete)
+                {
+                    this.CloseStreams();
+                }
```
Charles-Gagnon: This seems odd to do the cleanup here - wouldn't most things only call this request once? I would expect that after serialization is complete the streams would be automatically closed

Charles-Gagnon: Ok so looking at it more it seems like this will get multiple requests in batches for a single file (which makes sense). I guess this is fine but I'm not completely sold on the IsComplete name - IsLast seemed to make more sense to me. But IsLast isn't great either - maybe IsLastBatch? I'd also put a comment about keeping the streams open until we get told that all the data is transferred since it's otherwise unclear that this is a batched operation. 

---
In reply to: [309756726](https://github.com/microsoft/sqltoolsservice/pull/840#discussion_r309756726) [](ancestors = 309756726)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
```diff
@@ -0,0 +1,256 @@
...
+            RequestContext<SerializeDataResult> requestContext)
+        {
+            try
+            {
+                Validate.IsNotNull(nameof(serializeParams), serializeParams);
+                Validate.IsNotNullOrWhitespaceString("FilePath", serializeParams.FilePath);
+                DataSerializer serializer = null;
+                bool hasSerializer = inProgressSerializations.TryGetValue(serializeParams.FilePath, out serializer);
+                if (!hasSerializer)
+                {
```
Charles-Gagnon: Is SqlToolsService actually multithreaded? If so this has a small chance for a race condition - if there isn't a serializer when we set hasSerializer but then another thread comes in and makes one before AddOrUpdate is called then we'll end up having two serializers for the same file. And generally this kind of check is an anti-pattern anyways.

I'd just create the new serializer and use TryAdd (ignoring if it already exists), simpler and avoids the possible race condition

kevcunnane: So my current thinking here is that the client will wait on the result from each call before sending more across. I believe given this, it should be fine.
I think to fix, we'd want to separate into 2 messages (start and update) and fail on start if another serializer existed. That would handle cases, for example, where folks clicked save twice. Let me take a look at this and think on it - this may be more robust against multiple requests.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
kevcunnane: Ack, will consider renaming.

PR #841, opened 2019-08-01
Remove .Net Core 1.0 requirement for Loc and fix mac build

kevcunnane:
Use StringResourceTool to remove .Net Core 1.0 dependency for Loc
Fixed Mac build by copying the code used in internal repo for build.
- This now mostly matches internal, so expect this should work OK.


build.cake
```diff
@@ -554,12 +554,24 @@ Task("SetPackageVersions")
 Task("SRGen")
 	.Does(() =>
 {
+    // save current working directory to restore at end of task
+    var taskStartedWorkingDirectory = System.IO.Directory.GetCurrentDirectory();
+    var relativePathPrefix = "." + System.IO.Path.DirectorySeparatorChar.ToString();
```
Charles-Gagnon: You shouldn't need to add the separator char here - just using "." in Path.Combine should work fine.

build.cake
```diff
@@ -554,12 +554,24 @@ Task("SetPackageVersions")
...
+
     var projects = System.IO.Directory.GetFiles(sourceFolder, "*.csproj", SearchOption.AllDirectories).ToList();
     var locTemplateDir = System.IO.Path.Combine(sourceFolder, "../localization");
```
Charles-Gagnon: Hm, did you verify this works on macs? Anyways might be good to fix it up and just make it

```
Path.Combine(sourceFolder, "..", "localization");
```

build.cake
```diff
@@ -635,6 +648,9 @@ Task("SRGen")
             templateDoc.SaveAsResX(newPath.Replace("xlf","resx"));        
         }
     }
+
+    // restore the original working directory
+    System.IO.Directory.SetCurrentDirectory(taskStartedWorkingDirectory);
```
Charles-Gagnon: Put this in a try/finally

build.cake
kevcunnane: Good point - will fix. FYI I'm porting this over from @kenvanhyning 's fix in other repo, so we should likely update there too.

build.cake
kevcunnane: It's a mac-related fix 😄 so definitely tested there!

build.cake
kevcunnane: Will do

PR #842, opened 2019-08-04
Need to match the command field in the vscode language service client data contract

shengyfu:


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/Completion.cs
```diff
@@ -58,7 +58,7 @@ public class Command
         /// <summary>
         /// The identifier of the actual command handler, like `vsintellicode.completionItemSelected`.
         /// </summary>
-        public string CommandStr { get; set; }
+        public string command { get; set; }
```
Charles-Gagnon: You have other places in the code that reference this value - make sure you're updating them (and verifying the tests still pass)

Charles-Gagnon: Casing shouldn't matter so let's have this be capitalized in keeping with C# styling.

Charles-Gagnon: Let's also add Title and Tooltip as well to fully complete the vscode spec.

shengyfu: Because the class name is also Command, I had to use command for the field to avoid naming conflict.

shengyfu: Fixed the test.

shengyfu: Where are Title/Tooltip defined in the spec?

Charles-Gagnon: This is for the Command interface - which is what you're modifying here. 

https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts#L19

shengyfu: Added Title/Tooltip in the Command interface.

PR #845, opened 2019-08-08
Update dacfx to latest preview

udeeshagautam:
This brings in 
- Changes to fix SC bugs : https://github.com/microsoft/azuredatastudio/issues/6174 and https://github.com/microsoft/azuredatastudio/issues/6173
- AAD change for SC

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
@@ -116,6 +116,10 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
 
         public int CommandTimeout { get; set; } = 120;
 
+        public int LongRunningCommandTimeout { get; set; } = 0;
+
+        public int DatabaseLockTimeout { get; set; } = 60;
+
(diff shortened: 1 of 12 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -185,7 +191,7 @@ public class DeploymentOptions
                 ObjectType.LinkedServerLogins,
                 ObjectType.Endpoints,
                 ObjectType.ErrorMessages,
-                ObjectType.Filegroups,
+                ObjectType.Files,
                 ObjectType.Logins,
```
pensivebrian: Should we leave this to be explicit?

udeeshagautam: Both will be present in options allowing user to change... only that Files will be excluded and groups included by default (following SSDT defaults).

(Files should not be by default included because they in most cases fail the deployment - one will have to explicitly exclude the changes regarding db files for it to work. Earlier it used to work because Filegroup meant both - files and filegroups)

pensivebrian: I see. Thanks makes sense. Thanks!

PR #846, opened 2019-08-08
Fix serialization tests & don't block thread

kevcunnane:
- Fixed potential null ref when closing streams
- Always clean up serialization queue if an error occurs
- Stop blocking dispatcher thread by not awaiting task that processes the message
- Improved error logging in EventFlowValidator to help debug issues

My hope is this will result in the tests reliably passing now.
Will monitor in GCI to verify

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
@@ -20,5 +20,10 @@ namespace Microsoft.SqlTools.Hosting.Contracts
         /// Error message
         /// </summary>
         public string Message { get; set; }
+
+        public override string ToString()
+        {
+            return string.Format("Error(Code={0},Message='{1}'", Code, Message);
+        }
     }
 }
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
@@ -23,7 +23,7 @@ namespace Microsoft.SqlTools.Hosting.Contracts
 
         public override string ToString()
         {
-            return string.Format("Error(Code={0},Message='{1}'", Code, Message);
+            return $"Error(Code={Code},Message='{Message}'";
         }
     }
 }
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
@@ -23,7 +23,7 @@ namespace Microsoft.SqlTools.Hosting.Contracts
 
         public override string ToString()
         {
-            return $"Error(Code={Code},Message='{Message}'";
+            return $"Error(Code={Code},Message='{Message}')";
         }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
@@ -48,7 +48,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
             Task.Run(async () => {
                 await RunSerializeStartRequest(serializeParams, requestContext);
             }).ContinueWithOnFaulted(async t => await SendErrorAndCleanup(serializeParams?.FilePath, requestContext, t.Exception));
-            return Task.FromResult<bool>(true);
+            return Task.CompletedTask;
         }
 
         internal async Task RunSerializeStartRequest(SerializeDataStartRequestParams serializeParams, RequestContext<SerializeDataResult> requestContext)
@@ -91,6 +91,11 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                 {
                     DataSerializer removed;
                     inProgressSerializations.TryRemove(filePath, out removed);
+                    if (removed != null)
+                    {
+                        // Flush any contents to disk and remove the writer
+                        removed.CloseStreams();
+                    }
                 }
                 catch
                 {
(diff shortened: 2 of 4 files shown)


src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
```diff
@@ -20,5 +20,10 @@ public class Error
         /// Error message
         /// </summary>
         public string Message { get; set; }
+
+        public override string ToString()
+        {
+            return string.Format("Error(Code={0},Message='{1}'", Code, Message);
```
Charles-Gagnon: >( [](start = 39, length = 1)

Missing )

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
```diff
@@ -40,93 +41,106 @@ public override void InitializeService(IProtocolEndpoint serviceHost)
...
+            if (filePath != null)
+            {
+                try
+                {
+                    DataSerializer removed;
+                    inProgressSerializations.TryRemove(filePath, out removed);
+                }
+                catch
+                {
+                    // Do not care if there was an error removing this, must always delete if something failed
```
Charles-Gagnon: >must always delete if something failed [](start = 72, length = 38)

You're not actually deleting the file though - is that intentional?

alanrenmsft:
Charles pointed out that you are not actually deleting the file, looks good other than that.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
```diff
@@ -40,93 +41,106 @@ public override void InitializeService(IProtocolEndpoint serviceHost)
...
         /// Begin to process request to save a resultSet to a file in CSV format
         /// </summary>
-        internal async Task HandleSerializeStartRequest(SerializeDataStartRequestParams serializeParams,
+        internal Task HandleSerializeStartRequest(SerializeDataStartRequestParams serializeParams,
             RequestContext<SerializeDataResult> requestContext)
+        {
+            // Run in separate thread so that message thread isn't held up by a potentially time consuming file write
+            Task.Run(async () => {
+                await RunSerializeStartRequest(serializeParams, requestContext);
+            }).ContinueWithOnFaulted(async t => await SendErrorAndCleanup(serializeParams?.FilePath, requestContext, t.Exception));
```
Charles-Gagnon: So a potential problem is that now since this is truly async we'll immediately return  - at which point the caller may then call to continue serialization. But the start request may not actually have started/finished yet and thus we have a race condition if the continue serialization call starts trying to do stuff assuming that the start request has finished.

Same issue with the continue calls - those could then get out of order or start causing problems trying to write at the same time

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
```diff
@@ -40,93 +41,106 @@ public override void InitializeService(IProtocolEndpoint serviceHost)
...
+            return Task.FromResult<bool>(true);
```
ranasaria: For  Task with no return value expected it suggest returning Task.CompletedTask instead of Task.FromResult<bool>(true);

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
kevcunnane: So, this is dependent on the caller waiting on responses. That's what I've implemented in ADS - we wait on the result to come back, then send the next request. Since each one is awaited, we will get correct ordering. Right now I would treat this as "by design" because:
- the cost to implement differently would be a lot higher since you then need an ordered queue model to track
- most importantly since we're sending potentially large amounts of data, all of which are kept in memory until serialized, I explicitly do not want to encourage "send 5 messages in a row" and instead require "send message, wait until processed, send next message".

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
kevcunnane: Thanks! WIll change.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
Charles-Gagnon: Oh right, yeah my mistake. Got confused about the return value here being the indication that the request was done. Nevermind!

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
kevcunnane: Great question. I have a gap here for sure. On consideration, I will ensure the write stream is closed out in this case and intentionally leave file contents that have been written as-is. The thinking is that for most formats (CSV, JSON) users are going to be happier with half the written contents on an unexpected error than none.

src/Microsoft.SqlTools.Hosting/Hosting/Contracts/Error.cs
ranasaria: nit: suggest using string interpolation instead of string.Format, it is more modern and more readable as in:
return $"Error(Code={Code}, Message='{Message}'";

kevcunnane: Fixed both. Didn't know about string interpolation (show's I'm off my C# game in the last year or two) so thanks for recommending! We should consider updating across the codebase

ranasaria:
I suggest that leave a comment in the code about the stated ‘by design’ contract in this pull review thread.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
ranasaria: I suggest leaving a comment in the code about this stated 'by design' contract.

PR #848, opened 2019-08-13
Add execution options to the stream factory

aaomidi:
The output factory was never getting the new query options if it was made before.

Let me know if I'm making a mistake here!

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/IFileStreamFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/IFileStreamFactory.cs
@@ -18,5 +18,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
 
         void DisposeFile(string fileName);
 
+        SqlContext.QueryExecutionSettings QueryExecutionSettings { get; set; }
+
     }
 }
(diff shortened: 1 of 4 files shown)


kburtram:
@aaomidi the change looks fine conceptually to me, but I'm not sure what specific problem this is solving.  Are there cases where the query execution code isn't getting the user's settings?

aaomidi:
Yeah, the sqltoolsservice doesn't register most of the settings in: https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs

Except a few specific ones in random places:

https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs#L120

https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs#L524-L534

https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs#L540-L567

However, the settings that don't have a special condition written for them in the Query file don't actually get applied.

I ran into this issue with not being able to increase the XML with MaxXmlCharsToStore setting.

kburtram:
@aaomidi I'll merge this now based on discussion in DRI meeting yesterday.

PR #849, opened 2019-08-14
Use new DMV for big data cluster check

kevcunnane:


src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
```diff
@@ -47,7 +47,9 @@ @rc 			int
 ";
 
     public const string GetOsVersion = @"SELECT OSVersion = RIGHT(@@version, LEN(@@version)- 3 -charindex (' on ', LOWER(@@version)))";
-    public const string GetClusterEndpoints = @"IF OBJECT_ID (N'master.dbo.cluster_endpoint_info') IS NOT NULL
+    public const string GetClusterEndpoints = @"IF OBJECT_ID (N'sys.dm_cluster_endpoints') IS NOT NULL
```
Charles-Gagnon: So if the DMV doesn't exist then it looks like you're just going to return an empty result set instead of falling back to the old logic since it won't actually error out. If we really want to keep supporting old CTPs (is that actually necessary? Alan said we didn't need to do that for the dashboard) then you'll want to change this to take that into account.

Charles-Gagnon: Or at least I think it will..not actually sure the behavior off hand of calling .Read() when there isn't a result set. Maybe it'll work as intended?

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -869,6 +859,48 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
...
+            {
+                ExecuteReader(
+                    connection,
+                    SqlConnectionHelperScripts.GetClusterEndpoints,
+                    delegate (IDataReader reader)
+                    {
+                        while (reader.Read())
+                        {
+                            clusterEndpoints.Add(new ClusterEndpoint {
+                                ServiceName = reader.GetString(0),
```
Charles-Gagnon: So we have to also update all the calling code to correctly check all the options? Since new clusters will have these new properties filled out while old ones will have the old ones instead...we should definitely make sure we need to support older CTPs in this way and if so consider trying to make this easier to work with (do some parsing on our end here to fill in the IP/Port for example)

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
kevcunnane: Oh thanks! Will verify this errors as I need that to happen. Reason being, this would immediately break all existing CTP 3.2 customers and we don't want that. I think "support last month's release" is the acceptable limit for us. In my other review I'm deleting older-then-3.2 code support.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
kevcunnane: Yes, the calling code is on the Azure Data Studio side. I've put in the compensating logic there and verified it still works when falling back to the old data. Need to verify things work with then new shape too 😄 Reason I'm not doing more on this side is that this should be short-term, after 1 month we can just remove the older code and not do compensation.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
kevcunnane: Verified you were right and updated.

PR #854, opened 2019-08-22
fall back to master when connected to secondary

shueybubbles:
If the database name in the connection string doesn't match any database found in sys.databases, fall back to master for the database node in Object Explorer.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -41,7 +41,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             if (context == null)
             {
                 base.EnsureContextInitialized();
-                Database db = SmoObject as Database;
+                var db = SmoObject as Database;
                 if (db != null)
                 {
                     context.Database = db;
@@ -75,7 +75,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             }
             catch (Exception ex)
             {
-                string error = string.Format(CultureInfo.InvariantCulture, "Failed to get IsAccessible. error:{0} inner:{1} stacktrace:{2}",
+                var error = string.Format(CultureInfo.InvariantCulture, "Failed to get IsAccessible. error:{0} inner:{1} stacktrace:{2}",
                     ex.Message, ex.InnerException != null ? ex.InnerException.Message : "", ex.StackTrace);
                 Logger.Write(TraceEventType.Error, error);
                 ErrorMessage = ex.Message;
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -19,13 +19,22 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
         {
             Parent = serverNode;
             NodeValue = databaseName;
-            Database db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, this.NodeValue);
+            var db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, this.NodeValue);
             db.Refresh();
+            // If we got this far, the connection is valid. However, it's possible
+            // the user connected directly to the master of a readable secondary
+            // In that case, the name in the connection string won't be found in sys.databases
+            // We detect that here and fall back to master
+            if (db.State == SqlSmoState.Creating)
+            {
+                db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, "master");
+                db.Refresh();
+            }
             CacheInfoFromModel(db);
         }
 
         /// <summary>
-        /// Initializes the context and ensures that 
+        /// Initializes the context and sets its ValidFor property 
         /// </summary>
         protected override void EnsureContextInitialized()
         {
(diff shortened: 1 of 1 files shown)


ranasaria:
Is the fallback to master a feature of SSMS currently?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -19,21 +19,30 @@ public DatabaseTreeNode(ServerNode serverNode, string databaseName): this()
         {
             Parent = serverNode;
             NodeValue = databaseName;
-            Database db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, this.NodeValue);
+            var db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, this.NodeValue);
```
ranasaria: >var [](start = 12, length = 3)

+1 on this var usage. The convention has been to avoid the use of var, though personally for local method specific and specially loop specific variable there are ok from my point of view. Especially when followed by a new or a cast expression that makes type abundantly clear.

There are a lot of articles on internet discussing the pros and cons :) so I will not go into those here, but we will need to be consistent within our team.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -19,21 +19,30 @@ public DatabaseTreeNode(ServerNode serverNode, string databaseName): this()
...
         /// <summary>
-        /// Initializes the context and ensures that 
+        /// Initializes the context and sets its ValidFor property 
         /// </summary>
         protected override void EnsureContextInitialized()
         {
             if (context == null)
             {
                 base.EnsureContextInitialized();
                 Database db = SmoObject as Database;
```
ranasaria: >Database [](start = 16, length = 8)

if you var in previous location then perhaps you should var here as well :)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -19,21 +19,30 @@ public DatabaseTreeNode(ServerNode serverNode, string databaseName): this()
         {
             Parent = serverNode;
             NodeValue = databaseName;
-            Database db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, this.NodeValue);
+            var db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, this.NodeValue);
             db.Refresh();
+            // If we got this far, the connection is valid. However, it's possible
+            // the user connected directly to the master of a readable secondary
+            // In that case, the name in the connection string won't be found in sys.databases
+            // We detect that here and fall back to master
```
Charles-Gagnon: Might be worth mentioning that we "detect" that by the fact that SMO will leave the DB as "Creating"

shueybubbles:
SSMS doesn't have this issue because its object explorer root is not database-scoped. It always displays the full result of sys.databases query regardless of whether the connection string has an initial catalog.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
shueybubbles: I don't see that the code suggests we're detecting it any other way

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
shueybubbles: I try to use var whenever it's in a type-obvious situation but i also don't really mind inconsistencies across a code base with it. As long as the code is expressive, most of the time it's ok.

ranasaria:
Is this behavior change a good change then? I worry that this might surprise (not necessarily pleasantly) the user. Consider a scenario where a user has database1, database2 and master on her server. Further assume that that database1 is production, database2 is pre-production and master is her development test instance, so all there has same user database objects and schema. Now when she mistypes database1 and database2 in her connection strings to db1 and db2 or some such, she silently gets connected to master and she may not realize that this has happened for quite some time since a lot of child nodes on the database would have the object that she expects!

shueybubbles:
if "initial catalog=somename" is in the string, and "somename" doesn't exist on the server at all, the connection attempt itself will fail.
if "somename" refers to the readable secondary master, the connection will succeed, but queries to sys.databases won't return "somename".

shueybubbles:
ie we still get this dialog: System.Data.SqlClient.SqlException (0x80131904): Cannot open database "myfunky" requested by the login. The login failed.
Login failed for user 'user'.
   at System.Data.SqlClient.SqlInternalConnectionTds..ctor(DbConnectionPoolIdentity identity, SqlConnectionString connectionOptions, SqlCredential credential, Object providerInfo, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance,

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
Charles-Gagnon: Yeah it's probably fine but I'm not sure someone not familiar with SMO would look at this and not be confused why the state would even be "Creating".

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
kevcunnane: Yeah our rule had been "unless you can see the "new Xyz" statement don't use var, but here it's clear (RHS is a new Database statement).

PR #855, opened 2019-08-22
Added renaming and pinning agent notebook runs

aasimkhan30:
Added pinning and renaming notebook runs


src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1410,6 +1412,64 @@ internal async Task HandleUpdateAgentNotebookRequest(UpdateAgentNotebookParams p
...
+                {
+                    result.Success = false;
+                    result.ErrorMessage = e.ToString();
+
+                }
+                await requestContext.SendResult(result);
+            });
+        }
+
+        internal async Task HandleUpdateAgentNotebookRunPinRequest(UpdateAgentNotebookRunPinParams parameters, RequestContext<ResultStatus> requestContext)
```
anthonydresser: So we are considering pinned to be a attribute of the notebook and not of the user's preference?

PR #858, opened 2019-08-29
Added delete materialized run endpoints 

aasimkhan30:
I've added delete materialized run API. 
Also modified some logic for rename and pin to support notebook runs that don't have a materialized notebook metadata in the database.


aasimkhan30:
Hey @kburtram can you please take a look at this PR?

src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/AgentNotebookHelper.cs
```diff
@@ -620,26 +635,75 @@ UPDATE notebooks.nb_materialized
...
+            string targetDatabase)
+        {
+            string deleteMaterializedNotebookQuery =
+            @"
+            IF EXISTS
+            (SELECT * FROM notebooks.nb_materialized 
+            WHERE job_id = @jobId AND run_time = @startTime AND run_date = @startDate)
+            BEGIN
+                UPDATE notebooks.nb_materialized 
+                SET is_deleted = 1,
```
kburtram: so we don't want to actually delete the rows?  I think either design is fine, though, my first thought was that deleted would actually remove the data.

src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/AgentNotebookHelper.cs
```diff
@@ -620,26 +635,75 @@ UPDATE notebooks.nb_materialized
...
             updateMaterializedNotebookPinParams.Add(new SqlParameter("notebookPin", pin));
-            updateMaterializedNotebookPinParams.Add(new SqlParameter("materializedId", materializedId));
             await AgentNotebookHelper.ExecuteSqlQueries(
                 connInfo,
                 updateMaterializedNotebookPinQuery,
                 updateMaterializedNotebookPinParams,
                 targetDatabase);
         }
+
+        public static async Task DeleteMaterializedNotebook(
```
kburtram: Add a comment block describing the delete logic.

PR #862, opened 2019-09-06
Feat/notebookendpoints

aasimkhan30:
Added delete notebook endpoints and made some changes queries to handle space sensitive databases.


Charles-Gagnon:
>     /// SQL Agent Notebook templates params 

Looks like these doc comments should be updated - most of the ones below here seem to be inaccurate

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs:217 in 77f9b0f. [](commit_id = 77f9b0f43222d3c94d0f406b30adcd30a53b9fd0, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -256,4 +256,32 @@ public static readonly
...
+        public string TargetDatabase { get; set; }
+        public AgentNotebookHistoryInfo agentNotebookHistory { get; set; }
+    }
+
+    /// <summary>
+    /// SQL Agent Notebook materialized result
+    /// </summary>
+    public class DeleteNotebookMaterializedResult : ResultStatus
+    {
+        public string NotebookMaterialized { get; set; }
```
Charles-Gagnon: >NotebookMaterialized [](start = 22, length = 20)

Is this supposed to be the name of the notebook that was deleted? Suggest having the name be MaterializedNotebook then - right now this sounds like it's telling us whether the notebook was materialized (which makes no sense in this context)

src/Microsoft.SqlTools.ServiceLayer/Agent/Contracts/AgentNotebookRequest.cs
```diff
@@ -256,4 +256,32 @@ public static readonly
...
+    }
+
+    /// <summary>
+    /// SQL Agent Notebook materialized request type
+    /// </summary>
+    public class DeleteNotebookMaterializedRequest
+    {
+        public static readonly
+            RequestType<DeleteNotebookMaterializedParams, ResultStatus> Type =
+            RequestType<DeleteNotebookMaterializedParams, ResultStatus>.Create("agent/deletenotebookmaterialized");
```
Charles-Gagnon: >deletenotebookmaterialized [](start = 86, length = 26)

Same here - this seems like it should probably be agent/deletematerializednotebook?

src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/AgentNotebookHelper.cs
```diff
@@ -200,7 +200,7 @@ INNER JOIN
                 notebook.JobId,
                 notebook.TargetDatabase);
 
-           
+
         }
```
Charles-Gagnon: nit: get rid of extra lines here

src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/AgentNotebookHelper.cs
```diff
@@ -567,79 +556,143 @@ UPDATE notebooks.nb_template
...
                 updateMaterializedNotebookPinQuery,
                 updateMaterializedNotebookPinParams,
                 targetDatabase);
         }
+
+        /// <summary>
+        /// Delete a particular run of the job. Deletion mainly including clearing out the notebook,
+        /// and notebook-error. The API doesn't delete the row because some notebook runs that have job
+        /// error in them don't have an entry in the materialized table. For keeping track of those notebook
+        /// runs the entry is added into the table with is_delete set to 1.
```
Charles-Gagnon: Why do we want to keep track of these? If we're deleting them isn't that implying we don't care about them anymore?

aasimkhan30: The API only deletes the artifacts created by notebook job history and not the history of the notebook job. So if the job has an error there won't be any entry in the materialized table. Therefore, just to remove the job from the UI I'm setting the 'is_deleted to false.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1470,7 +1470,36 @@ internal async Task HandleUpdateAgentNotebookRunPinRequest(UpdateAgentNotebookRu
...
+                        connInfo,
+                        parameters.agentNotebookHistory,
+                        parameters.TargetDatabase);
+                    result.Success = true;
+                }
+                catch (Exception e)
+                {
+                    result.Success = false;
+                    result.ErrorMessage = e.ToString();
+
```
Charles-Gagnon: Normally you'd just do await result.SendError(e);

Charles-Gagnon: Make an issue to make these exceptions call SendError instead

---
In reply to: [321924046](https://github.com/microsoft/sqltoolsservice/pull/862#discussion_r321924046) [](ancestors = 321924046)

PR #864, opened 2019-09-10
Fix DW extract failing if connection is to master

kisantia:
Extract of a DW database was failing with “Error: Could not extract package from specified database. Catalog view ‘sysconfigures’ is not supported in this version” if the connection was to master rather than directly to the DW database. This fixes that by setting the database name to the DW database so that the connection string gets created with the DW database name instead of master, which is then passed to DacServices to perform the extract.


kisantia:
@udeeshagautam nope on-prem and Azure exports still work with this change :)

PR #865, opened 2019-09-11
Switch Tools Service to new SMO and Microsoft.Data.SqlClient driver

shueybubbles:
We have a new SQL client driver, Microsoft.Data.SqlClient. NetStandard SMO and NetCore Sql Tools Service will be our first consumers of this new driver, which supports UTF-8 collations more completely as well as some more features like Always Encrypted with enclaves.

I need to finish running integration tests before I submit.


src/Microsoft.SqlTools.CoreServices/Microsoft.SqlTools.CoreServices.csproj
```diff
@@ -1,5 +1,4 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="../../Common.props" />
```
kevcunnane: Why is nothing importing common.props now? Don't we need settings from this?

src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
```diff
@@ -16,11 +15,6 @@
     <RuntimeIdentifiers>win7-x64;win7-x86;ubuntu.14.04-x64;ubuntu.16.04-x64;centos.7-x64;rhel.7.2-x64;debian.8-x64;fedora.23-x64;opensuse.13.2-x64;osx.10.11-x64;linux-x64</RuntimeIdentifiers>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="System.Data.SqlClient" />
-  </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="System.Data.SqlClient" Version="4.6.0" />
```
kevcunnane: Does the SqlClient package reference now come in transitively via SMO?

src/Directory.Build.props
```diff
@@ -0,0 +1,10 @@
+<Project>
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```
kevcunnane: Just so I can understand - does this props file get auto-imported somehow by all other projects? I don't understand how this part works so would be nice to understand.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -826,17 +826,24 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
                 // Get BDC endpoints
                 if (!serverInfo.IsCloud)
                 {
-                    List<ClusterEndpoint> clusterEndpoints = new List<ClusterEndpoint>();
-                    serverInfo.Options.Add(ServerInfo.OptionClusterEndpoints, clusterEndpoints);
-
-                    try
+                    if (serverInfo.ServerMajorVersion >= 15)
```
kevcunnane: Thanks for adding this check, appreciate you doing this given we missed it.

src/Microsoft.SqlTools.CoreServices/Microsoft.SqlTools.CoreServices.csproj
shueybubbles: msbuild 15+ automatically imports Directory.Build.props and Directory.Build.targets files it finds in the current directory or a parent directory.

src/Directory.Build.props
shueybubbles: yes, Directory.Build.props and Directory.Build.targets are automatically included by msbuild/dotnet sdk.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -548,18 +548,20 @@ private void SetScriptBehavior(ScriptingOptions options)
...
-            map.Add(SqlScriptOptions.ScriptCompatabilityOptions.Script120Compat.ToString(), SqlServerVersion.Version120);
-            map.Add(SqlScriptOptions.ScriptCompatabilityOptions.Script110Compat.ToString(), SqlServerVersion.Version110);
-            map.Add(SqlScriptOptions.ScriptCompatabilityOptions.Script105Compat.ToString(), SqlServerVersion.Version105);
-            map.Add(SqlScriptOptions.ScriptCompatabilityOptions.Script100Compat.ToString(), SqlServerVersion.Version100);
-            map.Add(SqlScriptOptions.ScriptCompatabilityOptions.Script90Compat.ToString(), SqlServerVersion.Version90);
-
-            return map;
+            return new Dictionary<string, SqlServerVersion>
+            {
+                {SqlScriptOptions.ScriptCompatibilityOptions.Script150Compat.ToString(), SqlServerVersion.Version150},
```
kevcunnane: Wait is this why scripting External Tables was failing in some cases?

src/Microsoft.SqlTools.Credentials/Microsoft.SqlTools.Credentials.csproj
shueybubbles: no the <PackageReference> is in Directory.Build.props now

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
shueybubbles: I don't think so, I just noticed 150 wasn't here.
Eric's funky external table that doesn't script also fails to script in SSMS, I haven't had time to look into it.

kevcunnane: Well that's better than "it fails in ADS only". Have we looped in the folks who added syntax support to target fixing that scripting scenario?

PR #874, opened 2019-10-07
Create the file path on the tool service

aaomidi:
We need to create this path on the tooling side instead of ADS.

Fixes --- https://github.com/microsoft/azuredatastudio/pull/7523

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
@@ -129,6 +129,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Scripting
 
                 if (parameters.FilePath == null)
                 {
+                    // Create a temporary file to handle this request
                     parameters.FilePath = Path.Join(Path.GetTempPath(), $"{Convert.ToBase64String(Guid.NewGuid().ToByteArray()).Substring(0, 8)}-{DateTime.Now:yyyy-MM-dd_hh-mm-ss-fff}");
                 }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
@@ -130,7 +130,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Scripting
                 if (parameters.FilePath == null)
                 {
                     // Create a temporary and random path to handle this operation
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -126,6 +127,11 @@ public async Task HandleScriptExecuteRequest(ScriptingParams parameters, Request
                     }
                 }
 
+                if (parameters.FilePath == null)
+                {
+                    parameters.FilePath = Path.Join(Path.GetTempPath(), $"{Convert.ToBase64String(Guid.NewGuid().ToByteArray()).Substring(0, 8)}-{DateTime.Now:yyyy-MM-dd_hh-mm-ss-fff}");
```
llali: please add a comment to explain what the path would look like and reason for it.

aaomidi: A comment on the code or here?

llali: Also please explain in the PR how this is going to fix the original issue

llali: in the code. it's useful to have a sample in a comment, makes it more clear. also may add some comment about why you did that. trying to create a random file name?

aaomidi:
We're currently generating the file path for this operation on the frontend of ADS. This is causing issues for our exploration work and we need to generate it either on the mssql extension on ADS or the tools service. 

Moving this here allows me to get rid of the import 'os'; on the area of the app where it shouldn't have one.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
kburtram: @aaomidi any reason to not use [GetTempFileName](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=netcore-3.0)?

kburtram: @aaomidi double checking the docs, it looks like [GetRandomFileName](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrandomfilename?view=netcore-3.0) is the one I'm asking about.  GetTempFileName appears to actually create a file, which isn't what we want I think.

aaomidi:
Nope! I just don't know C# as well as I should @kburtram

PR #876, opened 2019-10-11
Adding Support for new engine edition

BranislavGrbicMDCS:
Adding supporting changes for new engine edition.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.tt
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.tt
@@ -309,6 +309,14 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 	   {
 			flags.Add("ValidForFlag.NotSqlDw");
        }
+	   if (validForStr.Contains("SqlOnDemand"))
+	   {
+			flags.Add("ValidForFlag.SqlOnDemand");
+	   }
+	   if (validForStr == "NotSqlDemand")
+	   {
+			flags.Add("ValidForFlag.NotSqlDemand");
+	   }
 	   if (validForStr == "All")
 	   {
 			flags.Add("ValidForFlag.All");
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.tt
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.tt
@@ -23,7 +23,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 <#
 	var directory = Path.GetDirectoryName(Host.TemplateFile);
 	string xmlFile = Path.Combine(directory, "SmoQueryModelDefinition.xml");
-    
+	
 	/////////
 	// Now generate all the Query methods
 	/////////
@@ -58,7 +58,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 		WriteLine("");
 
 		// Query impl
- 		WriteLine("public override  IEnumerable<SqlSmoObject> Query(SmoQueryContext context, string filter, bool refresh, IEnumerable<string> extraProperties)");
+		WriteLine("public override  IEnumerable<SqlSmoObject> Query(SmoQueryContext context, string filter, bool refresh, IEnumerable<string> extraProperties)");
 		WriteLine("{");
 		PushIndent(indent);
 
@@ -157,7 +157,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 		}
 		return typesList.ToArray();
 	}
-    
+	
 	public static XmlElement GetNodeElement(string xmlFile, string nodeName)
 	{
 		XmlDocument doc = new XmlDocument();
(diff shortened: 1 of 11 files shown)


test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ServerVersionHelperTests.cs
```diff
@@ -151,6 +160,17 @@ private void VerifyCalculateServerType(string serverVersion, SqlServerType expec
...
 
+        private void VerifyCalculateServerTypeForEngineEdition(int engineEdition, SqlServerType expected)
+        {
+            ServerInfo serverInfo = new ServerInfo
+            {
+                EngineEditionId = engineEdition
+            };
+
+            SqlServerType actual = ServerVersionHelper.CalculateServerType(serverInfo);
+            Assert.Equal(expected, actual);
```
shueybubbles: please add messages for asserts. These failures are annoyingly hard to debug without them.

BranislavGrbicMDCS: Per offline chat - unfortunately there isn't support for adding message for Assert.Equal.

BranislavGrbicMDCS: Per offline chat - added workaround since Assert.Equal doesn't support adding messages.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ServerVersionHelperTests.cs
```diff
@@ -168,7 +168,8 @@ private void VerifyCalculateServerTypeForEngineEdition(int engineEdition, SqlSer
             };
 
             SqlServerType actual = ServerVersionHelper.CalculateServerType(serverInfo);
-            Assert.Equal(expected, actual);
+            
+            Assert.True(actual.ToString().Equals(expected.ToString()), "Verify correct engine edition number");
```
shueybubbles: maybe print expected and actual values in the message

Charles-Gagnon: Yes, please do. The message is mostly useless if you don't also have the expected/actual values since you'd have to debug in to find that out then regardless.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.tt
```diff
@@ -309,6 +309,14 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 	   {
 			flags.Add("ValidForFlag.NotSqlDw");
        }
+	   if (validForStr.Contains("SqlOnDemand"))
+	   {
+			flags.Add("ValidForFlag.SqlOnDemand");
+	   }
+	   if (validForStr == "NotSqlDemand")
```
Charles-Gagnon: >== [](start = 20, length = 2)

You should be doing a .Contains here

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
```diff
@@ -70,7 +70,7 @@
     <Child Name="Security"/>
   </Node>
 
-  <Node Name="Tables" LocLabel="SR.SchemaHierarchy_Tables" BaseClass="ModelBased" Strategy="MultipleElementsOfType" ChildQuerierTypes="SqlTable" TreeNode="TableTreeNode">
+  <Node Name="Tables" LocLabel="SR.SchemaHierarchy_Tables" BaseClass="ModelBased" Strategy="MultipleElementsOfType" ChildQuerierTypes="SqlTable" TreeNode="TableTreeNode" ValidFor="NotSqlDemand">
```
Charles-Gagnon: >ValidFor [](start = 170, length = 8)

Hrm. If this was the first I'd say I'd rather see a NotValidFor property instead. But since it isn't I'll leave it up to you whether you want to do the extra work - feel free to leave it as is. Having to create separate Not* flags for each combination of "Not" is going to become annoying as we add more types though.

Charles-Gagnon:
>                 NodeValue = SR.SchemaHierarchy_ServerRoles,

How are server roles exposed for On Demand?

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeGenerator.cs:264 in f8eecc2. [](commit_id = f8eecc28be36229a82c0820024ca96ced54e2966, deletion_comment = False)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeGenerator.tt
```diff
@@ -391,6 +391,14 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 	   {
 			flags.Add("ValidForFlag.NotSqlDw");
        }
+	   if (validForStr.Contains("SqlOnDemand"))
```
Charles-Gagnon: Same with this file - mind cleaning it up and spacifying the doc?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ServerVersionHelperTests.cs
```diff
@@ -151,6 +160,18 @@ private void VerifyCalculateServerType(string serverVersion, SqlServerType expec
...
+        {
+            ServerInfo serverInfo = new ServerInfo
+            {
+                EngineEditionId = engineEdition
+            };
+
+            SqlServerType actual = ServerVersionHelper.CalculateServerType(serverInfo);
+            
+            Assert.True(actual.ToString().Equals(expected.ToString()), "Verify correct engine edition number");
+        }
```
Charles-Gagnon: Why are you calling ToString? You should be able to just directly compare with ==

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.tt
```diff
@@ -309,6 +309,14 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 	   {
 			flags.Add("ValidForFlag.NotSqlDw");
        }
+	   if (validForStr.Contains("SqlOnDemand"))
```
Charles-Gagnon: >   [](start = 1, length = 2)

nit: while you're here could you turn the spacings into tabs? It's inconsistent for some reason in this one

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModel.tt
BranislavGrbicMDCS: I was using ValidForFlag.All as a roll model for this. Since it includes all other flags except Sql On Demand it seems ok to check for equality instead of .contains

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ServerVersionHelperTests.cs
```diff
@@ -151,6 +160,19 @@ private void VerifyCalculateServerType(string serverVersion, SqlServerType expec
...
 
+        private void VerifyCalculateServerTypeForEngineEdition(int engineEdition, SqlServerType expected)
+        {
+            ServerInfo serverInfo = new ServerInfo
+            {
+                EngineEditionId = engineEdition
+            };
+
+            SqlServerType actual = ServerVersionHelper.CalculateServerType(serverInfo);
+            string errorString = "Verify server type based on Engine Edition. Actual value: " + actual + ", Expected value: " + expected;
```
Charles-Gagnon: nit - but interpolated strings make this a lot easier to work with 

$"Verify server type based on Engine Edition. Actual value: {actual}, Expected value: {expected}"

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
BranislavGrbicMDCS: Definitely it would be useful. Unfortunately I don't believe that I have bandwidth for this right away, but we can sync about this.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
```diff
@@ -85,6 +85,7 @@
       <Property Name="IsExternal" ValidFor="Sql2016|Sql2017|AzureV12"/>
     </Properties>
     <Child Name="SystemTables" IsSystemObject="1"/>
+    <Child Name="ExternalTables"/>
   </Node>
```
Charles-Gagnon: I agree with Kevin - I'd prefer to keep existing behavior where external tables are just normal tables under the tables folder with an extra (External) tag added

PR #881, opened 2019-10-23
Schema compare include/exclude changes

kisantia:
Two main changes to the previous include/exclude behavior:
1. Fail exclude request if dependencies are still included. Check dependencies before the actual exclude call so that the include state doesn't get into a weird state where DacFx remembers when the exclude was attempted but failed, so it it excludes it when the dependent entry is excluded.
2. Send back the affected differences from successful include/exclude requests so that the table in ADS can be updated with the correct include/exclude states.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
@@ -10,6 +10,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
+using System.Linq;
 
 namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
 {
@@ -70,16 +71,7 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                 {
                     IEnumerable<SchemaDifference> dependencies = this.ComparisonResult.GetExcludeDependencies(node);
 
-                    bool block = false;
-                    foreach (SchemaDifference entry in dependencies)
-                    {
-                        if (entry.Included)
-                        {
-                            block = true;
-                            break;
-                        }
-                    }
-
+                    bool block = dependencies.Any(d => d.Included);
                     if (block)
                     {
                         this.Success = false;
@@ -93,16 +85,7 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                 if (this.Success)
                 {
                     IEnumerable<SchemaDifference> dependencies = this.ComparisonResult.GetIncludeDependencies(node);
-
-                    this.ChangedDifferences = new List<DiffEntry>();
-                    if (dependencies != null)
-                    {
-                        foreach (SchemaDifference difference in dependencies)
-                        {
-                            DiffEntry diffEntry = SchemaCompareUtils.CreateDiffEntry(difference, null);
-                            this.ChangedDifferences.Add(diffEntry);
-                        }
-                    }
+                    this.ChangedDifferences = dependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
                 }
             }
             catch (Exception e)
@@ -145,10 +128,12 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                 // Don't need to check if included is the same when verifying if the difference is equal
                 if (prop.Name != "Included")
                 {
-                    result = result &&
-                        ((prop.GetValue(diffEntry) == null &&
+                    if(!((prop.GetValue(diffEntry) == null &&
                         prop.GetValue(entryFromDifference) == null) ||
-                        prop.GetValue(diffEntry).SafeToString().Equals(prop.GetValue(entryFromDifference).SafeToString()));
+                        prop.GetValue(diffEntry).SafeToString().Equals(prop.GetValue(entryFromDifference).SafeToString())))
+                    {
+                        return false;
+                    }
                 }
             }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
@@ -40,7 +40,6 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
 
         public List<DiffEntry> ChangedDifferences;
 
-
         public SchemaCompareIncludeExcludeNodeOperation(SchemaCompareNodeParams parameters, SchemaComparisonResult comparisonResult)
         {
             Validate.IsNotNull("parameters", parameters);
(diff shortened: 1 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -97,10 +138,13 @@ private bool IsEqual(SchemaDifference difference, DiffEntry diffEntry)
             System.Reflection.PropertyInfo[] properties = diffEntry.GetType().GetProperties();
             foreach (var prop in properties)
             {
-                result = result &&
-                    ((prop.GetValue(diffEntry) == null &&
-                    prop.GetValue(entryFromDifference) == null) ||
-                    prop.GetValue(diffEntry).SafeToString().Equals(prop.GetValue(entryFromDifference).SafeToString()));
+                if (prop.Name != "Included")
```
udeeshagautam: good catch if I understand this correctly. Can you please add a comment here too.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -58,7 +61,45 @@ public void Execute(TaskExecutionMode mode)
...
+                    this.ChangedDifferences = new List<DiffEntry>();
+                    if (dependencies != null)
+                    {
+                        foreach (SchemaDifference difference in dependencies)
+                        {
+                            DiffEntry diffEntry = SchemaCompareUtils.CreateDiffEntry(difference, null);
+                            this.ChangedDifferences.Add(diffEntry);
+                        }
+                    }
+                }
```
udeeshagautam: Please add a comment that exclude just returns false if dependencies found and include actually adds dependencies - and this is parity with SSDT.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -38,6 +38,9 @@ class SchemaCompareIncludeExcludeNodeOperation : ITaskOperation
 
         public bool Success { get; set; }
 
+        public List<DiffEntry> ChangedDifferences;
+
```
udeeshagautam: nit- extra line

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
kisantia: added

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
kisantia: added

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -58,7 +65,45 @@ public void Execute(TaskExecutionMode mode)
                     throw new InvalidOperationException(SR.SchemaCompareExcludeIncludeNodeNotFound);
                 }
 
+                // Check first if the dependencies will allow this if it's an exclude request
+                if (!this.Parameters.IncludeRequest)
+                {
+                    IEnumerable<SchemaDifference> dependencies = this.ComparisonResult.GetExcludeDependencies(node);
+
+                    bool block = false;
```
pensivebrian: This foreach could be a single linq statement: bool block = dependencies.Any(d => d.Included);

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -58,7 +65,45 @@ public void Execute(TaskExecutionMode mode)
...
+                    IEnumerable<SchemaDifference> dependencies = this.ComparisonResult.GetIncludeDependencies(node);
+
+                    this.ChangedDifferences = new List<DiffEntry>();
+                    if (dependencies != null)
+                    {
+                        foreach (SchemaDifference difference in dependencies)
+                        {
+                            DiffEntry diffEntry = SchemaCompareUtils.CreateDiffEntry(difference, null);
+                            this.ChangedDifferences.Add(diffEntry);
+                        }
```
pensivebrian: Could you linqify this?
this.ChangedDifferences = dependencies.Select(d => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList()

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -97,10 +142,14 @@ private bool IsEqual(SchemaDifference difference, DiffEntry diffEntry)
...
             foreach (var prop in properties)
             {
-                result = result &&
-                    ((prop.GetValue(diffEntry) == null &&
-                    prop.GetValue(entryFromDifference) == null) ||
-                    prop.GetValue(diffEntry).SafeToString().Equals(prop.GetValue(entryFromDifference).SafeToString()));
+                // Don't need to check if included is the same when verifying if the difference is equal
+                if (prop.Name != "Included")
+                {
+                    result = result &&
```
pensivebrian: >result [](start = 29, length = 6)

Seems like we can optimize this by returning as soon as result is false, rather than continuing to iterate over the foreach loop.

pensivebrian: reflection is SLOW if you didn't ready know

---
In reply to: [338769969](https://github.com/microsoft/sqltoolsservice/pull/881#discussion_r338769969) [](ancestors = 338769969)

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -38,6 +38,8 @@ class SchemaCompareIncludeExcludeNodeOperation : ITaskOperation
 
         public bool Success { get; set; }
 
+        public List<DiffEntry> ChangedDifferences;
```
Benjin: Any particular reason for explicit accessors in the other changes, but not here?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
Benjin: ♥ Linq

PR #884, opened 2019-10-25
Fix schema compare exclude behaviour

kisantia:
Previously, we were blocking exclude calls if there were dependencies that were still included since we assumed that the exclude call would fail. However, this resulted in unexpected exclude failures because primary keys and nonclustered indexes were getting returned as dependencies and causing exclude calls to be blocked when it shouldn't. We were doing this check before attempting the actual exclude call so that the difference wouldn't get into a partial exclude state.

The fix is to just let the exclude call fail and send back the list of dependencies that caused the call to fail and have the partial excluded state behavior. The partial exclude state means that when the dependent that was blocking the exclude is excluded, then the partially excluded difference is also excluded.

includeExcludeFix

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
@@ -39,7 +39,9 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
 
         public bool Success { get; set; }
 
-        public List<DiffEntry> Dependencies;
+        public List<DiffEntry> AffectedDependencies;
+        public List<DiffEntry> BlockingDependencies;
+
 
         public SchemaCompareIncludeExcludeNodeOperation(SchemaCompareNodeParams parameters, SchemaComparisonResult comparisonResult)
         {
@@ -68,11 +70,19 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
 
                 this.Success = this.Parameters.IncludeRequest ? this.ComparisonResult.Include(node) : this.ComparisonResult.Exclude(node);
 
-                // create list of dependencies of this schema difference
-                // if not successful, send exclude dependencies that caused it to fail
                 // if successful, send dependencies that might have been affected by this request
-                IEnumerable<SchemaDifference> dependencies = this.Success ? this.ComparisonResult.GetIncludeDependencies(node) : this.ComparisonResult.GetExcludeDependencies(node);
-                this.Dependencies = dependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
+                if(this.Success)
+                {
+                    IEnumerable<SchemaDifference> affectedDependencies = this.ComparisonResult.GetIncludeDependencies(node);
+                    this.AffectedDependencies = affectedDependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
+                }
+                else
+                {
+                    // if not successful, send exclude dependencies that caused it to fail
+                    IEnumerable<SchemaDifference> blockingDependencies = this.ComparisonResult.GetExcludeDependencies(node);
+                    blockingDependencies = blockingDependencies.Where(difference => difference.Included == node.Included);
+                    this.BlockingDependencies = blockingDependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
+                }
             }
             catch (Exception e)
             {
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
@@ -70,18 +70,27 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
 
                 this.Success = this.Parameters.IncludeRequest ? this.ComparisonResult.Include(node) : this.ComparisonResult.Exclude(node);
 
-                // if successful, send dependencies that might have been affected by this request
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -66,27 +66,13 @@ public void Execute(TaskExecutionMode mode)
...
                 this.Success = this.Parameters.IncludeRequest ? this.ComparisonResult.Include(node) : this.ComparisonResult.Exclude(node);
 
-                // create list of affected dependencies of this request
-                if (this.Success)
-                {
-                    IEnumerable<SchemaDifference> dependencies = this.ComparisonResult.GetIncludeDependencies(node);
-                    this.ChangedDifferences = dependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
-                }
+                // create list of dependencies of this schema difference
+                // if not successful, send exclude dependencies that caused it to fail
```
Benjin: Are we relying on the caller to interpret the `Dependencies` property correctly based on success?  If so, it'd probably be safer to set different properties depending out output, e.g. `this.AffectedDependencies` and `this.DependenciesToExclude`.

kisantia: True. I'll make them two separate properties

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -66,27 +66,13 @@ public void Execute(TaskExecutionMode mode)
...
+                // if successful, send dependencies that might have been affected by this request
+                IEnumerable<SchemaDifference> dependencies = this.Success ? this.ComparisonResult.GetIncludeDependencies(node) : this.ComparisonResult.GetExcludeDependencies(node);
```
udeeshagautam: not sure I understand this why are we deciding whether to get exclude dependencies or include based on success? technically we can have include operation that failed too - right? Am I missing something here?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -66,26 +68,20 @@ public void Execute(TaskExecutionMode mode)
...
-
                 this.Success = this.Parameters.IncludeRequest ? this.ComparisonResult.Include(node) : this.ComparisonResult.Exclude(node);
 
-                // create list of affected dependencies of this request
-                if (this.Success)
+                // if successful, send dependencies that might have been affected by this request
+                if(this.Success)
+                {
+                    IEnumerable<SchemaDifference> affectedDependencies = this.ComparisonResult.GetIncludeDependencies(node);
+                    this.AffectedDependencies = affectedDependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
```
pensivebrian: >null [](start = 137, length = 4)

nit: named parameter for readability

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
```diff
@@ -66,26 +68,20 @@ public void Execute(TaskExecutionMode mode)
...
+                }
+                else
                 {
-                    IEnumerable<SchemaDifference> dependencies = this.ComparisonResult.GetIncludeDependencies(node);
-                    this.ChangedDifferences = dependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
+                    // if not successful, send exclude dependencies that caused it to fail
+                    IEnumerable<SchemaDifference> blockingDependencies = this.ComparisonResult.GetExcludeDependencies(node);
+                    blockingDependencies = blockingDependencies.Where(difference => difference.Included == node.Included);
+                    this.BlockingDependencies = blockingDependencies.Select(difference => SchemaCompareUtils.CreateDiffEntry(difference, null)).ToList();
```
pensivebrian: >null [](start = 137, length = 4)

nit: named parameter for readability

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeNodeOperation.cs
kisantia: changed this in the next commit. If successful, sends back dependencies that might have been affected from tthe GetIncludedDependencies() call. If not successful, send back the blocking dependents which are from the GetExcludedDependencies() call.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -864,8 +864,10 @@ public async void SchemaCompareIncludeExcludeWithDependencies()
                 t2ExcludeOperation.Execute(TaskExecutionMode.Execute);
                 Assert.False(t2ExcludeOperation.Success, "Excluding Table t2 should fail because view v1 depends on it");
                 Assert.True(t2ExcludeOperation.ComparisonResult.Differences.Where(x => x.SourceObject != null && x.SourceObject.Name.Parts[1] == "t2").First().Included, "Difference Table t2 should still be included because the exclude request failed");
+                Assert.True(t2ExcludeOperation.BlockingDependencies.Count == 1, "There should be one dependency");
+                Assert.True(t2ExcludeOperation.BlockingDependencies[0].SourceValue[1] == "v1",  "Dependency should be View v1");
```
udeeshagautam: not added in this PR but can we please add a constraint in the script for include exclude to cover that for any changes in general

PR #882, opened 2019-10-24
Explicitly erroring out on some failing commands that were not already included

udeeshagautam:
Explicitly erroring out on some failing commands that were not already included.

For better errors in microsoft/azuredatastudio#7851, microsoft/azuredatastudio#7381

PR #887, opened 2019-10-30
Revert "Explicitly erroring out on some failing commands that were not already included"

udeeshagautam:
Reverts microsoft/sqltoolsservice#882


udeeshagautam:
Checking since the PR failure was to do with timeout and not related to current change

PR #888, opened 2019-10-30
Bump SMO Version

Charles-Gagnon:
Changes for fixing issues with SQL On Demand support


Charles-Gagnon:
Incoming changes are going to make this unnecessary

PR #889, opened 2019-11-04
Revert to SMO with older SQLClient driver

kburtram:
Reverting back to System.Data.SqlClient while the Idle Connection Resiliency bug is investigated.


kburtram:
This puts master inline with what we've shipped in ADS Nov releases.

PR #892, opened 2019-11-11
[DRAFT] Revert "Revert to SMO with older SQLClient driver"

aaomidi:
Reverts microsoft/sqltoolsservice#889
Forgot to mark this as draft...


Charles-Gagnon:
Is there a reason you're making an (intended) draft PR vs just queuing builds with your topic branch separately?

aaomidi:
@Charles-Gagnon Because when you click the revert button it I think forces it into a PR - not really important either way.

PR #895, opened 2019-11-24
Fix auto-complete inserting brackets around function names

Charles-Gagnon:
Fixes https://github.com/microsoft/azuredatastudio/issues/7570

Adding brackets around reserved names was added in https://github.com/microsoft/sqltoolsservice/pull/191 - but we shouldn't be doing this for all declaration types since there's some things - such as functions - that are valid in SELECT/FROM/etc. clause lists.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
@@ -54,8 +54,9 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion
 
             // Bracket quote valid SQL names that aren't specified as reserved keywords already
             if (delimitedIdentifier == null && !string.IsNullOrEmpty(DeclarationTitle) && 
-                !ValidSqlNameRegex.IsMatch(DeclarationTitle) &&
-                !AutoCompleteHelper.IsReservedWord(InsertText))
+                (!ValidSqlNameRegex.IsMatch(DeclarationTitle) ||
+                AutoCompleteHelper.IsReservedWord(InsertText) && )
+                // https://github.com/Microsoft/vscode-mssql/issues/473
             {
                 InsertText = WithDelimitedIdentifier(BracketedIdentifiers, DeclarationTitle);
             }
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
@@ -20,7 +20,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion
     {
         private static Regex ValidSqlNameRegex = new Regex(@"^[\p{L}_@#][\p{L}\p{N}@$#_]{0,127}$");
         private static DelimitedIdentifier BracketedIdentifiers = new DelimitedIdentifier { Start = "[", End = "]" };
-        private static DelimitedIdentifier FunctionIdentifiers = new DelimitedIdentifier { Start = "", End = "()" };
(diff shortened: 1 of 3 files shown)


test/Microsoft.SqlTools.ServiceLayer.UnitTests/LanguageServer/SqlCompletionItemTests.cs
```diff
@@ -477,19 +181,18 @@ public void LabelShouldIncludeQuotedIdentifiersGivenTokenWithQuotedIdentifiers()
         }
 
         [Fact]
-        public void InsertTextShouldIncludeBracketGivenReservedName()
```
llali: >InsertTextShouldIncludeBracketGivenReservedName [](start = 20, length = 47)

still for reserve names we need bracket right? do we need a test for that scenario as well?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
```diff
@@ -48,14 +49,38 @@ public SqlCompletionItem(string declarationTitle, DeclarationType declarationTyp
...
+                {
+                    case DeclarationType.Server:
+                    case DeclarationType.Database:
+                    case DeclarationType.Table:
+                    case DeclarationType.Column:
+                    case DeclarationType.View:
+                    case DeclarationType.Schema:
+                        // Only quote if we need to - i.e. if this isn't a valid name (has characters that need escaping such as [) 
+                        // or if it's a reserved word
+                        if (!ValidSqlNameRegex.IsMatch(DeclarationTitle) || AutoCompleteHelper.IsReservedWord(InsertText)) {
```
kburtram: brace character `{` should be on its own line in C#

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/SqlCompletionItem.cs
```diff
@@ -48,14 +49,38 @@ public SqlCompletionItem(string declarationTitle, DeclarationType declarationTyp
...
+                            InsertText = WithDelimitedIdentifier(BracketedIdentifiers, DeclarationTitle);
+                        }
+                        break;
+                    case DeclarationType.BuiltInFunction:
+                    case DeclarationType.ScalarValuedFunction:
+                    case DeclarationType.TableValuedFunction:
```
kburtram: should there be a `default` case?

Charles-Gagnon: We already default InsertText to be the DeclarationTitle above so having a separate default here isn't necessary I think

test/Microsoft.SqlTools.ServiceLayer.UnitTests/LanguageServer/SqlCompletionItemTests.cs
Charles-Gagnon: Only if it's one of the types we specified - for example server/database/table/etc. 

We should only be bracket quoting things if it makes sense to do so, in these cases for identifiers that happen to share a name with a reserved word. 

I updated the bracket quoted test to check the reserved names too though since that should be checking that.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/LanguageServer/SqlCompletionItemTests.cs
```diff
@@ -13,303 +14,6 @@ namespace Microsoft.SqlTools.ServiceLayer.UnitTests.LanguageServer
 {
     public class SqlCompletionItemTests
     {
-        private static readonly string[] ReservedWords = new string[]
```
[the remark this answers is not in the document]
Charles-Gagnon: I don't know why they'd be different. My guess is that these weren't kept up to date with changes made to the original list - which is part of the reason I'm deleting it.

I don't know of a good reason to keep a separate list since we want these tests to be checking against that list of reserved words anyways.

PR #898, opened 2019-12-03
SqlCmd Connect/On Error/Include commands support

udeeshagautam:
Adding support for SqlCmd Connect, On Error, Include. 
This targets https://github.com/microsoft/azuredatastudio/issues/7381 but it is incremental and we still need other commands like Exec, Out, !! etc.


udeeshagautam:
@kburtram please take a look whenever you get a chance.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ConnectSqlCmdCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+
+
+        /// <summary>
+        /// attempts to establish connection with given params
+        /// </summary>
+        /// <returns>returns the connection object is successful, throws otherwise</returns>
+        public DbConnection Connect()
+        {
+            //create SqlConnectionInfo object
+            SqlConnectionInfo tempConnectionInfo = new SqlConnectionInfo();
```
ranasaria: nit: change the variable name to connectionInfo. I do not see anything temporary about the connection info.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ConnectSqlCmdCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+            if (ci == null || ci.ServerName == null)
+            {
+                return null;
+            }
+
+            IDbConnection conn = null;
+            try
+            {
+                string connString = ci.ConnectionString;
+                connString += ";Pooling=false"; //turn off connection pooling (this is done in other tools so following the same pattern)
```
ranasaria: why not set the SqlConnectionInfo.Pooled property instead before using ci.ConnectionString out. Is there a reason why this can't be set in the calling function (i.e in the Connect() method.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/BatchParserSqlCmd.cs
```diff
@@ -105,17 +105,78 @@ public override void SetVariable(PositionStruct pos, string name, string value)
...
+                return BatchParserAction.Abort;
+            }
+
+            filename.GetText(resolveVariables: true, text: out newFilename, lineInfo: out lineInfo);
+            string resolvedFileNameWithFullPath = GetFilePath(newFilename);
+
+            if (!File.Exists(resolvedFileNameWithFullPath))
+            {
+                stream = null;
+                return BatchParserAction.Abort;
```
ranasaria: Should we RaiseScriptError before we abort?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
```diff
@@ -501,7 +502,7 @@ private void DoScriptExecution(bool isBatchParser)
             }
             else
             {
-                ExecuteBatchInternal(script, /* num */ 1, /* lineNumber */ 0);
+                ExecuteBatchInternal(script, /* num */ 1, /* lineNumber */ 0, /* sqlcmdCommand required for parsing only*/ null);
```
ranasaria: Suggest using named parameters (comments like 'required for parsing only' can still be kept)

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -1,63 +1,63 @@
...
-    mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with 
-            : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
-            : and then encoded with base64 encoding.
-
-    mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array 
-            : using a System.ComponentModel.TypeConverter
-            : and then encoded with base64 encoding.
+  <!--
```
ranasaria: Is this just whitespace change?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ConnectSqlCmdCommand.cs
udeeshagautam: This code is copied from the current windows code (similar in SSDT/SSMS) for handling these commands and then trimmed that's why some things are the way they are, I will try to change and see the effect for this and other suggestions.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
udeeshagautam: no no the actual change is on line 325 for <data name="SqlCmdExitOnError" xml:space="preserve">. Not sure why the sopace change is showing like that seems like auto alining in VS. Will see if I can revert the spacing

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/BatchParserSqlCmd.cs
udeeshagautam: Validated that the error is raised in caller depending on abort result - if we raise here end user will get two errors

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ConnectSqlCmdCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+            return dbConnection;
+        }
+
+        /// <summary>
+        /// called when we need to establish new connection for batch executio as a
+        /// result of "connect" command processing
+        /// </summary>
+        /// <param name="ci"></param>
+        /// <returns></returns>
+        private DbConnection AttemptToEstablishCurConnection(SqlConnectionInfo ci)
```
kburtram: Can we use `ConnectionService::OpenSqlConnection` to open the SqlConnection?  Ideally the iteractions with SqlClient to open new connections are consolidated there to centralize that logic.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
```diff
@@ -522,6 +530,88 @@ private void ParseSetvar(Token setvarToken)
...
+                {
+                    password = ResolveVariables(LookaheadToken, 0, null);
+                    Accept();
+                    Accept(LexerTokenType.Whitespace);
+                }
+            }
+            return password;
+        }
+
+        private void ConnectAction()
```
kburtram: couldn't see where this is being used?

src/Microsoft.SqlTools.ManagedBatchParser/Localization/sr.resx
```diff
@@ -233,4 +233,16 @@
...
+  <data name="SqlCmd_PathInvalid" xml:space="preserve">
+    <value>Path contains invalid characters.</value>
+    <comment></comment>
+  </data>
+  <data name="SqlCmd_PathLong" xml:space="preserve">
+    <value>Path too long.</value>
+    <comment></comment>
+  </data>
+  <data name="SqlCmd_FailedInclude" xml:space="preserve">
+    <value>Could not find included file.</value>
```
kburtram: should the error message include the file name that couldn't be found?  or do we expect that to generally be obvious to the user?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -445,7 +464,11 @@ private async Task ExecuteInternal()
             catch (Exception e)
             {
                 HasErrored = true;
-                if (e is OperationCanceledException)
+                if (e is SqlCmdException || CancelledBySqlCmd)
+                {
+                    await BatchMessageSent(new ResultMessage("An error was encountered during execution of batch.Exiting.", false, null));
```
kburtram: don't need to localize this error message?  also missing a ' ' after the '.'

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/QueryExecution/ExecuteTests.cs
```diff
@@ -115,9 +115,10 @@ public void DatabaseChangesWhenCallingUseDatabase()
             Assert.True(multipleElapsedTime > elapsedTime);
         }
 
-        public Query CreateAndExecuteQuery(string queryText, ConnectionInfo connectionInfo, IFileStreamFactory fileStreamFactory)
+        public static Query CreateAndExecuteQuery(string queryText, ConnectionInfo connectionInfo, IFileStreamFactory fileStreamFactory, bool IsSqlCMD = false)
```
kburtram: nit: I think according to .Net naming conventions parameter should be named `IsSqlCmd`

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ConnectSqlCmdCommand.cs
udeeshagautam: So currently two things
1. The Batchparse --> Service layer reverse ref - but this can be solved by moving the OpenSqlConnection to ConnectionServiceCore
2. The second is a little weird that we need ConnectionInfo object for OpenSqlConnection which is owner URI based and is cached (it seems) but here we only want temporary connection for the length of batch execution and the overall query connection should remain same.

I will keep in mind for later in refactoring can help here - but for now will leave this one

PR #899, opened 2019-12-04
Support active directory pwd authentication

aaomidi:
Adds support for a new authentication model - username/password authentication with AAD.


aaomidi:
https://github.com/microsoft/azuredatastudio/pull/8565

PR #902, opened 2019-12-17
Bump dacfx nuget package for user fix

kisantia:
Bump DacFx nuget package to get fix for https://github.com/microsoft/azuredatastudio/issues/8715. Generate script and deploy were failing for schema compare if there was a user change included.


kisantia:
@kburtram no dependency version changes. Only two small bug fixes are the difference between this and the previous DacFx nuget package.

PR #905, opened 2020-01-02
DacFxService changes to stabilize ADS tests

kisantia:
This changes behavior of DacFx ExecuteOperation() to return the result after the operation completes instead of after it is started. This doesn't change how the ADS Dacpac wizard will behave because it didn't care about the result since it was shown in the Task Viewlet. This change will enable the ADS DacFx tests to be stable since the tests won't have to keep checking if the database and tables are created, which would often interfere with the DacFx deployment/import.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -67,7 +67,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
                 if (connInfo != null)
                 {
                     ExportOperation operation = new ExportOperation(parameters, connInfo);
-                    await ExecuteOperation(operation, parameters, SR.ExportBacpacTaskName, requestContext);
+                    ExecuteOperation(operation, parameters, SR.ExportBacpacTaskName, requestContext);
                 }
             }
             catch (Exception e)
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -221,21 +221,35 @@ await requestContext.SendResult(new GenerateDeployPlanRequestResult()
...
 
-        private async Task ExecuteOperation(DacFxOperation operation, DacFxParams parameters, string taskName, RequestContext<DacFxResult> requestContext)
+        private void ExecuteOperation(DacFxOperation operation, DacFxParams parameters, string taskName, RequestContext<DacFxResult> requestContext)
         {
-            SqlTask sqlTask = null;
-            TaskMetadata metadata = TaskMetadata.Create(parameters, taskName, operation, ConnectionServiceInstance);
-
-            // put appropriate database name since connection passed was to master
-            metadata.DatabaseName = parameters.DatabaseName;
+            Task.Run(async () =>
```
udeeshagautam: we don't need await here?

alanrenmsft: if you do await here, it will block the main handler thread.

udeeshagautam: closed offline

PR #908, opened 2020-01-13
Enable column encryption setting

Xtrimmer:
In ADS within the advanced properties of the connection details dialog, there is an option to enable or disable column encryption for the Always Encrypted feature.

image

Changing this option is currently a NOOP.

This change-set enables the functionality of this option by "_wiring it up_" in sqltoolsservice.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -1144,8 +1144,17 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
             }
             if (!string.IsNullOrEmpty(connectionDetails.ColumnEncryptionSetting))
             {
-                connectionBuilder.ColumnEncryptionSetting =  connectionDetails.ColumnEncryptionSetting == "Enabled" 
-                ? SqlConnectionColumnEncryptionSetting.Enabled : SqlConnectionColumnEncryptionSetting.Disabled;
+                switch (connectionDetails.ColumnEncryptionSetting.ToUpper())
+                {
+                    case "ENABLED":
+                        connectionBuilder.ColumnEncryptionSetting = SqlConnectionColumnEncryptionSetting.Enabled;
+                        break;
+                    case "DISABLED":
+                        connectionBuilder.ColumnEncryptionSetting = SqlConnectionColumnEncryptionSetting.Disabled;
+                        break;
+                    default:
+                        throw new ArgumentException(SR.ConnectionServiceConnStringInvalidColumnEncryptionSetting(connectionDetails.ColumnEncryptionSetting));
+                }
             }
             if (connectionDetails.Encrypt.HasValue)
             {
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -1142,6 +1142,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                         throw new ArgumentException(SR.ConnectionServiceConnStringInvalidAuthType(connectionDetails.AuthenticationType));
                 }
             }
+            if (!string.IsNullOrEmpty(connectionDetails.ColumnEncryptionSetting))
+            {
+                connectionBuilder.ColumnEncryptionSetting =  connectionDetails.ColumnEncryptionSetting == "Enabled" 
+                ? SqlConnectionColumnEncryptionSetting.Enabled : SqlConnectionColumnEncryptionSetting.Disabled;
+            }
             if (connectionDetails.Encrypt.HasValue)
             {
                 connectionBuilder.Encrypt = connectionDetails.Encrypt.Value;
@@ -1313,6 +1318,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                 ConnectTimeout = builder.ConnectTimeout,
                 CurrentLanguage = builder.CurrentLanguage,
                 DatabaseName = builder.InitialCatalog,
+                ColumnEncryptionSetting = builder.ColumnEncryptionSetting.ToString(),
                 Encrypt = builder.Encrypt,
                 FailoverPartner = builder.FailoverPartner,
                 LoadBalanceTimeout = builder.LoadBalanceTimeout,
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
@@ -535,8 +535,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
             if (ServerName != other.ServerName
                 || AuthenticationType != other.AuthenticationType
                 || UserName != other.UserName
-                || AzureAccountToken != other.AzureAccountToken
-                || ColumnEncryptionSetting != other.ColumnEncryptionSetting)
+                || AzureAccountToken != other.AzureAccountToken)
             {
                 return false;
             }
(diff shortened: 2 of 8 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -519,7 +535,8 @@ public bool IsComparableTo(ConnectionDetails other)
             if (ServerName != other.ServerName
                 || AuthenticationType != other.AuthenticationType
                 || UserName != other.UserName
-                || AzureAccountToken != other.AzureAccountToken)
+                || AzureAccountToken != other.AzureAccountToken
+                || ColumnEncryptionSetting != other.ColumnEncryptionSetting)
```
llali: if this is here then other settings should be here too. I think compare only compares connection details and authentication properties

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -99,6 +99,22 @@ public string AuthenticationType
             }
         }
 
+        /// <summary>
+        /// Gets or sets a value that specifies that Always Encrypted functionality is enabled in a connection.
+        /// </summary>
+        public string ColumnEncryptionSetting
```
llali: Shouldn't this be be Boolean line "Encrypt"?

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
Xtrimmer: When this setting is not present in the `IsComparableTo()` method, I found that adjusting this setting in the Change Connection dialog in ADS did not change anything. That is, if I set an column encryption enabled connection to column encryption disabled then the connection stays column encryption enabled. It appears that when a user attempts to change the connection in this way, then this method determines whether anything has really changed and if so, changes the connection, otherwise, it is like a NOOP.
Once I added this setting to the method then things started working as expected.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
Xtrimmer: I thought so at first but it actually gets sent over from ADS in the JSON RPC request as a string value.

llali: it's because how it's configured here:  [https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs](https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs)

---
In reply to: [366025319](https://github.com/microsoft/sqltoolsservice/pull/908#discussion_r366025319) [](ancestors = 366025319)

llali: we can keep it as is since smo also gets the value of this as "enabled"/"disabled" not boolean

---
In reply to: [366494485](https://github.com/microsoft/sqltoolsservice/pull/908#discussion_r366494485) [](ancestors = 366494485,366025319)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1142,6 +1142,11 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
                         throw new ArgumentException(SR.ConnectionServiceConnStringInvalidAuthType(connectionDetails.AuthenticationType));
                 }
             }
+            if (!string.IsNullOrEmpty(connectionDetails.ColumnEncryptionSetting))
+            {
+                connectionBuilder.ColumnEncryptionSetting =  connectionDetails.ColumnEncryptionSetting == "Enabled" 
+                ? SqlConnectionColumnEncryptionSetting.Enabled : SqlConnectionColumnEncryptionSetting.Disabled;
```
llali: you can use enum parser to covert string to enum

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
llali: this looks like a bug to me. we have to check all the settings here then. the scenario is when user connected to a db and wants to change the connection in the dialog. changing settings don't make the connection to reset. we should fix that here and check all the settings.

---
In reply to: [366024645](https://github.com/microsoft/sqltoolsservice/pull/908#discussion_r366024645) [](ancestors = 366024645)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionDetailsTests.cs
```diff
@@ -148,6 +151,7 @@ public void ConnectionDetailsOptionsShouldBeDefinedInConnectionProviderOptions()
             details.MaxPoolSize = expectedForInt + index++;
             details.MinPoolSize = expectedForInt + index++;
             details.PacketSize = expectedForInt + index++;
+            details.ColumnEncryptionSetting = expectedForStrings + index++;
```
llali: please add a test specifically for this new setting

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: And make sure you're handling case sensitivity properly - i.e. both enabled and Enabled I would expect to be valid

Xtrimmer: I thought about doing this at first, but wanted to make sure that we are not throwing an `ArgumentException` in the case that we get a value that is a name, but not one of the named constants defined for the enumeration. Since I wasn't certain that the value was guaranteed to be one of the named constants defined for the enumeration, I decided to go with a safe option to enable when the value matched and disable in all other cases. I can certainly use the `Enum.Parse()` method if this is the preferred way.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionDetailsTests.cs
Xtrimmer: Done. Tests added to ConnectionServiceTests.cs

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
kburtram: This is likely to have undesirable side-effects, since our principle has been the connection identity only contain fields which are visible to the user in the Servers viewlet.  Otherwise, difference connections with different properties will appear identical to the user and this will cause confusion.  If we want to add new properties to the connection identity then we need to also do the work to fix the Server UX to display these properties when there is a possibility of ambiguity.

I would recommend to remove this change from this PR, and open an issue for the change connection bug (which likely is not specific to Column Encryption).  We can discuss how to fix the issue for all properties, but that should be a largely update which address the related issue holistically.

Xtrimmer: I removed this change and submitted the [change connection bug](https://github.com/microsoft/sqltoolsservice/issues/913) as recommended.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Xtrimmer: I changed the way this was assigned so that it is not case sensitive and so that an invalid value throws the appropriate exception

PR #912, opened 2020-01-15
Clean up solution

Charles-Gagnon:
- Use global.json to control what version of .NET SDK is used for building so local builds also use the version used for pipeline builds
- Add some missing items to the sln so they are viewable through VS
- Delete extra project file

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -20,9 +20,9 @@ steps:
     ArtifactName: source
 
 - task: UseDotNet@2
-  displayName: 'Use .NET Core sdk 2.2.108'
+  displayName: 'Use defined .NET Core sdk'
   inputs:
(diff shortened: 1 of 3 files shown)


abist:
I cannot build the project after this change. Are we supposed to configure something?

Charles-Gagnon:
What error are you getting? Are you sure you have the correct version of the .NET SDK installed? You have to do that manually still per the instructions [here](https://github.com/Microsoft/sqltoolsservice/wiki#prerequisites-to-build-the-repository).

abist:
Does it have to be the specific version i.e. "2.2.108"? Why are we forcing a specific version now? This wasn't the case earlier.

Charles-Gagnon:
Followed up offline, but for context here - yes you need to install the specific version now.

PR #915, opened 2020-01-28
Revert "Support active directory pwd authentication"

aaomidi:
No point having code we can't use: https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/482


Charles-Gagnon:
@aaomidi This still something you want to get in?

PR #917, opened 2020-01-28
Fix formatter options for select statement

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/850

extension side change - https://github.com/microsoft/vscode-mssql/pull/1578

Note that this bug is in ADS too, so will need to add this flag to ADS as well.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Formatter/Impl/SqlSelectClauseFormatter.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Formatter/Impl/SqlSelectClauseFormatter.cs
@@ -23,7 +23,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Formatter
         private NewLineSeparatedListFormatter NewLineSeparatedListFormatter { get; set; }
 
         internal SqlSelectClauseFormatter(FormatterVisitor visitor, SqlSelectClause codeObject)
-            : base(visitor, codeObject, visitor.Context.FormatOptions.PlaceEachReferenceOnNewLineInQueryStatements)
+            : base(visitor, codeObject, visitor.Context.FormatOptions.PlaceSelectStatementReferenceOnNewLine || 
+                visitor.Context.FormatOptions.PlaceEachReferenceOnNewLineInQueryStatements)
         {
             NewLineSeparatedListFormatter = new NewLineSeparatedListFormatter(visitor, codeObject, true);
         }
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/Formatter/Impl/SqlSelectClauseFormatter.cs
```diff
@@ -23,7 +23,8 @@ internal class SqlSelectClauseFormatter : CommaSeparatedListFormatter
         private NewLineSeparatedListFormatter NewLineSeparatedListFormatter { get; set; }
 
         internal SqlSelectClauseFormatter(FormatterVisitor visitor, SqlSelectClause codeObject)
-            : base(visitor, codeObject, visitor.Context.FormatOptions.PlaceEachReferenceOnNewLineInQueryStatements)
+            : base(visitor, codeObject, visitor.Context.FormatOptions.PlaceSelectStatementReferenceOnNewLine || 
+                visitor.Context.FormatOptions.PlaceEachReferenceOnNewLineInQueryStatements)
```
Charles-Gagnon: Hm, I'm not sure I like the overlapping settings. What if someone has set PlaceSelectStatementReferenceOnNewLine to false but still wants to have the FROM clause statements on separate lines? Setting PlaceEachReferenceOnNewLineInQueryStatements to true in that case would result in both sets of clauses being on new lines.

abist: PlaceEachReferenceOnNewLineInQueryStatements means placing each reference on new line. Not respecting that sounds like a bug to me. But if we want the select statements only be controlled by the PlaceSelectStatementReferenceOnNewLine, I can change it. I think then we should change the wording of the user settings on ADS and extension side both to state that select statements aren't affected by this setting.

Charles-Gagnon: I agree that leaving the wording as it is might be confusing. That's why I didn't really like the overlap of the settings. I think if we're going to have that we should have separate settings for each. Because right now I would consider the behavior I outlined above just as much of a bug as what you described.

It's not clear to me what the best approach is though. I would suggest bringing this up in a tech sync or something and see what other opinions are before we just go and start adding more settings. Ideally this model is something we could easily extend to other statement types too. What about UPDATE, INSERT, DELETE, etc? It would be good to at least try to consider adding configurations for the other ones as well in the future when we make these - which may not necessarily involve having a separate setting for each one.

PR #919, opened 2020-02-05
Enable Always Encrypted enclave connection parameters

Xtrimmer:
This pull request "wires up" the two new advanced connection security properties; Attestation Protocol and Enclave Attestation URL. These new properties will be used in conjunction with the Always Encrypted property (formerly known as Column Encryption) to support the [Always Encrypted with secure enclaves feature](https://docs.microsoft.com/en-us/sql/relational-databases/security/encryption/always-encrypted-enclaves?view=sql-server-ver15)

image

[A corresponding pull request](https://github.com/microsoft/azuredatastudio/pull/9062) has been submitted to the [Azure Data Studio project](https://github.com/microsoft/azuredatastudio) which add these options to the UI.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
@@ -119,8 +119,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                     new ConnectionOption
                     {
                         Name = "columnEncryptionSetting",
-                        DisplayName = "Column encryption setting",
-                        Description = "Default column encryption setting for all the commands on the connection",
+                        DisplayName = "Always Encrypted",
+                        Description = "Enables or disables Always Encrypted for the connection",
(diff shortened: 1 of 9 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
```diff
@@ -119,16 +119,36 @@ internal static ConnectionProviderOptions BuildConnectionProviderOptions()
                     new ConnectionOption
                     {
                         Name = "columnEncryptionSetting",
-                        DisplayName = "Column encryption setting",
-                        Description = "Default column encryption setting for all the commands on the connection",
+                        DisplayName = "Always Encrypted",
+                        Description = "Enables or disables Always Encrypted for the connection",
```
Charles-Gagnon: Where is this new description coming from? No issues with the change - just not sure I understand the context for the change here.

Xtrimmer: Our Program Manager wanted it to match the verbiage that exists in SSMS for user experience parity.

PR #920, opened 2020-02-19
Allow the arm baseURI to be set dynamically

aaomidi:
This is work required to make https://github.com/microsoft/azuredatastudio/pull/9141 work

There were some formatting changes too.


src/Microsoft.SqlTools.ResourceProvider.Core/Contracts/Account.cs
```diff
@@ -58,7 +58,55 @@ public IEnumerable<Tenant> Tenants
...
+        public string DisplayName;
+
+        /// <summary>
+        /// ID of the provider
+        /// </summary>    
+        public string Id;
+        /// <summary>
+        /// Settings for the provider itself
+        /// </summary>
+        public ProviderSettingsObject Settings;
```
Charles-Gagnon: Is there a reason this is in its own object? Seems odd to have a settings class contain its own settings var

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/AzureResourceManager.cs
```diff
@@ -55,12 +55,19 @@ public Task<IAzureResourceManagementSession> CreateSessionAsync(IAzureUserAccoun
             CommonUtil.CheckForNull(subscriptionContext, "subscriptionContext");
             try
             {
+                string armEndpoint = subscriptionContext.UserAccount.UnderlyingAccount.Properties.ProviderSettings?.Settings?.ArmResource?.Endpoint;
+                Uri armUri = null;
+                if (armEndpoint != null)
+                {
+                    armUri = new Uri(armEndpoint);
```
Charles-Gagnon: Consider for these having it fall back to using the credentials if parsing the URI fails

src/Microsoft.SqlTools.ResourceProvider.Core/Contracts/Account.cs
aaomidi: It's just to mimic the object we get from ADS. I didn't want to change it __too much__.

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/AzureResourceManager.cs
aaomidi: I think I'm doing that here: https://github.com/microsoft/sqltoolsservice/pull/920/files/28765aa33ebcf5b8f59e9ae3806e37c4851120a2#diff-68c0bc2a8a7ae81b805166ef253c3696R65

Charles-Gagnon: I'm saying if it throws an exception while parsing (since that's what'll happen if armEndpoint isn't a valid URI)

aaomidi: Ah I see, that'd be an extremely weird case but I'll plan for it too.

aaomidi:
I'm going to merge this today if no one has any objections :)

smartguest:
So far seems good, no issues so far

PR #924, opened 2020-02-26
Fix string usage for build breaks

abist:
Fix build breaks since the SR gen tool was breaking with newly added strings


abist:
@Charles-Gagnon The files were updated by the tool, I just changed the string usage in the file

alanrenmsft:
I scheduled a test build and it passed, merging the changes now.

PR #928, opened 2020-02-28
Fix for deleting text columns

smartguest:
When using EditData, deleting rows that have text values in them causes the whole table to be dropped due to improper parsing. This commit will fix the issue by adding a proper clause, as well as making the delete command to be more consistent with SMSS


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -65,15 +65,42 @@ public override DbCommand GetCommand(DbConnection connection)
...
+
             command.CommandText = commandText;
             command.Parameters.AddRange(where.Parameters.ToArray());
-
             return command;
         }
 
+        private Boolean checkIfValid(DbConnection connection, WhereClause where)
+        {
+            string formatString = "SELECT COUNT(*) FROM {0} {1}";
```
aaomidi: String.format could lead to a sql injection, you should use SqlParameters.

You can find examples of them https://github.com/microsoft/sqltoolsservice/blob/427f0c11e8d0c1db0a81f4c6023215001dd0e32d/src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/AgentNotebookHelper.cs

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -65,15 +65,42 @@ public override DbCommand GetCommand(DbConnection connection)
 
             // Return a SqlCommand with formatted with the parameters from the where clause
             WhereClause where = GetWhereClause(true);
-            string commandText = GetCommandText(where.CommandText);
-
             DbCommand command = connection.CreateCommand();
+            string commandText = "";
+            if (!this.checkIfValid(connection, where))
+            {
+                commandText = " DECLARE @error NVARCHAR(100) = N'INVALID DELETE: rows with duplicate or text values are not currently supported';" + Environment.NewLine +
```
aaomidi: i18n?

smartguest:
@Charles-Gagnon I will remove the text table filter that was in RowEditBase and the reason for the duplicate thing is that a table containing rows that are duplicates will cause those rows to be deleted as they share the same row values as the row that is to be deleted. When you attempt to delete a row in SMSS for a table with duplicates, it notifies the user that it must be changed first.

smartguest:
Tests need to be rewritten to account for the change in handling

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
```diff
@@ -181,34 +181,32 @@ protected WhereClause GetWhereClause(bool parameterize)
...
-                    }
-                    else
-                    {
-                        // General case is to just use the value from the cell
-                        if (parameterize)
+                        // Add a parameter and parameterized clause component
+                        // NOTE: We include the row ID to make sure the parameter is unique if
+                        //       we execute multiple row edits at once.
+                        string paramName = $"@Param{RowId}{col.Ordinal}";
+                        if (col.DbColumn.DataTypeName.Equals("TEXT", StringComparison.OrdinalIgnoreCase) ||
```
Charles-Gagnon: Why did you get rid of the byte[] type check?

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
```diff
@@ -181,34 +181,32 @@ protected WhereClause GetWhereClause(bool parameterize)
...
-                            //       we execute multiple row edits at once.
-                            string paramName = $"@Param{RowId}{col.Ordinal}";
-                            cellDataClause = $"= {paramName}";
-                            SqlParameter parameter = new SqlParameter(paramName, col.DbColumn.SqlDbType)
-                            {
-                                Value = cellData.RawObject
-                            };
-                            output.Parameters.Add(parameter);
+                            // Special cases for TEXT/NTEXT types
+                            cellDataClause = $"LIKE {paramName}";
```
Charles-Gagnon: Why are you using LIKE here? Without things such as wildcards this should behave the same as just using = directly

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
```diff
@@ -181,34 +181,32 @@ protected WhereClause GetWhereClause(bool parameterize)
                 }
                 else
                 {
-                    if (cellData.RawObject is byte[] ||
-                        col.DbColumn.DataTypeName.Equals("TEXT", StringComparison.OrdinalIgnoreCase) ||
-                        col.DbColumn.DataTypeName.Equals("NTEXT", StringComparison.OrdinalIgnoreCase))
+                    if (parameterize)
                     {
-                        // Special cases for byte[] and TEXT/NTEXT types
-                        cellDataClause = "IS NOT NULL";
```
Charles-Gagnon: Did you ever look into the history of this to try and figure out why this was done in the first place? It's such a weird thing to do that there might be a good reason to do it like this instead of something similar to what you're doing.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
smartguest: attempting to use = apparently does not work for strings when I tested it out

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
smartguest: I think that using equals should handle the comparison of byte arrays just fine, unless I'm wrong about that.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
smartguest: https://github.com/microsoft/sqltoolsservice/commit/795eba3da6419a22c33fc04419648b3a942c7a80#diff-1f1765bd0d92616e0edf269c197b8422 

According to the first version of this class, it apparently existed from the beginning. I can only guess that it was done because using "=" did not work on those types exactly due to incompatibility and so this was a placeholder.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
Charles-Gagnon: Ah interesting - yeah the types aren't compatible https://blog.sqlauthority.com/2012/12/04/sql-server-fix-error-402-the-data-types-ntext-and-varchar-are-incompatible-in-the-equal-to-operator/

The suggestion there that I generally agree with is to use CONVERT over LIKE. But in this case I think it's fine to just use LIKE to keep it simpler.

Charles-Gagnon: I'd suggest adding a comment explaining why we have to special case these though (just mention that the types aren't compatible with n/varchar so direct comparison doesn't work)

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
Charles-Gagnon: Hm, I'm not sure. I'd suggest making sure to test it out first. varbinary I expect probably works fine so perhaps this logic is for the image type? https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver15 Make sure you test that to verify.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowEditBase.cs
benrr101: This was how the code from SSMS was written. I can't remember all the details exactly, but I think it had something to do with the possibility of multiple gig values for binary/full-text columns.

Charles-Gagnon:
@smartguest Is this something you plan on completing at some point? If not I'd suggest just closing the PR for now

smartguest:
@Charles-Gagnon Do you know what the test failure was exactly? It's been a while since I last worked on this. I've seemed to have addressed the test that was having issues during my last few commits. I'll have to test it again, before giving the go ahead.

Charles-Gagnon:
You'll need to target this against main so it'll run the checks properly - do that and we'll see if it's still happening.

smartguest:
@Charles-Gagnon Pipeline passed without any test failures.

smartguest:
@Charles-Gagnon Here's the pipeline with the results: https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=111106&view=results

Charles-Gagnon (approved):
Just to verify - you've gone through and done various testing yourself to make sure things are still working as expected now with the fix?

smartguest:
I tested it out, deleting Text table and varchar tables seem to work fine so far. No issues.

PR #931, opened 2020-03-04
Warning for multiple delete

smartguest:
As requested by @kburtram , this commit will run a select query to determine if multiple rows will be deleted in the following delete query. If yes, the delete operation will be cancelled, else it will proceed. 

The use of Parametrization for the command so far has run into several issues as the Where Clause cannot be dynamically inserted, neither is the table name. Although string.Format is no longer used.


src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +80,27 @@ public override DbCommand GetCommand(DbConnection connection)
             return command;
         }
 
+        private bool verifyCommand(WhereClause where, string input, DbConnection connection){
+            bool verifyStatus = true;
```
kburtram: generally I'd recommend defaulting to false and only setting true once you've verified the condition your checking.  In this case I think the return value will be correct, though if you invert the condition it makes it easier to conclude we don't ever incorrectly return a true verification status.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -516,12 +516,24 @@ private async Task CommitEditsInternal(DbConnection connection, Func<Task> succe
...
+                        using (DbCommand editCommand = editOperation.GetCommand(connection))
+                        using (DbDataReader reader = await editCommand.ExecuteReaderAsync())
+                        {
+                            // Apply the changes of the command to the result set
+                            await editOperation.ApplyChanges(reader);
+                        }
+                    }
+                    catch (Exception e)
+                    {
+                        if (e.ToString().Contains("delete"))
```
kburtram: can you create a new Exception-derived type and catch just that.  This isn't the correct pattern for detecting a particular type of exception.

smartguest:
I will write tests as soon as I can

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -516,12 +516,21 @@ private async Task CommitEditsInternal(DbConnection connection, Func<Task> succe
...
+                        {
+                            // Apply the changes of the command to the result set
+                            await editOperation.ApplyChanges(reader);
+                        }
+                    }
+                    catch (DeleteError e)
+                    {
+                        //clear EditCache to allow for deletion of other rows.
+                        EditCache.TryRemove(editOperation.RowId, out RowEditBase xe);
+                        throw e;
```
Charles-Gagnon: Just use `throw;` so it doesn't reset the stack

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -15,13 +15,34 @@
 
 namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
 {
+    /// <summary>
+    /// Represents a row that should be deleted. This will generate a DELETE statement
```
Charles-Gagnon: This summary doesn't make sense for this class

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,33 @@ public override DbCommand GetCommand(DbConnection connection)
             return command;
         }
 
+        private bool VerifyCommand(WhereClause where, string input, DbConnection connection)
+        {
+            bool verifyStatus = false;
+            DbCommand command = connection.CreateCommand();
+            command.CommandText = input;
+            command.Parameters.AddRange(where.Parameters.ToArray());
+            DbDataReader reader = command.ExecuteReader();
```
Charles-Gagnon: Use 
```
using(DbCommand command = connection.CreateCommand()) {
...
using(DbDataReader reader = command.ExecuteReader()) { ...
``` 
here, that will ensure that the command and reader are safely closed and disposed when done. Right now if an error occurs while reading you'll never clean up those resources.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,33 @@ public override DbCommand GetCommand(DbConnection connection)
             return command;
         }
 
+        private bool VerifyCommand(WhereClause where, string input, DbConnection connection)
```
Charles-Gagnon: Please add a summary comment for this method - the name alone isn't clear enough to know what it's doing.

Although I'd also consider either renaming this method completely or making it more generic since right now the logic is only checking if the query returns exactly 1 row and the name alone isn't enough to indicate that.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -131,6 +184,11 @@ protected override int CompareToSameType(RowEditBase rowEdit)
             return RowId.CompareTo(rowEdit.RowId) * -1;
         }
 
+        private string GetVerifyText(string whereText)
+        {
+            return VerifyStatement + AssociatedObjectMetadata.EscapedMultipartName + whereText;
```
Charles-Gagnon: `return $"{VerifyStatement}{AssociatedObjectMetadata.EscapedMultipartName}{whereText}";`

smartguest:
Currently attempting to write a test, it is currently not complete at the moment

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -131,6 +189,11 @@ protected override int CompareToSameType(RowEditBase rowEdit)
             return RowId.CompareTo(rowEdit.RowId) * -1;
         }
 
+        private string GetVerifyText(string whereText)
+        {
+            return $"{VerifyStatement}{AssociatedObjectMetadata.EscapedMultipartName}{whereText}";;
```
Charles-Gagnon: Extra ; here

smartguest:
So far I cannot figure out how to actually execute the query within the code, but I did include a check to make sure it was generated correctly and I have tested it on my own tables without any issues.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,38 @@ public override DbCommand GetCommand(DbConnection connection)
...
+                    try
+                    {
+                        while (reader.Read())
+                        {
+                            if (reader.GetInt32(0) != 1)
+                            {
+                                verifyStatus = true;
+                            }
+                        }
+                        reader.Close();
```
Charles-Gagnon: You don't need to call close here - dispose will close it automatically

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,38 @@ public override DbCommand GetCommand(DbConnection connection)
...
+                command.Parameters.AddRange(where.Parameters.ToArray());
+                using (DbDataReader reader = command.ExecuteReader())
+                {
+                    try
+                    {
+                        while (reader.Read())
+                        {
+                            if (reader.GetInt32(0) != 1)
+                            {
+                                verifyStatus = true;
```
Charles-Gagnon: I'd just return true here and get rid of this var completely (returning false at the end of the method)

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,38 @@ public override DbCommand GetCommand(DbConnection connection)
...
+                            }
+                        }
+                        reader.Close();
+                    }
+                    catch
+                    {
+                        //Likely means there was nothing found that matched the query.
```
Charles-Gagnon: I don't think this statement is true - that should be a rare case and it's more expected that a network error or something would happen.

Regardless I'd just remove the comment - I think it's clear enough that we just want to default to false if we can't determine what it is from the query itself.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -101,6 +159,14 @@ public override string GetScript()
             return GetCommandText(GetWhereClause(false).CommandText);
         }
 
+        /// <summary>
+        /// Generates a WHERE statement to verify the row delete is unique.
+        /// </summary>
+        /// <returns>String of the WHERE statement</returns>
+        public string GetVerifyScript(){
```
Charles-Gagnon: Please format this file - the spacing here is off for example.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/RowDeleteTests.cs
```diff
@@ -200,6 +200,36 @@ public async Task RevertCell()
...
+
+            // If: I generate a verify command
+            String verifyCommand = rd.GetVerifyScript();
+
+             // Then:
+            // ... The command should not be null
+            Assert.NotNull(verifyCommand);
+
+            // ... It should be formatted into an where script
+            string regexTest = @"SELECT COUNT \(\*\) FROM (.+) WHERE (.+)";
```
Charles-Gagnon: So this is better than nothing, but the point of the testing would be to verify that an error is thrown if the query comes back with more or less than 1 row. 

What kind of issues were you having? Can you show the code you had that you were having issues with?

smartguest: The issue I'm currently having is creating a dataset that has multiple rows (with the same data) and deleting it with the appropriate response, I can't seem to get the query to execute properly with the TestDbDataReader (as it simply returns the next row unaltered). I would like to see an example of a test that actually executes queries on a dataset that is altered.

smartguest: For reference here is my original code:
`
[Fact]
        public async Task GetVerifyQuery(){
            // Setup: create result set to be worked on.
            Common.TestDbColumnsWithTableMetadata data = new Common.TestDbColumnsWithTableMetadata(false, false, 0, 0);
            var rs = await Common.GetResultSet(data.DbColumns, false);
            var testResultSet = PreCreateRowResultSet(data.DbColumns, rs.GetRow(0));
            var newRowReader = new TestDbDataReader(new [] {testResultSet}, false);
            RowCreate rc = new RowCreate(1, rs, data.TableMetadata);
            await rc.ApplyChanges(newRowReader);


            var x = rs.GetRow(0);
            var y = rs.GetRow(1);
            Assert.Equal(2, rs.RowCount);

            RowDelete rd = new RowDelete(1, rs, data.TableMetadata);
            testResultSet = PostCreateRowResultSet(data.DbColumns, rs.GetRow(0));
            // ... Mock db connection for building the command
            var mockConn = new TestSqlConnection(new [] {testResultSet});
            // If: I attempt to get a command for the edit
            DbCommand cmd = rd.GetCommand(mockConn);

            Assert.True(true);

        }
`

smartguest:
I have written a test that uses a simulated select query result that has a number that is not 1, and it successfully is able to throw an exception.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -15,13 +15,34 @@
 
 namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
 {
+    /// <summary>
+    /// An error indicating that a delete action will delete multiple rows.
+    /// </summary>
+    public class DeleteError : Exception
```
kburtram: Could you rename this class to better follow .Net naming conventions for exception classes?  For example, maybe something like `EditDataDeleteException` or something like that.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,35 @@ public override DbCommand GetCommand(DbConnection connection)
             return command;
         }
 
+        /// <summary>
+        /// Runs a query using the where clause to determine if duplicates are found (causes issues when deleting).
+        /// </summary>
+        private bool checkWhereDuplicate(WhereClause where, string input, DbConnection connection)
```
kburtram: please following .Net naming conventions, which says functions should be PascalCase, so `CheckWhereDuplicate` in this case.  Though personally I think a name like `CheckForDuplicateDeleteRows` would be more "self documenting".

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,35 @@ public override DbCommand GetCommand(DbConnection connection)
...
+            {
+                command.CommandText = input;
+                command.Parameters.AddRange(where.Parameters.ToArray());
+                using (DbDataReader reader = command.ExecuteReader())
+                {
+                    try
+                    {
+                        while (reader.Read())
+                        {
+                            if (reader.GetInt32(0) != 1)
```
kburtram: it looks like if there are multiple rows and any of them contain a 1 this method will return true.  I'd recommend breaking or returning false, if this condition returns false to make the intent of the code more explicit.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +100,35 @@ public override DbCommand GetCommand(DbConnection connection)
...
+                            {
+                                return true;
+                            }
+                        }
+                    }
+                    catch
```
kburtram: it may be nice to add the exception message to the log to help potential troubleshooting in the future.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
smartguest: this is for running the select query on the rows, since GetScalar is not recognized by the tests, I have to use GetInt32 to handle it

Charles-Gagnon: What do you mean "not recognized"? And by GetScalar do you mean ExecuteScalar? If a test is using this test object https://github.com/microsoft/sqltoolsservice/blob/master/test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/TestObjects.cs#L163 then yes, it looks like another case of that feature not being implemented. If that's the case you should implement that instead of using a less-ideal approach in the production code.

Charles-Gagnon:
@smartguest Any progress on this PR?

smartguest:
@Charles-Gagnon Ideally it should be complete according to what I remember, but I haven't received any approval so I haven't been able to merge it yet.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -66,6 +88,11 @@ public override DbCommand GetCommand(DbConnection connection)
             // Return a SqlCommand with formatted with the parameters from the where clause
             WhereClause where = GetWhereClause(true);
             string commandText = GetCommandText(where.CommandText);
+            string verifyText = GetVerifyText(where.CommandText);
+            if (!CheckForDuplicateDeleteRows(where, verifyText, connection))
+            {
+                throw new EditDataDeleteException("This action will delete more than one row!");
```
kburtram, line 94: typically error string shouldn't contain `!` characters.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -74,6 +101,38 @@ public override DbCommand GetCommand(DbConnection connection)
...
+            {
+                command.CommandText = input;
+                command.Parameters.AddRange(where.Parameters.ToArray());
+                using (DbDataReader reader = command.ExecuteReader())
+                {
+                    try
+                    {
+                        while (reader.Read())
+                        {
+                            //If the count of the row is
```
kburtram, line 120: looks like the comment go cut off.  Style note: please add space ` ` after the `//`.

PR #933, opened 2020-03-09
Add  date_bucket key word

Satyem-MS:
Add key word for date_bucket intrinsic for sql tools.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
@@ -115,6 +115,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             "datepart",
             "datetime",
             "datetime2",
+            "date_bucket",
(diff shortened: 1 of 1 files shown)


shueybubbles:
> Looks fine here but let's wait for a new SMO update with the changes you made there to verify everything works correctly.
> 
> @shueybubbles Do we have an updated SMO package with those changes?
> 
> You should also update https://github.com/microsoft/azuredatastudio/blob/master/extensions/sql/syntaxes/sql.tmLanguage.json#L206 and https://github.com/microsoft/vscode-mssql/blob/master/syntaxes/SQL.plist#L342 for the full language support.

the SMO package now has a dependency on the separate sqlparser package. So we can push just a new sqlparser and add a packagereference to it in the tools service, or we can build a new SMO with a dependency on the new sqlparser package version and just update tools service directory.build.props with the new SMO version to get both.

Charles-Gagnon:
@shueybubbles Let's just update SMO (+ SqlParser) for now. Splitting the dependencies apart may be useful in the future though so if you update that package as well I can look into doing that change down the line.

PR #934, opened 2020-03-11
Always Encrypted Azure Key Vault Support

Xtrimmer:
In the last Iteration we added support for Always Encrypted with secrets stored as local certificates. This change adds support for secrets stored in Azure Key Vault (AKV).

In the event that a user executes a select query, and one or more columns are encrypted using the Always Encrypted feature, and where the decryption key exists in AKV, then an access token will be requested under the authority of a Linked Azure Account and provided to the Microsoft.Data.SqlClient Driver so that it can acquire the key from AKV and use it to decrypt the query results data. 

This is half of the solution with the other half of the [changes made in Azure Data Studio](https://github.com/microsoft/azuredatastudio/pull/9556).


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
+{
+    class RequestSecurityTokenParams 
+    {
+        /// <summary>
+        /// Gets or sets authority.
```
Charles-Gagnon: It would be nice to have these actually say what "authority", "resource" and "scope" are in this context. It's not too hard to figure out given the class name but currently these comments are all but worthless and it would be useful for people that don't necessarily know the finer details of the protocol.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Gets or sets scope.
+        /// </summary>
+        public string Scope { get; set; }
+    }
+
+    class RequestSecurityTokenResponse
+    {
+        /// <summary>
+        /// Gets or sets userName.
```
Charles-Gagnon: And this should definitely be at least fixed to be the correct name (but the above comment applies here too)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -115,6 +116,26 @@ public ConnectionService()
             var defaultQueue = new ConnectedBindingQueue(needsMetadata: false);
             connectedQueues.AddOrUpdate("Default", defaultQueue, (key, old) => defaultQueue);
             this.LockedDatabaseManager.ConnectionService = this;
+            
+            SqlColumnEncryptionAzureKeyVaultProvider sqlColumnEncryptionAzureKeyVaultProvider = new SqlColumnEncryptionAzureKeyVaultProvider(AzureActiveDirectoryAuthenticationCallback);
+            SqlConnection.RegisterColumnEncryptionKeyStoreProviders(customProviders: new Dictionary<string, SqlColumnEncryptionKeyStoreProvider>(capacity: 1, comparer: StringComparer.OrdinalIgnoreCase)
```
Charles-Gagnon: This isn't going to work with how the tests are currently set up to Mock the connection service which will create multiple instances of this. You can probably get around that by putting this in a static constructor instead.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public string Token { get; set; }
+    }
+
+    /// <summary>
+    /// ConnectionComplete notification mapping entry
```
Charles-Gagnon: Update

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -20,6 +20,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.0.0" />
```
Charles-Gagnon: This is causing problems with the version of System.Net.Http that the tests are using. Should be fine to just update those to the newer version.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Xtrimmer: Creating a static constructor for this code chunk should work if I can make this member property static. Is this okay or do you think that this will this cause problems too? 
``` csharp
internal static IProtocolEndpoint ServiceHost { get; set; }
```
The reason is because, while I am instantiating and setting the KeyVaultProvider object on the SqlConnection statically, the callback function that requests the token from ADS depends on the ServiceHost ProtocolEndpoint which is not static.

Xtrimmer: Looking at it more, A better option may be to simply grab a singleton instance of ServiceHost from within the callback function and use that rather than the ConnectionService's ServiceHost member property. I'll do that instead.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
Xtrimmer: Comments updated.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
Xtrimmer: Fixed.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
Xtrimmer: Fixed.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Xtrimmer: Fixed.

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
Xtrimmer: Updated the test versions.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -23,6 +24,8 @@
 using Microsoft.SqlTools.Utility;
 using System.Diagnostics;
 
+using HostingService = Microsoft.SqlTools.ServiceLayer.Hosting.ServiceHost;
```
aaomidi: We were discussing that you won't need to do this, is that still the plan?

Charles-Gagnon: Am I forgetting some context here - where was that being discussed?

aaomidi: This was a discussion I had with @Xtrimmer. Essentially you can use the Singleton to get the Connection Service and access the ServiceHost from there.

Xtrimmer: Oh yea, updated now.

PR #935, opened 2020-03-17
Save Kusto results in different formats

gupta1:
Saves results in csv, excel, json and xml formats.


Charles-Gagnon:
@gupta1 Please make sure to delete your branches after you're done merging in a PR, thanks!

image

(I've done it for you this time, just as a future reminder)

PR #936, opened 2020-03-21
Show file location for dacfx extract and export

kisantia:
Show file location for extract and export operations because it's more useful than showing the server and database name. Intermediate fix for https://github.com/microsoft/azuredatastudio/issues/9057
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -228,9 +228,18 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
                 try
                 {
                     TaskMetadata metadata = TaskMetadata.Create(parameters, taskName, operation, ConnectionServiceInstance);
+                    if (operation.GetType() == typeof(ExtractOperation) || operation.GetType() == typeof(ExportOperation))
+                    {
+                        // show package file path in task viewlet instead of server and database name
+                        metadata.DatabaseName = null;
+                        metadata.ServerName = parameters.PackageFilePath;
+                    }
+                    else
+                    {
+                        // put appropriate database name since connection passed was to master
+                        metadata.DatabaseName = parameters.DatabaseName;
+                    }
 
-                    // put appropriate database name since connection passed was to master
-                    metadata.DatabaseName = parameters.DatabaseName;
                     SqlTask sqlTask = SqlTaskManagerInstance.CreateTask<SqlTask>(metadata);
 
                     await sqlTask.RunAsync();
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -228,9 +228,18 @@ private void ExecuteOperation(DacFxOperation operation, DacFxParams parameters,
                 try
                 {
                     TaskMetadata metadata = TaskMetadata.Create(parameters, taskName, operation, ConnectionServiceInstance);
+                    if (operation.GetType() == typeof(ExtractOperation) || operation.GetType() == typeof(ExportOperation))
+                    {
+                        // show package file path in task viewlet instead of server and database name
+                        metadata.DatabaseName = null;
+                        metadata.ServerName = parameters.PackageFilePath;
```
udeeshagautam, line 235: absolutely like the direction. I am feeling can we not have both pieces of info from which db to which file or is that too much for single line?

udeeshagautam: @chlafreniere (since you opened the issue :) ) do you have any preference?

Charles-Gagnon, line 235: I'm not a fan of overloading the field like this - what happens if someone on the ADS side tries to use this expecting it to be a valid server name all the time? If we're going to do this I'd rather see a new field added, something like `TargetDestination` (since it should be generic across many operations)

kisantia: Yeah that's makes sense. How should ADS taskRederer handle this new field? Show the TargetDestination if there aren't values for server and database? Or should it have priority over those?

Charles-Gagnon: I'd say default to TargetDestination and if it's undefined/empty then fall back to what it was previously.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -117,7 +117,7 @@ public async Task HandleExtractRequest(ExtractParams parameters, RequestContext<
                     // Set connection details database name to ensure the connection string gets created correctly for DW(extract doesn't work if connection is to master)
                     connInfo.ConnectionDetails.DatabaseName = parameters.DatabaseName;
                     ExtractOperation operation = new ExtractOperation(parameters, connInfo);
-                    ExecuteOperation(operation, parameters, SR.ExtractDacpacTaskName, requestContext);
+                    ExecuteOperation(operation, parameters, SR.ExtractDacpacTaskName, requestContext, parameters.PackageFilePath);
```
Charles-Gagnon, line 120: You're already passing in the whole parameters object so passing in package file path additionally seems unnecessary

PR #938, opened 2020-03-25
Add var to control whether we push symbols to nuget

Charles-Gagnon:
This was publishing the symbol package to nuget for every single build, including PR validation builds. We should only do this for official product builds so adding this in so we can skip the step for other builds using this pipeline definition.

We should probably investigate how necessary doing this actually is - I'm not even sure what this feed is for. But currently this is causing PR validation builds to fail so doing this as a quick fix to unblock that.


aaomidi:
Let's just make sure we allow it to push during the scheduled builds.

PR #946, opened 2020-04-11
Add AQL Assessment service

aguzev:
This cannot go in until after #943 since it depends on [SMO 160.2004021](https://www.nuget.org/packages/Microsoft.SqlServer.SqlManagementObjects/160.2004021.0). 

[SQL Assessment API](https://docs.microsoft.com/en-us/sql/sql-assessment-api/sql-assessment-api-overview) provides a mechanism to evaluate the configuration
of SQL Server for best practices. SQL Assessment API gives a list
of recommended actions to improve SQL Server performance or security.

The SQL Assessment service is used by the expected SQL Assessment
feature of Azure Data Studio. 

SqlAssessmentService forwards JSONRPC calls to SQL Assessment engine
and wraps results as a response.

`assessment/getAssessmentItems` returns a set of checks 
applicable to a given target.

`assessment/invoke` returns a set of recommendations
for improving SQL Server instance or database configurations. 

`assessment/generateScript` returns a T-SQL script for storing
an assessment result set to a SQL data table.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
@@ -1768,10 +1768,6 @@
     <value>Could not find the schema compare session to cancel</value>
     <comment></comment>
   </data>
-  <data name="SqlAssessmentOperationExecuteCalledTwice" xml:space="preserve">
-    <value>A SQL Assessment operation&apos;s Execute method should not be called more than once</value>
-    <comment></comment>
-  </data>
   <data name="SqlAssessmentGenerateScriptTaskName" xml:space="preserve">
     <value>Generate SQL Assessment script</value>
     <comment></comment>
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
@@ -1768,4 +1768,25 @@
     <value>Could not find the schema compare session to cancel</value>
     <comment></comment>
   </data>
+  <data name="SqlAssessmentOperationExecuteCalledTwice" xml:space="preserve">
+    <value>A SQL Assessment operation&apos;s Execute method should not be called more than once</value>
+    <comment></comment>
+  </data>
+  <data name="SqlAssessmentGenerateScriptTaskName" xml:space="preserve">
+    <value>Generate SQL Assessment script</value>
+    <comment></comment>
+  </data>
+  <data name="SqlAssessmentQueryInvalidOwnerUri" xml:space="preserve">
+    <value>Not connected to a server</value>
+    <comment></comment>
+  </data>
+  <data name="SqlAssessmentConnectingError" xml:space="preserve">
+    <value>Cannot connect to the server</value>
+    <comment></comment>
+  </data>
+  <data name="SqlAssessmentUnsuppoertedEdition" xml:space="preserve">
+    <value>Unsupported engine edition {0}</value>
+    <comment>.
+ Parameters: 0 - editionCode (int) </comment>
+  </data>
 </root>
--- a/src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Text;
 using System.Threading;
+using Microsoft.SqlTools.ServiceLayer.Management;
 using Microsoft.SqlTools.ServiceLayer.SqlAssessment.Contracts;
 using Microsoft.SqlTools.ServiceLayer.TaskServices;
 using Microsoft.SqlTools.Utility;
@@ -17,9 +18,11 @@ namespace Microsoft.SqlTools.ServiceLayer.SqlAssessment
     /// <summary>
     /// Generates a script storing SQL Assessment results to a table.
     /// </summary>
-    internal sealed class GenerateScriptOperation : ITaskOperation
+    internal sealed class GenerateScriptOperation : ITaskOperation, IDisposable
     {
-        private CancellationTokenSource cancellation = null;
+        private readonly CancellationTokenSource cancellation = new CancellationTokenSource();
+
+        private bool disposed = false;
 
         /// <summary>
         /// Gets the unique id associated with this instance.
@@ -58,15 +61,6 @@ namespace Microsoft.SqlTools.ServiceLayer.SqlAssessment
         /// </exception>
         public void Execute(TaskExecutionMode mode)
         {
-            // This method is not re-enterable as a trade-off for avoiding
-            // overcomplicated multiple CancellationTokenSources management
-            if (cancellation != null)
-            {
-                throw new InvalidOperationException(SR.SqlAssessmentOperationExecuteCalledTwice);
-            }
-
-            cancellation = new CancellationTokenSource();
-
             try
             {
                 var scriptText = GenerateScript(Parameters, cancellation.Token);
@@ -80,21 +74,16 @@ namespace Microsoft.SqlTools.ServiceLayer.SqlAssessment
                 ErrorMessage = e.Message;
                 Logger.Write(TraceEventType.Error, string.Format(
                     CultureInfo.InvariantCulture,
-                    "SQL Assessment generate script operation failed with exception {0}",
+                    "SQL Assessment: generate script operation failed with exception {0}",
                     e.Message));
 
                 throw;
-            }
-            finally
-            {
-                var c = Interlocked.Exchange(ref cancellation, null);
-                c?.Dispose();
             }
         }
 
         public void Cancel()
         {
-            cancellation?.Cancel();
+            cancellation.Cancel();
         }
 
         #region Helpers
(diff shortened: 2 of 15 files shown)


src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/SqlAssessmentService.cs
```diff
@@ -0,0 +1,463 @@
...
+                if (connection.State == ConnectionState.Closed)
+                {
+                    connection.Open();
+                }
+
+                // Obtain target server details
+                var cmd = ((DbConnection)connection).CreateCommand();
+                var server = new SqlObjectLocator();
+
+                const string PlatformQueryText = @"SELECT [host_platform] FROM sys.dm_os_host_info";
```
Charles-Gagnon, line 163: Assuming these queries are actually necessary for the assessment request you should use a common helper like ReliableConnectionHelper.GetServerVersion for this.

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/SqlAssessmentService.cs
```diff
@@ -0,0 +1,463 @@
...
+        /// Internal for testing
+        /// </remarks>
+        internal async Task<List<AssessmentResultItem>> InvokeSqlAssessment(SqlObjectLocator target)
+        {
+            var result = new List<AssessmentResultItem>();
+            var resultsList = await Engine.GetAssessmentResultsList(target);
+            Logger.Write(TraceEventType.Information, $"Got {resultsList.Count} SQL Assessment results.");
+            foreach (var r in resultsList)
+            {
+                var nameStart = r.TargetPath.LastIndexOf('/') + 1;
```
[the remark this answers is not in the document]
aguzev: It is not a file path but will remove this line.

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/SqlAssessmentService.cs
aguzev: I had to extend the helper to return ServerName.

Directory.Build.props
```diff
@@ -1,8 +1,9 @@
 <Project>
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
-    <SmoPackageVersion>160.2001141.0-preview</SmoPackageVersion>
-    <SqlClientPackageVersion>1.1.0</SqlClientPackageVersion>
+    <SmoPackageVersion>160.2004021.0</SmoPackageVersion>
```
Charles-Gagnon, line 4: Merge in lastest from master - these changes should be in there already

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlAssessment/SqlAssessmentServiceTests.cs
```diff
@@ -0,0 +1,170 @@
...
+                        });
+            }
+
+            return response;
+        }
+
+        private void AssertInfoPresent(AssessmentItemInfo item)
+        {
+
+            Assert.False(string.IsNullOrWhiteSpace(item.CheckId), nameof(AssessmentItemInfo.CheckId));
```
Charles-Gagnon, line 163: Suggest adding a message to these asserts so if they fail it's clear what the issue is (since the default message doesn't usually say what values it was checking)

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/SqlAssessmentService.cs
```diff
@@ -0,0 +1,449 @@
...
+        /// returned by SERVERPROPERTY('EngineEdition').
+        /// </summary>
+        /// <param name="representation">
+        /// A number returned by SERVERPROPERTY('EngineEdition').
+        /// </param>
+        /// <returns>
+        /// Returns a <see cref="SqlEngineEdition"/>
+        /// corresponding to the <paramref name="representation"/>.
+        /// </returns>
+        private static SqlEngineEdition GetEngineEdition(int representation)
```
Charles-Gagnon, line 405: What is this SqlEngineEdition and why is this cast necessary?

aguzev: This is a value used by the assessment engine. I'm replacing this conversion function with a less cryptic one.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -1768,4 +1768,25 @@
...
+    <comment></comment>
+  </data>
+  <data name="SqlAssessmentConnectingError" xml:space="preserve">
+    <value>Cannot connect to the server</value>
+    <comment></comment>
+  </data>
+  <data name="SqlAssessmentUnsuppoertedEdition" xml:space="preserve">
+    <value>Unsupported engine edition {0}</value>
+    <comment>.
+ Parameters: 0 - editionCode (int) </comment>
```
Charles-Gagnon, line 1790: Did you add this comment yourself? That period seems wrong - if you did could you remove it and just have the comment on one line?

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// Execute a task
+        /// </summary>
+        /// <param name="mode">Task execution mode (e.g. script or execute)</param>
+        /// <exception cref="InvalidOperationException">
+        /// The method has been called twice in parallel for the same instance.
+        /// </exception>
+        public void Execute(TaskExecutionMode mode)
+        {
+            // This method is not re-enterable as a trade-off for avoiding
+            // overcomplicated multiple CancellationTokenSources management
```
Charles-Gagnon, line 62: Did you see an instance of this happening? I haven't spent a lot of time in this code but a quick look through makes me wonder whether re-entry is even possible since we create a new operation every time a request comes in.

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                sb.Append(scriptPrologue);
+                foreach (var item in generateScriptParams.Items)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    if (item.Kind == AssessmentResultItemKind.Note)
+                    {
+                        sb.Append("\r\n('")
+                            .Append(Escape(item.DisplayName)).Append("','")
```
Charles-Gagnon, line 136: Instead of a whole bunch of these append calls it'd be easier and faster just to just use string interpolation

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
```diff
@@ -0,0 +1,164 @@
...
+                    }
+                }
+
+                sb.Length -= 1;
+            }
+
+            return sb.ToString();
+        }
+
+        private static string Escape(string original)
```
Charles-Gagnon, line 157: We already have a helper method for this [https://github.com/Microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/Management/Common/Utils.cs#L547](https://github.com/Microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/Management/Common/Utils.cs#L547)

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/SqlAssessmentService.cs
```diff
@@ -0,0 +1,512 @@
...
+            };
+
+            await ConnectionService.Connect(connectParams);
+
+            var connection = await ConnectionService.Instance.GetOrOpenConnection(taskUri, ConnectionType.Query);
+
+            try
+            {
+                const string PlatformQueryText = @"SELECT [host_platform] FROM sys.dm_os_host_info";
+
```
Charles-Gagnon, line 267: Why not add this to the other helper method (with the same logic to check for server version first)? Again this seems like a generally useful thing to have

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
aguzev: No, this method is called once per instance. That's why I haven't implemented a facility for multiple CancellationTokenSources. But there is no guarantee. I used an exception to catch possible bugs in the future. I'd rather ensure the invariant of `cancellation != null` with GenerateScriptOperation constructor, but ITaskOperation was not IDisposable enough to ensure proper disposal of the CancellationTokenSource.  I'd be glad to remove this check if everyone calling ITaskOperation.Execute was informed of the consequences.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
aguzev: No, it was generated. We can see it on many other lines with parameterized strings.  [1626](https://github.com/microsoft/sqltoolsservice/blob/cf6eec47d3a195f906706668ae1ee327494c2936/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx#L1626)

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
Charles-Gagnon: Other existing operations just create a cancellation token on operation creation. I'd say just do the same here so at least we're consistent - the current way the code is set up I don't see the need to try and make any special cases for trying to handle that.

aguzev: They are not calling the CancellationTokenSource.Dispose method. Looks odd.

aguzev: When was this method called? [SchemaCompareGenerateScriptOperation.Dispose](https://github.com/microsoft/sqltoolsservice/blob/cf6eec47d3a195f906706668ae1ee327494c2936/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareGenerateScriptOperation.cs#L95)

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
aguzev: master branch contains the dot as well. [sr.resx#L320](https://github.com/microsoft/sqltoolsservice/blob/0a53ff9ff7241156a88c1311e352042480bdc1db/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx#L320)

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/SqlAssessmentService.cs
aguzev: Moved to `ReliableConnectionHelper.GetServerHostInfo`.

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
aguzev: Replaced with string interpolation. Surprisingly, it was not faster, but much more readable.

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/GenerateScriptOperation.cs
aguzev: Made `GenerateScriptOperation` IDisposable.

PR #721, opened 2018-10-30
Feat/result streaming

ranasaria:
This changes adds the following two notifications from the results processing within a batch. These new notifications allows a consumer to stream results from a resultset instead of getting them all at once after the entire resultset has been fetched.
1. ResultsAvailable
This is issued after at least 1 row has been fetched for this resultset.
2. ResultsUpdated
This is issued periodically as more rows are available on this resultset. The final send of this notification when all rows have been fetched has the property 'Complete' set to true in the ResultSummary object.

PR #950, opened 2020-04-22
Reduce the verbosity of logging

aaomidi:
Addresses https://github.com/microsoft/azuredatastudio/issues/10108

The content logging was introduced in a previous PR:

https://github.com/microsoft/sqltoolsservice/commit/6dd9a4b5f1a202755cacd51762da5318da4eb0f5#diff-5795c8aed9d883c5d8451bd71d88a49c

https://github.com/microsoft/sqltoolsservice/pull/721


aaomidi:
It might make sense to give users an option (with informed consent) to log PII. But I don't think we've ever had to use that?

Charles-Gagnon:
Yeah, in the end though I think we can just get have a policy not to log any PII for now. That's the simplest and safest and if we need more verbose logging in the future we can figure it out then.

PR #953, opened 2020-04-29
Add Always Encrypted Parameterization Functionality

Xtrimmer:
**Problem**
In Azure Data Studio, queries that insert or update encrypted columns are not supported. Additionally, Selecting data while filtering on encrypted column values in the WHERE clause is not supported.

**Why?**
Azure Data Studio sends the query in the exact same text that was originally typed into the query window by the user, including the values of the data to be inserted, updated, or filtered upon in a SELECT query. As a result, the query fails with an encryption scheme mismatch error because SQL Server expects the values correlating to encrypted columns to be encrypted, not in plain text.

**Solution:**
Implement parameterization of T-SQL variables with inline declarations. Expose a checkbox to enable parameterization. The design needs to include the ability for a user to receive notifications about which variables gets parameterized and how (and error messages).

**This PR**
On the STS side, I have introduced a few classes that will encapsulate almost all the logic to do both query parameterization and return Intellisense items back to the ADS UI. The entry point for both lies in the `SqlParameterizer.cs` class. 
1. The `Parameterize()` method is an extension method on the `DbCommand` object that extracts the Command Text from within and converts any candidate variables into `SqlParameter` objects.
2. The `CodeSense()` method will take in a sql script and return a collection of `ScriptFileMarker` objects that will get returned to ADS for UI display as Intellisense.

In addition, on the ADS side, there will be a User Setting feature flag where one can enable and disable this functionality. That work can be found in [this PR for ADS](https://github.com/microsoft/azuredatastudio/pull/10182).


src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.IO;
+using Microsoft.SqlServer.TransactSql.ScriptDom;
+using System.Linq;
+using Microsoft.SqlTools.ServiceLayer.AutoParameterizaition.Exceptions;
+using System.Data.Common;
+using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.AutoParameterizaition
+{
+    public static class SqlParameterizer
```
Charles-Gagnon, line 16: So I assume a lot of this logic is copied from the SSMS implementation? Have you looked into what it would take to have this stuff be in a common module that both use instead of having all this duplicated code?

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
```diff
@@ -0,0 +1,128 @@
...
+        private static readonly IList<ScriptFileMarker> EmptyCodeSenseItemList = Enumerable.Empty<ScriptFileMarker>().ToList();
+
+        public static SqlScriptGenerator GetScriptGenerator()
+        {
+            return new Sql150ScriptGenerator();
+        }
+
+        public static TSqlParser GetTSqlParser(bool initialQuotedIdentifiers)
+        {
+            return new TSql150Parser(initialQuotedIdentifiers);
```
Charles-Gagnon, line 28: So I remember leaving this comment in the original implementation in SSMS - but isn't this going to cause problems as time goes on and new syntax is added to newer versions? Using a hardcoded parser version like this means it'll fail to parse that until we go in and update this.

Can't you just fetch the latest parser version available and use that? That was as we update the parser lib it'll keep itself up to date

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
```diff
@@ -0,0 +1,41 @@
...
+        public ParameterizationFormatException(MessageHelper.MessageType type, string variableName, string sqlDataType, string cSharpDataType, string literalValue, int lineNumber, Exception e)
+            : base(MessageHelper.GetLocalizedMessage(type, variableName, sqlDataType, literalValue), e)
+        {
+            LineNumber = lineNumber;
+            LiteralValue = literalValue;
+            SqlDatatype = sqlDataType;
+            CSharpDataType = cSharpDataType;
+            CodeSenseMessage = this.Message;
+            LogMessage = MessageHelper.GetLocaleInvariantMessage(type, variableName, sqlDataType) + ", Literal Value: " + literalValue + ", On line: " + lineNumber + ", CSharp DataType: " + cSharpDataType;
+            TelemetryMessage = MessageHelper.GetLocaleInvariantMessage(type, variableName, sqlDataType);
```
Charles-Gagnon, line 38: Is the variable name the generated variable name? I guess that's ok to send - although is it actually going to be useful to have that? In general we shouldn't send anything that isn't going to provide value so make sure everything here is useful in some way

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
Xtrimmer: That is a good point and something I'll ask Brian about. If there is a way to get the latest parser from DacFx then I will make use of it.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
Xtrimmer: There are some differences. For example in SSMS, this class is not static nor it the parameterization an extension method on the DbCommand object. Also, the code sense returns an ADS specific ScriptFileMarker instead of whatever SSMS uses for the UI to display information. Other than that, the core is basically the same stuff. I will take a look and see what it will take to extract and centralize the similar logic. I am sure we could adapt to the differences.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using Microsoft.SqlTools.ServiceLayer.AutoParameterizaition.Helpers;
+using System;
+
+namespace Microsoft.SqlTools.ServiceLayer.AutoParameterizaition.Exceptions
+{
+    /// <summary>
+    /// ParameterizationFormatException is used to surface format exceptions encountered in the TSQL batch to perform auto-parameterization of literals
+    /// </summary>
+    class ParameterizationFormatException : FormatException
```
Benjin, line 14: class scoping

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
```diff
@@ -0,0 +1,41 @@
...
+        public readonly int LineNumber;
+        public readonly string LiteralValue;
+        public readonly string CodeSenseMessage;
+        public readonly string LogMessage;
+        public readonly string TelemetryMessage;
+        public readonly string SqlDatatype;
+        public readonly string CSharpDataType;
+
+        public ParameterizationFormatException(MessageHelper.MessageType type, string variableName, string sqlDataType, string cSharpDataType, string literalValue, int lineNumber)
+            : this(type, variableName, sqlDataType, cSharpDataType, literalValue, lineNumber, null)
```
Benjin, line 25: named parameter for the magic constant

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
```diff
@@ -0,0 +1,41 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.ServiceLayer.AutoParameterizaition.Helpers;
+using System;
```
Benjin, line 7: System.* before other usings

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
```diff
@@ -0,0 +1,722 @@
...
+using Microsoft.SqlServer.TransactSql.ScriptDom;
+using Microsoft.SqlTools.ServiceLayer.AutoParameterizaition.Exceptions;
+using Microsoft.SqlTools.ServiceLayer.AutoParameterizaition.Helpers;
+using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Data.SqlTypes;
+using System.Globalization;
+using System.Linq;
```
Benjin, line 16: using ordering

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
```diff
@@ -0,0 +1,722 @@
...
+            if (node == null)
+            {
+                return;
+            }
+
+            CurrentScalarExpression = node;
+
+            if (node is Literal literal)
+            {
+
```
Benjin, line 133: remove empty line

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
```diff
@@ -0,0 +1,722 @@
...
+                }
+
+                base.ExplicitVisit(node); //let the base class finish up
+                return;
+            }
+
+
+            if (node is ParenthesisExpression parenthesisExpression)
+            {
+
```
Benjin, line 168: and here

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
```diff
@@ -0,0 +1,722 @@
...
+                    {
+                        parseSuccessful = false;
+                        string sqlDataTypeString = GetSqlDataTypeStringOneParameter(SqlDbType.VarBinary, sqlDataTypeParameters);
+                        HandleError(MessageHelper.MessageType.ERROR_MESSAGE, literal, VariableName, sqlDataTypeString, C_SHARP_BYTE_ARRAY, literalValue, literal.StartLine, e);
+                    }
+                    break;
+
+
+                //Integer literals of form 24.0 will not be supported
+                case SqlDataTypeOption.BigInt:
```
Benjin, line 277: My gut feeling is that some of these (e.g. BigInt/Int/SmallInt and Money/SmallMoney) can be consolidated.  Maybe worth looking into?

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
Benjin: (and elsewhere; there's a VS setting that will do this automatically, dunno about VS Code)

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
```diff
@@ -0,0 +1,128 @@
...
+using Microsoft.SqlTools.ServiceLayer.Workspace.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.AutoParameterizaition
+{
+    public static class SqlParameterizer
+    {
+        private const int maxStringLength = 300000; // Approximately 600 Kb
+        private static readonly IList<ScriptFileMarker> EmptyCodeSenseItemList = Enumerable.Empty<ScriptFileMarker>().ToList();
+
+        public static SqlScriptGenerator GetScriptGenerator()
```
Benjin, line 21: OoC, does this codebase allow inline definitions yet?

`public static SqlScriptGenerator GetScriptGenerator() => new Sql150ScriptGenerator();`

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
```diff
@@ -0,0 +1,128 @@
...
+                ScriptFileMarker maxStringLengthCodeSenseItem = new ScriptFileMarker
+                {
+                    Level = ScriptFileMarkerLevel.Error,
+                    Message = SR.ScriptTooLarge(maxStringLength, CurrentScriptlength),
+                    ScriptRegion = new ScriptRegion
+                    {
+                        StartLineNumber = 1,
+                        StartColumnNumber = 1,
+                        EndLineNumber = 2,
+                        EndColumnNumber = 1
```
Benjin, line 78: Are these the default values for when a string is too long again?  Add a comment explaining what they are, or someone might think `EndLineNumber = 2` is a mistake given the rest are 1

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
Xtrimmer: I asked the DacFx folks and they told me that there is no method to get the latest version of the parser/scriptor currently.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
Xtrimmer: The variable name is the user declared variable name. For example the following query...
```sql
DECLARE @Number int = 'ABCDEFG'

SELECT * FROM [dbo].[Table]
WHERE [N] = @Number
GO
```
...will display the following message. If there are multiple variable/paramers then the variable name will identify which one is causing the problem.
`Unable to Convert @Number to a Microsoft.Data.SqlClient.SqlParameter object. The specified literal cannot be converted to Int(System.Data.SqlDbType)., Literal Value: ABCDEFG, On line: 2, CSharp DataType: Int32`

Charles-Gagnon: Ok. My default answer to anything user-generated is "no, we can't send that up as telemetry".

If having the name is important for the data you want to gather let's have a quick chat about what we can do.

You can still capture stuff like the types - just not anything user-generated. 

(not I'm not talking about the LogMessage here. That's fine as-is. I'm referring to the TelemetryMessage field)

Xtrimmer: You know what? This `TelemetryMessage` is not even used anywhere now so the simplest thing to do would be to remove it alltogether.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.AutoParameterizaition.Exceptions
+{
+    /// <summary>
+    /// ParameterizationFormatException is used to surface format exceptions encountered in the TSQL batch to perform auto-parameterization of literals
```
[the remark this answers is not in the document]
Xtrimmer: Added the detail to the docstring

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
Xtrimmer: added scope.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
Xtrimmer: renamed variable from e to exception and added the parameter hint to the function call

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/Exceptions/ParameterizationFormatException.cs
Xtrimmer: fixed

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
Xtrimmer: fixed here and in other files.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
Xtrimmer: fixed.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
Xtrimmer: fixed

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/ScalarExpressionTransformer.cs
Xtrimmer: I was thinking about whether I should decompose this switch/case to a collection of independent command patterned classes for each datatype that is acquired by some factory method. I debated whether this would be overkill or not?

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
Xtrimmer: Let's try it and see! Converted to expression body definitions.

src/Microsoft.SqlTools.ServiceLayer/AutoParameterizaition/SqlParameterizer.cs
Xtrimmer: added comment to explain magic numbers

PR #955, opened 2020-05-01
Enable win10 arm builds

aaomidi:
This enables sqltoolsservice to be published for arm machines as well.

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -179,6 +179,20 @@ steps:
     includeRootFolder: false
     archiveFile: '$(Build.SourcesDirectory)/artifacts/package/Microsoft.SqlTools.ServiceLayer-win-x86-netcoreapp2.2.zip'
 
+- task: ArchiveFiles@1
+  displayName: 'Archive windows10 arm 32 bit build'
+  inputs:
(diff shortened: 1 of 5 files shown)


anthonydresser:
do we actually have a use case for this?

aaomidi:
It's been a feature request here: https://github.com/microsoft/azuredatastudio/issues/1774

I just thought I'd start on the sqltoolsservice side since the ads side shouldn't be terribly difficult considering I've seen repos like this do it: https://github.com/stevedesmond-ca/vscode-arm

aaomidi:
The reason I opened the PR is also to see what the team generally thinks about just having these built & published and then used later on.

Charles-Gagnon:
Yeah I remembered the people asking about that right after I said that. Fine with experimentation - do you have any idea how complicated this would be to add on the ADS side?

aaomidi:
I don't suspect it being that difficult, we have some node modules that might give us some problems

(Looking at you Kerberos)

But I think just getting a test build manually built out by hand and validating it shouldn't be terribly difficult)

aaomidi:
It'd be pretty cool to see ADS on something like a chromebook or Windows 10X devices

PR #956, opened 2020-05-04
Find credentials api

aaomidi:
- [x] Windows
- [x] Linux
- [ ] OSX
- [ ] Expose API protocol
- [ ] Update Data Protocol Client
- [ ] Update ADS

This API would allow us to mass delete the user's cached password to a certain service. We should be doing this when we uninstall the application & also give a way for the user to clear out their password.


aaomidi:
I need help in trying to implement the OSX password searching.

According to the docs here: https://developer.apple.com/documentation/security/keychain_services/keychain_items?language=objc

The way to do it is using https://developer.apple.com/documentation/security/1398306-secitemcopymatching?language=objc

I'm not entirely sure how to express CFDictionaryRef and the variable return type (either Array or single Element) in C#.

Any help in this space would be greatly appreciated.

aaomidi:
Essentially I need help in integrating that API call into this file: https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.Credentials/Credentials/OSX/Interop.Security.cs

src/Microsoft.SqlTools.Credentials/Credentials/OSX/OSXCredentialStore.cs
```diff
@@ -45,14 +46,42 @@ public bool Save(Credential credential)
             return result;
         }
 
+        public bool GetCredentials(string credentialPattern, out List<Credential> credentials)
+        {
+            var query = new List<(string, string)>
+            {
+                ("kSecClass", "kSecClassGenericPassword"),
+                ("kSecAttrService", credentialPattern),
```
Charles-Gagnon, line 54: So what is the pattern you're planning on using here? Let's be careful not to accidently be deleting other credentials that don't belong to us. Maybe it'd be enough to list the credentials to delete for the user to accept before doing the actual delete?

aaomidi: Nothing in SqlToolsService, here it will support any generic pattern.

aaomidi:
Unfortunately it seems upon further investigation we're not going to be able to bring this functionality to OSX.

So, OSX generic passwords come with multiple fields, but the ones that are important for us here are

- "Service"
- "Account"
- "Password"

Currently, SqlToolsService stores the key of the password in Service. OSX does not allow partial Service mapping, and doesn't seem to allow listing all credentials so we can filter out the ones with the Service strings that match our needs.

The correct way to store passwords would've been to put "Microsoft.SqlTools" as the Service string (hardcoded so no one can accidentally change that), and then each account string would've been the key to store and lookup passwords by. This would allow us to match all credentials stored with Microsoft.SqlTools key as the service string and from there allow us to delete any password that was under that service.

Changing this would break saved passwords for existing users, so I don't feel comfortable doing that.

PR #957, opened 2020-05-05
Fix for list databases service API for contained databases 

aasimkhan30:
This PR fixes microsoft/azuredatastudio#9361 in ADS where the list database fails when the user doesn't have access to the master DB. 

-Changed Raw SQL Query to SMO objects for list db
-Added edge cases contaiend and azure DB
- Now list db will only show databases which users have access to.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
@@ -63,8 +63,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         {
             ConnectionDetails connectionDetails = connectionInfo.ConnectionDetails.Clone();
 
-            // Connect to master
-            connectionDetails.DatabaseName = "master";
             using (var connection = connectionFactory.CreateSqlConnection(ConnectionService.BuildConnectionString(connectionDetails), connectionDetails.AzureAccountToken))
             {
                 connection.Open();


src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+                            )
                         {
-                            results.Add(this.CreateItem(reader));
+                            Console.WriteLine(exc.Message);
+                        }
+                        else
+                        {
+                            throw exc;
                         }
-                        // Put system databases at the top of the list
```
Charles-Gagnon, line 84: Your code doesn't seem to be doing this sorting any more

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
-                connection.Open();
-                ListDatabasesResponse response = new ListDatabasesResponse();
-                using (DbCommand command = connection.CreateCommand())
+                if (db.Name != serverConnection.CurrentDatabase)
                 {
-                    command.CommandText = this.QueryText;
-                    command.CommandTimeout = 15;
-                    command.CommandType = CommandType.Text;
-                    using (var reader = command.ExecuteReader())
+                    if(serverConnection.IsContainedAuthentication){ //Check if the user is a contained user. If yes, then don't add the database.
```
Charles-Gagnon, line 80: Why are we always skipping this? If a user connects directly to the database then I'd expect this to at least return that DB in the list

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+            List<T> result = new List<T>();
+            ListDatabasesResponse response = new ListDatabasesResponse();
+
+
+            foreach (SMO.Database db in SQLServer.Databases)
             {
-                connection.Open();
-                ListDatabasesResponse response = new ListDatabasesResponse();
-                using (DbCommand command = connection.CreateCommand())
+                if (db.Name != serverConnection.CurrentDatabase)
```
Charles-Gagnon, line 78: Why are you skipping the current DB? We want those to show up in the list still.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+                        dbServerConnection.ExecuteNonQuery(String.Format(CultureInfo.InvariantCulture ,"use [{0}]", Util.EscapeString(db.Name , ']')));
+                        result.Add(CreateItem(db));
+                    }
+                    catch (Exception exc)
                     {
-                        List<T> results = new List<T>();
-                        while (reader.Read())
+                        SqlException sqlExc = exc.InnerException as SqlException;
+                        if (sqlExc != null
+                            && (sqlExc.Number == 916 //The server principal is not able to access the database under the current security context.
```
Charles-Gagnon, line 99: I think these errors need some more thought. There's a whole bunch of reasons someone might not be able to access a database (one common one we hit is if the database is restoring).

If we're doing this we should really be looking at filtering out any databases we can't access for any reason - although that requires some more thought and testing before just implementing.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+                                || sqlExc.Number == 911) //Database does not exist. Make sure that the name is entered correctly.
+                            )
                         {
-                            results.Add(this.CreateItem(reader));
+                            Console.WriteLine(exc.Message);
```
Charles-Gagnon, line 103: Don't write to console, that isn't very helpful here. Use the output logger and make sure to add context so we know where the error is coming from

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
-                    command.CommandText = this.QueryText;
-                    command.CommandTimeout = 15;
-                    command.CommandType = CommandType.Text;
-                    using (var reader = command.ExecuteReader())
+                    if(serverConnection.IsContainedAuthentication){ //Check if the user is a contained user. If yes, then don't add the database.
+                        continue;
+                    }
+
+                    ServerConnection dbServerConnection;
+                    dbServerConnection = serverConnection.GetDatabaseConnection(db.Name);
```
Charles-Gagnon, line 85: You can just run the query on the database directly and it'll handle switching as needed

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+
 
-            // Connect to master
-            connectionDetails.DatabaseName = "master";
-            using (var connection = connectionFactory.CreateSqlConnection(ConnectionService.BuildConnectionString(connectionDetails), connectionDetails.AzureAccountToken))
+            List<T> result = new List<T>();
+            ListDatabasesResponse response = new ListDatabasesResponse();
+
+
+            foreach (SMO.Database db in SQLServer.Databases)
```
Charles-Gagnon, line 76: @shueybubbles Remind me - does SMO already filter out inaccessible DBs from the DB list for a server?

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
Charles-Gagnon, line 80: Format this file please

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
     {
         private static readonly string[] SystemDatabases = new string[] { "master", "model", "msdb", "tempdb" };
 
-        public abstract string QueryText { get; }
-
         public ListDatabasesResponse HandleRequest(ISqlConnectionFactory connectionFactory, ConnectionInfo connectionInfo)
         {
             ConnectionDetails connectionDetails = connectionInfo.ConnectionDetails.Clone();
```
Charles-Gagnon, line 66: Doesn't seem like ConnectionDetails is being used anymore?

Charles-Gagnon:
Also - is there an issue for this that you're fixing? It's good to have context on specific issues people are having and how this is addressing them.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+            var serverConnection = ConnectionService.OpenServerConnection(connectionInfo);
+
+            SMO.Server SQLServer = new SMO.Server(serverConnection);
```
shueybubbles, line 69: local variables are usually lower case named, maybe just use "server"

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+
 
-            // Connect to master
-            connectionDetails.DatabaseName = "master";
-            using (var connection = connectionFactory.CreateSqlConnection(ConnectionService.BuildConnectionString(connectionDetails), connectionDetails.AzureAccountToken))
+            List<T> result = new List<T>();
```
shueybubbles, line 72: var is your friend

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+
+                    ServerConnection dbServerConnection;
+                    dbServerConnection = serverConnection.GetDatabaseConnection(db.Name);
+
+                    SqlExecutionModes originalMode = serverConnection.SqlExecutionModes;
+                    dbServerConnection.SqlExecutionModes = SqlExecutionModes.ExecuteSql;
+
+                    try
+                    {
+                        dbServerConnection.ExecuteNonQuery(String.Format(CultureInfo.InvariantCulture ,"use [{0}]", Util.EscapeString(db.Name , ']')));
```
shueybubbles, line 92: I don't understand why we don't just return all the names from server.Databases
All this other contained auth stuff is overkill and doesn't handle all the reasons why someone might fail to connect to a database anyway.
Besides , "use" isn't supported in Azure sql database etc. Also there can be hundreds of databases on a server, why would we loop through them and slow down the experience?

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -139,18 +158,18 @@ protected override void SetResponse(ListDatabasesResponse response, DatabaseInfo
...
         }
 
-        protected virtual void SetProperties(DbDataReader reader, DatabaseInfo databaseInfo)
+        protected virtual void SetProperties(SMO.Database database, DatabaseInfo databaseInfo)
         {
-            databaseInfo.Options[ListDatabasesRequestDatabaseProperties.Name] = reader["name"].ToString();
-            databaseInfo.Options[ListDatabasesRequestDatabaseProperties.State] = reader["state"].ToString();
-            databaseInfo.Options[ListDatabasesRequestDatabaseProperties.SizeInMB] = reader["size"].ToString();
+            databaseInfo.Options[ListDatabasesRequestDatabaseProperties.Name] = database.Name;
+            databaseInfo.Options[ListDatabasesRequestDatabaseProperties.State] = database.State.ToString();
```
shueybubbles, line 171: before accessing server.Databases collection, either:
1. Use server.SetDefaultInitFields(typeof(Database), new [] {nameof(`<property1>`), nameof(`<property2>`) etc}) 
or 
2. Use server.Databases.ClearAndInitialize(new [] {nameof(`<property1>`), nameof(`<property2>`) 

where the `<property>` names are the ones you access here. This will dramatically improve performance

shueybubbles: on further thought...we should be very careful about fetching any properties besides Name for Azure databases in particular.
Azure has the potential to create major headaches here, as fetching most properties of each database requires logging in to the database. Compounding the problem is that if the connection string was to "master", sys.databases returns all the databases on the server, even ones that may be paused or that the user doesn't have access to. So fetching any property besides Name has side effects like:
1. Waking up paused databases
2. Throwing SqlException when login fails to the database.
3. Taking a looong time to log in to low SLO databases etc

It may be better to defer such property fetches until they are actually needed.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
aasimkhan30: I did check it out. From my testing, it returned all the databases that are present in sys.databases for the connection.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +61,66 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
     {
         private static readonly string[] SystemDatabases = new string[] { "master", "model", "msdb", "tempdb" };
 
-        public abstract string QueryText { get; }
-
         public ListDatabasesResponse HandleRequest(ISqlConnectionFactory connectionFactory, ConnectionInfo connectionInfo)
         {
             ConnectionDetails connectionDetails = connectionInfo.ConnectionDetails.Clone();
+            var serverConnection = ConnectionService.OpenServerConnection(connectionInfo);
```
alanrenmsft, line 67: the connection is not being disposed properly

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
shueybubbles: SMO has some exception handlers in place to handle failure to fetch the Edition of each database in the list, so it sets Edition to Unknown on those it can't login to, but it has no way of excluding such DBs from the list.

alanrenmsft, line 76: I have a concern about the performance here, with this change, you have changed the required server roundtrips from 1 to 1+x (number of dbs). that is what i tried to avoid when i was working on this.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -4,192 +4,177 @@
...
+			foreach (SMO.Database db in server.Databases)
+			{
+				results.Add(CreateItem(db));
+			}
+
+			results = results.Where(s => SystemDatabases.Any(x => this.NameMatches(x, s))).Concat(
+						 results.Where(s => SystemDatabases.All(x => !this.NameMatches(x, s)))).ToList();
+
+			SetResponse(response, results.ToArray());
+			serverConnection.Disconnect();
```
shueybubbles, line 83: should this be in a finally block?

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -4,192 +4,177 @@
...
+	}
+
+	/// <summary>
+	/// Standalone SQL Server database detail handler
+	/// </summary>
+	class SqlServerDatabaseDetailHandler : BaseDatabaseDetailHandler
+	{
+		protected override void setDefaultServerInitProperties(SMO.Server server)
+		{
+			server.SetDefaultInitFields(typeof(Database), new[] { nameof(Database.LastBackupDate) });
```
shueybubbles, line 159: call base.set too

Charles-Gagnon:
@aasimkhan30  If you're fixing an issue it's helping to use the #fixes tag so the issue is auto-closed after the PR is merged in 

https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -4,223 +4,177 @@
...
+	}
+
+	/// <summary>
+	/// database names handler
+	/// </summary>
+	class DatabaseNamesHandler : ListDatabaseRequestHandler<string>
+	{
+		protected override void setDefaultServerInitProperties(SMO.Server server)
+		{
+			server.SetDefaultInitFields(typeof(Database), new[] { nameof(Database.Name) });
```
Charles-Gagnon, line 101: Did you test this? This shouldn't be necessary - names are one of the properties that are always fetched (since the server needs to know about them to populate the collection)

shueybubbles:
> @aasimkhan30 you have resolved a bunch of comments without answering them, please keep them open until you addressed them. regarding to my previous comment, you are still using SMO to get the database information, which will be at least 1+X server roundtrips. we chatted about this offline, are you still working on this?

Even for details, SMO would make 1 query since he's using SetDefaultInitFields. The exception would be if one of the properties is only accessible from within the database context itself, in which case he'd have to be making the loop anyway.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -4,223 +4,177 @@
...
+			databaseInfo.Options[ListDatabasesRequestDatabaseProperties.LastBackup] = (database.LastBackupDate == default(DateTime)) ? "" : database.LastBackupDate.ToString("yyyy-MM-dd hh:mm:ss");
+		}
+	}
+
+	/// <summary>
+	/// SQL DB database detail handler
+	/// </summary>
+	class SqlDBDatabaseDetailHandler : BaseDatabaseDetailHandler
+	{
+		protected override void setDefaultServerInitProperties(SMO.Server server)
```
Charles-Gagnon, line 174: Method names in C# are capitalized

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -4,223 +4,177 @@
...
+
+			// Using this function to set the default init properties of the server so that those things can be fetched in a single query.
+			setDefaultServerInitProperties(server);
+
+			foreach (SMO.Database db in server.Databases)
+			{
+				results.Add(CreateItem(db));
+			}
+
+			results = results.Where(s => SystemDatabases.Any(x => this.NameMatches(x, s))).Concat(
```
Charles-Gagnon, line 79: Put the comment back in about sorting the system names to the top

alanrenmsft:
I see, you are leveraging the SMO init properties now. have you tested whether the state, size and lastbackup done in just one query? @aasimkhan30

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
alanrenmsft, line 174: you don't need to override this method here, also instead of a method it is more natural for me to just have a property, InitialDatabaseProperties

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
alanrenmsft: previously i left a comment about putting the serverConection into a using statement.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
aasimkhan30: Yeah, the names are fetched in a single query. I'll remove this

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
aasimkhan30: Unfortunately, I cannot use the using statement for serverConnection. I'll add that in a finally block.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
aasimkhan30: I guess, I won't have to do this if I'm using overridden properties instead of this method.

aasimkhan30:
> I see, you are leveraging the SMO init properties now. have you tested whether the state, size and lastbackup done in just one query? @aasimkhan30

@alanrenmsft I have checked it with the database name fetching on a live database using SQL profiler. It uses only one query for that. 

Can you tell me an entry point for the other apis on ADS. If there are none I can probably write some integration tests to do the same.

aasimkhan30:
> @aasimkhan30 If you're fixing an issue it's helping to use the #fixes tag so the issue is auto-closed after the PR is merged in
> 
> https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

@Charles-Gagnon I've updated the pull request to point to the issue.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -57,43 +60,43 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
             {
-                connection.Open();
-                ListDatabasesResponse response = new ListDatabasesResponse();
-                using (DbCommand command = connection.CreateCommand())
+                SMO.Server server = new SMO.Server(serverConnection);
+
+                // Using this function to set the default init properties of the server so that those things can be fetched in a single query.
+                server.SetDefaultInitFields(typeof(Database), DefaultServerInitProperties);
+
+                foreach (SMO.Database db in server.Databases)
```
Charles-Gagnon, line 79: Currently you're changing the behavior here - previously we were only listing databases where state was ONLINE. You should add that back in unless we decide we want to display offline DBs as well.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -199,28 +188,6 @@ protected override void SetProperties(DbDataReader reader, DatabaseInfo database
     /// </summary>
     class SqlDBDatabaseDetailHandler : BaseDatabaseDetailHandler
```
Charles-Gagnon, line 189: This class is pointless now - I'd just get rid of it and make the base class a regular class instead of abstract.

Charles-Gagnon:
@aasimkhan30 This still isn't going to auto-close the issue. Read the article I linked, there's very specific keywords you have to put before the issue link to have Github auto-close the issue.

e.g. 

Fixes #<issue#>
Resolves #<issue#>
Close #<issue#>

Charles-Gagnon:
@aasimkhan30 There are failing tests now as well, please fix those

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -58,16 +58,21 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
-            // Connect to master
-            connectionDetails.DatabaseName = "master";
             using (var connection = connectionFactory.CreateSqlConnection(ConnectionService.BuildConnectionString(connectionDetails), connectionDetails.AzureAccountToken))
             {
                 connection.Open();
+
+                // Connect to master if not a contained user 
+                if (!connectionFactory.CheckIfContainedAuthorization(connection))
+                {
+                    connectionDetails.DatabaseName = "master";
```
shueybubbles, line 73: why are you forcing a connection to master in Azure?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -84,7 +82,7 @@ public void CanCancelConnectRequest()
             var mockFactory = new Mock<ISqlConnectionFactory>();
             mockFactory.Setup(factory => factory.CreateSqlConnection(It.IsAny<string>(), It.IsAny<string>()))
                 .Returns(mockConnection.Object);
-
+            mockFactory.Setup(factory => factory.CheckIfContainedAuthorization(mockConnection.Object)).Returns(false);
```
shueybubbles, line 85: add a case where it returns true

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
aasimkhan30: I though that line was present for back compatibility. If it is not necessary, I will just get rid of this line.

src/Microsoft.SqlTools.ServiceLayer/Connection/ISqlConnectionFactory.cs
```diff
@@ -16,5 +18,15 @@ public interface ISqlConnectionFactory
...
+
+        /// <summary>
+        /// Safely casts DbConnection into SqlConnection
+        /// </summary>
+        SqlConnection ConvertDBConnectionToSqlConnection(DbConnection connection);
+
+        /// <summary>
+        /// Check if authorization is for a contained database user.
+        /// </summary>
+        bool CheckIfContainedAuthorization(DbConnection connection);
```
kburtram, line 30: I don't think this is the correct interface for these methods.  Possibly the Convert method could be in a factory interface, but certainly the check method belongs.  Both the convert and check operations aren't creating anything, so problem neither belongs in a factory and should be in a helper\utils class.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
kburtram: this was my question as well.  I'd say if no one knows why we need to set to master, perhaps we should just not do that and test on the lowest supported server version for backcompat (SQL Server 2014).

src/Microsoft.SqlTools.ServiceLayer/Connection/SqlConnectionFactory.cs
```diff
@@ -24,5 +26,36 @@ public DbConnection CreateSqlConnection(string connectionString, string azureAcc
...
+            return sqlConn;
+        }
+
+        /// <summary>
+        /// Check if authorization is for a contained database user.
+        /// </summary>
+        public bool CheckIfContainedAuthorization(DbConnection connection)
+        {
+            SqlConnection sqlConnection = this.ConvertDBConnectionToSqlConnection(connection);
+            return new ServerConnection(sqlConnection).IsContainedAuthentication;
```
kburtram, line 58: should `sqlConnection` be null  checked before being used to create a ServerConnection?

src/Microsoft.SqlTools.ServiceLayer/Connection/SqlConnectionFactory.cs
```diff
@@ -24,5 +26,36 @@ public DbConnection CreateSqlConnection(string connectionString, string azureAcc
             RetryPolicy commandRetryPolicy = RetryPolicyFactory.CreateDefaultConnectionRetryPolicy();
             return new ReliableSqlConnection(connectionString, connectionRetryPolicy, commandRetryPolicy, azureAccountToken);
         }
+        /// <summary>
+        /// Safely casts DbConnection into SqlConnection
+        /// </summary>
+        public SqlConnection ConvertDBConnectionToSqlConnection(DbConnection connection)
+        {
+            // Get a connection to the database for SMO purposes
```
Charles-Gagnon, line 34: This method itself has nothing to do with SMO so I would just omit the comments here mentioning SMO (or at least make them more generic) - it's up to the callers what they do with the value returned.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/TestObjects.cs
```diff
@@ -282,5 +283,16 @@ public DbConnection CreateSqlConnection(string connectionString, string azureAcc
...
         }
+
+        public SqlConnection ConvertDBConnectionToSqlConnection(DbConnection connection)
+        {
+            return new SqlConnection(connection.ConnectionString);
+        }
+
+        public bool CheckIfContainedAuthorization(DbConnection connection)
+        {
+            //Returning false to mathc the previous behaviour
```
Charles-Gagnon, line 294: This comment doesn't explain anything. What previous behavior? Why does it matter for a test object what the "previous" behavior was?

aasimkhan30:
I have checked the code on SQLserver 2008 and 2014. I think it is safe to remove the forced master connection line. Let me know if you there is something that I'm missing.

PR #958, opened 2020-05-08
Updating DacFx nuget to one with SqlTasks

udeeshagautam:
This is to update dacfx Nuget and ensure targets files gets published along with tools service. 
This nuget contain the sqltasks dll and targets.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
+++ b/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.0.0" />
-    <PackageReference Include="Microsoft.SqlServer.DacFx" Version="150.4763.1-preview" />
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.0.0" />
-    <PackageReference Include="Microsoft.SqlServer.DacFx" Version="150.4763.1-preview" />
+    <PackageReference Include="Microsoft.SqlServer.DacFx" Version="150.4779.1-preview" GeneratePathProperty="true" />
```
Charles-Gagnon, line 24: Instead of having multiple references through a couple different projects I'd suggest using a single var defined in https://github.com/microsoft/sqltoolsservice/blob/master/Directory.Build.props like we do for SMO

udeeshagautam: good point, so i see we have smo version in  top level directory build props and then ref in src\directory build props and test\ build props, we can have all in top level only right? will add all in one if thats okay

PR #963, opened 2020-05-23
Prevent ArgumentNullException in refresh request

### LINKED ISSUE #962: ArgumentNullException in Connection Refresh
Issue Type: <b>Bug</b>

```c#
Error running command mssql.refreshObjectExplorerNode: System.ArgumentNullException: Value cannot be null.
Parameter name: key
  at System.Collections.Concurrent.ConcurrentDictionary`2.ThrowKeyNullException()
  at System.Collections.Concurrent.ConcurrentDictionary`2.TryGetValue(TKey key, TValue& value)
  at Microsoft.SqlTools.ServiceLayer.ObjectExplorer.ObjectExplorerService.HandleRefreshRequest(RefreshParams refreshParams, RequestContext`1 context) in D:\a\1\s\src\Microsoft.SqlTools.ServiceLayer\ObjectExplorer\ObjectExplorerService.cs:line 245.
```

Extension version: 1.9.0
VS Code version: Code - Insiders 1.46.0-insider (2591ede95a52400d5d22989b16706f9b8b4831da, 2020-05-22T21:13:15.709Z)
OS version: Windows_NT x64 10.0.19628

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz (8 x 2808)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>opengl: enabled_on<br>protected_video_decode: enabled<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>viz_display_compositor: enabled_on<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.88GB (4.53GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details>

xsoheilalizadeh:
Fixes #962

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
@@ -243,7 +243,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer
 
                 string uri = refreshParams.SessionId;
                 ObjectExplorerSession session = null;
-                if (!sessionMap.TryGetValue(uri, out session))
+                if (string.IsNullOrEmpty(uri) || !sessionMap.TryGetValue(uri, out session))
                 {
                     Logger.Write(TraceEventType.Verbose, $"Cannot expand object explorer node. Couldn't find session for uri. {uri} ");
                     await serviceHost.SendEvent(ExpandCompleteNotification.Type, new ExpandResponse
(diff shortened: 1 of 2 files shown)


Charles-Gagnon:
In general being defensive like this seems fine but a couple of things :

1. Did you investigate why this is happening? I would not expect the URI to ever be null/empty in a real scenario so I'm curious how this is happening
2. If we do decide to do this it would seem reasonable to update the other request handlers that take in a URI as well - there shouldn't be anything special about refresh such that it'd be the only one that needs it that I know of

xsoheilalizadeh:
> Did you investigate why this is happening? I would not expect the URI to ever be null/empty in a real scenario so I'm curious how this is happening

I thought the same thing, but I considered the null value as an invalid value. I wasn't able to reproduce that because my STS knowledge is not enough yet. 

> If we do decide to do this it would seem reasonable to update the other request handlers that take in a URI as well - there shouldn't be anything special about refresh such that it'd be the only one that needs it that I know of

I'll try to reproduce It to find out the situation that this issue happens.

PR #967, opened 2020-06-05
Update schema compare default IgnoreWhitespace option

kisantia:
There was an issue found in ADS where when comparing the dacpac of AdventureWorks created from a sql project with the database, there were differences being found when there shouldn't be, but SSDT was showing no differences as expected. This is because SSDT has the option IgnoreWhitespace=true as default, but in SqlToolsService we were setting this to false, causing different behavior when comparing in ADS. After this change, comparing the dacpac created from AdventureWorks sqlproj in ADS doesn't show any differences with the database with the default options. 

There's still the other issue that there are whitespace differences with the dacpac created in ADS, but there aren't for the dacpac created in SSDT. This PR fixes https://github.com/microsoft/azuredatastudio/issues/10612 and I opened https://github.com/microsoft/azuredatastudio/issues/10777 to track the other problem for DML Triggers.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
@@ -225,7 +225,6 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
             options.DropRoleMembersNotInSource = true;
             options.IgnoreKeywordCasing = false;
             options.IgnoreSemicolonBetweenStatements = false;
-            options.IgnoreWhitespace = false;
 
             System.Reflection.PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
 
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
@@ -377,7 +377,6 @@ END
             dacOptions.DropRoleMembersNotInSource = true;
             dacOptions.IgnoreKeywordCasing = false;
             dacOptions.IgnoreSemicolonBetweenStatements = false;
-            dacOptions.IgnoreWhitespace = false;
 
             SchemaCompareTestUtils.CompareOptions(deployOptions, dacOptions);
         }


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -225,7 +225,6 @@ public DeploymentOptions()
             options.DropRoleMembersNotInSource = true;
             options.IgnoreKeywordCasing = false;
             options.IgnoreSemicolonBetweenStatements = false;
-            options.IgnoreWhitespace = false;
```
SakshiS-harma, line 228: Wouldn't this be a problem with the comment that's mentioned in the function:  "Adding these defaults to ensure behavior similarity with other tools. Dacfx and SSMS import/export wizards use these defaults."? If not DacFx, I feel it might be a problem with SSMS (based on the mentioned comment, didn't test it).

kisantia: This was originally done because there are different defaults for the different tools and we wanted to move towards having one set of defaults. However, we didn't make any other effort to have a common set of default values. In this case, I think it's better to match the default of SSDT for this option so that the comparison will have the same behavior.

SakshiS-harma: Sounds good. Thanks!

udeeshagautam:
some checks seem failing - please take a look

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -225,7 +225,6 @@ public DeploymentOptions()
             options.DropRoleMembersNotInSource = true;
             options.IgnoreKeywordCasing = false;
             options.IgnoreSemicolonBetweenStatements = false;
```
udeeshagautam, line 227: anything needed on Tool Service side?

kisantia: this is the only change needed for SqlToolsService. ADS gets the default options from here, so nothing needs to be updated there.

kisantia:
@udeeshagautam The failure isn't related to this change. Charles sent out a PR to fix it, so he said it's ok to ignore the failure :)

PR #969, opened 2020-06-08
Fix TargetName for assessment items (#968)

### LINKED ISSUE #968: Double server name in assessment item TargetName
TargetName property of items returned by `assessment/getAssessmentItems` for a SQL Server instance contains the server name repeated twice with a slash between.  

A single server name is expected. 

image

aguzev:
* Remove the second appearance of server name in TargetName property
* Replace slash with colon as database name separator in TargetName to avoid ambiguity
Fixes #968


Charles-Gagnon:
@aguzev I'm going to merge this in now so the next build has these changes.

Charles-Gagnon:
Let me know if this is something you'd like to get fixed for the June release - we'll need to do an ask mode request and the earlier we do that the better. 

(my suggestion here is that unless this is a huge issue for users that you document it as a known issue and then either encourage people to use insiders for the time being or just let them know it'll be fixed in the next release. Porting over more fixes this late in the release signoff cycle is something we usually discourage)

alexprotsenko:
@Charles-Gagnon we're good to present this fix in the next release cycle. Thanks for taking care of this

PR #970, opened 2020-06-08
Fix test notebook name casing

Charles-Gagnon:
Was causing build failures on case sensitive OS's


chlafreniere:
I believe this is in the agent extension. Hmm, doesn't look like this is referenced anywhere, correct?

Charles-Gagnon:
https://github.com/microsoft/sqltoolsservice/blob/master/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Agent/AgentTestUtils.cs#L427

This was mostly to fix the build errors that the Code Scanning pipeline was having.

PR #974, opened 2020-06-12
Bump to latest .Net Core SDK 3.1.301

kburtram:


kburtram:
We might as well go directly to latest current 3.1 SDK if we need to make build\release updates anyways.

PR #979, opened 2020-06-15
Changing folder names to netcoreapp3.1 to fix pipeline issues

aasimkhan30:
The pipeline build were failing because they were trying to find dlls in dotnetcoreapp2.2


build.cake
```diff
@@ -584,7 +584,7 @@ Task("SRGen")
                 continue;
             }
 
-            var srgenPath = System.IO.Path.Combine(toolsFolder, "Microsoft.Data.Tools.StringResourceTool", "lib", "netcoreapp2.2", "srgen.dll");
+            var srgenPath = System.IO.Path.Combine(toolsFolder, "Microsoft.Data.Tools.StringResourceTool", "lib", "netcoreapp3.1", "srgen.dll");
```
Charles-Gagnon, line 587: FYI this isn't going to work since the 2.0.0 version of the tool wasn't compiled for netcore 3.1. I have a PR out fixing this already with an updated version of the tool.

test/CodeCoverage/codecoverage.bat
```diff
@@ -58,7 +58,7 @@ dotnet restore %REPOROOT%\test\Microsoft.SqlTools.ManagedBatchParser.Integration
 dotnet build %REPOROOT%\test\Microsoft.SqlTools.ManagedBatchParser.IntegrationTests\Microsoft.SqlTools.ManagedBatchParser.IntegrationTests.csproj %DOTNETCONFIG% 
 
 SET TEST_SERVER=localhost
-SET SQLTOOLSSERVICE_EXE=%REPOROOT%\src\Microsoft.SqlTools.ServiceLayer\bin\Integration\netcoreapp2.2\win7-x64\MicrosoftSqlToolsServiceLayer.exe
```
Charles-Gagnon, line 61: Let's take the opportunity to move this into a var at the top of the file instead.

build.cake
aasimkhan30: I'll wait for your build to pass and be merged in main. After that, I will pull and add the changes to my branch as well.

test/CodeCoverage/codecoverage.bat
```diff
@@ -1,5 +1,7 @@
 SET WORKINGDIR=%~dp0
 SET REPOROOT=%WORKINGDIR%..\..
+SET SQLTOOLSSERVICE_EXE=%REPOROOT%\src\Microsoft.SqlTools.ServiceLayer\bin\Integration\netcoreapp3.1\win7-x64\MicrosoftSqlToolsServiceLayer.exe
```
Charles-Gagnon, line 3: I would instead just make a var for the netcoreapp3.1 part and then re-use that in all the places that value is used. For now that's fine since everything is compiled against the same version.

PR #983, opened 2020-06-16
2485 Create Kusto Unit Tests

JustinMDotNet:


Charles-Gagnon:
So my first reaction is that this is way too massive of a PR to realistically review in a timely manner. Could you please scale this back and just start with the few core code changes + a few tests showing how those are used? Then after that's in you can go through and add tests for each area one by one so reviewers are able to keep context on what the changes are targeting and not have to understand the entire testing suite in one go.

JustinMDotNet:
Closing PR as it's become stale and work is being done as part of Task 3278 - https://github.com/microsoft/sqltoolsservice/pull/1057

PR #985, opened 2020-06-19
2485 Add Kusto Unit Test project

JustinMDotNet:
…rowserTests.cs. Added AssemblyInfo to MIcrosoft.Kusto.ServiceLayer


test/Microsoft.Kusto.ServiceLayer.UnitTests/Microsoft.Kusto.ServiceLayer.UnitTests.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>netcoreapp2.2</TargetFramework>
```
Charles-Gagnon, line 5: STS has been updated to netcore3.1 - you should merge up to main and then update this to match the rest

test/Microsoft.Kusto.ServiceLayer.UnitTests/FileBrowser/FileBrowserTests.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Microsoft.Kusto.ServiceLayer.UnitTests.FileBrowser
+{
+    /// <summary>
+    /// File browser unit tests
+    /// </summary>
+    public class FileBrowserTests
+    {
+        [Fact]
+        public void CreateFileBrowserOperationTest()
```
Charles-Gagnon, line 19: So the test looks fine for what it is - but I guess the first question I have is do you actually need this filebrowser code for your stuff? And is it any different than the normal ServiceLayer? Because if you do and it isn't then I would rather see you make it so that the code is shared between the projects instead of duplicating it.

test/Microsoft.Kusto.ServiceLayer.UnitTests/Microsoft.Kusto.ServiceLayer.UnitTests.csproj
Charles-Gagnon: And that being said - what I would suggest is you separately do the merge from main and get that merged into your feature branch first so this PR stays concise.

.vscode/launch.json
```diff
@@ -26,6 +40,15 @@
             "requireExactSource": false,
             "justMyCode": false,
             "enableStepFiltering": false
+        },
+        {
+            "name": "Kusto Core Attach",
```
kburtram, line 45: this looks exactly the same as `.NET Core Attach` task

JustinMDotNet:
Closing this PR as the unit tests are being done in a different branch.

PR #986, opened 2020-06-22
Update DacFx and MDS to same version

udeeshagautam:
Update DacFx and MDS to same version. Fixing a test (unrelated).

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -17,9 +17,9 @@
 		<PackageReference Update="Microsoft.Azure.Management.Sql" Version="1.41.0-preview" />
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.0.0-preview3.20122.2"/>
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.0.0"/>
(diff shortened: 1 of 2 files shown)


Packages.props
```diff
@@ -17,9 +17,9 @@
 		<PackageReference Update="Microsoft.Azure.Management.Sql" Version="1.41.0-preview" />
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.0.0-preview3.20122.2"/>
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.0.0"/>
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="160.2004021.0" />
-		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="150.4787.1-preview" GeneratePathProperty="true" />
+		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="150.4822.1-preview" GeneratePathProperty="true" />
```
Benjin, line 22: What does the DacFx version being listed as a preview mean?  Just that it's not a version that went through the full release process for https://www.microsoft.com/en-us/download/details.aspx?id=55114?

udeeshagautam: Yes, they can have additional changes than the last eleased version. The nugets used are from here : https://www.nuget.org/packages/Microsoft.SqlServer.DACFx

PR #988, opened 2020-06-24
2747 Fixed IsUnknownType error

JustinMDotNet:


src/Microsoft.Kusto.ServiceLayer/QueryExecution/Contracts/DbColumnWrapper.cs
```diff
@@ -176,8 +173,7 @@ public DbColumnWrapper()
         /// Logic taken from SSDT determination of unknown columns. It may not even be possible to
         /// have "unknown" column types with the .NET Core SqlClient.
         /// </remarks>
-        public bool IsUnknownType => DataType == typeof(object) &&
-                                     DataTypeName.Equals(UnknownTypeName, StringComparison.OrdinalIgnoreCase);
+        public bool IsUnknownType => DataType == typeof(object) && DataTypeName?.ToLower() == UnknownTypeName;
```
gupta1, line 176: Can you add a comment here explaining the fix or some helpful tip so that we don't end up reverting later  on?

JustinMDotNet:
Closing PR and reopening due to incorrect commit messages.

PR #995, opened 2020-07-21
Add support for getting DacFx deploy options from a publish profile

kisantia:
If the publish profile has any deploy options specified, those will be used. Otherwise, the default deploy options will be returned.


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -212,19 +212,23 @@ public class DeploymentOptions
...
-            // Adding these defaults to ensure behavior similarity with other tools. Dacfx and SSMS import/export wizards use these defaults.
-            // Tracking the full fix : https://github.com/microsoft/azuredatastudio/issues/5599
-            options.AllowDropBlockingAssemblies = true;
-            options.AllowIncompatiblePlatform = true;
-            options.DropObjectsNotInSource = true;
-            options.DropPermissionsNotInSource = true;
-            options.DropRoleMembersNotInSource = true;
-            options.IgnoreKeywordCasing = false;
-            options.IgnoreSemicolonBetweenStatements = false;
+            if (setDefaults)
```
kisantia, line 220: I added this for now for tests because a profile without any options will return the default options without setting these. Since we're changing these defaults for deploy, they should also be the defaults for reading from a profile, but I wasn't sure of a good way to do that. One option is to check the profile file to see if these options are set, otherwise set these default values for those that aren't.

kisantia: removed this and decided to set the default values for these options if they aren't specified in the publish profile.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -12,6 +12,7 @@
 using System.Collections.Concurrent;
 using Microsoft.Data.SqlClient;
 using System.Threading.Tasks;
+using Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts;
```
Benjin, line 15: While you're here, can you sort the usings properly?  There's a VS setting to do it for you automatically (system first, then the rest alphabetically)

Benjin: Same for other files, please!

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -214,6 +216,7 @@ public class DeploymentOptions
 
         public DeploymentOptions()
         {
+
```
Benjin, line 219: intentional?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -241,6 +244,52 @@ public DeploymentOptions()
...
+        /// <summary>
+        /// create deployment options from the options in a publish profile.xml
+        /// </summary>
+        /// <param name="options">options created from the profile</param>
+        /// <param name="profilePath"></param>
+        public DeploymentOptions(DacDeployOptions options, string profilePath)
+        {
+            // check if defaults need to be set if they aren't specified in the profile
+            string contents = File.ReadAllText(profilePath);
```
Benjin, line 259: Move loading from file out of constructor so it can made async

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -241,6 +244,52 @@ public DeploymentOptions()
...
+            if (!contents.Contains("<AllowDropBlockingAssemblies>"))
```
Benjin, line 260: Are these all the different settings that might appear in a deploy profile, or might more get added later?  I'm iffy on a series of very similar-looking checks; would it be better/more scalable to loop over a list of tuples?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -241,6 +244,52 @@ public DeploymentOptions()
...
+            {
+                options.AllowDropBlockingAssemblies = true;
+            }
+            if (!contents.Contains("<AllowIncompatiblePlatform>"))
+            {
+                options.AllowIncompatiblePlatform = true;
+            }
+            if (!contents.Contains("<DropObjectsNotInSource>"))
+            {
+                options.DropObjectsNotInSource = true;
```
Benjin, line 270: Double-checking that this (and the rest of the `true` crew) are actually supposed to be true here, 'cause they feel backward...

> If the profile _doesn't_ say DropObjectsNotInSource, then set DropObjectsNotInSource to true... which then drops them.

Am I reading that right?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -532,6 +537,78 @@ public async void GenerateDeployScriptWithSqlCmdVariables()
...
+        /// Verify that options can get retrieved from publish profile
+        /// </summary>
+        [Fact]
+        public async void GetOptionsFromProfile()
+        {
+            DeploymentOptions expectedResults = new DeploymentOptions();
+            expectedResults.ExcludeObjectTypes = null;
+            expectedResults.IncludeCompositeObjects = true;
+            expectedResults.BlockOnPossibleDataLoss = true;
+            expectedResults.AllowIncompatiblePlatform = true;
```
Benjin, line 550: Block initializer:
```
new DeploymentOptions()
{
    ExcludeObjectTypes = null,
    IncludeCompositeObjects = true,
    ...
}
```

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -532,6 +537,78 @@ public async void GenerateDeployScriptWithSqlCmdVariables()
...
+        {
+            System.Reflection.PropertyInfo[] deploymentOptionsProperties = expected.GetType().GetProperties();
+            foreach (var v in deploymentOptionsProperties)
+            {
+                var defaultP = v.GetValue(expected);
+                var actualP = v.GetValue(actual);
+
+                if (v.Name == "ExcludeObjectTypes")
+                {
+                    Assert.True((defaultP == null && actualP == null) || ((defaultP as ObjectType[]).Length == (actualP as ObjectType[]).Length), "Number of excluded objects is different not equal");
```
Benjin, line 601: Can use null chaining operator to make more readable/concise:

`(defaultP as ObjectType[])?.Length == (actualP as ObjectType[])?.Length`

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -532,6 +537,78 @@ public async void GenerateDeployScriptWithSqlCmdVariables()
...
+                }
+                else
+                {
+                    Assert.True((defaultP == null && actualP == null) || (defaultP == null && (actualP as string) == string.Empty) || defaultP.Equals(actualP), $"Actual Property from Service is not equal to default property for { v.Name}, Actual value: {actualP} and Default value: {defaultP}");
```
Benjin, line 605: extra space before `v.name`

Benjin: Github, when can I leave comments on specific chunks of text rather than full lines?!

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
kisantia: Yup the default for this setting is false, but when deploying we want to dropping the objects not in source to be the default. These are the same values that DacFx and SSMS deploy wizard set as the defaults for these settings.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
kisantia: These are the only settings that need to be checked. DacFx and SSMS deploy wizard change these settings from the defaults that DacDeployOptions gets created with, so those defaults also need to get set here if they aren't specified in the profile so that the deployment behavior is consistent.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
kisantia: nope

PR #998, opened 2020-07-23
Update DacFx deploy and generate script with options

kisantia:
Update DacFx deploy and generate script requests to accept deployment options


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxserviceTests.cs
```diff
@@ -537,6 +540,147 @@ public async void GenerateDeployScriptWithSqlCmdVariables()
...
+                string dacpacPath = InitialExtract(service, sourceDb, result);
+
+                // generate script to deploy the created dacpac with options
+                var generateScriptParams = new GenerateDeployScriptParams
+                {
+                    PackageFilePath = dacpacPath,
+                    DatabaseName = targetDb.DatabaseName,
+                    DeploymentOptions = new DeploymentOptions()
+                    {
+                        DropObjectsNotInSource = false,
```
udeeshagautam, line 647: should we check once with true once with false so we know this really takes effect?

PR #999, opened 2020-07-24
Fix Null Pointer exception in test

Charles-Gagnon:
I don't know why null is something someone would actually pass here but just in case...

The test in question : https://github.com/microsoft/sqltoolsservice/blob/main/test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/CommandOptionsTests.cs#L52

Will also be following up with making sure PRs are running validation builds that run the unit tests since that would have caught this in the PR.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Utility/ServiceLayerCommandOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Utility/ServiceLayerCommandOptions.cs
@@ -27,7 +27,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Utility
         {
             for (int i = 0; i < args.Length; ++i)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Utility/ServiceLayerCommandOptions.cs
```diff
@@ -27,7 +27,7 @@ public ServiceLayerCommandOptions(string[] args) : base(args.Where(arg => !servi
         {
             for (int i = 0; i < args.Length; ++i)
             {
-                string arg = args[i].ToLowerInvariant();
+                string arg = args[i]?.ToLowerInvariant();
```
kburtram, line 30: I assume args is coming from Main() and are there for "guaranteed" not to be null.  But the check should hurt anything.

PR #1003, opened 2020-07-27
Setting up siggen class

aasimkhan30:


arslan9955:
looks good,

src/Microsoft.InsightsGenerator/SigGen.cs
```diff
@@ -8,8 +8,8 @@ public SignatureGenerator(DataArray table)
         {
             this.table = table;
         }
-        
-        public Learn()
+
+        public SignatureGeneratorResult Learn()
```
arslan9955, line 12: qq, we would just call a bunch of methods from this function and add insights to the results right?

aasimkhan30: Yeah, we will create one method for each kind of insight(Avg, mean, top3, etc.) and call them from the learn method. 
These methods will keep adding insights to result.insight Dictionary.

PR #1004, opened 2020-07-28
Insights data transformation logic

kburtram:


src/Microsoft.InsightsGenerator/DataTransformation.cs
```diff
@@ -3,13 +3,132 @@
...
+            }
+
+            bool containsDateTime = columnInfo.ContainsKey(DataArray.DataType.DateTime);
+            string[] labels = new string[columnCount];
+            if (containsDateTime)
+            {
+                List<ColumnInfo> dateColumns = columnInfo[DataArray.DataType.DateTime];
+                for (int i = 0; i < dateColumns.Count; ++i)
+                {
+                    labels[dateColumns[i].ColumnIndex] = "input_t_" + i;
```
aasimkhan30, line 61: Can we just have an enum for this to avoid parsing the column names for the next workflow?

kburtram: we could use an enum I suppose.  but the string was in the spec 😄 let's chat in the Teams channel what we want to do here.

src/Microsoft.InsightsGenerator/DataTransformation.cs
```diff
@@ -3,13 +3,132 @@
...
+                }
+            }
+
+            if (columnInfo.ContainsKey(DataArray.DataType.String))
+            {
+                int startingIndex = 0;
```
kburtram, line 67: this code block should assign the input_g label to the column with the lowest distinctValue count

src/Microsoft.InsightsGenerator/DataTransformation.cs
```diff
@@ -3,13 +3,132 @@
...
+                List<ColumnInfo> stringColumns = columnInfo[DataArray.DataType.String];
+                if (stringColumns.Count > 1)
+                {
+                    labels[stringColumns[startingIndex].ColumnIndex] = "input_g_0";
+                    ++startingIndex;
+                }
+
+                for (int i = startingIndex; i < stringColumns.Count; ++i)
+                {
+                    labels[stringColumns[i].ColumnIndex] = "slicer_" + i;
```
kburtram, line 77: instead of `for (int i = startingIndex; i < stringColumns.Count; ++i)` this should be `for (int i = 0; i < stringColumns.Count - startingIndex; ++i)`

src/Microsoft.InsightsGenerator/DataTransformation.cs
```diff
@@ -3,13 +3,132 @@
...
+
+                long longValue;
+                double doubleValue;
+                if (long.TryParse(firstValueString, out longValue) || double.TryParse(firstValueString, out doubleValue))
+                {
+                    return DataArray.DataType.Number;
+                }
+    
+                DateTime dateValue;
+                if (DateTime.TryParse(firstValueString, out dateValue))
```
kburtram, line 125: this DateTime parsing could be more robust for various datetime formats

src/Microsoft.InsightsGenerator/DataTransformation.cs
kburtram: Here are some of the notes on what the expect output column names would be in various cases:
```
Input
 - Time
 - Groups
Slicer
 - 1...N
Output
 - 1..N

Time
String
Number

-> Input_t, Slicer, Output


Time
String ->input_g
String -> slicer
Number

-> Input_t, Input_g, Slicer, Output

Time -> input_t_0
String -> slicer_0
String -> slicer_1
String -> input_g
String -> slicer_2
Number -> output_0
Number -> output_1
Time -> input_t_1
```

PR #1005, opened 2020-07-28
Added rules engine base implementation

gadudhbh:


src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        public class ColoumnHeaders
+        {
+            public List<string> singleHashValues;
+            public List<string> doubleHashValues;
+            public int templateID;
+        }
+
+
+        public static ColoumnHeaders TemplateParser(string templateFile)
```
jiarima, line 22: Typo: Column

src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+
+            string rules = null;
+            //rulesgenerator
+            return rules;
+        }
+        public static string RulesChecking(string singleHashHeaders, string doublehashHeaders)
+        {
+            string template = null;
+            //if(singleHashHeaders!=null && doublehashHeaders!=null && RulesGeneratorFromInput(singleHashHeaders, doublehashHeaders).Equals(RulesGeneratorFromTemplate()){
```
jiarima, line 72: I'm not 100% sure, but I thought we need to check if all the columns (I believe columns are from doubleHashHeaders) exists in the template, and not the vaues (singleHashHeaders) but need to confirm with Hermineh

src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -0,0 +1,87 @@
...
+                        }
+                    }
+                }
+
+            }
+
+            return ch;
+        }
+
+        public static Dictionary<int, string> RulesGeneratorFromTemplate()
```
gadudhbh, line 54: type of rule, category.

src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            string template = null;
+            //if(singleHashHeaders!=null && doublehashHeaders!=null && RulesGeneratorFromInput(singleHashHeaders, doublehashHeaders).Equals(RulesGeneratorFromTemplate()){
+            //from dictionary mapping select template id
+            //}
+
+            return template;
+        }
+
+        public static void Main(string[] args)
```
kburtram, line 79: this is a class library, so probably shouldn't have a Main method.  I'd recommend using unit tests to verify functionality, such as is done in this PR https://github.com/microsoft/sqltoolsservice/pull/1007.

src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Threading.Tasks;
+
+using System.Diagnostics;
+using System.IO;
+
+namespace Microsoft.InsightsGenerator
+{ 
+    public class RulesEngine
+    {
+        public class ColoumnHeaders
```
kburtram, line 14: Standard .Net coding conventions discourage having public fields, and recommends using properties for this purpose.  Also, public properties should use a capital letter for first char following PascalCase naming convention.  For example:

```C#
public List<string> SingleHashValues { get; set; }
// etc..
```

src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
```
kburtram, line 1: please sort using statements and only include usings that are needed.

kburtram, line 1: please add MIT license header block to this file similar to other files in repo.

src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public static Dictionary<int, string> RulesGeneratorFromTemplate()
+        {
+            Dictionary<int, string> Rules_templateID = new Dictionary<int, string>();
+            string rules = null;
+            ColoumnHeaders header = TemplateParser(@"template_16.txt");
+            return Rules_templateID;
+        }
+
+        public static Dictionary<int, string> RulesGeneratorFromInput(string singleHashHeaders, string doublehashHeaders)
```
gadudhbh, line 62: transformer=double hash values, mapping of actual column name and template double hash values, update template with this. 
siggen=list->single hash values- name, value

src/Microsoft.InsightsGenerator/RulesEngine.cs
gadudhbh: category=1

PR #1006, opened 2020-07-28
Nara/workflow

NaraVen:
Add a queue processor for workflow. The workflow begins when the first input is sent not when the application starts.
The input and output object has to be entered. I will do that a spart fo the next PR but for now I ant thsi reviewed so that all of us are in sync.

### DIFF
--- a/src/Microsoft.InsightsGenerator/Workflow.cs
+++ b/src/Microsoft.InsightsGenerator/Workflow.cs
@@ -1,0 +1,105 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.InsightsGenerator
+{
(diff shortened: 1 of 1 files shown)


src/Microsoft.InsightsGenerator/Workflow.cs
```diff
@@ -0,0 +1,105 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.InsightsGenerator
+{
+    using System;
```
kburtram, line 11: could you please put all the using together at the top of the file and sort them?  that's the style convention we've been using so far in the repo, so it would be good to be consistent.

NaraVen: you mean outside the namespace and sorted right?

NaraVen: done

PR #1009, opened 2020-07-29
Added new Kusto ServiceLayer

gupta1:
- Added a new Kusto ServiceLayer just like we have SqlToolsServiceLayer to enable Kusto experiences in ADS.
- All the Kusto code is in DataSource directory. The rest of the code calls DataSource and that is the abstraction which we will be maturing up in the future to make it easier to add new datasources.
-  There is some code which still contains SQL functionality. Since its isolated code, our plan is to keep doing clean up after merging it to main.


gupta1:
> I have some big concerns about the amount of duplicated code in here - you really should spend some time going through and removing the parts you don't need and consolidating the stuff that you do that's shared. It'll make things a lot easier for future improvements/fixes.

Sure @Charles-Gagnon , we plan to clean up this code in next few iterations.

PR #1011, opened 2020-07-29
Armemon/insights2

arslan9955:
added max and min insights per slice and testing


arslan9955:
@aasimkhan30 I noticed there are two sum methods (one uses double.parse and the other uses long.parse). do we need both?

aasimkhan30:
> @aasimkhan30 I noticed there are two sum methods (one uses double.parse and the other uses long.parse). do we need both?

I will get rid of the long one.

PR #1014, opened 2020-07-29
Aasim/insights/insight methods

aasimkhan30:


src/Microsoft.InsightsGenerator/SigGen.cs
```diff
@@ -102,35 +356,58 @@ private double CalculateColumnMax(object[][] rows, int colIndex)
...
+        }
+
+        public Object[] DeepCloneRow(object[] row)
+        {
+            return row.Select(a => a).ToArray();
+        }
+
+        public double Percentage(double value, double sum)
+        {
+            return Math.Round((double)((value / sum) * 100), 2);
```
arslan9955, line 410: don't need explicit conversion, i thnk, since both are already doubles

PR #1015, opened 2020-07-29
Added all the templates

jiarima:
Added a template processor
Added a method to find matched template

### DIFF
--- a/src/Microsoft.InsightsGenerator/RulesEngine.cs
+++ b/src/Microsoft.InsightsGenerator/RulesEngine.cs
@@ -134,6 +134,10 @@ namespace Microsoft.InsightsGenerator
             foreach (var list in singleHashHeaders)
             {
                 topHeaderList.Add("#" + list.First());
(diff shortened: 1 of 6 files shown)


src/Microsoft.InsightsGenerator/RulesEngine.cs
```diff
@@ -3,81 +3,200 @@
...
 using System.Text.RegularExpressions;
 
 namespace Microsoft.InsightsGenerator
 {
     public class RulesEngine
     {
-        public class ColumnHeaders
+        public static List<Template> Templates;
+
+        public static List<string> TopListHashHeaders = new List<string>{ "#top", "#average", "#averageSlice", "#topPerslice" , "#bottom"};
```
gadudhbh, line 20: not used?

jiarima: It's used to check if the text replacement needs extra step!

PR #1017, opened 2020-07-29
-Added logic for Insights Generator Service Handler

aasimkhan30:


src/Microsoft.SqlTools.ServiceLayer/InsightsGenerator/InsightsGeneratorService.cs
```diff
@@ -45,6 +46,18 @@ public void InitializeService(ServiceHost serviceHost)
...
+
+            QueryExecution.Contracts.SubsetParams subsetParams = new QueryExecution.Contracts.SubsetParams{
+                OwnerUri = parameters.OwnerUri
+            }; 
+
+            QueryExecution.Contracts.ResultSetSubset result = await service.InterServiceResultSubset(subsetParams);
+            
+            Microsoft.InsightsGenerator.Workflow insightWorkFlowInstance = Microsoft.InsightsGenerator.Workflow.Instance();
+
+            insightWorkFlowInstance.IngestRules(result.Rows);
```
NaraVen, line 59: do we need a result.Rows.to_array() or does it come in the DataArray format already. Appologize haven't used ti recently so I don't remember how it looks.

aasimkhan30: I can probably send a DataArray to the IngestRules

NaraVen: This PR uses dataarray : https://github.com/microsoft/sqltoolsservice/pull/1019

kburtram, line 59: note sure if we need to copy the array before sending to avoid it being modified while we use it.  i.e, if it's locked.  But this is fine for PoC.  Also, we may need to do some reshaping to get the data into a structure that we can process in insights engine and also work with graph library.

aasimkhan30:
@NaraVen Please pull the workflow tests  from this branch.

PR #1022, opened 2020-07-29
Add service for converting a notebook to sql and back

Charles-Gagnon:
Logic to parse the content and return the Notebook JSON/SQL queries will be added later

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/Contracts/NotebookConvertContracts.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/Contracts/NotebookConvertContracts.cs
@@ -1,0 +1,62 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlTools.Utility;
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/Contracts/NotebookConvertContracts.cs
```diff
@@ -0,0 +1,62 @@
...
+            RequestType<ConvertNotebookToSqlParams, ConvertNotebookToSqlResult> Type =
+            RequestType<ConvertNotebookToSqlParams, ConvertNotebookToSqlResult>.Create("notebookconvert/convertnotebooktosql");
+    }
+
+    public class ConvertSqlToNotebookParams : GeneralRequestDetails
+    {
+        /// <summary>
+        /// The ClientUri of the SQL Query file we're converting
+        /// </summary>
+        public string ClientUri { get; set; }
```
kburtram, line 41: I think we typically call it ownerUri.

Charles-Gagnon: For some reason I decided to call it ClientUri in STS..although I can't remember why exactly. Mostly cause it came from the client I guess? The use here is even more incorrect anyways - the method I'm calling (GetFile) has the parameter as filePath but it takes in URI's too... 

https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Utility/ResolvedFile.cs#L25

Anyways I agree we should settle on something here - but in this case I wanted to stick with ClientUri for consistencies sake until we could look at how these are used across the app in its entirety.

PR #1023, opened 2020-07-29
Nara/remove queing

NaraVen:
removed queue
added data transform
added test cases with sample data

### DIFF
--- a/src/Microsoft.InsightsGenerator/Workflow.cs
+++ b/src/Microsoft.InsightsGenerator/Workflow.cs
@@ -14,7 +14,7 @@ namespace Microsoft.InsightsGenerator
     {
         private readonly ConcurrentQueue<object> _siggenQueue = new ConcurrentQueue<object>();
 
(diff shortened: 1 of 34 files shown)


src/Microsoft.InsightsGenerator/Workflow.cs
```diff
@@ -12,105 +12,47 @@ namespace Microsoft.InsightsGenerator
...
-                //Get the signature result
-                SignatureGenerator siggen = new SignatureGenerator(rulesData);
-                SignatureGeneratorResult result = siggen.Learn();
-                // call the rules engine processor
-                if (result?.Insights == null)
+                try
                 {
-                    Console.WriteLine("Failure in generating insights, Input not recognized!");
+                    DataTransformer transformer = new DataTransformer();
+                    rulesData = transformer.Transform(rulesData);
```
kburtram, line 33: technically you don't need to assign the return code since the input is modified by reference, the return value is mostly to support fluent API calling style

NaraVen: done

PR #1032, opened 2020-07-30
Bug fixes for hack insights

aasimkhan30:


src/Microsoft.InsightsGenerator/DataTransformation.cs
```diff
@@ -78,29 +78,29 @@ private string[] GetColumnLabels(DataArray array, out DataArray.DataType[] colum
             {
                 if (columnInfo.ContainsKey(DataArray.DataType.String))
                 {
-                    int minDistinctValue = -1;
-                    int minColumnIndex = -1;
-
+                    int maxDistinctValue = Int16.MaxValue;
```
kburtram, line 81: should this be Int32.MaxValue?  also, a common pattern for max is to set to the minimum possible value and then not have the extra check in the loop assuming all values will be greater.

src/Microsoft.InsightsGenerator/DataTransformation.cs
```diff
@@ -78,29 +78,29 @@ private string[] GetColumnLabels(DataArray array, out DataArray.DataType[] colum
...
-                            minColumnIndex = i;
+                            maxDistinctValue = stringColumns[i].DistinctValues;
+                            maxColumnIndex = i;
+                            maxColumnLabelIndex = stringColumns[i].ColumnIndex;
                         }
                     }
                     
-                    labels[stringColumns[minColumnIndex].ColumnIndex] = "input_g_0";
-
+                    labels[maxColumnLabelIndex] = "input_g_0";
```
kburtram, line 95: I thought the expected logic is that the string column with the least amount of distinct values should be the grouping column.  Why are we making this change?

src/Microsoft.InsightsGenerator/DataTransformation.cs
aasimkhan30: Yeah, I worked with Nara yesterday to change the logic of the slicer column select. The thought process behind this was that the categories(slicers) should have less distinct values than the input

kburtram: that makes sense.  thanks!

PR #1035, opened 2020-07-31
Add an ADO yaml file for integration tests

shueybubbles:
To run our integration tests in ADO, I've made a few changes:
1. Added a json file to the secure file library in our ADO subscription containing a sql connection to use
2. Updated the debug build to produce full symbols to enable vstest to measure code coverage
3. Added a runsettings file to turn on code coverage for product DLLs.

image

### DIFF
--- a/azure-pipelines/integrationtests.yml
+++ b/azure-pipelines/integrationtests.yml
@@ -4,7 +4,7 @@ pool:
   - task: DownloadSecureFile@1
     displayName: 'Download secure file'
     inputs:
-      secureFile: 'aec0999c-ce6e-49b4-a386-edd5729f1b0f'
+      secureFile: 'testsettings.json'
   
   - task: VisualStudioTestPlatformInstaller@1
     displayName: 'Visual Studio Test Platform Installer'
--- a/azure-pipelines/integrationtests.yml
+++ b/azure-pipelines/integrationtests.yml
@@ -1,0 +1,52 @@
+pool:
+    name: SqlToolsTestAgentPool
+  steps:
+  - task: DownloadSecureFile@1
+    displayName: 'Download secure file'
+    inputs:
+      secureFile: 'aec0999c-ce6e-49b4-a386-edd5729f1b0f'
+  
+  - task: VisualStudioTestPlatformInstaller@1
+    displayName: 'Visual Studio Test Platform Installer'
+    inputs:
+      versionSelector: latestStable
+  
+  - task: UseDotNet@2
+    displayName: 'Use .NET Core sdk '
+    inputs:
+      useGlobalJson: true
+  
+  - task: DotNetCoreCLI@2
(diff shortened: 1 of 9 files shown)


udeeshagautam:
Thanks @shueybubbles for doing this!

azure-pipelines/testrun.runsettings
```diff
@@ -6,11 +6,19 @@
           <CodeCoverage>
             <ModulePaths>
               <Include>
-                <ModulePath>Microsoft\.SqlTools.*\.dll$</ModulePath>
-                <ModulePath>Microsoft\.SqlTools.*\.exe$</ModulePath>
+                <ModulePath>.*SqlTools.*</ModulePath>
```
udeeshagautam, line 9: not sure I understand this change. We dont want to include dlls/exes from child directories?

azure-pipelines/integrationtests.yml
```diff
@@ -0,0 +1,52 @@
...
+      TargetFolder: '$(Build.SourcesDirectory)\test\Microsoft.SqlTools.ServiceLayer.IntegrationTests\bin\Debug\netcoreapp3.1'
+  
+  - task: DotNetCoreCLI@2
+    displayName: 'Run integration tests'
+    inputs:
+      command: test
+      projects: '**/Microsoft.SqlTools.ServiceLayer.IntegrationTests.csproj'
+      arguments: '--no-build'
+      testRunTitle: 'SqlToolsService Integration Tests'
+    enabled: false
```
Charles-Gagnon, line 38: Not enabled? Given the below task is this necessary?

azure-pipelines/integrationtests.yml
```diff
@@ -0,0 +1,52 @@
...
+  
+  - task: VSTest@2
+    displayName: 'Run integration tests with code coverage'
+    inputs:
+      testAssemblyVer2: '**/bin/Debug/netcoreapp3.1/Microsoft.SqlTools.ServiceLayer.IntegrationTests.dll'
```
Charles-Gagnon, line 43: `**/bin/Debug/**/Microsoft.SqlTools.ServiceLayer.IntegrationTests.dll`? One less thing to update if we update the netcore version

azure-pipelines/testrun.runsettings
shueybubbles: It seems the covrun tools use the full module path, not just the name, so each entry has to start with ".*". The "dll" and "exe" suffixes are not really needed since the tool only looks at those kinds of modules anyway.
The heuristic "any dll whose name contains SqlTools" seems like a reasonable starting point and if we find noise in the data we can tweak the filter later.

azure-pipelines/integrationtests.yml
shueybubbles: i was experimenting with both

Issue #667, opened 2018-07-28
Please add messages to asserts and add tracing to product code


shueybubbles:
I've had some tests fail in the appveyor run as part of a pull request, but there was zero information in the log to identify the issue.
Eg, this assert on not null provides no information on what is null:
  ```
Location[] locations = scripter.GetSqlObjectDefinition(objectName, schemaName, objectType);
            if (shouldReturnValidResult)
            {
                Assert.NotNull(locations);
                Cleanup(locations);
            }
            else
            {
                Assert.Null(locations);
            }
```
Also, GetSqlObjectDefinition has zero tracing. Since the failure doesn't reproduce readily on a dev machine figuring out how to make these tests reliable in the cloud will be difficult without appropriate tracing.

PR #1037, opened 2020-08-02
Convert most tools service tests to nunit

shueybubbles:
To help resolve issue #667 - I have not been happy with the state of our tests, especially as I plan to start working more in this project myself.
I am advocating for this move to nUnit because it has a much richer constraint model for asserts. Benefits include:
1. Much more information being written to the log on an assertion failure. It will tell you where a string comparison failed and show spans of text on either side of the fail point.
2. There's much less chance of swapping expected and actual by mistake when using Assert.That
3. the Does/Is/Has constraint classes lead to much more composable and readable assertion construction.
4. All the assert varieties in nunit provide a "message" parameter. 

It's easy to tell how readily we've been misusing xunit just based on the dozens of build warnings that xunit's static analysis plugin reports during a build. If it's harder to write a test correctly than incorrectly, we should examine our practices.

I believe the test log should provide as much information about the failure as possible to help reduce time to fix, and nunit's philosophy much more closely aligns to that than does xunit.

I verified with these changes that all the tests pass.

For this change I followed several steps. First I flipped "expected" and "actual" where needed for all the Assert.Equal calls that had it wrong. Then I converted each project in its own commit. When I had to manually edit some asserts beyond a find/replace I tried to convert some of the comments into assert messages.
I did not try to convert every single assert into the nunit constraint model, as it wouldn't provide much value.
Also, I removed the custom code in  the TestDriver project that was loading test functions based on the xunit FactAttribute and running them. @llali please consider whether your performance test suite can simply invoke mstest instead of using our customer test driver.


test/Microsoft.SqlTools.Hosting.UnitTests/ProtocolTests/MessageReaderTests.cs
```diff
@@ -10,51 +10,52 @@
...
         
         #region ReadMessage Tests
 
-        [Theory]
-        [InlineData(512)]    // Buffer size can fit everything in one read
-        [InlineData(10)]     // Buffer size must use multiple reads to read the headers
-        [InlineData(25)]     // Buffer size must use multiple reads to read the contents
-        public async Task ReadMessageSingleRead(int bufferSize)
+        [Test]
+        //[InlineData(512)]    // Buffer size can fit everything in one read
```
Charles-Gagnon, line 55: remove

test/Microsoft.SqlTools.ServiceLayer.TestDriver.Tests/StressTests.cs
```diff
@@ -167,12 +169,12 @@ public async Task TestQueryExecutionService()
...
-                    Assert.NotNull(queryResult.BatchSummaries);
-                    Assert.NotEmpty(queryResult.BatchSummaries);
-                    Assert.NotNull(queryResult.BatchSummaries[0].ResultSetSummaries);
-                    Assert.NotNull(queryResult.BatchSummaries[1].ResultSetSummaries);
-                    Assert.NotNull(queryResult.BatchSummaries[2].ResultSetSummaries);
-                    Assert.NotNull(queryResult.BatchSummaries[3].ResultSetSummaries);
+                    Assert.That(queryResult.BatchSummaries, Is.Not.Null, "queryResult.BatchSummaries");
+                    Assert.That(queryResult.BatchSummaries.Select(b => b.ResultSetSummaries), Has.Exactly(4).Not.Null, "ResultSetSummaries in the queryResult");
+                    //Assert.NotNull(queryResult.BatchSummaries[0].ResultSetSummaries);
+                    //Assert.NotNull(queryResult.BatchSummaries[1].ResultSetSummaries);
```
llali, line 175: should we remove these if replaced by those two lines above?

llali:
@shueybubbles I cannot find the status of current perf tests in the lab  so maybe we haven't run the tests for a while. If we're not planning to run them, maybe we don't need to worry about your change breaking perf tests. @kburtram or  @chlafreniere might know.  If we're still running them, we just need to change the executable and parameters in a config file.

PR #1039, opened 2020-08-06
Changes required for Kusto manage dashboard

gupta1:
- Added AdminService, Metadataservice to the Kusto.ServiceLayer. Required for 2 method: fetching dashboard "database" properties and MetadataService fetches database child metadata. Both are used to fetch details for Database manage view.

- Added changes to include database details when calling listdatabases method. These details for Kusto will include database name and database size.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Data;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Kusto.ServiceLayer.Utility;
 using Microsoft.Kusto.ServiceLayer.Admin.Contracts;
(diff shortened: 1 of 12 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
```diff
@@ -419,5 +446,69 @@ public static DataSourceObjectMetadata CreateFolderMetadata(DataSourceObjectMeta
...
+        {
+            switch (dataSourceType)
+            {
+                case DataSourceType.Kusto:
+                    {
+                        var databaseDetails = new List<DatabaseInfo>();
+
+                        foreach(var dbDetail in clusterDBDetails)
+                        {
+                            DatabaseInfo databaseInfo = new DatabaseInfo();
```
qifahs, line 460: It would be good to validate that the ObjectMetadata is of type Db.

src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
```diff
@@ -419,5 +446,69 @@ public static DataSourceObjectMetadata CreateFolderMetadata(DataSourceObjectMeta
                     throw new ArgumentException($"Unsupported data source type \"{dataSourceType}\"", nameof(dataSourceType));
             }
         }
+
+        public static List<DatabaseInfo> ConvertToDatabaseInfo(DataSourceType dataSourceType, List<DataSourceObjectMetadata> clusterDBDetails)
+        {
+            switch (dataSourceType)
```
qifahs, line 452: Do we even need this param. I don't see anything specific to Kusto being done here. Just a conversion from DataSourceObjectMetadata to DatabaseInfo.

src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
```diff
@@ -419,5 +446,69 @@ public static DataSourceObjectMetadata CreateFolderMetadata(DataSourceObjectMeta
...
+                    }
+
+                default:
+                    throw new ArgumentException($"Unsupported data source type \"{dataSourceType}\"", nameof(dataSourceType));
+            }
+        }
+
+        public static List<ObjectMetadata> ConvertToObjectMetadata(DataSourceType dataSourceType, List<DataSourceObjectMetadata> dbChildDetails)
+        {
+            switch (dataSourceType)
```
qifahs, line 476: Same thing here. Not sure if we need dataSourceType.

PR #1040, opened 2020-08-07
GetConnectionString API fix to not change the cache only return the value

udeeshagautam:
Get connection string call was changing the connection info (ref of the object) received from cache. Changing it to just get and make changes to only the returned string..

In ref to bug : https://github.com/microsoft/azuredatastudio/issues/11697


udeeshagautam:
@Charles-Gagnon @Karl please let me know if you already have the changes for RowDeleteTests - I can revert those. Just added to ensure PR validation goes through.

Charles-Gagnon:
@smartguest FYI on the tests

Thanks @udeeshagautam !

PR #1041, opened 2020-08-10
Fix build break in Edit Data tests

kburtram:


kburtram:
Looks like @udeeshagautam beat me to this fix.  Thanks!

PR #1042, opened 2020-08-10
Switch pdb type back to portable

kburtram:
The full pdb's break the VS Code "Debug test" functionality since the Windows pdbs aren't compatible with that debugger.  What we've done for code coverage runs previously is we switch the pdb in the code coverage script while collecting those metrics.  See https://github.com/microsoft/sqltoolsservice/blob/main/test/CodeCoverage/codecoverage.bat#L25 for example.


kburtram:
Merging to unblock dev workflows using vscode.

shueybubbles:
I haven't looked at the PR yet, but can we just make the pdb type parameterizable with an msbuild /p variable? The ado build using vstest would pass the variable to use full.

Sent from my T-Mobile 4G LTE Device
Get Outlook for Android<https://aka.ms/ghei36>

PR #1043, opened 2020-08-10
full symbols for visual studio or code coverage

shueybubbles:
this tweaks the reversion to portable PDBs to use full pdbs when building debug inside visual studio or when setting /p:CodeCoverageBuild=true.
Modifies the yml file for integrationtests ADO to use the variable.
I had tried setting debugtype in the props file instead of the targets file, but Roslyn has some other props file which sets the pdb type to portable after directory.build.props processed.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,8 +1,8 @@
 <Project>
   <Sdk Name="Microsoft.Build.CentralPackageVersions" Version="2.0.52" />
   <!-- Need full debugtype for code coverage -->
-	<PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-		<DebugType>portable</DebugType>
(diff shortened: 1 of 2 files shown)


Directory.Build.targets
```diff
@@ -1,8 +1,8 @@
 <Project>
   <Sdk Name="Microsoft.Build.CentralPackageVersions" Version="2.0.52" />
   <!-- Need full debugtype for code coverage -->
-	<PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-		<DebugType>portable</DebugType>
+	<PropertyGroup Condition="'$(Configuration)' == 'Debug' and ('$(BuildingInsideVisualStudio)' != '' or '$(CodeCoverageBuild)' == 'true')">
```
Charles-Gagnon, line 4: Don't you still need to define the case for it being portable? Or does it default to that already if not specified?

shueybubbles: portable is the default. It's assigned in directory.build.props, too, just to be sure.

PR #1049, opened 2020-08-14
Multiple Tables support in case queries return multiple tables

rajeshka:
- Multiple Tables are now returned if query results contain them


src/Microsoft.Kusto.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -369,6 +367,16 @@ private async Task ExecuteOnce(ReliableDataSourceConnection conn, CancellationTo
 
                 } while (reader.NextResult());
 
+                if (resultSets.Count > 3)
```
qifahs, line 370: Please add a comment as to why you are removing the last 3 results.

src/Microsoft.Kusto.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -369,6 +367,16 @@ private async Task ExecuteOnce(ReliableDataSourceConnection conn, CancellationTo
...
+                {
+                    resultSets.RemoveRange(resultSets.Count - 3, 3);
+                }
```
qifahs, line 373: QueryExecution is not the right place to do this. For the Kusto.ServiceLayer we are localizing all Kusto specific changes to the DataSource folder. Kusto specific logic should not be sprinkled across the rest of the code which we've kept generic.

A great place to add this is in the KustoResultsReader.cs.

rajeshka: Done

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoResultsReader.cs
```diff
@@ -10,13 +12,15 @@ public KustoResultsReader(IDataReader reader)
...
-        /// from the IDataReader.
+        /// Kusto returns atleast 4 results tables - QueryResults(sometimes more than one), QueryProperties, QueryStatus and Query Results Metadata Table. 
+        /// When returning query results we need to trim off the last 3 tables as we want the caller to only read results table. 
         /// </summary>
-        /*public override bool NextResult()
+        public void SanitizeResults(List<ResultSet> resultSets)
         {
-            return false;
-        }*/
+            if (resultSets.Count > 3)
```
qifahs, line 20: Can't this entire logic be contained in this file. The result set can be trimmed in the constructor or in NextResults. Without any changes needed in QueryExecution. This whole class was created so that we could override the results and return as needed, without changing the rest of the code.

rajeshka: We need changes to QueryExecution as the logic there send the tables to ADS as it reads. We want to read all the results and trim them out before sending the results to ADS.

Charles-Gagnon:
I'm going to close this to clean up stale PRs - please reopen once you're clean up the merge conflicts if you plan on finishing this up at some point!

PR #1050, opened 2020-08-14
added task to clear diagnostics when file closed

lucyzhang929:
This PR fixes https://github.com/microsoft/azuredatastudio/issues/11689.
I added a new Task that clears language diagnostics when a file is closed.


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -701,6 +704,36 @@ public async Task HandleDidChangeTextDocumentNotification(ScriptFile[] changedFi
...
+            try
+            {
+                // if not in the preview window and diagnostics are enabled then clear diagnostics
+                if (!IsPreviewWindow(scriptFile)
+                    && CurrentWorkspaceSettings.IsDiagnosticsEnabled)
+                {
+                    await DiagnosticsHelper.ClearScriptDiagnostics(scriptFile.ClientUri, eventContext);
+                }
+
+                await Task.FromResult(true);
```
Charles-Gagnon, line 728: Since you're not returning anything why is this here?

PR #1053, opened 2020-08-20
3278 Kusto Unit Test Refactor

JustinMDotNet:
Refactored classes to use dependency injection through constructors and code to interface instead of implementations.


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoResultsReader.cs
```diff
@@ -4,19 +4,8 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
 {
     internal class KustoResultsReader : DataReaderWrapper
     {
-        public KustoResultsReader(IDataReader reader)
-            : base(reader)
+        public KustoResultsReader(IDataReader reader) : base(reader)
         {
         }
-
-        /// <summary>
```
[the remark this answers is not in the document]
JustinMDotNet: Reverted it back.

src/Microsoft.Kusto.ServiceLayer/Scripting/Scripter.cs
```diff
@@ -3,73 +3,34 @@
...
+using Microsoft.SqlServer.Management.Sdk.Sfc;
+using System.Text;
 
 namespace Microsoft.Kusto.ServiceLayer.Scripting
 {
-	internal partial class Scripter
+    [Export(typeof(IScripter))]
+    public class Scripter : IScripter
     {
-		private void Initialize()
```
JustinMDotNet, line 14: Removed unused function

src/Microsoft.Kusto.ServiceLayer/Admin/AdminService.cs
```diff
@@ -64,15 +56,15 @@ public static AdminService Instance
         /// <summary>
         /// Initializes the service instance
         /// </summary>
-        public void InitializeService(ServiceHost serviceHost)
+        public void InitializeService(IProtocolEndpoint serviceHost)
```
qifahs, line 59: Replaced by a completely different interface. Is this expected?

JustinMDotNet: The ServiceHost implements the IProtocolEndpoint. I just changed the param to it's interface instead of the ServiceHost. I no longer need the change so I'll revert it.

PR #1054, opened 2020-08-21
Fix SerializeDataStartRequestParams initialization

tamasvajk:


kburtram:
It looks like we have this same bug in the SQL project as well.  I'll send a PR for that duplicate issue.  Thanks!

PR #1055, opened 2020-08-24
3491 Kusto Execute Function support

JustinMDotNet:
Added GenerateExecuteFunctionScript to IDataSource, DataSourceBase, and KustoDataSource. Added ExecutionFunction to Scripter and IScripter.


src/Microsoft.Kusto.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -166,6 +170,19 @@ private string GenerateScriptAlter(IDataSource dataSource, UrnCollection urns)
             return string.Empty;
         }
 
+        private string GenerateScriptExecute(IDataSource dataSource, UrnCollection urns)
```
gupta1, line 173: We can refactor this one to use the same function.

PR #1057, opened 2020-08-26
3278 Kusto Unit Tests - Part 1

JustinMDotNet:

PR #1061, opened 2020-08-28
Adding Kusto Build Definition

joberume:
Adding the steps to compile Kusto project + Test Project + Copying binaries to SQL Tools Service to package the same library

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -45,15 +45,33 @@ steps:
     projects: '$(Build.SourcesDirectory)/src/Microsoft.SqlTools.ServiceLayer'
     arguments: '--configfile $(Build.SourcesDirectory)/nuget.config'
 
+- task: DotNetCoreCLI@1
+  displayName: 'dotnet restore src/Microsoft.Kusto.ServiceLayer'
+  inputs:
+    command: restore
+    projects: '$(Build.SourcesDirectory)/src/Microsoft.Kusto.ServiceLayer'
+    arguments: '--configfile $(Build.SourcesDirectory)/nuget.config'
+
 - task: DotNetCoreCLI@2
   displayName: 'dotnet build src/Microsoft.SqlTools.ServiceLayer'
   inputs:
     projects: '$(Build.SourcesDirectory)/src/Microsoft.SqlTools.ServiceLayer'
 
 - task: DotNetCoreCLI@2
+  displayName: 'dotnet build src/Microsoft.Kusto.ServiceLayer'
+  inputs:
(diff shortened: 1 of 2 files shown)


azure-pipelines/build.yml
```diff
@@ -69,13 +87,35 @@ steps:
...
 
+- task: DotNetCoreCLI@1
+  displayName: 'dotnet test test/Microsoft.Kusto.ServiceLayer.UnitTests'
+  inputs:
+    command: test
+    projects: test/Microsoft.Kusto.ServiceLayer.UnitTests
+    arguments: '--logger "trx;LogFileName=xunit.trx"'
+
+- task: CopyFiles@2
+  displayName: 'Copy Kusto Service Layer binaries'
```
Charles-Gagnon, line 111: You should just add Kusto as a project reference to STS if you're going to do this I would say.

But first - how big are these binaries? We want to avoid adding in a bunch of extra stuff that most people won't be using if we can avoid it.

src/Microsoft.SqlTools.ServiceLayer/SqlAssessment/Contracts/AssessmentRequest.cs
```diff
@@ -1,129 +1,129 @@
...
-        /// </summary>
-        public bool Success { get; set; }
-
-        /// <summary>
-        /// Gets or sets an status message for the operation.
-        /// </summary>
-        public string ErrorMessage { get; set; }
-    }
-}
+//
```
Charles-Gagnon, line 1: Please undo all these other changes so we can easily see what's actually changed

azure-pipelines/build.yml
joberume: Only 3 net files get copied over. The projects are almost identical as far as dependencies except for Kusto dlls, so the copying just skips over most of them.

Charles-Gagnon: Yes, but by doing this you make it so that if anyone tries to build/deploy from anything else without also doing this step they're not going to get the right results.

Ideally one should be able to easily build from a local enlistment and get the same result as in the lab. Manually piecing together deployments like this is likely to cause confusion and issues in the future.

Charles-Gagnon: And you should still run this change by all the major stakeholders (and provide exact DLL sizes) before doing a change like this - we should avoid unnecessary bloat when possible.

joberume: Got it, makes sense. Will change it up. I switched this back to draft, I will roll that suggestion at the following PR #1057

Charles-Gagnon:
@joberume I don't think this one is necessary anymore right? I'm going to close it now, feel free to reopen if you think it's still needed 😄

PR #1066, opened 2020-09-02
Insights generator project

kburtram:
This adds the initial iteration of the InsightsGenerator enhanced accessibility for visualizations assembly.


kburtram:
@aasimkhan30 let's go through the process of getting the InsightGenerator project merged to main branch in here and ads repos.  I think the bulk of the remaining work for initial check-in will be in how we surface these insight strings in the UI.

src/Microsoft.SqlTools.ServiceLayer/Localization/transXliff/sr.de.xlf
```diff
@@ -2321,13 +2321,13 @@
       </trans-unit>
       <trans-unit id="ExportBacpacTaskName">
         <source>Export bacpac</source>
-        <target state="translated">BACPAC exportieren</target>
+        <target state="translated">BACPAC-Datei exportieren</target>
```
kburtram, line 2324: not sure why these random loc string are in this PR.  I'll remove these.

aasimkhan30:
@kburtram everything looks good. However, the way the rule engine currently works needs to be changed to generate better 'insights'. I have had a chat with Hermineh on the changes that needs to be made. Will work on it when I get some free cycles.

aasimkhan30 (approved):

PR #1067, opened 2020-09-02
3647 Kusto Refresh Logic

JustinMDotNet:


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
```diff
@@ -347,38 +347,43 @@ private IEnumerable<DataSourceObjectMetadata> GetDatabaseMetadata(bool includeSi
         {
             if (_databaseMetadata == null)
             {
-                CancellationTokenSource source = new CancellationTokenSource();
```
JustinMDotNet, line 350: Moved all this logic into `SetDatabaseMetadata` function

src/Microsoft.Kusto.ServiceLayer/DataSource/IDataSource.cs
```diff
@@ -70,11 +70,6 @@ public interface IDataSource : IDisposable
         /// <returns>Metadata for all children.</returns>
         IEnumerable<DataSourceObjectMetadata> GetChildObjects(DataSourceObjectMetadata parentMetadata, bool includeSizeDetails = false);
 
-        /// <summary>
-        /// Refresh object list for entire cluster.
-        /// </summary>
-        void Refresh();
```
JustinMDotNet, line 76: Removed function that is not called outside KustoDataSource

src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
```diff
@@ -77,9 +77,6 @@ public async Task<T> ExecuteScalarQueryAsync<T>(string query, CancellationToken
         /// <inheritdoc/>
         public abstract IEnumerable<DataSourceObjectMetadata> GetChildObjects(DataSourceObjectMetadata parentMetadata, bool includeSizeDetails = false);
 
-        /// <inheritdoc/>
-        public abstract void Refresh();
```
JustinMDotNet, line 81: Removed function from interface and abstract since it's not called outside KustoDataSource

src/Microsoft.Kusto.ServiceLayer/DataSource/Metadata/FolderMetadata.cs
```diff
@@ -5,6 +5,6 @@
     /// </summary>
     public class FolderMetadata : DataSourceObjectMetadata
     {
-        public DataSourceObjectMetadata ParentMetadata { get; set; }
```
JustinMDotNet, line 8: Renamed since this contains the database metadata and not the immediate parent.

src/Microsoft.Kusto.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
```diff
@@ -43,27 +43,17 @@ public class TreeNode : IComparable<TreeNode>
...
             DataSource = dataSource;
             ObjectMetadata = objectMetadata;
             NodeValue = objectMetadata.Name;
         }
 
-        /// <summary>
-        /// Constructor that accepts a label to identify the node
-        /// </summary>
-        /// <param name="value">Label identifying the node</param>
-        public TreeNode(string value, IDataSource dataSource, DataSourceObjectMetadata objectMetadata)
```
JustinMDotNet, line 59: Removed unused constructor

src/Microsoft.Kusto.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
```diff
@@ -133,7 +123,7 @@ public object BuildingMetadataLock
         /// for many nodes such as the server, the display label will be different
         /// to the value.
         /// </summary>
-        public string Label {
```
JustinMDotNet, line 136: Changed visibility to align with usage

src/Microsoft.Kusto.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
```diff
@@ -290,17 +280,6 @@ public IList<TreeNode> GetChildren()
             return new ReadOnlyCollection<TreeNode>(children);
         }
 
-        /// <summary>
-        /// Adds a child to the list of children under this node
-        /// </summary>
-        /// <param name="newChild"><see cref="TreeNode"/></param>
-        public void AddChild(TreeNode newChild)
```
JustinMDotNet, line 297: Removed unused function

src/Microsoft.Kusto.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
```diff
@@ -368,33 +347,30 @@ protected virtual void PopulateChildren(bool refresh, string name, CancellationT
...
 
-        protected IEnumerable<TreeNode> ExpandChildren(TreeNode parent, bool refresh, string name, bool includeSystemObjects, CancellationToken cancellationToken)
+        protected IEnumerable<TreeNode> ExpandChildren(TreeNode parent, bool refresh, string name,
+            bool includeSystemObjects, CancellationToken cancellationToken)
         {
-            List<TreeNode> allChildren = new List<TreeNode>();
-
             try
             {
-                OnExpandPopulateNonFolders(allChildren, parent, refresh, name, cancellationToken);
```
JustinMDotNet, line 377: Changed to return value rather than ref parameter.

src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
gupta1: We may need it here because its a method any other data store we later add will need to implement as well. Something to double check.

gupta1 (approved):

src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
qifahs: Agree. That looks like an override for any other DataSource we implement.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
qifahs: SetDatabaseMetadata is doing the same thing. Is it being extracted for testing purposes? If you need to extract it out, the name is a little misleading too. Maybe call it InternalGetMetadata. We are not setting any database info on Kusto.

src/Microsoft.Kusto.ServiceLayer/DataSource/Metadata/FolderMetadata.cs
qifahs: What about the nested folder case? I think we should leave this generic for adding other datastores. It might be database for Kusto. But it could be completely something else for other stores. Better to leave it as "ParentMetadata" and let the store implementation decide what the parent refers to. The Parent here is meant to be the parent in the hierarchy for the object explorer.

test/Microsoft.Kusto.ServiceLayer.UnitTests/ObjectExplorer/SmoModel/ServerNodeTests.cs
```diff
@@ -0,0 +1,160 @@
...
+using Microsoft.Kusto.ServiceLayer.DataSource;
+using Microsoft.Kusto.ServiceLayer.DataSource.Metadata;
+using Microsoft.Kusto.ServiceLayer.ObjectExplorer.DataSourceModel;
+using Microsoft.Kusto.ServiceLayer.ObjectExplorer.Nodes;
+using Microsoft.Kusto.ServiceLayer.Utility;
+using Microsoft.SqlTools.Extensibility;
+using Moq;
+using NUnit.Framework;
+
+namespace Microsoft.Kusto.ServiceLayer.UnitTests.ObjectExplorer.SmoModel
```
qifahs, line 15: It's interesting that we still have a SmoModel. This is very Sql specific. If we need this interface, probably needs to be renamed to DataSourceModel?

qifahs (changes requested):

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
JustinMDotNet: I extracted it because I'm using the 'Set' functionality in the Refresh method. I have the Refresh method simply resetting the data but not returning it.

qifahs (approved):

PR #1069, opened 2020-09-08
3702 Fixed Kusto Cancel Query Button

JustinMDotNet:
Changed CancelQuery to use ExecuteControlCommand instead of ExecuteQuery in KustoDataSource. Fixed runtime error in NodePathGenerator for loading TreeNodeDefinition.xml


src/Microsoft.Kusto.ServiceLayer/ObjectExplorer/DataSourceModel/NodePathGenerator.cs
```diff
@@ -24,7 +24,7 @@ internal static void Initialize()
             }
 
             var assembly = typeof(ObjectExplorerService).Assembly;
-            var resource = assembly.GetManifestResourceStream("Microsoft.Kusto.ServiceLayer.ObjectExplorer.SmoModel.TreeNodeDefinition.xml");
```
JustinMDotNet, line 27: This fixes a run time error caused by the folder being renamed from SmoModel to DataSourceModel

qifahs (approved):

PR #1070, opened 2020-09-08
stop getting os version by parsing string

alanrenmsft:
this PR fixes https://github.com/microsoft/azuredatastudio/issues/11312

noticed this issue while testing the Azure SQL Edge instances, we have using the following query to get the version of the hosting os:
```SQL
SELECT OSVersion = RIGHT(@@version, LEN(@@version)- 3 -charindex (' on ', LOWER(@@version)))
```
It is relying on the 'on' keyword in the string returned by running the select @@version query, unfortunately, this keyword is not there for Azure SQL Edge instance. 

Fix: use the host information provided by the DMVs

Examples:
SQL Server on Linux : Ubuntu 16.04
SQL Server on Windows:
  if major version <= 13 (SQL Server 2016) - Windows 6.5
  otherwise - Windows Server 2019 Standard 10.0

I worked with Matteo on this, and SSMS is also using this way to display the OS version.

### DIFF
--- a/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
@@ -714,28 +714,34 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
         /// <param name="connection">The connection</param>
         public static ServerHostInfo GetServerHostInfo(IDbConnection connection)
         {
-            // SQL Server 2016 and below does not provide sys.dm_os_host_info
+            var hostInfo = new ServerHostInfo();
+            // SQL Server 2016 and earlier versions does not provide sys.dm_os_host_info and we know the host OS can only be Windows.
             if (!Version.TryParse(ReadServerVersion(connection), out var hostVersion) || hostVersion.Major <= 13)
             {
-                return new ServerHostInfo
-                {
-                    Platform = "Windows"
-                };
+                ExecuteReader(
+                    connection,
+                    SqlConnectionHelperScripts.GetHostWindowsVersion,
+                    reader =>
+                    {
+                        reader.Read();
+                        hostInfo.Platform = "Windows";
+                        hostInfo.Release = reader[0].ToString();
+                    });
             }
-
-            var hostInfo = new ServerHostInfo();
-            ExecuteReader(
-                connection,
-                SqlConnectionHelperScripts.GetHostInfo,
-                reader =>
-                {
-                    reader.Read();
-                    hostInfo.Platform = reader[0].ToString();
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -706,36 +706,42 @@ public static ServerInfo GetServerVersion(SqlConnectionStringBuilder csb, string
...
+            // SQL Server 2016 and earlier versions does not provide sys.dm_os_host_info and we know the host OS can only be Windows.
             if (!Version.TryParse(ReadServerVersion(connection), out var hostVersion) || hostVersion.Major <= 13)
             {
-                return new ServerHostInfo
-                {
-                    Platform = "Windows"
-                };
+                ExecuteReader(
+                    connection,
+                    SqlConnectionHelperScripts.GetHostWindowsVersion,
```
udeeshagautam, line 723: in this case "!Version.TryParse(ReadServerVersion(connection), out var hostVersion)" if GetHostWindowsVersion runs it will cause failure - right? Should we try catch it?

alanrenmsft: good catch, I am catching the error and ignore it, that falls back to the previous behavior of only setting the platform to  Windows.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -706,36 +706,49 @@ public static ServerInfo GetServerVersion(SqlConnectionStringBuilder csb, string
...
+                catch
                 {
-                    reader.Read();
-                    hostInfo.Platform = reader[0].ToString();
-                    hostInfo.Distribution = reader[1].ToString();
-                    hostInfo.Release = reader[2].ToString();
-                    hostInfo.ServicePackLevel = reader[3].ToString();
-                });
-
+                    // Ignore the error and only set the Platform to Windows by default
```
udeeshagautam, line 735: may be a console log? just for ease of debugging in case...

udeeshagautam (approved):

aaomidi (approved):

abist (approved):

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
alanrenmsft: oops, I only saw the approved message and didn't see this one, I guess it is ok not to have it, at debug time, the intellisense will show the exception...

PR #1071, opened 2020-09-10
Fix suggestions after keyword

gupta1:
Fixes intellisense after keywords like summarize, limit etc. The position index starts from 1. Changing character position to follow that.


JustinMDotNet (approved):
Looks good

PR #1073, opened 2020-09-10
3643 Intellisense fix for spaces

JustinMDotNet:
Enhanced GetAutoCompleteSuggestions to escape spaces in the insertText when generating list of suggestions. Changed detail to include kind.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
@@ -432,8 +432,7 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
                 var insertText = KustoQueryUtils.EscapeName(autoCompleteItem.DisplayText);
                 var completionKind = KustoIntellisenseHelper.CreateCompletionItemKind(autoCompleteItem.Kind);
                 completions.Add(AutoCompleteHelper.CreateCompletionItem(label, autoCompleteItem.Kind.ToString(),
-                    insertText,
-                    completionKind, scriptDocumentInfo.StartLine, scriptDocumentInfo.StartColumn,
+                    insertText, completionKind, scriptDocumentInfo.StartLine, scriptDocumentInfo.StartColumn,
                     textPosition.Character));
             }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
```diff
@@ -429,7 +429,12 @@ public override async Task<IEnumerable<T>> ExecuteControlCommandAsync<T>(string
             foreach (var autoCompleteItem in completion.Items)
             {
                 var label = autoCompleteItem.DisplayText;
-                completions.Add(AutoCompleteHelper.CreateCompletionItem(label, label + " keyword", label, KustoIntellisenseHelper.CreateCompletionItemKind(autoCompleteItem.Kind), scriptDocumentInfo.StartLine, scriptDocumentInfo.StartColumn, textPosition.Character));
+                var insertText = KustoQueryUtils.EscapeName(autoCompleteItem.DisplayText);
```
gupta1, line 432: Can we use label here?

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
```diff
@@ -429,7 +429,12 @@ public override async Task<IEnumerable<T>> ExecuteControlCommandAsync<T>(string
...
+                var completionKind = KustoIntellisenseHelper.CreateCompletionItemKind(autoCompleteItem.Kind);
```
gupta1, line 433: Is completionKind same as autoCompleteItem.Kind?

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
```diff
@@ -429,7 +429,12 @@ public override async Task<IEnumerable<T>> ExecuteControlCommandAsync<T>(string
...
+                completions.Add(AutoCompleteHelper.CreateCompletionItem(label, autoCompleteItem.Kind.ToString(),
+                    insertText,
+                    completionKind, scriptDocumentInfo.StartLine, scriptDocumentInfo.StartColumn,
```
gupta1, line 436: nit: formatting is a bit off here

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
JustinMDotNet: Completion Kind is the kind from the database, it gets translated to the AutoCompleteItem Kind when sent to ADS. I didn't use that because it translates Table -> File which seems odd for the AutoComplete suggestions.

JustinMDotNet: I just moved this out of the function and into it's own variable to make it easier to read.

gupta1 (approved):

PR #1074, opened 2020-09-11
Fix for Kusto control command with comments or spaces

gupta1:
Fix for Kusto control command with comments or spaces.
Currently Kusto control command like ".show databases" with trailing spaces or comments will fail


JustinMDotNet (approved):
Looks good

PR #1075, opened 2020-09-12
Fix for displaying multiple table

rajeshka:


src/Microsoft.Kusto.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -658,12 +625,30 @@ private async Task SendCurrentResults()
...
             }
         }
+        #endregion
+
+        #region Private Helper Methods
+        /// <summary>
+        /// Sends the ResultsUpdated message if the number of rows has changed since last send.
+        /// </summary>
+        /// <param name="stateInfo"></param>
+        private void SendResultAvailableOrUpdated(object stateInfo = null)
```
JustinMDotNet, line 646: Is this `stateInfo` still needed?

rajeshka: Yes it is a TimerCallback

namespace System.Threading
{
    //
    // Summary:
    //     Represents the method that handles calls from a System.Threading.Timer.
    //
    // Parameters:
    //   state:
    //     An object containing application-specific information relevant to the method
    //     invoked by this delegate, or null.
    public delegate void TimerCallback(object? state);
}

PR #1076, opened 2020-09-14
3715 Default database name for Kusto

JustinMDotNet:
Refactored GetDatabaseName to load database names from kusto server when no database name available in connection string.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
@@ -140,8 +140,7 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
             {
                 var rows = reader.ToEnumerable();
                 var row = rows?.FirstOrDefault();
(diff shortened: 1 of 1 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoDataSource.cs
```diff
@@ -117,14 +117,32 @@ private static string GetClusterName(string connectionString)
...
+            CancellationTokenSource source = new CancellationTokenSource();
+            CancellationToken token = source.Token;
+
+            string query = ".show databases | project DatabaseName";
+
+            using (var reader = ExecuteQuery(query, token))
+            {
+                var rows = reader.ToEnumerable();
+                var row = rows?.FirstOrDefault();
+                var tableName = row?[0].ToString() ?? string.Empty;
```
qifahs, line 143: Good thung that the failover case is an empty string. You could have a new Kusto cluster where databases have not yet been created.

qifahs (approved):

PR #1077, opened 2020-09-17
Change default kusto query to get 10 rows

gupta1:


src/Microsoft.Kusto.ServiceLayer/Scripting/Scripter.cs
```diff
@@ -22,7 +22,7 @@ public string SelectFromTableOrView(IDataSource dataSource, Urn urn)
             // TODOKusto: Extract into the Kusto folder.
             selectQuery.Append($"{KustoQueryUtils.EscapeName(urn.GetAttribute("Name"))}");
             selectQuery.Append($"{KustoQueryUtils.StatementSeparator}");
-            selectQuery.Append("limit 1000");
+            selectQuery.Append("limit 10");
```
qifahs, line 25: This should be 'take' not 'limit'

qifahs (changes requested):

src/Microsoft.Kusto.ServiceLayer/Scripting/Scripter.cs
gupta1: Changed to "take 10"

qifahs (approved):

PR #1079, opened 2020-09-19
Result handling for Kusto results

qifahs:
Better fix for #1075.

Reverted the previous change to return tables for multiple results. The previous change makes it difficult to write code for multiple queries separated by blank lines. Each result has to be separately trimmed. The changes to ResultSet.cs were not needed. 

Added code to KustoResultsReader.cs to trim results and package only the displayable tables into it's own IDataReader. The new changes are only in the KustoResultsReader.cs and DataReaderWrapper.cs. The rest of the changes are to revert the previous change which was reverted using 'git revert'.


rajeshka (approved):
I was apprehensive about doing work in the constructor. Seems like that is the only way.

gupta1 (approved):

PR #1082, opened 2020-09-23
Fix for multiple parameter error.

smartguest:
Clears the first use of the parameter (used for error checking with multiple delete), so the where clause parameter can be used again.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
@@ -111,9 +111,9 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData.UpdateManagement
             {
                 command.CommandText = input;
                 command.Parameters.AddRange(where.Parameters.ToArray());
-                using (DbDataReader reader = command.ExecuteReader())
+                try
                 {
-                    try
+                    using (DbDataReader reader = command.ExecuteReader())
                     {
                         while (reader.Read())
                         {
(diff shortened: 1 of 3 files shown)


alanrenmsft:
could you please explain a bit more on the root cause? also it seems you can have a finally block to do the clean up.

smartguest:
@alanrenmsft The parameters used in the where clause were used by a command to check for multiple delete errors, they were put into the command's parameters and associated with them. 

If you try to use them again, the code throws an error saying the parameters are already associated with a command parameter. This is what happened when I tried to add the same parameters into the actual delete command. 

The solution was to clear the parameters of the error check command so they could be reused again by the actual command.

This link showed me where the cause of the error was: https://stackoverflow.com/questions/7837762/the-sqlparameter-is-already-contained-by-another-sqlparametercollection-does-u

alanrenmsft:
ok, you should put a 
``` C#
try
{
    using (DbDataReader reader = command.ExecuteReader())
    ...
}
finally
{
    command.parameters.clear();
}

```

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -128,6 +128,9 @@ private bool CheckForDuplicateDeleteRows(WhereClause where, string input, DbConn
                     {
                         Logger.Write(TraceEventType.Error, ex.ToString());
                     }
+                    finally{
```
alanrenmsft, line 131: Please see my comments about where to add the try-finally block, if you add it here and ExecuteReader fails, the parameter collection won't be cleared.

alanrenmsft:
looking at the code, I think your code can be simplified to:
```C#
        private bool CheckForDuplicateDeleteRows(WhereClause where, string input, DbConnection connection)
        {
            int count = 0;
            using (DbCommand command = connection.CreateCommand())
            {
                command.CommandText = input;
                command.Parameters.AddRange(where.Parameters.ToArray());
                try
                {
                    count = Convert.ToInt32(command.ExecuteScalar());
                }
                finally
                {
                    command.Parameters.Clear();
                }
            }
            return count == 1;
        }
```
The execution exception should not be ignored, we should just let it bubble up and fail the request.

alanrenmsft:
regarding to adding some code back to make the test case pass, unit tests were created to make sure the code is behaving as expected, if we need to adjust the code to:
1. use more appropriate way, ExecuteScalar instead of ExecuteReader
2. Surface the error instead of ignore it, because it doesn't make sense to continue if the command execution fails
then we need to update the test cases to match the new behavior.

Charles-Gagnon (changes requested):
Take a look at Alan's suggested fix - I would prefer that way of doing it as well.

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -105,32 +105,23 @@ public override DbCommand GetCommand(DbConnection connection)
         /// Runs a query using the where clause to determine if duplicates are found (causes issues when deleting).
         /// If no duplicates are found, the check passes, else it returns false;
         /// </summary>
-        private bool CheckForDuplicateDeleteRows(WhereClause where, string input, DbConnection connection)
+          private bool CheckForDuplicateDeleteRows(WhereClause where, string input, DbConnection connection)
```
alanrenmsft, line 108: please fix the spacing/indention here

src/Microsoft.SqlTools.ServiceLayer/EditData/UpdateManagement/RowDelete.cs
```diff
@@ -107,30 +107,21 @@ public override DbCommand GetCommand(DbConnection connection)
         /// </summary>
         private bool CheckForDuplicateDeleteRows(WhereClause where, string input, DbConnection connection)
         {
+            int count = 0;
```
Charles-Gagnon, line 110: This is unnecessary - just make line 117 

`return Convert.ToInt32(command.ExecuteScalar()) <= 1`

I would also suggest 

1. Renaming the method to HasDuplicateRows (more descriptive name)
1. Change the logic to check for `>1`
1. Change the check on line 92 to be `if(HasDuplicateRows(...))`

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/TestObjects.cs
```diff
@@ -160,7 +160,36 @@ public override int ExecuteNonQuery()
 
         public override object ExecuteScalar()
         {
-            throw new NotImplementedException();
+            // Tailored for Row Delete tests
```
Charles-Gagnon, line 163: I don't think this comment is necessary. It's not important that edit data is the only thing using this currently and should have no impact on the implementation here. 

I would suggest adding a comment explaining what this code is doing since it's not directly clear what value this would be returning.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Utility/TestObjects.cs
```diff
@@ -158,9 +158,50 @@ public override int ExecuteNonQuery()
...
         }
 
+        /// <summary>
+        /// Function to check for duplicates in Data's rows.
+        /// Returns the max duplicate count for a unique row value (int).
+        /// </summary>
         public override object ExecuteScalar()
         {
-            throw new NotImplementedException();
+            if (Data != null)
```
Charles-Gagnon, line 167: I'm not super pleased with having this kind of specific logic tied here. I'd suggest creating a new class TestEditDataSqlCommand that extends TestSqlCommand and then implements this function there.

Charles-Gagnon (approved):

alanrenmsft (approved):

alanrenmsft:
@smartguest I am going to merge this and produce a new build of STS

PR #1085, opened 2020-09-29
3490 Kusto Connection Refresh Fix

JustinMDotNet:
Related Issue: https://github.com/microsoft/azuredatastudio/issues/11824

Injected OwnerUri into KustClient to store for token refreshing. Removed UpdateAzureToken from IDataSource, DataSourceBase, and KustoDataSource. Removed logic for retrying queries related to Unauthorized datasource in Batch and Query. Changed ScriptingService, ScriptingScriptOperation, and ScriptAsScriptingOperation to take DataSource in the constructor instead of datasourcefactory. Changed ScriptingService to inject ConnectionService through InitializeService function.


JustinMDotNet:
Issue: https://github.com/microsoft/azuredatastudio/issues/11824

src/Microsoft.Kusto.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -118,13 +113,6 @@ public override void Execute()
                     });
                 }
             }
-            finally
-            {
-                if (disconnectAtDispose && DataSource != null)
-                {
-                    DataSource.Dispose();
-                }
-            }
```
JustinMDotNet, line 127: Don't dispose of the datasource because it's a singleton from the ConnectionService

src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -284,8 +284,7 @@ internal void RefreshAzureToken(string ownerUri)
             var response = Instance.ServiceHost.SendRequest(SecurityTokenRequest.Type, requestMessage, true).Result;
             existingConnection.UpdateAzureToken(response.Token);
             
-            existingConnection.TryGetConnection(ConnectionType.Query, out var reliableDataSourceConnection);
-            reliableDataSourceConnection.GetUnderlyingConnection().UpdateAzureToken(response.Token);
```
JustinMDotNet, line 288: Doing this work within the KustoClient rather finding the client through the connection

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -36,8 +41,10 @@ public class KustoClient : IKustoClient
         public string ClusterName { get; }
         public string DatabaseName { get; private set; }
 
-        public KustoClient(string connectionString, string azureAccountToken)
+        public KustoClient(string connectionString, string azureAccountToken, string ownerUri)
         {
+            _ownerUri = ownerUri;
+            _retryCount = 1;
```
JustinMDotNet, line 47: Default to retry once. Saving OwnerUri within the Client so it's able to refresh the token easily.

src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -453,7 +452,7 @@ private async Task<ConnectionCompleteParams> TryOpenConnection(ConnectionInfo co
                 string connectionString = BuildConnectionString(connectionInfo.ConnectionDetails);
 
                 // create a sql connection instance
-                connection = connectionInfo.Factory.CreateDataSourceConnection(connectionString, connectionInfo.ConnectionDetails.AzureAccountToken);
+                connection = connectionInfo.Factory.CreateDataSourceConnection(connectionString, connectionInfo.ConnectionDetails.AzureAccountToken, connectionInfo.OwnerUri);
```
JustinMDotNet, line 455: Pass down OwnerUri to add to KustoClient

src/Microsoft.Kusto.ServiceLayer/DataSource/DataSourceBase.cs
```diff
@@ -89,8 +89,6 @@ public async Task<T> ExecuteScalarQueryAsync<T>(string query, CancellationToken
 
         public abstract string GenerateExecuteFunctionScript(string functionName);
 
-        public abstract void UpdateAzureToken(string azureToken);
```
JustinMDotNet, line 92: DataSource no longer handles refreshing the AzureToken

src/Microsoft.Kusto.ServiceLayer/HostLoader.cs
```diff
@@ -91,7 +90,7 @@ private static void InitializeRequestHandlersAndServices(ServiceHost serviceHost
             QueryExecutionService.Instance.InitializeService(serviceHost);
             serviceProvider.RegisterSingleService(QueryExecutionService.Instance);
 
-            ScriptingService.Instance.InitializeService(serviceHost, scripter, dataSourceFactory);
```
JustinMDotNet, line 94: No longer creating connection within ScriptingService so we need the ConnectionService to be able to access existing Kusto connections

src/Microsoft.Kusto.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -267,12 +263,6 @@ public async Task Execute(ReliableDataSourceConnection conn, CancellationToken c
             {
                 await DoExecute(conn, cancellationToken);
             }
-            catch (DataSourceUnauthorizedException)
-            {
-                // Rerun the query once if unauthorized
-                _retryCount--;
-                throw;
-            }
```
JustinMDotNet, line 275: Not bubbling up token refresh logic so this can be removed

src/Microsoft.Kusto.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -391,11 +390,6 @@ private async Task ExecuteInternal()
                     await QueryCompleted(this);
                 }
             }
-            catch (DataSourceUnauthorizedException)
-            {
-                ConnectionService.Instance.RefreshAzureToken(editorConnection.OwnerUri);
-                await ExecuteInternal();
-            }
```
JustinMDotNet, line 398: Not bubbling up token refresh logic so this can be removed

src/Microsoft.Kusto.ServiceLayer/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -24,20 +24,15 @@ public class ScriptAsScriptingOperation : SmoScriptingOperation
...
-            DataSource = _dataSourceFactory.Create(DataSourceType.Kusto, this.Parameters.ConnectionString,
-                azureAccountToken);
             _scripter = scripter;
         }
 
-        internal IDataSource DataSource { get; set; }
-
-        private string serverName;
-        private string databaseName;
-        private bool disconnectAtDispose = false;
```
JustinMDotNet, line 39: These properties aren't needed since we already have an existing datasource

src/Microsoft.Kusto.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -22,40 +21,18 @@ namespace Microsoft.Kusto.ServiceLayer.Scripting
     /// </summary>
     public sealed class ScriptingService : IDisposable
     {
-        private const int ScriptingOperationTimeout = 60000;
```
JustinMDotNet, line 25: Unused property

src/Microsoft.Kusto.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -22,40 +21,18 @@ namespace Microsoft.Kusto.ServiceLayer.Scripting
...
 
         private bool disposed;
         
         private IScripter _scripter;
-        private IDataSourceFactory _dataSourceFactory;
-
-        /// <summary>
-        /// Internal for testing purposes only
-        /// </summary>
-        internal static ConnectionService ConnectionServiceInstance
```
JustinMDotNet, line 44: This is being injected in InitializeService instead of being initialized within the property

src/Microsoft.Kusto.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -108,36 +87,35 @@ private async Task HandleListObjectsRequest(ScriptingListObjectsParams parameter
...
                     }
                     else
                     {
                         throw new Exception("Could not find ConnectionInfo");
                     }
                 }
 
+                SmoScriptingOperation operation;
+                var datasource = _connectionService.GetOrOpenConnection(parameters.OwnerUri, ConnectionType.Default)
+                    .Result.GetUnderlyingConnection();
```
JustinMDotNet, line 111: Get existing datasource and pass it down instead of creating new ones.

src/Microsoft.Kusto.ServiceLayer/Scripting/SmoScriptingOperation.cs
```diff
@@ -73,17 +72,10 @@ protected virtual void SetCommonEventProperties(ScriptingEventParams parameters)
             parameters.OperationId = this.OperationId;
         }
 
-        protected string GetServerNameFromLiveInstance(string connectionString, string azureAccessToken)
+        protected string GetServerNameFromLiveInstance()
         {
-            string serverName = string.Empty;
-            
-            using(var dataSource = _dataSourceFactory.Create(DataSourceType.Kusto, connectionString, azureAccessToken))
```
JustinMDotNet, line 80: Use existing datasource instead of creating a new one _just_ to get the server name

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -187,7 +195,14 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
             }
             catch (KustoRequestException exception) when (exception.FailureCode == 401) // Unauthorized
             {
-                throw new DataSourceUnauthorizedException(exception);
+                if (_retryCount <= 0)
+                {
+                    throw;
+                }
+
+                _retryCount--;
```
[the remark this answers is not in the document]
qifahs: Or just have the parent function call ExecuteQueryInternal once, if the right exception is thrown, call it one more time. It will just make the retry be simple.

For the long term, I think we should just have multiple retry logic on any calls going over the wire to the server. There are many reasons the call could fail and retries with an some waits in the middle will help our code be more reliable for temporary network issues.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -75,8 +82,9 @@ private void Initialize(string clusterName, string databaseName, string azureAcc
             _kustoAdminProvider = KustoClientFactory.CreateCslAdminProvider(stringBuilder);
         }
 
-        public void UpdateAzureToken(string azureAccountToken)
+        private void RefreshAzureToken()
         {
+            string azureAccountToken = ConnectionService.Instance.RefreshAzureToken(_ownerUri);
```
qifahs, line 87: There is too much threading of this parameter through all the classes which do not need it. I am thinking if we can avoid it in some manner. The ConnectionService has all the ConnectionInfos. The ConnectionInfo has the list of active ReliableDataSources->DataSource. I wonder if we can just call RefreshAzureToken and pass it a reference to the IDataSource we are calling from, and ConnectionService can match the ConnectionInfo it needs to operate on.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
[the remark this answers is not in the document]
qifahs: Checkout the RetryPolicy used in ReliableDataSourceConnection. I think we should use it in all calls we make to Kusto. I think a better retry policy can be done in a separate checkin. Noting it here, as I find retries missing in general. 

For this change, it is not really a retry but refresh of the token. Another thing to consider for token refresh is, why is this only in ExecuteQuery? It should be in all function which call back to Kusto. All calls to Kusto should eventually had a retry, maybe using RetryPolicy, and also handle token refresh if needed.

qifahs (changes requested):

src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -786,9 +784,10 @@ public ListDatabasesResponse ListDatabases(ListDatabasesParams listDatabasesPara
             {
                 throw new Exception(SR.ConnectionServiceListDbErrorNotConnected(owner));
             }
-            ConnectionDetails connectionDetails = info.ConnectionDetails.Clone();
 
-            IDataSource dataSource = OpenDataSourceConnection(info);
+            info.TryGetConnection("Default", out ReliableDataSourceConnection connection);
+            IDataSource dataSource = connection.GetUnderlyingConnection();
```
JustinMDotNet, line 789: Use existing connection instead of creating a new one

src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -161,6 +161,11 @@ public void RemoveAllConnections()
         public void UpdateAzureToken(string token)
         {
             ConnectionDetails.AzureAccountToken = token;
+
+            foreach (var connection in _connectionTypeToConnectionMap.Values)
+            {
+                connection.UpdateAzureToken(token);
+            }
```
JustinMDotNet, line 168: Need to update all the ReliableDataSourceConnections under the ConnectionInfo ^

qifahs (approved):

gupta1 (approved):

PR #1089, opened 2020-10-01
Fix for multiple queries separated by space for Kusto

gupta1:
Since Kusto doesnt understand and executes multiple queries separated by space directly. Fix is to separately send all the queries after parsing the entire query text from editor/notebook

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
@@ -173,22 +173,29 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
             clientRequestProperties.SetOption(ClientRequestProperties.OptionNoTruncation, true);
             cancellationToken.Register(() => CancelQuery(clientRequestProperties.ClientRequestId));
 
-            var kustoCodeService = new KustoCodeService(query);
-            var minimalQuery = kustoCodeService.GetMinimalText(MinimalTextKind.RemoveLeadingWhitespaceAndComments);
+            var script = CodeScript.From(query, GlobalState.Default);
+            List<IDataReader> origReaders = new List<IDataReader>();
 
-            try
+            foreach (var codeBlock in script.Blocks)
             {
-                IDataReader origReader = _kustoQueryProvider.ExecuteQuery(
-                    KustoQueryUtils.IsClusterLevelQuery(minimalQuery) ? "" : databaseName,
-                    minimalQuery,
-                    clientRequestProperties);
+                var minimalQuery = codeBlock.Service.GetMinimalText(MinimalTextKind.RemoveLeadingWhitespaceAndComments);
                 
-                return new KustoResultsReader(origReader);
+                try
+                {
+                    IDataReader origReader = _kustoQueryProvider.ExecuteQuery(
+                        KustoQueryUtils.IsClusterLevelQuery(minimalQuery) ? "" : databaseName,
+                        minimalQuery,
+                        clientRequestProperties);
+                    
+                    origReaders.Add(origReader);
+                }
+                catch (KustoRequestException exception) when (exception.FailureCode == 401) // Unauthorized
+                {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -173,22 +173,29 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
             clientRequestProperties.SetOption(ClientRequestProperties.OptionNoTruncation, true);
             cancellationToken.Register(() => CancelQuery(clientRequestProperties.ClientRequestId));
 
-            var kustoCodeService = new KustoCodeService(query);
-            var minimalQuery = kustoCodeService.GetMinimalText(MinimalTextKind.RemoveLeadingWhitespaceAndComments);
+            var script = CodeScript.From(query, GlobalState.Default);
+            List<IDataReader> origReaders = new List<IDataReader>();
 
-            try
+            foreach (var codeBlock in script.Blocks)
```
qifahs, line 179: Instead of going sequential, we should do this in parallel. Send all queries in parallel and wait for all to return.

qifahs: When doing in parallel, you might have to make sure that the results are put back in order. So using an array for the number of results might work out better. Each result can go into its query's index.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoResultsReader.cs
```diff
@@ -11,26 +11,31 @@ internal class KustoResultsReader : DataReaderWrapper
...
-            // Trim results
-            if(results.Count > 3) results.RemoveRange(results.Count - 3, 3);
-
             // Create a DataReader for the trimmed set
             _resultDataSet = new DataSet();
-            foreach(var result in results)
+            for(int i = 0; i < results.Count; i++)
             {
-                _resultDataSet.Tables.Add(result);
+                results[i].TableName = "Table_" + i;
```
qifahs, line 37: Not sure how this Table name shows up in the UX. Please check on this and do something which is on par with the SQL experience.

qifahs (changes requested):

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoResultsReader.cs
gupta1: This name doesn't show anywhere in UX, its just for identifying it with unique name in_resultDataSet since all the separate queries return tables with name Table_0

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -173,22 +173,29 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
             clientRequestProperties.SetOption(ClientRequestProperties.OptionNoTruncation, true);
             cancellationToken.Register(() => CancelQuery(clientRequestProperties.ClientRequestId));
 
-            var kustoCodeService = new KustoCodeService(query);
-            var minimalQuery = kustoCodeService.GetMinimalText(MinimalTextKind.RemoveLeadingWhitespaceAndComments);
+            var script = CodeScript.From(query, GlobalState.Default);
+            IDataReader[] origReaders = new IDataReader[script.Blocks.Count];
 
-            try
+            Parallel.ForEach(script.Blocks, (codeBlock, state, index) =>
```
JustinMDotNet, line 179: Should we pass the cancellation token into this Parallel.Foreach incase the user cancels?

qifahs (approved):

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
gupta1: Looks like in previous implementation, its not passed. So, it should not regress anything. We can discuss and target that in a separate PR too.

PR #1090, opened 2020-10-06
sqlod changes for cloud servertype

alljamziMS:
The ServerType of SQL OD in SMO is being changed to cloud. This is a minimal set of changes needed in sqltoolservice in order to keep everything working within ADS

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
 using System.Linq;
+using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlTools.ServiceLayer.Admin.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -25,9 +26,9 @@ public class ListDatabasesRequestDatabaseProperties
     /// </summary>
     static class ListDatabaseRequestHandlerFactory
     {
-        public static IListDatabaseRequestHandler getHandler(bool includeDetails, bool isSqlDB)
+        public static IListDatabaseRequestHandler getHandler(bool includeDetails, bool isSqlDB, ConnectionInfo connectionInfo = null)
         {
-            if (!includeDetails)
+            if (!includeDetails || (connectionInfo != null && connectionInfo.EngineEdition == DatabaseEngineEdition.SqlOnDemand))
```
Charles-Gagnon, line 31: `connectionInfo?.EngineEdition == DatabaseEngineEdition.SqlOnDemand`

Charles-Gagnon (changes requested):
Don't you need to update the SMO version too?

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
BranislavGrbicMDCS: @Charles-Gagnon - there is another STS change on the way. 
I'll be merging the changes so there is only one review.

The comment is resolved in the PR that is on the way.

Charles-Gagnon:
@BranislavGrbicMDCS @alljamziMS Are you planning on finishing this at some point? If not it'd be great if you would close this to clean up stale PRs

PR #1096, opened 2020-10-14
3875 Kusto Retry Logic

JustinMDotNet:
Changed ExecuteQueryAsync and ExecuteControlCommandAsync to return correct types in IKustoClient. Added RetryLimit to KustoClient. Added check to skip database and function schema when no databaseName available. Added logic for retry on ExecuteQuery and ExecuteControlCommand

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/IKustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/IKustoClient.cs
@@ -22,14 +22,15 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
         /// <summary>
         /// Executes a query or command against a kusto cluster and returns a sequence of result row instances.
         /// </summary>
-        Task<IEnumerable<T>> ExecuteControlCommandAsync<T>(string command, bool throwOnError, CancellationToken cancellationToken);
+        Task ExecuteControlCommandAsync(string command, bool throwOnError, CancellationToken cancellationToken);
 
         /// <summary>
         /// Executes a query.
         /// </summary>
         /// <param name="query">The query.</param>
         /// <returns>The results.</returns>
-        Task<IDataReader> ExecuteQueryAsync(string query, CancellationToken cancellationToken, string databaseName = null);
+        Task<IEnumerable<T>> ExecuteQueryAsync<T>(string query, CancellationToken cancellationToken,
+            string databaseName = null);
 
         /// <summary>
         /// Executes a Kusto control command.
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/IKustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/IKustoClient.cs
@@ -3,6 +3,7 @@ using System.Data;
 using System.Threading;
 using System.Threading.Tasks;
 using Kusto.Language;
+using Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection;
 
 namespace Microsoft.Kusto.ServiceLayer.DataSource
 {
@@ -39,7 +40,7 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
         /// <param name="retryCount"></param>
         void ExecuteControlCommand(string command, int retryCount = 1);
 
-        void UpdateDatabase(string databaseName);
+        void UpdateDatabase(string databaseName, RetryPolicy commandRetryPolicy);
         
         void Dispose();
     }
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
@@ -38,6 +38,8 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
         public string ClusterName { get; }
         public string DatabaseName { get; private set; }
 
+        private const int RetryLimit = 60;
+
         public KustoClient(string connectionString, string azureAccountToken, string ownerUri)
         {
             _ownerUri = ownerUri;
@@ -49,21 +51,23 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
 
         private GlobalState LoadSchemaState()
         {
-            CancellationTokenSource source = new CancellationTokenSource();
-
             IEnumerable<ShowDatabaseSchemaResult> tableSchemas = Enumerable.Empty<ShowDatabaseSchemaResult>();
             IEnumerable<ShowFunctionsResult> functionSchemas = Enumerable.Empty<ShowFunctionsResult>();
 
-            Parallel.Invoke(() =>
+            if (!string.IsNullOrWhiteSpace(DatabaseName))
             {
-                tableSchemas = ExecuteControlCommandAsync<ShowDatabaseSchemaResult>(
-                    $".show database {DatabaseName} schema",
-                    false, source.Token).Result;
-            }, () =>
-            {
-                functionSchemas = ExecuteControlCommandAsync<ShowFunctionsResult>(".show functions", false,
-                    source.Token).Result;
-            });
+                var source = new CancellationTokenSource();
+                Parallel.Invoke(() =>
+                {
+                    tableSchemas =
+                        ExecuteQueryAsync<ShowDatabaseSchemaResult>($".show database {DatabaseName} schema",
+                            source.Token, DatabaseName).Result;
+                }, () =>
+                {
+                    functionSchemas =
+                        ExecuteQueryAsync<ShowFunctionsResult>(".show functions", source.Token, DatabaseName).Result;
+                });
+            }
 
             return KustoIntellisenseHelper.AddOrUpdateDatabase(tableSchemas, functionSchemas,
                 GlobalState.Default,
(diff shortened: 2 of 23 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -49,21 +51,23 @@ public KustoClient(string connectionString, string azureAccountToken, string own
 
         private GlobalState LoadSchemaState()
         {
-            CancellationTokenSource source = new CancellationTokenSource();
-
             IEnumerable<ShowDatabaseSchemaResult> tableSchemas = Enumerable.Empty<ShowDatabaseSchemaResult>();
             IEnumerable<ShowFunctionsResult> functionSchemas = Enumerable.Empty<ShowFunctionsResult>();
 
-            Parallel.Invoke(() =>
+            if (!string.IsNullOrWhiteSpace(DatabaseName))
```
JustinMDotNet, line 57: Added check to not attempt showing database information if database name is empty

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -49,21 +51,23 @@ public KustoClient(string connectionString, string azureAccountToken, string own
...
             {
-                tableSchemas = ExecuteControlCommandAsync<ShowDatabaseSchemaResult>(
-                    $".show database {DatabaseName} schema",
-                    false, source.Token).Result;
-            }, () =>
-            {
-                functionSchemas = ExecuteControlCommandAsync<ShowFunctionsResult>(".show functions", false,
-                    source.Token).Result;
-            });
+                var source = new CancellationTokenSource();
```
qifahs, line 59: It does not seem right to create a cancellation token here, as the user will not be able to use this to cancel anything. This should come through the call which is initializing state. I think we should use the RetryPolicy which is set up in the ReliableDataSourceConnection. There might be a cancellation token being passed in when this operation is invoked.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -183,36 +189,40 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
...
                 });
-                
+
                 return new KustoResultsReader(origReaders);
             }
-            catch (AggregateException exception) 
-                when (retryCount > 0 &&
-                      exception.InnerException is KustoRequestException innerException 
+            catch (AggregateException exception)
+                when (retryCount < RetryLimit &&
```
qifahs, line 196: I think we should use the RetryPolicy which is setup for the ReliableDataSourceConnection. Maybe pass it in here. It can handle the retries, etc. This way the user can override our policy. If the user does not pass in a retry policy we can enforce our default by creating a RetryPolicy which reflect what you are trying to do here.

qifahs: There are various types of retry policies defined in RetryPolicy.cs, including ExponentialDelayRetryPolicy.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -221,11 +231,20 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
...
+            try
+            {
+                var resultReader = await Task.Run(() => ExecuteQuery(query, cancellationToken, databaseName), cancellationToken);
+                var results = KustoDataReaderParser.ParseV1(resultReader, null);
+                var tableReader = results[WellKnownDataSet.PrimaryResult].Single().TableData.CreateDataReader();
+                return new ObjectReader<T>(tableReader);
+            }
+            catch (Exception)
+            {
+                return null;
```
qifahs, line 246: It will be good to add telemetry on these failures so that we can trace what went wrong.

qifahs (changes requested):

src/Microsoft.Kusto.ServiceLayer/DataSource/IKustoClient.cs
```diff
@@ -22,14 +22,15 @@ public interface IKustoClient
         /// <summary>
         /// Executes a query or command against a kusto cluster and returns a sequence of result row instances.
         /// </summary>
-        Task<IEnumerable<T>> ExecuteControlCommandAsync<T>(string command, bool throwOnError, CancellationToken cancellationToken);
+        Task ExecuteControlCommandAsync(string command, bool throwOnError, CancellationToken cancellationToken);
```
JustinMDotNet, line 25: Changed Command to return a task instead of an IEnumerable because it shouldn't be returning data since it's not querying.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -146,7 +148,7 @@ private string GetClusterName(string connectionString)
             return kcsb;
         }
 
-        private ClientRequestProperties GetCLientRequestProperties(CancellationToken cancellationToken){
+        private ClientRequestProperties GetClientRequestProperties(CancellationToken cancellationToken){
```
JustinMDotNet, line 151: Fixed capitalization.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -170,18 +173,22 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
...
+                    if (!string.IsNullOrEmpty(minimalQuery))
+                    {
+                        // Query is empty in case of comments
                         IDataReader origReader;
-                        var clientRequestProperties = GetCLientRequestProperties(cancellationToken);
+                        var clientRequestProperties = GetClientRequestProperties(cancellationToken);
 
-                        if(minimalQuery.StartsWith(".") && !minimalQuery.StartsWith(".show")){
+                        if (minimalQuery.StartsWith(".") && !minimalQuery.StartsWith(".show"))
+                        {
```
JustinMDotNet, line 184: Fixed formatting here

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -193,44 +200,40 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
...
                 return ExecuteQuery(query, cancellationToken, databaseName, retryCount);
             }
         }
-        
+
         /// <summary>
         /// Executes a query or command against a kusto cluster and returns a sequence of result row instances.
         /// </summary>
-        public async Task<IEnumerable<T>> ExecuteControlCommandAsync<T>(string command, bool throwOnError,
+        public async Task ExecuteControlCommandAsync(string command, bool throwOnError,
```
JustinMDotNet, line 228: Changed function to execute command and not execute query.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -193,44 +200,40 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
...
             CancellationToken cancellationToken)
         {
             try
             {
-                var resultReader = await ExecuteQueryAsync(command, cancellationToken, DatabaseName);
-                var results = KustoDataReaderParser.ParseV1(resultReader, null);
-                var tableReader = results[WellKnownDataSet.PrimaryResult].Single().TableData.CreateDataReader();
-                return new ObjectReader<T>(tableReader);
+                await Task.Run(() => ExecuteControlCommand(command), cancellationToken);
```
JustinMDotNet, line 233: Changed function to execute command and not execute query since that's the function name.

src/Microsoft.Kusto.ServiceLayer/DataSource/ReliableDataSourceConnection.cs
```diff
@@ -224,6 +215,34 @@ public Task OpenAsync(CancellationToken token)
...
+        public string GenerateAlterFunctionScript(string functionName)
+        {
+            return _commandRetryPolicy.ExecuteAction(() => _dataSource.GenerateAlterFunctionScript(functionName));
+        }
+
+        public string GenerateExecuteFunctionScript(string functionName)
+        {
+            return _commandRetryPolicy.ExecuteAction(() => _dataSource.GenerateExecuteFunctionScript(functionName));
+        }
+
```
JustinMDotNet, line 245: Added the above functions to not expose the Datasource to the rest of the project.

src/Microsoft.Kusto.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -860,15 +858,17 @@ internal Hover GetHoverItem(TextDocumentPosition textDocumentPosition, ScriptFil
 
             ScriptDocumentInfo scriptDocumentInfo = new ScriptDocumentInfo(textDocumentPosition, scriptFile, scriptParseInfo);
 
-            if(connInfo != null){
+            if (connInfo != null)
```
JustinMDotNet, line 861: Fixed formatting and pulled SchemaState from the connection instead of dataSource

JustinMDotNet:
Closing PR to be reworked at a later date.

PR #1101, opened 2020-10-19
Fix for queries with comments separated by new lines

gupta1:
Fix for Kusto queries separated by newline
query example:

// Comment to explain

ab | take  1

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
@@ -146,16 +146,20 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
             return kcsb;
         }
 
-        public IDataReader ExecuteQuery(string query, CancellationToken cancellationToken, string databaseName = null, int retryCount = 1)
-        {
-            ValidationUtils.IsArgumentNotNullOrWhiteSpace(query, nameof(query));
-
+        private ClientRequestProperties GetCLientRequestProperties(CancellationToken cancellationToken){
             var clientRequestProperties = new ClientRequestProperties
             {
                 ClientRequestId = Guid.NewGuid().ToString()
             };
             clientRequestProperties.SetOption(ClientRequestProperties.OptionNoTruncation, true);
             cancellationToken.Register(() => CancelQuery(clientRequestProperties.ClientRequestId));
+
+            return clientRequestProperties;
+        }
+
+        public IDataReader ExecuteQuery(string query, CancellationToken cancellationToken, string databaseName = null, int retryCount = 1)
+        {
+            ValidationUtils.IsArgumentNotNullOrWhiteSpace(query, nameof(query));
 
             var script = CodeScript.From(query, GlobalState.Default);
             IDataReader[] origReaders = new IDataReader[script.Blocks.Count];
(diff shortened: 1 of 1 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -167,21 +167,23 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
...
-                    }
-                    else{
-                        origReader = _kustoQueryProvider.ExecuteQuery(
-                            KustoQueryUtils.IsClusterLevelQuery(minimalQuery) ? "" : databaseName,
-                            minimalQuery,
-                            clientRequestProperties);
-                    }
 
-                    origReaders[index] = origReader;
+                    if(!string.IsNullOrEmpty(minimalQuery)){        // Query is empty in case of comments
```
qifahs, line 171: For the empty DataReader in origReaders, does KustoResultsReader handle it correctly. I am assuming that it will be undefined/null and so KustoDataReader skips it. But please do check on this.

qifahs (approved):

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
gupta1: Sending the entire text in query editor to mimic Kusto Explorer behavior when trying to run only comments.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -161,28 +161,39 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
             IDataReader[] origReaders = new IDataReader[script.Blocks.Count];
             try
             {
+                var numOfQueries = 0;
```
JustinMDotNet, line 164: Isn't the number of queries just the length of `script.Blocks`?

gupta1: Not really, even comments are considered as blocks in Kusto api

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -161,28 +161,39 @@ public IDataReader ExecuteQuery(string query, CancellationToken cancellationToke
...
                     var minimalQuery =
                         codeBlock.Service.GetMinimalText(MinimalTextKind.RemoveLeadingWhitespaceAndComments);
                     
-                    IDataReader origReader;
-                
-                    if(minimalQuery.StartsWith(".") && !minimalQuery.StartsWith(".show")){
-                        origReader = _kustoAdminProvider.ExecuteControlCommand(
-                            KustoQueryUtils.IsClusterLevelQuery(minimalQuery) ? "" : databaseName,
-                            minimalQuery,
-                            clientRequestProperties);
```
qifahs, line 175: I think we need a new clientRequestProperties for each call to .Execute...(). We should pull that from the top of this function into this loop. Or create helper functions which will initialize clentRequestProperties and call the .Execute...() functions.

qifahs (changes requested):

qifahs (approved):

PR #1103, opened 2020-10-27
Add a parent process ID argument to ServiceLayer so that we can exit gracefully if the parent process exits.

corivera:
When starting SqlToolsService from a C# app, the service process doesn't exit if its parent process exits. This change adds an optional "parent-pid" argument to the ServiceLayer process so that we can periodically check if the parent process is still running, and gracefully exit if it already exited. This change is needed for the SQL kernel work I'm doing in dotnet-interactive.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Program.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Program.cs
@@ -77,10 +77,13 @@ namespace Microsoft.SqlTools.ServiceLayer
 
         private static void CheckParentStatusLoop(Process parent)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Program.cs
```diff
@@ -66,5 +74,17 @@ internal static void Main(string[] args)
                 Logger.Close();
             }
         }
+
+        private static void CheckParentStatusLoop(Process parent)
+        {
+            while (true)
```
Charles-Gagnon, line 80: Let's log that we're starting up this loop and when it exits as well

Charles-Gagnon (approved):

PR #1105, opened 2020-10-29
Added additional code path if user.UniqueId not set

cssuh:


abist (approved):

abist:
Make sure the latest mssql release uses the latest sqltoolsservice version that has this fix

PR #1113, opened 2020-10-31
Fix logic for result 

cssuh:
Fixes https://github.com/microsoft/vscode-mssql/issues/1779
The flow was not entering the `result == null` portion since it is initially set to `Enumerable.Empty<IAzureUserAccountSubscriptionContext>()`


abist:
Could you write a comment on why this change was required?

abist (approved):

PR #1116, opened 2020-11-06
Update release pipeline to validate build artifacts hash against current source hash

Charles-Gagnon:
This is the release pipeline portion of the changes made to the build scripts

This will prevent us accidently releasing a build with a higher version number but older artifacts - this pipeline is currently set up to assume that the build being released is the latest and so this is just enforcing that.

In the future if we decide we want to support release older builds then we can add that functionality but so far I haven't seen that being necessary so this should be fine as it is.

I also updated it to not use the pre-release tag, there isn't really a point in marking this as pre-release and it just makes finding the build slightly more annoying.

### DIFF
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -10,7 +10,7 @@ stages:
       inputs:
         azureSubscription: 'ClientToolsInfra_670062 (88d5392f-a34f-4769-b405-f597fc533613)'
         KeyVaultName: 'ado-secrets'
-        SecretsFilter: 'github-distro-mixin-password,ado-crossplatbuildscripts-password'
+        SecretsFilter: 'github-distro-mixin-password,ado-crossplatbuildscripts-password,ado-crossplatbuildscripts-builds-password'
     - powershell: |
         git clone https://$(ado-crossplatbuildscripts-password)@dev.azure.com/mssqltools/_git/CrossPlatBuildScripts
         cd CrossPlatBuildScripts/AutomatedReleases/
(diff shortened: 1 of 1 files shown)


cssuh (approved):

azure-pipelines/release.yml
```diff
@@ -50,11 +50,11 @@ stages:
       displayName: 'PowerShell Script'
       inputs:
         filePath: '$(System.DefaultWorkingDirectory)/CrossPlatBuildScripts/AutomatedReleases/sqltoolsserviceRelease.ps1'
-        arguments: '-workspace $(System.DefaultWorkingDirectory)/sqltoolsservice -minTag v3.0.0.0 -target main -tagFormat release'
+        arguments: '-workspace $(System.DefaultWorkingDirectory)/sqltoolsservice -minTag v3.0.0.0 -target main -tagFormat release -isPrerelease $false -artifactsBuildId $(Build.TriggeredBy.BuildId)'
         workingDirectory: '$(System.DefaultWorkingDirectory)/sqltoolsservice'
       env:
         GITHUB_DISTRO_MIXIN_PASSWORD: $(github-distro-mixin-password)
-
+        ADO_CROSSPLATBUILDSCRIPTS_PASSWORD: $(ado-crossplatbuildscripts-password)
```
alanrenmsft, line 57: it is not very obvious what this password is for, is it a personal access token?

Charles-Gagnon: Yeah, need it to access build info

alanrenmsft (approved):
the change looks fine to me, what's wrong with the code scanning build, it failed for all the commits in this PR.

Charles-Gagnon:
Not sure, error doesn't seem related though so I'll look into it later if it keeps happening.

PR #1118, opened 2020-11-09
Remove schedule from pipeline

Charles-Gagnon:
Having a default job here made it so that if we didn't want it to run on a schedule (say for adhoc runs) we couldn't do that (no schedule in ADO means it defaults to the yaml defined schedule).

Instead we'll just not define it here and instead define it in the ADO pipelines that use it.

### DIFF
--- a/azure-pipelines/build-and-release.yml
+++ b/azure-pipelines/build-and-release.yml
@@ -1,12 +1,6 @@
 trigger: none
 pr: none
-
-schedules:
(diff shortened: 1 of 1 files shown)


alanrenmsft:
looks good, I've done the same for ADS pipelines a while back, missed STS...

alanrenmsft (approved):

abist (approved):

PR #1120, opened 2020-11-11
Fix nunit package reference

Charles-Gagnon:
Fixing these errors that started popping up recently. 

```
 /home/runner/work/sqltoolsservice/sqltoolsservice/test/Microsoft.SqlTools.Hosting.UnitTests/Microsoft.SqlTools.Hosting.UnitTests.csproj : error : The package reference 'nunit' must have a version defined in '/home/runner/work/sqltoolsservice/sqltoolsservice/Packages.props'.
```

Not sure why this only started happening now and why case would make a difference but everything else looks correct.


Charles-Gagnon:
Merging this in now since it fixes some of the issues. Will follow up with other fixes if we decide to further pursue getting this working again.

PR #1123, opened 2020-11-19
13159 Kusto Aria Database Name Fix

JustinMDotNet:
Created SetDatabaseName in KustoClient to take database name from ( ). Changed ListDatabases in Kusto ConnnectionService to set DatabaseNames in a different format when the PrettyName doesn't match the name.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
@@ -793,7 +793,8 @@ namespace Microsoft.Kusto.ServiceLayer.Connection
             ListDatabasesResponse response = new ListDatabasesResponse();
 
             // Mainly used by "manage" dashboard
-            if(listDatabasesParams.IncludeDetails.HasTrue()){
+            if (listDatabasesParams.IncludeDetails.HasTrue())
+            {
                 IEnumerable<DataSourceObjectMetadata> databaseMetadataInfo = dataSource.GetChildObjects(objectMetadata, true);
                 List<DatabaseInfo> metadata = MetadataFactory.ConvertToDatabaseInfo(databaseMetadataInfo);
                 response.Databases = metadata.ToArray();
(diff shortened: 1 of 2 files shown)


src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -793,7 +793,8 @@ public ListDatabasesResponse ListDatabases(ListDatabasesParams listDatabasesPara
             ListDatabasesResponse response = new ListDatabasesResponse();
 
             // Mainly used by "manage" dashboard
-            if(listDatabasesParams.IncludeDetails.HasTrue()){
+            if (listDatabasesParams.IncludeDetails.HasTrue())
+            {
```
JustinMDotNet, line 797: Slight formatting change

src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -802,7 +803,13 @@ public ListDatabasesResponse ListDatabases(ListDatabasesParams listDatabasesPara
...
 
             IEnumerable<DataSourceObjectMetadata> databaseMetadata = dataSource.GetChildObjects(objectMetadata);
-            if(databaseMetadata != null) response.DatabaseNames = databaseMetadata.Select(objMeta => objMeta.PrettyName).ToArray();
+            if (databaseMetadata != null)
+            {
+                response.DatabaseNames = databaseMetadata
+                    .Select(objMeta => objMeta.PrettyName == objMeta.Name ? objMeta.PrettyName : $"{objMeta.PrettyName} ({objMeta.Name})")
+                    .ToArray();
+            }
+
```
JustinMDotNet, line 812: Changed the way we render the database name when the names don't match. If they're the same then it'll render `DatabaseName` as usual but if they're different then it'll render as `Pretty Name (DatabaseName)` because we need the real name to query.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -42,11 +43,23 @@ public KustoClient(string connectionString, string azureAccountToken, string own
...
 
+        private void SetDatabaseName(string databaseName)
+        {
+            var regex = new Regex(@"(?<=\().+?(?=\))");
+            
+            DatabaseName = regex.IsMatch(databaseName)
+                ? regex.Match(databaseName).Value
+                : databaseName;
+        }
+
```
JustinMDotNet, line 62: If the databaseName has `( )` then we want the real database name which is listed in the parenthesis, otherwise we can use the normal database name.

qifahs (approved):

gupta1 (approved):

PR #1126, opened 2020-11-24
Splitting out default DacFx options for Publish and Schema Compare

Benjin:
Consolidating logic that sets default from ADS side to Tools Service: https://github.com/microsoft/azuredatastudio/blob/7bb4d000730f6c24a9cb421f641c97fbda6f35b2/extensions/sql-database-projects/src/dialogs/publishDatabaseDialog.ts#L222

Prep work for (possibly) a more complete consolidation of default logic in DacFx.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
@@ -313,7 +313,7 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
         {
             DeploymentOptions result = new DeploymentOptions();
 
-            result.ExcludeObjectTypes = (ObjectType[])result.ExcludeObjectTypes.Where(x => x != ObjectType.DatabaseScopedCredentials); // re-include database-scoped credentials
+            result.ExcludeObjectTypes = result.ExcludeObjectTypes.Where(x => x != ObjectType.DatabaseScopedCredentials).ToArray(); // re-include database-scoped credentials
 
             return result;
         }
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
@@ -304,12 +304,12 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts
             }
         }
 
-        internal static DeploymentOptions GetDefaultSchemaCompareOptions()
+        public static DeploymentOptions GetDefaultSchemaCompareOptions()
         {
             return new DeploymentOptions();
         }
 
-        internal static DeploymentOptions GetDefaultPublishOptions()
+        public static DeploymentOptions GetDefaultPublishOptions()
         {
             DeploymentOptions result = new DeploymentOptions();
 
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
@@ -2,7 +2,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
+using System;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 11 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -297,5 +303,19 @@ public void SetOptions(DacDeployOptions options)
                 }
             }
         }
+
+        public static DeploymentOptions GetDefaultSchemaCompareOptions()
```
udeeshagautam, line 307: I like this idea of having clear apis, few things to look closely 

1. why are we having deply option != SC option (is it cause its different today in SSDT?)
2. are we having SC option in ADS == SC option in SSDT? I think that should be the case to start with yeah?
3. Depoly option for project vs for dacpac ext - will thhey be same?

4. more on API - We do have an this api for SC - right? and do we plan with this change to use the new api for deploy/extract too?

udeeshagautam (commented):
i am good to sign off but want to test the new api with ADS and ensure no behavior regression before we actually use it

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
```diff
@@ -2,7 +2,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
+using System;
```
udeeshagautam, line 5: nit : do you want to move this file to a common folder of sorts?

Benjin: Yeah, that thought crossed my mind but I don't see one already established.  Think it's worth making a new Common folder?  The other possibility is to put it with the DacFxService folder since SchemaCompareService is built on DacFx so the hierarchy makes a little more sense that way.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
Benjin: This change as is shouldn't change anything so far as differences (or similarities) between ADS SC, ADS Publish, or SSDT.  All it's currently doing is moving* this logic from a previous bugfix from ADS to Tools Service where it makes more sense: https://github.com/microsoft/azuredatastudio/blob/7bb4d000730f6c24a9cb421f641c97fbda6f35b2/extensions/sql-database-projects/src/dialogs/publishDatabaseDialog.ts#L222

* technically it isn't moving it yet, it's duplicating the logic so the ADS side is a no-op, and I'll remove that logic from ADS once we update Tools Service in that branch.

udeeshagautam: makes sense

Benjin: 1. We hit a bug in Publish a while back where DatabaseScopedCredentials needed to be included by default, but weren't because we were using the SchemaCompare defaults.  I _think_ @kisantia said that she'd copied the SC defaults from SSDT, so we shouldn't change, so I added it just for the publish flow.  I think this was an askmode bug, so we did the fix just within the extension where we wouldn't delay the release rather than having to port a new version of Tools Service.

2.  I think so?

3. This change shouldn't affect the Dacpac Publish flow; if it was already different from Project Publish, it'll continue to be.  Having said that, I think they _should_ be the same...  See thoughts below.

4. Which API for SC/Deploy/Extract?  Probably also related to thoughts below.

Thoughts:
I think we should audit what the defaults currently are for the different ways to trigger DacFx ops (SSDT/ADS/SqlPacakge x Schema Compare/Publish) so we can see where the differences are, and decide whether they're intentional or are drift due to having defaults getting set in different places.  Then we can consolidate the logic for determining defaults for different operations and eliminate the drift.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/DeploymentOptions.cs
udeeshagautam: yeah i dot think there is one already,  am okay with how you want arrange this.

udeeshagautam (approved):

alanrenmsft (approved):

Benjin:
All tested through ADS (schema compare, dacpac wizard, project publish)

PR #1138, opened 2020-12-11
Added service for model management for machine learning extension

llali:
Moving the code from ml extension, to query, import, delete, update and download model. 
Added unit tests for the service and integration tests for model operations.


src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelManagementService.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// <summary>
+        /// Handles configure model table request
+        /// </summary>
+        /// <param name="parameters">Request parameters</param>
+        /// <param name="requestContext">Request Context</param>
+        public async Task HandleConfigureModelTableRequest(ConfigureModelTableRequestParams parameters, RequestContext<ConfigureModelTableResponseParams> requestContext)
+        {
+            Logger.Write(TraceEventType.Verbose, "HandleConfigureModelTableRequest");
+            ConfigureModelTableResponseParams response = new ConfigureModelTableResponseParams
+            {
```
Charles-Gagnon, line 201: Why are these on extra lines (and here at all)?

src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelManagementService.cs
```diff
@@ -0,0 +1,247 @@
...
+                    using (IDbConnection dbConnection = ConnectionService.OpenSqlConnection(connInfo))
+                    {
+                        response = operation(dbConnection, parameters, response);
+                        response.Success = true;
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                response.ErrorMessage = e.Message;
```
Charles-Gagnon, line 240: You don't want to SendError instead? That'll make it throw on the extension side which is usually what we want to happen so we don't have constantly be checking the response state. (this goes for all the other places you're setting success to false as well)

src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelOperations.cs
```diff
@@ -0,0 +1,416 @@
...
+		}
+
+		private TResult WithDbChange<T, TResult>(IDbConnection connection, T request, Func<T, TResult> operation) where T : ModelRequestBase
+		{
+			string currentDb = connection.Database;
+			if (connection.Database != request.DatabaseName)
+			{
+				connection.ChangeDatabase(request.DatabaseName);
+			}
+			TResult result = operation(request);
```
Charles-Gagnon, line 221: You should set it back to the original in a finally block - otherwise an error in the operation will leave it in the wrong state

src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelOperations.cs
```diff
@@ -0,0 +1,416 @@
...
+			return $@"
+        {ModelSelectColumns}
+        FROM {GetThreePartsTableName(dbName, tableName, schemaName)}
+		WHERE model_name not like 'MLmodel' and model_name not like 'conda.yaml'
+		ORDER BY model_id";
+		}
+
+		private static string GetConfigTableVerificationQuery(string dbName, string tableName, string schemaName)
+		{
+			string twoPartsTableName = GetTwoPartsTableName(CUtils.EscapeStringSQuote(tableName), CUtils.EscapeStringSQuote(schemaName));
```
Charles-Gagnon, line 273: I don't think escaping the single quote here is correct - you're already bracketing it and so this will just unnecessarily escape single quotes (which would likely break the script)

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelOperations.cs
llali: hmm I thought so to but my test failed when I only escaped with bracket.

src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelManagementService.cs
llali: yeah I put it back to send error

src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelOperations.cs
Charles-Gagnon: What "test"?

What's the final query look like when you have a table with a ' in its name?

llali: this test I added to ModelOperationsTests in this PR.
 [Test]
        public void VerifyImportTableShouldReturnTrueGivenTableNameThatNeedsEscaping()
        {
            bool expected = true;
            bool actual = VerifyModelOperation((dbConnection, databaseName , tableName) =>
            {
                ModelOperations modelOperations = new ModelOperations();
                ModelRequestBase request = new ModelRequestBase
                {
                    DatabaseName = databaseName,
                    SchemaName = "dbo",
                    TableName = tableName
                };
                modelOperations.ConfigureImportTable(dbConnection, request);
                return modelOperations.VerifyImportTable(dbConnection, request);
            }, null, "models[]'");

            Assert.AreEqual(expected, actual);
        }

---
In reply to: [542915398](https://github.com/microsoft/sqltoolsservice/pull/1138#discussion_r542915398) [](ancestors = 542915398)

llali: this line was failing: IF EXISTS (SELECT * FROM syscolumns WHERE ID=OBJECT_ID('[dbo].[models[]]') AND NAME='model_name')

---
In reply to: [542916949](https://github.com/microsoft/sqltoolsservice/pull/1138#discussion_r542916949) [](ancestors = 542916949,542915398)

Charles-Gagnon: Ah, yeah I didn't see that it was also being single quote escaped. Yeah that looks fine then

src/Microsoft.SqlTools.ServiceLayer/ModelManagement/ModelOperations.cs
llali: good point. thanks

---
In reply to: [542822384](https://github.com/microsoft/sqltoolsservice/pull/1138#discussion_r542822384) [](ancestors = 542822384)

PR #1141, opened 2020-12-22
Fix default initialization response for V2 hosting.

nahk-ivanov:
It turns out that `vscode-languageclient` npm package requires workspace capabilities to be present in the reponse, otherwise feature intialization fails, as it tries to blindly access (otherwise undefined) values.

This change addresses the problem by initializing all workspace-related capabilities with their default values in V2 hosting libraries.

### DIFF
--- a/src/Microsoft.SqlTools.DataProtocol.Contracts/ServerCapabilities/ServerCapabilities.cs
+++ b/src/Microsoft.SqlTools.DataProtocol.Contracts/ServerCapabilities/ServerCapabilities.cs
@@ -4,7 +4,10 @@
 //
 
 namespace Microsoft.SqlTools.DataProtocol.Contracts.ServerCapabilities
-{    
+{
+    /// <summary>
+    /// Defines common server capabilities.
+    /// </summary>
     public class ServerCapabilities
     {
         /// <summary>
@@ -107,12 +110,11 @@ namespace Microsoft.SqlTools.DataProtocol.Contracts.ServerCapabilities
         /// </summary>
         /// TODO: Union type
         public bool? TypeDefinitionProvider { get; set; }
-        
+
         /// <summary>
-        /// Options specific to workspaces the server supoorts. Can be <c>null</c> to indicate the
-        /// server does not support workspace requests.
+        /// Options specific to workspaces the server supports.
         /// </summary>
-        public WorkspaceCapabilities Workspace { get; set; }
+        public WorkspaceCapabilities Workspace { get; } = new WorkspaceCapabilities();
         
         /// <summary>
         /// Whether the server provides workpace symbol support. Can be <c>null</c>
--- a/src/Microsoft.SqlTools.DataProtocol.Contracts/ServerCapabilities/Workspace.cs
+++ b/src/Microsoft.SqlTools.DataProtocol.Contracts/ServerCapabilities/Workspace.cs
@@ -9,14 +9,20 @@ using Newtonsoft.Json.Linq;
 
 namespace Microsoft.SqlTools.DataProtocol.Contracts.ServerCapabilities
 {
+    /// <summary>
+    /// Defines workspace capabilities supported by the server.
+    /// </summary>
     public class WorkspaceCapabilities
     {
         /// <summary>
         /// Options specific to workspace folders the server supports
         /// </summary>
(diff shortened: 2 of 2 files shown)


nahk-ivanov:
For reference: https://github.com/microsoft/vscode-languageserver-node/blob/998933dca41b70f85e963f31f225220b54044548/client/src/common/workspaceFolders.ts#L82

We could probably consider it a bug in `vscode` language client package, but I think it might be a required capability for them. We hit this because we decided to reuse their language client infrastructure, even though we are not building language clients with any of our extensions.

@Charles-Gagnon

Charles-Gagnon:
Can you include more information on what problem this is fixing?

And I'm not sure on the exact usage here - but I'm hesitant to completely remove the ability for external callers to set these capabilities. Do we know for sure that nobody is setting these or may want to set them in the future?

nahk-ivanov:
The problem is pretty straight-forward:

image

The line of code in the workspace folders feature that I've linked blindly tries to access the objects hierarchy. It seems like these are default (aka "built-in") features and client will have hard times trying to override these (if we want to do that at all).

I understand your concern, but this change doesn't really "remove the ability for external callers to set these capabilities". They can set them perfectly fine, the only change is that they won't be able to inherit to add additional properties, without defining a custom `InitializeResponse`. I don't think there is a way for us to know anything for sure, so they will just have to deal with it. I was told everything V2 is "an abandoned refactoring attempt and unlikely to be used in the future direction of the SQL Tools Service project", so I'd assume we are on the safer side with any changes here. Given that it is a breaking change - it most likely deserves a version bump under normal circumstances, but in current situation I don't think it matters.

There are few ways how we can deal with it in a less distractive fashion:
1) Keep the setter and throw an exception, if new `value` is null. Still initialize with default.
2) It seems like (I did not confirm in the code) that if you simply don't generate the property (`{}` vs `{"workspace": null}`) - it will skip that code path. So we specify the `NullValueHandling.Ignore` for just this property, but I feel like this would be a weird contract.

I like the current approach as it better matches the client-side code (this property is required to be set) and the check is at compile time, rather than runtime. I'm fine with (2), except for the fact that is a bizarre contract, unless done globally on the serializer, but we will also need to confirm that it is actually working in all scenarios by tracing the code where it bails out on missing property.

Thoughts?

Charles-Gagnon:
How did you run into this issue though? As far as I know we've never seen this before so I'm curious what's making it happen now.

By removing the setter though you're making it so things using this ServerCapabilities class aren't able to set their own WorkspaceCapabilities object. Again, I don't really know this code well enough to say for sure what that'd be used for but we definitely shouldn't be changing functionality like this unless it's really needed.

So yeah I'd prefer that if you go with the default approach that you also keep the setter in case something else is using it. I don't really think we need to change the logic to throw if null is passed in since that wasn't there before. But given what you've found I wouldn't be opposed to it either.

nahk-ivanov:
I just used default `ServiceHost` and let it set [InitializeResponse to default](https://github.com/microsoft/sqltoolsservice/blob/86a4e3f5148c24011f3b955f0b83e72070b19a28/src/Microsoft.SqlTools.Hosting.v2/ServiceHost.cs#L182). **Who** do you exactly mean when saying "we've never seen this before"? This particular codebase does not seem to be using V2. My POC hit this issue on the very first run. From the top of my head, there are two ways how this might have worked in another project that is using V2:
- It sets custom handler for initialization request that uses custom response without workspace property to begin with and my (2) assumption is true.
- It overrides built-in features collection on the client to remove workspace feature.

I could simply update `ServiceHost` to populate `InitializeResponse` with:
```csharp
new InitializeResponse
{
    Capabilities = new ServerCapabilities
    {
        Workspace = new WorkspaceCapabilities
        {
            WorkspaceFolders = new WorkspaceFolderCapabilities(),
        },
    },
};
```
But as I already mentioned - this will not enforce the contract at compile time.

> we definitely shouldn't be changing functionality like this

This appears to be a bug (client-server contract definition is wrong), rather than a change in functionality.
If we confirm that it is indeed the contract (property is expected to not be present or it is set "in full"), then we could use "NullValueHandling.Ignore" on just this property. Do you know who are the stakeholders here? Can you please add them as reviewers?

PR #1142, opened 2020-12-29
Update version of sqlclient in Package props file.

HarshiniCoder:
Updated the Sqlclient version from 2.0.0 to 2.1.1. 
Due to some bug in SqlDataClient version 2.1.0 changed to 2.1.1. 
There are some recent bug fixes in Microsoft.data.sqlclient which we need to port sqltools to ADS
All the ServiceLayer tests passed with the new version 2.1.1.

We need this version to be updated to a version that is 2.1.0 or higher because we need to test ADS on Linux since 2.1.0 is the version that has the required changes. 

image

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -17,7 +17,7 @@
 		<PackageReference Update="Microsoft.Azure.Management.Sql" Version="1.41.0-preview" />
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.1.0"/>
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.1.1"/>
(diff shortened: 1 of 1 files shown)


aasimkhan30 (approved):

alanrenmsft:
@HarshiniCoder , the change was previously reverted by Karl, https://github.com/microsoft/sqltoolsservice/commit/83da69c0634ebca0b08d2283a1b6c42b72b2acca, do you remember why?

Charles-Gagnon (changes requested):
This version (2.1.0) is broken for Macs using kerberos auth - see https://github.com/dotnet/SqlClient/issues/825#issuecomment-749873102

So if we need to update this we should be using 2.1.1.

@HarshiniCoder could you please include more details about what "bug fixes" are included here that you're making this PR for?

Xtrimmer (approved):

Charles-Gagnon (approved):
Ran some basic connection tests on all platforms (including Mac using Kerberos) to verify no regressions with this version and everything looked good so merging this in now. Will follow up with doing some more thorough validate once we have the official build published.

@kburtram @cssuh FYI

PR #1144, opened 2021-01-06
Dev/harpa/sqldataclient

HarshiniCoder:
Updated the Sqlclient version from 2.0.0 to 2.1.1.
Due to some bugs in 2.1.0 changed to 2.1.1
There are some recent bug fixes in Microsoft.data.sqlclient which we need to port sqltools to ADS
All the ServiceLayer tests passed with the new version 2.1.1.


HarshiniCoder:
Duplicate

Issue #1146, opened 2021-01-08
Before and After Batches are ignoring errors


Charles-Gagnon:
Currently we ignore any errors that occur during Before and After batches. One thing this affects is query settings - if an error occurs setting the settings then we just ignore that error and continue on - which means that none of the query settings are actually applied. This can be dangerous if the user is expecting certain settings to be correctly applied.

We should likely be doing what SSMS does and treat that as an error condition (at the very least for the query settings). Possibly with a setting to ignore those errors as well. 

https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs#L435

(the default option for OnErrorAction on the Execute call is to ignore : https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs#L273)

PR #1147, opened 2021-01-08
Remove invalid SET option for SqlOnDemand

Charles-Gagnon:
Fixes https://github.com/microsoft/sqltoolsservice/issues/1145

The issue here is that it appears SqlOnDemand doesn't support the NOEXEC SET option : 

```
Msg 15869, Level 16, State 2, Line 3
NOEXEC is not supported for SET.
```

So the simple fix is just to ignore this setting when connected to an On Demand server. Doing this results in the query settings being applied correctly.

Filed a separate issue about the fact that we ignore errors that occur while setting the query settings. The fix here solves the issue but this other issue should be addressed at some point so we aren't silently ignoring query settings if we run into another similar situation in the future : https://github.com/microsoft/sqltoolsservice/issues/1146

(also cleaned up some unused options in the Kusto code I noticed while investigating this)


Charles-Gagnon:
@BranislavGrbicMDCS Could you confirm that NOEXEC isn't supported by On Demand? Is there a list somewhere of the supported options to make sure we aren't missing any more?

alanrenmsft (approved):

BranislavGrbicMDCS:
> @BranislavGrbicMDCS Could you confirm that NOEXEC isn't supported by On Demand? Is there a list somewhere of the supported options to make sure we aren't missing any more?

I've verified with PMs, NOEXEC is not supported by OnDemand.

BranislavGrbicMDCS (approved):

PR #1155, opened 2021-02-02
Consolidate to one Nuget feed

Charles-Gagnon:
Consolidate nuget.config to one config - this includes putting the checked in packages in that feed. 

Also includes some cleanup that came up while I was fixing this - such as using the latest nuget and dotnet task in the build script.


abist (approved):

azure-pipelines/build.yml
```diff
@@ -18,38 +18,13 @@ steps:
...
-- task: DotNetCoreCLI@1
-  displayName: 'dotnet restore src/Microsoft.Kusto.ServiceLayer'
-  inputs:
-    command: restore
-    projects: '$(Build.SourcesDirectory)/src/Microsoft.Kusto.ServiceLayer'
-    arguments: '--configfile $(Build.SourcesDirectory)/nuget.config'
-
 - task: DotNetCoreCLI@2
-  displayName: 'dotnet build src/Microsoft.SqlTools.ServiceLayer'
+  displayName: 'dotnet restore sqltoolsservice.sln'
```
alanrenmsft, line 22: is .sln a newly supported target for restore and build operation? wondering why we we doing not doing that before

alanrenmsft (approved):

azure-pipelines/build.yml
Charles-Gagnon: No it isn't. No clue why we weren't doing it before - maybe whoever made it originally only cared about building specific projects. But I'd rather we just simplify this since we should be building everything anyways.

PR #1161, opened 2021-02-20
Kusto NoAuth authentication

JustinMDotNet:
Added username and password to DataSourceConnectionDetails. Refactored KustoClient>GetKustoConnectionStringBuilder to accept no username or password for no credentials authentication.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
@@ -74,9 +74,6 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
 
         private KustoConnectionStringBuilder GetKustoConnectionStringBuilder(DataSourceConnectionDetails connectionDetails)
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -74,9 +74,6 @@ private void RefreshAuthToken()
 
         private KustoConnectionStringBuilder GetKustoConnectionStringBuilder(DataSourceConnectionDetails connectionDetails)
         {
-            ValidationUtils.IsTrue<ArgumentException>(!string.IsNullOrWhiteSpace(connectionDetails.UserToken),
```
Charles-Gagnon, line 77: Don't you still want this check if the auth type is `AzureMFA` or `dstsAuth`?

Charles-Gagnon (approved):

Charles-Gagnon:
Looks like you have a test failure though - make sure to fix that before merging in.

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
JustinMDotNet: Moved this into the mapping function.

PR #1168, opened 2021-03-03
Fix authentication error in integration tests pipeline

Charles-Gagnon:
There's a bunch of failures now - https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=95371&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7bfb7e0f-549d-5929-c603-6e8046613b9e

Not sure if these are actually new and people just haven't been fixing them or something else, but I'll be coming in and looking at these separately so that we can get this up and running again.

### DIFF
--- a/azure-pipelines/integrationtests.yml
+++ b/azure-pipelines/integrationtests.yml
@@ -7,17 +7,19 @@ steps:
     displayName: 'Download secure file'
     inputs:
       secureFile: 'testsettings.json'
-  
+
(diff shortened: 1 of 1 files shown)


alanrenmsft (approved):

Charles-Gagnon:
It actually looks like these have been failing for quite a while - we just don't have anything that makes these failures visible currently...

PR #1169, opened 2021-03-04
Fixed false positive for Kusto connections

JustinMDotNet:
Removed catch block around ExecuteQueryAsync to allow exception to bubble up and be returned to ADS.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
@@ -195,20 +195,16 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
         /// Executes a query.
         /// </summary>
         /// <param name="query">The query.</param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="databaseName"></param>
         /// <returns>The results.</returns>
-        public async Task<IEnumerable<T>> ExecuteQueryAsync<T>(string query, CancellationToken cancellationToken, string databaseName = null)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -195,20 +195,16 @@ public async Task ExecuteControlCommandAsync(string command, bool throwOnError,
...
+            string databaseName = null)
         {
-            try
-            {
-                var resultReader = ExecuteQuery(query, cancellationToken, databaseName);
-                var results = KustoDataReaderParser.ParseV1(resultReader, null);
-                var tableReader = results[WellKnownDataSet.PrimaryResult].Single().TableData.CreateDataReader();
-                return await Task.FromResult(new ObjectReader<T>(tableReader));
-            }
-            catch (Exception)
```
JustinMDotNet, line 208: Removed catch block so exceptions will be bubbled up and returned to the ADS client.

qifahs: Do we have retries in the ExecuteQuery path. That will make this code path resilient to intermittent network/service issues.

qifahs (approved):

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
JustinMDotNet: We have a single retry happening inside the ExecuteQuery function.

PR #1177, opened 2021-03-16
bump migration assessment nuget

cssuh:
This will allow me to get the latest changes for the migration assessment api

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -23,7 +23,7 @@
 		<PackageReference Update="Microsoft.Azure.Kusto.Data" Version="9.0.4" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Language" Version="9.0.4"/>
 		<PackageReference Update="Microsoft.SqlServer.Assessment"  Version="[1.0.305]" />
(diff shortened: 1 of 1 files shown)


aasimkhan30 (approved):

aasimkhan30 (changes requested):
Pipeline agents are not able to find this package in the mssqltools feed. You will probably have to add a reference from msdata to mssqltools for this version.

Charles-Gagnon:
Where is this package coming from? As Aasim pointed out you haven't pushed it to the mssqltools feed so the builds will fail until that happens.

But also we can't have packages in that feed anyways since this is an open source project. Can you ask the Migration team if they can get this package published on Nuget.org? Because otherwise we'll need to go back to just checking in the package directly in the repo which isn't great.

Charles-Gagnon:
I'm also curious how you even built this with the package missing from the feed. You should always be building and testing locally before updating dependencies - did you have the package in your local nuget cache or something already?

cssuh:
Yep my mistake, I thought I could just bump the version.

PR #1182, opened 2021-03-26
Fix Sql Auth for Migration Assessments

cssuh:
Fixed SQL Auth Migration Assessments - passing connectionString to DmaEngine instead of SqlObjectLocator

Also verified that it still works for Windows Auth

### DIFF
--- a/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
@@ -56,7 +56,7 @@ SELECT [name], [description], [endpoint], [protocol_desc] FROM .[sys].[dm_cluste
 END TRY
 BEGIN CATCH
 DECLARE @endpoint VARCHAR(max)
-select @endpoint = CONVERT(VARCHAR(max),SERVERPROPERTY('ControllerEndpoint'))
+select @endpoint = ISNULL(CONVERT(VARCHAR(max),SERVERPROPERTY('ControllerEndpoint')), '')
 SELECT 'controller' AS name, 'Cluster Management Service' AS description, @endpoint as endpoint, SUBSTRING(@endpoint, 0, CHARINDEX(':', @endpoint))
 END CATCH
 ";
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
```diff
@@ -56,7 +56,7 @@ @rc 			int
 END TRY
 BEGIN CATCH
 DECLARE @endpoint VARCHAR(max)
-select @endpoint = CONVERT(VARCHAR(max),SERVERPROPERTY('ControllerEndpoint'))
+select @endpoint = ISNULL(CONVERT(VARCHAR(max),SERVERPROPERTY('ControllerEndpoint')), '')
```
Charles-Gagnon, line 59: Did you check whether this is changing whether we consider a server a BDC or not? I think this change might be breaking that so that it thinks servers are BDC's when they actually aren't. 

Easy way to tell is for your test server do you see an HDFS node under OE?

Charles-Gagnon (changes requested):

aasimkhan30 (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -179,9 +178,9 @@ public bool TryGetData(string column, out object value)
             }
         }
 
-        internal async Task<List<MigrationAssessmentInfo>> GetAssessmentItems(SqlObjectLocator target)
+        internal async Task<List<MigrationAssessmentInfo>> GetAssessmentItems(SqlObjectLocator target, string connectionString = "")
```
Charles-Gagnon, line 181: Is an empty string a valid value for DmaEngine?

aasimkhan30:
@cssuh, please work with Charles on the BDC issue. Thanks!

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
cssuh: Screen Shot 2021-03-25 at 5 23 13 PM
This is what I see under OE, is this correct?

Charles-Gagnon: You have to expand the server node. You should NOT be seeing the HDFS node for your servers

image

cssuh: yep, there is no HDFS node.

cssuh: HDFS

kburtram (approved):
Looks fine to me.  Please get @Charles-Gagnon to sign-off or remove changes requested prior to merging.

Charles-Gagnon (approved):

PR #1187, opened 2021-04-13
Default Kusto database name

JustinMDotNet:
Added check to set the databaseName to the first database on the cluster when no database is provided

This PR resolves this issue:
https://github.com/microsoft/azuredatastudio/issues/15102

ADO Related Bug
https://mssqltools.visualstudio.com/ADS%20LiveSite%20Support/_workitems/edit/5580

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
                 return;
             }
             
-            var dataReader = ExecuteQuery(".show databases schema | order by DatabaseName asc | take 1", new CancellationToken());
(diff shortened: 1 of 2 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/KustoClient.cs
```diff
@@ -53,6 +53,15 @@ private void Initialize(DataSourceConnectionDetails connectionDetails)
...
+
+            if (DatabaseName != "NetDefaultDB")
+            {
+                return;
+            }
+            
+            var dataReader = ExecuteQuery(".show databases | top 1 by DatabaseName | project DatabaseName", new CancellationToken());
+            var databaseName = dataReader.ToEnumerable().Select(row => row["DatabaseName"]).FirstOrDefault();
+            DatabaseName = databaseName?.ToString() ?? "";
         }
```
qifahs, line 65: Please check that this query works for Aria clusters too.

PR #1188, opened 2021-04-14
SchemaCompare Unknown errors display fix with DacFx errors

ssreerama:
1. Schema Comparing failures doesn't display error messages, instead popping up unknown error notification banner.
2. Have added code to read the error message from DacFx.
3. Issue was logged in ADS Repo: https://github.com/microsoft/azuredatastudio/issues/14107

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
@@ -10,6 +10,7 @@ using Microsoft.SqlTools.Utility;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 
 namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
@@ -119,14 +120,13 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                     }
                 }
 
-                // HashSet contains the set of errors that are stopping the schema compare
-                // Error Message support single message only, so breaking the functionality with single error each time
-                HashSet<string> errorsSet = ComparisonResult.GetAllErrorsList();
-                foreach (string errMsg in errorsSet)
+                // Appending the set of errors that are stopping the schema compare to the ErrorMessage 
+                var errorsList = ComparisonResult.GetErrors().Select(e => e.Message).Distinct().ToList();
+                foreach (string errMsg in errorsList)
                 {
-                    ErrorMessage = errMsg;
-                    break;
+                    ErrorMessage += errMsg + "; ";
                 }
+                ErrorMessage = ErrorMessage.Trim().TrimEnd(';');
             }
             catch (Exception e)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -118,6 +118,15 @@ public void Execute(TaskExecutionMode mode)
...
+
+                // HashSet contains the set of errors that are stopping the schema compare
+                // Error Message support single message only, so breaking the functionality with single error each time
+                HashSet<string> errorsSet = ComparisonResult.GetAllErrorsList();
+                foreach (string errMsg in errorsSet)
+                {
+                    ErrorMessage = errMsg;
+                    break;
+                }
             }
```
udeeshagautam, line 130: Since GetAllErrorsList() function was added in recent DacFx please do not check in in without DacFx update

udeeshagautam (changes requested):
Lets since on Dacfx update

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
udeeshagautam: We can - but it scripts correctly and is allowed with space too.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -118,6 +118,15 @@ public void Execute(TaskExecutionMode mode)
                         this.Differences.Add(diffEntry);
                     }
                 }
+
+                // HashSet contains the set of errors that are stopping the schema compare
+                // Error Message support single message only, so breaking the functionality with single error each time
```
kisantia, line 123: could you append the error messages so not just the last error message will be returned?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
ssreerama: sure, I would definitely wait for DacFx merge first!

udeeshagautam: so GetErrors is already present?

ssreerama: Yes, GetErrors() is already present in DacFx!

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -118,6 +119,14 @@ public void Execute(TaskExecutionMode mode)
                         this.Differences.Add(diffEntry);
                     }
                 }
+
+                // Appending the set of errors that are stopping the schema compare to the ErrorMessage 
+                var errorsList = ComparisonResult.GetErrors().Select(e => e.Message).Distinct().ToList();
+                foreach (string errMsg in errorsList)
```
kisantia, line 125: should be able to use string.Join() instead of foreach

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -118,6 +119,14 @@ public void Execute(TaskExecutionMode mode)
...
+                {
+                    ErrorMessage += errMsg + "; ";
```
kisantia, line 127: newline might be easier to read than ; if there are multiple error messages

udeeshagautam:
@ssreerama can you please run SqltoolsService all tests **for schema compare** and validate they pass.

udeeshagautam (approved):
looks good... just run the tests for Sc in sqltoolsservice and should be good

kisantia (approved):

PR #1189, opened 2021-04-14
Escape file URIs

Charles-Gagnon:
This is for https://github.com/microsoft/azuredatastudio/issues/10259 and https://github.com/microsoft/azuredatastudio/issues/14044

Taking the changes in those PRs isn't enough to fix the issue - while it does make it so that the URI comes over with spaces encoded (e..g `My%20File` instead of `My File`) the document are registered with fully-escaped URIs. So this means that a file with a path like `C:\my\path\with%20space\query.sql` that has an actual % in its path will be cached as something like `file:\\c:\my\path\with%2520space\query.sql` and so we'll fail to find the file in certain cases.

Unfortunately this functionality isn't broken for everything - some URIs come in as fully-escaped file URIs (such as the intellisense completion requests) but some come in as the owner URIs (which often aren't escaped). This means we can't modify GetFile directly since it doesn't know which type of URI it is - they'll both look the same except for the level of escaping done.

I'm open to suggestions here since this is clearly not ideal - in particular I'm trying to decide if this is the correct place to do this kind of change or whether I should try to keep it on the ADS side. But doing it on the ADS side would mean changing the URIs passed in to *any* provider which would probably break any other customers who have their own provider services unless they also made changes on their end. 

I'm still doing local testing but wanted to get this out early to get any feedback - currently don't see any issues with either Notebooks or SQL files.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
@@ -18,11 +18,13 @@ using Microsoft.SqlTools.ServiceLayer.Test.Common;
 using Microsoft.SqlTools.ServiceLayer.Test.Common.RequestContextMocking;
 using Microsoft.SqlTools.ServiceLayer.UnitTests.Utility;
 using Microsoft.SqlTools.ServiceLayer.Workspace;
-using Moq;
 using NUnit.Framework;
 
 namespace Microsoft.SqlTools.ServiceLayer.UnitTests.QueryExecution.Execution
 {
+
+    public class InvalidParams : ExecuteRequestParamsBase { }
+
     public class ServiceIntegrationTests
     {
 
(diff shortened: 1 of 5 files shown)


test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/Execution/ServiceIntegrationTests.cs
```diff
@@ -131,12 +133,12 @@ public void GetSqlTextFromInvalidType()
             // ... Mock up an implementation of ExecuteRequestParamsBase
             // ... Create a query execution service without a connection service or workspace
             //     service (we won't execute code that uses either
-            var mockParams = new Mock<ExecuteRequestParamsBase>().Object;
+            var invalidParams = new InvalidParams() { OwnerUri = "" };
```
Charles-Gagnon, line 136: This was failing because the mocked object had a null OwnerUri, but given that we assume OwnerUri is a valid string in many other places in the code already it seemed reasonable to just set it to empty here instead of updating the function to handle a null Uri.

alanrenmsft (approved):

kburtram (approved):

PR #1192, opened 2021-04-16
Move unused forked code to external directory

kburtram:
This PR moves the unused "V2" fork code out of the src directory to external as a first step in removing this code from this repo.  This is the initial step in refactoring the codebase without the added cluster of the deprecated codebase.


Charles-Gagnon (approved):

Charles-Gagnon:
You need to remove those from the sln as well

aasimkhan30:
The build is failing. You need to update the solution file

kburtram:
Yep, using the CI builds to catch build scenarios I don't typically do locally (I typically run `dotnet` commands directly on desired csproj).  Next step after moving these projects and getting only what we currently used building is to go through the forked code and see if anything should be integrated back into the main assemblies.  After that there are some refactoring needed to make it more easy to add new C# data providers, such as isolated contracts and perhaps adding interfaces to various services and simplifying setting up handlers, basically to remove some of the trivial boilerplate in create new providers.

PR #1194, opened 2021-04-22
update the JSON-RPC API document:

bugaosuni59:
1.add some connection and query API
	2.add some use case for some useful API


Charles-Gagnon:
This looks excellent, thank you for doing this!

PR #1196, opened 2021-04-27
Bump SMO and SqlClient driver

kburtram:
Update SMO and MS.Data.SqlClient to latest public packages on Nuget.  I want these updates to support on-going DW work.

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -17,8 +17,8 @@
 		<PackageReference Update="Microsoft.Azure.Management.Sql" Version="1.41.0-preview" />
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.0.0"/>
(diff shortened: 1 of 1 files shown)


kburtram:
@Charles-Gagnon I like the GitHub action you added with validation steps!  I'll go through those steps.  Any other concerns with bumping to these versions of SMO & SqlClient?

Charles-Gagnon:
Nope, in fact I was planning on updating the SMO version once the new version was released anyways so thanks for doing that for me! 😄 

If you could just document here what testing you've done so we know what for sure works that will help with verification later on (and then this next release definitely make sure to do a much more robust test pass of connection scenarios).

Charles-Gagnon:
I did ping David and Matteo about a thing they were planning on doing regarding taking the SqlParser dependency out of SMO...but it looks like that hasn't been finished yet. But let's make sure to verify with them in case something was missed.

Charles-Gagnon (approved):

kburtram:
@Charles-Gagnon I've run through basic testing with this bump.  I think the most efficient approach here is to merge this and pull it into ADS, then run further validation on that build.  Otherwise, I'll need to produce private bundles for each platform, etc.  I'll update this PR with results of further testing and if serious issues are found we can revert.

PR #1197, opened 2021-04-27
Fix OE with dbpool connection

kburtram:
This PR handles some special-case code paths to avoid exceptions with dbpool  connections.  Some of the issues may be upstream in SMO, so it may be better to resolve the issues there rather than workaround them here.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -8,6 +8,7 @@ using System;
 using System.Diagnostics;
 using System.Globalization;
 using System.Threading;
+using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlServer.Management.Smo;
 using Microsoft.SqlTools.Utility;
 
@@ -25,7 +26,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             // the user connected directly to the master of a readable secondary
             // In that case, the name in the connection string won't be found in sys.databases
             // We detect that here and fall back to master
-            if (db.State == SqlSmoState.Creating)
+            if (db.State == SqlSmoState.Creating && !IsDWGen3(db))
             {
                 db = new Database(serverNode.GetContextAs<SmoQueryContext>().Server, "master");
                 db.Refresh();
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ValidForFlag.cs
```diff
@@ -21,6 +21,7 @@ public enum ValidForFlag
         Sql2017 = 0x40,
         SqlDw = 0x80,
         SqlOnDemand = 0x100,
+        AzureSqlDWGen3 = 0x200,
```
Charles-Gagnon, line 24: You don't want to add this to AllAzure or All?

Although it looks like you're not using this at all - you just use the SqlDw flag instead. Is that intentional?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SqlServerType.cs
```diff
@@ -90,7 +91,14 @@ public static SqlServerType CalculateServerType(ServerInfo serverInfo)
             }
             else if (serverInfo.IsCloud)
             {
-                serverType = SqlServerType.AzureV12;
+                if (serverInfo.EngineEditionId == 6 && serverVersion.StartsWith("12", StringComparison.Ordinal))
```
Charles-Gagnon, line 94: Should use `DatabaseEngineEdition.SqlDataWarehouse` instead of hardcoded 6 here

Charles-Gagnon, line 94: Oh that's fun...we only know about versions up to 2017 in here? 🤦‍♂️ (not related to your PR)

Charles-Gagnon (approved):
Make sure to sync up with @shueybubbles on this, we should definitely be trying to fix this in SMO if we can.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -75,12 +76,28 @@ public bool IsAccessible(SmoQueryContext context)
...
+                    ErrorMessage = ex.Message;
+                    return false;
+                }
+                
             }
         }
+
+        private bool IsDWGen3(Database db)
+        {
+            return db != null
```
shueybubbles, line 98: the null check seems superfluous

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -75,12 +76,28 @@ public bool IsAccessible(SmoQueryContext context)
             }
             catch (Exception ex)
             {
-                var error = string.Format(CultureInfo.InvariantCulture, "Failed to get IsAccessible. error:{0} inner:{1} stacktrace:{2}",
-                    ex.Message, ex.InnerException != null ? ex.InnerException.Message : "", ex.StackTrace);
-                Logger.Write(TraceEventType.Error, error);
-                ErrorMessage = ex.Message;
-                return false;
+                // IsAccessible is not set of DW Gen3 so exception is expected in this case
```
shueybubbles, line 79: IsAccessible doesn't really make any sense in Azure. AFAICT if the current connection is to master, has_dbaccess('user database name') always returns null regardless of whether 'user database name' refers to a DW or a DB. 

If you are connecting to the user database  directly, the fact that SqlConnection.Open succeeds means you have access to the database, so why are we bothering to check this property?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ValidForFlag.cs
kburtram: @Charles-Gagnon I'll try adding to AllAzure and All, it makes sense to be there.  I reuse the SqlDW flag since that seemed to have the correct set of SMO properties that worked with the test server I have.  The other flags were trying to read properties that hadn't been set.

shueybubbles: What's the difference between SqlDw and AzureSqlDWGen3?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
kburtram: @shueybubbles I guess with dbpool connections "use master" doesn't work, so the connection will be to the user DB.  I'll see if there is a more generic way to handle this exception that takes into account what you said above.  We don't see this exception connecting to Azure DB normally, so either the property is set and doesn't mean anything or we don't go down this code path.  I'll debug and see what's up.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ValidForFlag.cs
kburtram: @shueybubbles Gen3 supports DB pools and SqlDw does not.  I'll see if there is a way to get rid of this flag and only use SqlDw but there were some issues when I tried that previously while debugging.

PR #1198, opened 2021-04-27
DRAFT - Azure monitor extension

JustinMDotNet:


Charles-Gagnon:
Does this need to be in this repo? It would be nice to have external components in their own repos instead of just continuing to put them in here. Much less risk of different projects breaking other projects in the repo (especially when using lots of shared Nuget packages).

Charles-Gagnon:
And if this is a draft you can just use the Draft PR functionality as well so it doesn't show up as an actual PR until it's ready to review.

JustinMDotNet:
Closing to reopen as draft.

Charles-Gagnon:
FYI you can convert an existing PR to a draft (and vice versa)

image

PR #1199, opened 2021-04-28
Fixing a notebook history bug in SQL Agent

aasimkhan30:
When we join notebook histories to agent job histories, the key used for joining in notebooks is made from concatenated rundate and runtime and for agent jobs it is a DateTime object. 

The problem arises when the runtime of the notebook is before 10am. In this case there is a loss of leading zero in the runtime of notebooks as it is stored as int in the sql table.

Example: A notebook history with a run date of 20210428 and runtime of 9:33:00 am will have a key like '2021042893300' whereas the agent job will have a runtime of 20210428093300 (The extra leading zero is before 9). 
This was causing a problem in the 'JOIN' operation I was doing in the edited method. 
 
This PR fixes https://github.com/microsoft/azuredatastudio/issues/15200

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
@@ -1551,7 +1551,9 @@ namespace Microsoft.SqlTools.ServiceLayer.Agent
                 DataTable materializedNotebookTable = await AgentNotebookHelper.GetAgentNotebookHistories(connInfo, jobId, targetDatabase);
                 foreach (DataRow materializedNotebookRow in materializedNotebookTable.Rows)
                 {
-                    string materializedRunDateTime = materializedNotebookRow["run_date"].ToString() + materializedNotebookRow["run_time"].ToString();
+                    // Adding a leading zero if the for run times less than 10am. Example: 9:10:00 will become 09:10:00 
+                    string notebookRuntime = (materializedNotebookRow["run_time"].ToString().Length == 5) ? "0" + materializedNotebookRow["run_time"].ToString() : materializedNotebookRow["run_time"].ToString();
+                    string materializedRunDateTime = materializedNotebookRow["run_date"].ToString() + notebookRuntime;
                     notebookHistoriesDict.Add(materializedRunDateTime, materializedNotebookRow);
                 }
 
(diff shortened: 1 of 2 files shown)


abist:
Any reason why we can't use the DateTime object in Notebooks jobs too? This seems like a hack and might show up in other cases too

aasimkhan30:
Unfortunately, that will need some modification of the sql table where the notebook histories are stored. That will be a much bigger change won't be easily back compatible. This is the only place where we use this type of 'JOIN' and I don't see this bug affecting other areas

aasimkhan30:
Reworking the fix based on a blog article shared by @abist

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1551,7 +1551,9 @@ string targetDatabase
                 DataTable materializedNotebookTable = await AgentNotebookHelper.GetAgentNotebookHistories(connInfo, jobId, targetDatabase);
                 foreach (DataRow materializedNotebookRow in materializedNotebookTable.Rows)
                 {
-                    string materializedRunDateTime = materializedNotebookRow["run_date"].ToString() + materializedNotebookRow["run_time"].ToString();
+                    // Adding a leading zero if the run time is before than 10am. Example: 9:10:00 will become 09:10:00 
+                    string notebookRuntime = (materializedNotebookRow["run_time"].ToString().Length == 5) ? "0" + materializedNotebookRow["run_time"].ToString() : materializedNotebookRow["run_time"].ToString();
```
Charles-Gagnon, line 1555: Your example doesn't match the logic here....`9:10:00` has 7 characters so wouldn't add the leading zero. Does it actually come in as `9:10`?

Charles-Gagnon (approved):

abist (approved):

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
aasimkhan30: Sorry, if I was not clear about it but it returned as 91000 and I am converting it to 091000.

aasimkhan30:
@Charles-Gagnon and @abist added a better fix.

src/Microsoft.SqlTools.ServiceLayer/Agent/Jobs/AgentNotebookHelper.cs
```diff
@@ -671,6 +670,7 @@ UPDATE notebooks.nb_materialized
                 notebook = '',
                 notebook_error = ''
                 WHERE 
+                materialized
```
Charles-Gagnon, line 673: This syntax doesn't seem correct - is this missing something?

abist-test (approved):

abist (approved):

abist:
Please ignore the review from the test account, I was trying to automate something

PR #1200, opened 2021-04-29
DRAFT - Azure Monitor Extension

JustinMDotNet:


kburtram (changes requested):
I see this is a draft, but please separate out the changes to create a contracts assembly from the azure monitor code.  the contracts assembly should be shared with sqltoolsservice so should be review on its own.

PR #1201, opened 2021-04-30
Kusto Special Character Database Fix

JustinMDotNet:
Removed Parallel.Invoke from KustoIntellisenseClient.cs > LoadSchema. Added EscapeName from databaseName to LoadSchema and GetTableInfos

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/Intellisense/KustoIntellisenseClient.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/Intellisense/KustoIntellisenseClient.cs
@@ -42,10 +42,15 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource.Intellisense
             if (!string.IsNullOrWhiteSpace(databaseName))
             {
                 var source = new CancellationTokenSource();
-                var tableQuery = $".show database {KustoQueryUtils.EscapeName(databaseName)} schema";
-                
-                tableSchemas = _kustoClient.ExecuteQueryAsync<ShowDatabaseSchemaResult>(tableQuery, source.Token, databaseName).Result;
-                functionSchemas = _kustoClient.ExecuteQueryAsync<ShowFunctionsResult>(".show functions", source.Token, databaseName).Result;
+                Parallel.Invoke(() =>
+                    {
+                        var tableQuery = $".show database {KustoQueryUtils.EscapeName(databaseName)} schema";
+                        tableSchemas = _kustoClient.ExecuteQueryAsync<ShowDatabaseSchemaResult>(tableQuery, source.Token, databaseName).Result;
+                    },
+                    () =>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/Intellisense/KustoIntellisenseClient.cs
```diff
@@ -42,16 +42,10 @@ private GlobalState LoadSchemaState(string databaseName, string clusterName)
...
-                                .Result;
-                    },
-                    () =>
-                    {
-                        functionSchemas = _kustoClient.ExecuteQueryAsync<ShowFunctionsResult>(".show functions", source.Token, databaseName).Result;
-                    });
+                var tableQuery = $".show database {KustoQueryUtils.EscapeName(databaseName)} schema";
+                
+                tableSchemas = _kustoClient.ExecuteQueryAsync<ShowDatabaseSchemaResult>(tableQuery, source.Token, databaseName).Result;
+                functionSchemas = _kustoClient.ExecuteQueryAsync<ShowFunctionsResult>(".show functions", source.Token, databaseName).Result;
```
JustinMDotNet, line 48: Removed Parallel.Invoke since both threads are acting against ExecuteQueryAsync which is a single instance and is still being executed with the thread blocked.

src/Microsoft.Kusto.ServiceLayer/DataSource/Intellisense/KustoIntellisenseClient.cs
```diff
@@ -332,7 +326,7 @@ public LanguageServices.Contracts.CompletionItem[] GetAutoCompleteSuggestions(Sc
             foreach (var autoCompleteItem in completion.Items)
             {
                 var label = autoCompleteItem.DisplayText;
-                var insertText = autoCompleteItem.Kind == CompletionKind.Table
+                var insertText = autoCompleteItem.Kind == CompletionKind.Table || autoCompleteItem.Kind == CompletionKind.Database
```
JustinMDotNet, line 329: Added Database here to escape databases in autocomplete when they have special characters in the name.

src/Microsoft.Kusto.ServiceLayer/DataSource/Intellisense/KustoIntellisenseClient.cs
qifahs: Even though the ICslQueryProvider.ExecuteQuery is synchronous, it might be issuing the queries in parallel underneath. We have another parallel execution in the KustoClient class, and I remember that made a difference when we issued multiple parallel queries. Better to be parallel on our side for perf reasons, unless we are calling code which is not thread safe.

qifahs (changes requested):

qifahs (approved):

PR #1202, opened 2021-05-03
Adding windows account validation api for sql migration extensions. 

aasimkhan30:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/ValidateWindowsAccount.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/ValidateWindowsAccount.cs
@@ -21,6 +21,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
 
     public class ValidateWindowsAccountRequest
     {
-        public static readonly RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult> Type = RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult>.Create("migration/validatewindowsaccount");
+        public static readonly RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult> Type = RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult>.Create("migration/validateWindowsAccount");
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/ValidateWindowsAccount.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/ValidateWindowsAccount.cs
@@ -1,0 +1,26 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
+{
+    public class ValidateWindowsAccountRequestParams
+    {
+        public string Username { get; set; }
+        public string Password { get; set; }
+    }
+
+    public class ValidateWindowsAccountResult
+    {
+        public bool Success { get; set; }
+        public string ErrorMessage { get; set; }
+    }
+
+    public class ValidateWindowsAccountRequest
+    {
+        public static readonly RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult> Type = RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult>.Create("migration/validatewindowsaccount");
+    }
+}
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
@@ -11,7 +11,6 @@ using Microsoft.SqlServer.Management.Assessment;
 using Microsoft.SqlServer.Management.Assessment.Checks;
 using Microsoft.SqlServer.Migration.Assessment.Common.Contracts.Models;
 using Microsoft.SqlServer.Migration.Assessment.Common.Engine;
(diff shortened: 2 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/ValidateWindowsAccount.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public class ValidateWindowsAccountResult
+    {
+        public bool Success { get; set; }
+        public string ErrorMessage { get; set; }
+    }
+
+    public class ValidateWindowsAccountRequest
+    {
+        public static readonly RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult> Type = RequestType<ValidateWindowsAccountRequestParams, ValidateWindowsAccountResult>.Create("migration/validatewindowsaccount");
```
Charles-Gagnon, line 24: Message names should be camelCased (validateWindowsAccount)

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -248,5 +251,47 @@ public void Dispose()
...
+        internal async Task HandleValidateWindowsAccountRequest(
+            ValidateWindowsAccountRequestParams parameters,
+            RequestContext<ValidateWindowsAccountResult> requestContext)
+        {
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                int separator = parameters.Username.IndexOf("\\");
+                string domainName = (separator > -1) ? parameters.Username.Substring(0, separator) : string.Empty;
+                string userName = (separator > -1) ? parameters.Username.Substring(separator + 1, parameters.Username.Length - separator - 1) : string.Empty;
```
Charles-Gagnon, line 268: If the username doesn't contain a `\\` then you should probably return an error that specifically says that. Right now you'll call the method with empty values which is unnecessary and the error returned would probably be misleading. 

Are you validating the username on the client side is in the form `domain\\username` as well? Would be nice to save some time.

Charles-Gagnon (approved):
Is there a valid scenario where the machine the user is running this from isn't able to logon via the specified user but the machine that's hosting the migration process thing would? For example if they're on different domains.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
aasimkhan30: The users are not allowed to enter the usernames. It is generated by code on the client side and is not editable.  (Third inputbox in the screenshot).

Charles-Gagnon: Does that code ensure that there is always a valid domain/username?

Regardless I think you should either : 

1. Remove the check and just require/assume that the value passed in is valid
2. Keep the check but add the return error message if it's not valid

I think #2 would be best in case you end up using this elsewhere or start allowing people to enter in their own credentials

aasimkhan30:
@Charles-Gagnon. ADS is getting the username from current integrated auth connection. I am assuming they should be able to logon using the username from the connection. 

Let me know if there is a corner case I am missing here.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
Charles-Gagnon: Also - are you just not going to support users who want to do the migration using a different account than the one they're using ADS with? Seems like a pretty common scenario - especially in corporate environments.

aasimkhan30: It was an ask from the PM to force the username from the connection

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -248,5 +252,57 @@ public void Dispose()
...
+
+        [DllImport("advapi32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+        public static extern bool LogonUser(String lpszUsername, String lpszDomain, String lpszPassword,
+       int dwLogonType, int dwLogonProvider, out SafeAccessTokenHandle phToken);
+
+        internal async Task HandleValidateWindowsAccountRequest(
+            ValidateWindowsAccountRequestParams parameters,
+            RequestContext<ValidateWindowsAccountResult> requestContext)
+        {
+            var domainUserRegex = new Regex(@"^[A-Za-z0-9\\\._-]{7,}$");
```
Charles-Gagnon, line 264: Why min 7 characters?

Something like this seems better : 

`^(?<domain>[A-Za-z0-9\._-]*)\\(?<username>[A-Za-z0-9\._-]*)$`

(and then you can use the groups from the result instead of substring below for the domain/username.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -248,5 +252,57 @@ public void Dispose()
...
+            // Checking if the username string is in 'domain\name' format
+            if (!domainUserRegex.Match(parameters.Username).Success)
+            {
+                await requestContext.SendResult(new ValidateWindowsAccountResult()
```
Charles-Gagnon, line 268: SendError is generally preferred - which will cause an exception to be thrown with the error message

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/ValidateWindowsAccount.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class ValidateWindowsAccountRequestParams
+    {
+        public string Username { get; set; }
+        public string Password { get; set; }
+    }
+
+    public class ValidateWindowsAccountResult
+    {
+        public bool Success { get; set; }
```
Charles-Gagnon, line 18: As mentioned below - we have SendError for if the request wasn't able to be completed. So generally I would say that these aren't necessary - if the request completes without throwing then it means it passed.

PR #1203, opened 2021-05-04
Adding network share validation in sql migration.

aasimkhan30:


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -259,17 +261,13 @@ public void Dispose()
 
         internal async Task HandleValidateWindowsAccountRequest(
             ValidateWindowsAccountRequestParams parameters,
-            RequestContext<ValidateWindowsAccountResult> requestContext)
+            RequestContext<bool> requestContext)
         {
             var domainUserRegex = new Regex(@"^[A-Za-z0-9\\\._-]{7,}$");
```
Charles-Gagnon, line 266: This should use the ValidateWindowsDomainUsername method right? And that would probably be best to return the result so you can get the username/domain from the match results directly instead of parsing the string below.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -290,19 +288,98 @@ await requestContext.SendResult(new ValidateWindowsAccountResult()
...
+            if (!ValidateUNCPath(parameters.Username))
+            {
+                await requestContext.SendError("Invalid network share path. Example: \\\\Servername.domainname.com\\Backupfolder");
+                return;
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                int separator = parameters.Username.IndexOf("\\");
+                string domainName = (separator > -1) ? parameters.Username.Substring(0, separator) : string.Empty;
```
Charles-Gagnon, line 323: You've already validated that the username is in this format so these checks are unnecessary.

And anyways - as above if you return the match results from the ValidateWindowsDomainUsername method then you can just use that instead of substringing it.

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
aasimkhan30: Sorry these are for 2 different screens. 

1. We are using the validate windows username for this screen. (To get the password for the integrated login username).
image

2. File share validation is done in this page. We verify whether the ADS client is able to connect to the remote file share using the information provided in this page. It can be the same account from 1 but there is a high chance that the user can setup a file share on remote machine). 
image

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -93,6 +94,7 @@ public void InitializeService(ServiceHost serviceHost)
             this.ServiceHost = serviceHost;
             this.ServiceHost.SetRequestHandler(MigrationAssessmentsRequest.Type, HandleMigrationAssessmentsRequest);
             this.ServiceHost.SetRequestHandler(ValidateWindowsAccountRequest.Type, HandleValidateWindowsAccountRequest);
+            this.ServiceHost.SetRequestHandler(ValidateNetworkFileShareRequest.Type, HandleValidateNetworkFileShareRequest);
```
caohai, line 97: Not sure if I'm missing anything here but I'm a bit confused that why do we need both `ValidateWindowsAccountRequest` and `ValidateNetworkFileShareRequest` for network share validation. I assume as long as the `{UNC, username, password}` can access the network share, we should be good, is this correct?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -290,19 +286,98 @@ await requestContext.SendResult(new ValidateWindowsAccountResult()
...
+            ValidateNetworkFileShareRequestParams parameters,
+            RequestContext<bool> requestContext)
+        {
+            if (!ValidateWindowsDomainUsername(parameters.Username))
+            {
+                await requestContext.SendError("Invalid user name format. Example: Domain\\username");
+                return;
+            }
+
+            if (!ValidateUNCPath(parameters.Username))
```
caohai, line 312: `parameter.Path`? 
Not sure if it would make more sense to update `Path` field in param def to something like `UNCPath`, just my 2 cents

caohai:
😄 It's great to know that the validation can be done on service side. Are you able to run an end-to-end test for this network share through the validation api locally?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
aasimkhan30: These validations are used in 2 separate places in the extension. 

Validate windows account is used in this screen. We are trying to get the windows account password for the Integrated sql server connection used in migration. This has nothing do with File share. 
image

Validate network file share is used in the screen where the user provides an SMB file share that contains the sql server backup files 
image

caohai: This makes sense now. It might be better to leave the changes related to windows auth to a separate PR but no biggie.

aasimkhan30:
> 😄 It's great to know that the validation can be done on service side. Are you able to run an end-to-end test for network share validation through this service side validation Api locally?

Yeah, I have tried it locally and it seems to work. There can be some false negatives in this API (A situation where ADS is not able to access file share however, the backend is able to access it).  However, for most users of the extension it will work as expected.

caohai (approved):

PR #1205, opened 2021-05-05
update DacFx version

kisantia:


udeeshagautam:
Tagging here so we dont forget - we need to check following passes after this update
https://github.com/microsoft/azuredatastudio/issues/14655

udeeshagautam (approved):

kburtram (approved):

Issue #1206, opened 2021-05-06
review all the request handlers and make sure nothing is running directly on the main thread.


alanrenmsft:

alanrenmsft:
https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs#L151 is blocking

Charles-Gagnon:
So @alanrenmsft the concern I had with just returning TaskCompleted for that instance was I just want us to make sure that we're ok with having each message handler determine whether they're blocking or non-blocking and to need to do that by spinning off a separate thread and then just returning immediately. Right now all messages are blocking which I'm not sure is really made clear anywhere so mistakes like this can happen. Might be worth looking at a couple of other options : 

1. Make messages non-blocking by default. I'm not even sure why message handlers are blocking in the first place - are there any scenarios that we need that which isn't just solved by the fact that on the calling side it needs to wait for a response to be sent anyways
2. Have some sort of flag to the set handler method that tells it whether to be blocking or not - just to make it more clear

PR #1207, opened 2021-05-06
use new thread to handle requests

alanrenmsft:
this PR fixes: https://github.com/microsoft/azuredatastudio/issues/15262

I observed that the getQueryRows is taking longer than normal when switching query editors and after debugging in the underlying vscodelanguageclient package, I noticed a request to list database names is actually blocking the getQueryRows request. the code that triggers the list database name request is introduced by: https://github.com/microsoft/azuredatastudio/pull/14946/files#diff-8cef5d0d3d162f3a0bdef0c58b4e865dedb986f676fb3cbb9714b88dc37f904bR810

root cause: the listDatabaseName request is being handled directly on the main thread which will block the handling of other requests.

fix: use a new thread to handle the request.

I also created a task to review all the request handlers and make sure we don't have any other handlers like this.: https://github.com/microsoft/sqltoolsservice/issues/1206

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -427,8 +427,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                 if (!ConnectionService.IsDbPool(connectionInfo.ConnectionDetails.DatabaseName))
                 {
                     connectionInfo.ConnectionDetails.DatabaseName = connection.Database;
-                }                
-                
+                }
+
                 if (!string.IsNullOrEmpty(connectionInfo.ConnectionDetails.ConnectionString))
                 {
                     // If the connection was set up with a connection string, use the connection string to get the details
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1048,21 +1048,24 @@ private void RunConnectRequestHandlerTask(ConnectParams connectParams)
...
+                {
+                    ListDatabasesResponse result = ListDatabases(listDatabasesParams);
+                    await requestContext.SendResult(result);
+                }
+                catch (Exception ex)
+                {
+                    await requestContext.SendError(ex.ToString());
+                }
+            });
+            return Task.CompletedTask;
```
Charles-Gagnon, line 1068: Why are you returning Task.CompletedTask instead of the return value of Task.Run?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
alanrenmsft: If I return the task.run, it will still block the main thread.

image
https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs#L304

 I haven't reviewed all the request handlers, there might be request types that we actually want to block the main thread, maybe that is why it is designed to return a task.

alanrenmsft: for example: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs#L151
looks like it is on a different thread, but it is actually blocking during my testing.

alanrenmsft: as mentioned in the pr description, I've created a task to review all the handlers after may release.

Charles-Gagnon:
Also you didn't fix the Kusto version of this method either. @JustinMDotNet @qifahs FYI

Code-DJ:
Could this be the reason for all the random freezing issues like https://github.com/microsoft/azuredatastudio/issues/15230?

PR #1209, opened 2021-05-11
Update Assessment nuget

cssuh:
Updating Assessment Nuget to version 1.0.20210510.492


Charles-Gagnon (commented):
Why are you adding 3 different versions of the package?

PR #1211, opened 2021-05-13
Fix intellisense parsing for Azure and MI connections

Charles-Gagnon:
For https://github.com/microsoft/azuredatastudio/issues/12228

We weren't properly checking the engine type - so it was marking Azure databases as having a version of 120 (12), which meant that there was a lot of perfectly valid syntax that was being marked as invalid. 

Fixed it so that the logic works for MI as well (see discussion below) and will hopefully similar other weird cases. 

Also added a few newer engine versions that have been added since this code was written. I'm going to follow up with other folks to see if there's a better way we can handle this (such as SqlParser providing this kind of helper function) since maintaining this is going to be an annoyance otherwise.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
@@ -123,7 +123,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             get
             {
                 return this.IsConnected 
-                    ? GetTransactSqlVersion(this.ServerVersion)
+                    ? GetTransactSqlVersion(this.DatabaseEngineType, this.ServerVersion)
                     : TransactSqlVersion.Current;
             }
         }
@@ -136,7 +136,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             get
             {
                 return this.IsConnected
-                    ? GetDatabaseCompatibilityLevel(this.ServerVersion)
+                    ? GetDatabaseCompatibilityLevel(this.DatabaseEngineType, this.ServerVersion)
                     : DatabaseCompatibilityLevel.Current;
             }
         }
@@ -165,8 +165,12 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         /// Gets the database compatibility level from a server version
         /// </summary>
         /// <param name="serverVersion"></param>
-        private static DatabaseCompatibilityLevel GetDatabaseCompatibilityLevel(ServerVersion serverVersion)
+        private static DatabaseCompatibilityLevel GetDatabaseCompatibilityLevel(DatabaseEngineType engineType, ServerVersion serverVersion)
         {
+            if(engineType == DatabaseEngineType.SqlAzureDatabase)
+            {
+                return DatabaseCompatibilityLevel.Azure;
+            }
             int versionMajor = Math.Max(serverVersion.Major, 8);
 
             switch (versionMajor)
@@ -183,6 +187,10 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                     return DatabaseCompatibilityLevel.Version120;
                 case 13:
                     return DatabaseCompatibilityLevel.Version130;
+                case 14:
+                    return DatabaseCompatibilityLevel.Version140;
+                case 15:
+                    return DatabaseCompatibilityLevel.Version150;
                 default:
                     return DatabaseCompatibilityLevel.Current;
             }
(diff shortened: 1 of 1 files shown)


Xtrimmer (approved):

shueybubbles:
oh BTW - Managed Edition should be treated as version 15 for syntax purposes.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
```diff
@@ -192,8 +200,13 @@ private static DatabaseCompatibilityLevel GetDatabaseCompatibilityLevel(ServerVe
         /// Gets the transaction sql version from a server version
         /// </summary>
         /// <param name="serverVersion"></param>
-        private static TransactSqlVersion GetTransactSqlVersion(ServerVersion serverVersion)
+        private static TransactSqlVersion GetTransactSqlVersion(DatabaseEngineType engineType, ServerVersion serverVersion)
```
shueybubbles, line 203: We could make this somewhat more resilient to SQL editions going rogue (eg MI) by getting the actual compatibility level of the database and using max(version, actualcompatibilitylevel) for non-azure instances.

shueybubbles:
I think putting a function like this on ServerConnection or Server in SMO would be fine. I wouldn't want to couple SqlParser to more concepts related to this algebra. I also wouldn't want to couple SMO to SqlParser, so whatever type or enum is used as the return value from the function has to be a base type

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
Charles-Gagnon: Alright, took a shot at that. @shueybubbles what do you think?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
```diff
@@ -162,68 +180,68 @@ public ParseOptions ParseOptions
...
-            if (engineType == DatabaseEngineType.SqlAzureDatabase)
+            if (server.DatabaseEngineType == DatabaseEngineType.SqlAzureDatabase)
             {
                 return DatabaseCompatibilityLevel.Azure;
             }
-            int versionMajor = Math.Max(serverVersion.Major, 8);
 
-            switch (versionMajor)
+            // Get the actual compat level of the database we're connected to
+            switch (server.Databases[server.ConnectionContext.DatabaseName].CompatibilityLevel)
```
shueybubbles, line 194: might want to SetDefaultInitFields(typeof(Database), nameof(Database.CompatibilityLevel)) first

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
```diff
@@ -162,68 +180,68 @@ public ParseOptions ParseOptions
...
-            if (engineType == DatabaseEngineType.SqlAzureDatabase)
+            if (server.DatabaseEngineType == DatabaseEngineType.SqlAzureDatabase)
             {
                 return TransactSqlVersion.Azure;
             }
 
-            int versionMajor = Math.Max(serverVersion.Major, 9);
-
-            switch (versionMajor)
+            // Use the compat level of the database we're connected to for determing what verison
```
shueybubbles, line 228: determining

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
```diff
@@ -162,68 +180,68 @@ public ParseOptions ParseOptions
...
+            // of T-SQL to support
+            switch (server.Databases[server.ConnectionContext.DatabaseName].CompatibilityLevel)
```
shueybubbles, line 230: the compat level of the database doesn't limit the DDL syntax. We should use the Math.Max equivalent of (serverVersion.Major, CompatibilityLevel) because there are now SQL editions whose DDL version is higher than their engine version. Such a check will miss an MI instance set to a lower compat level but the alternative is to start checking for specific editions and we don't know a priori when a given edition will start supporting new DDL versions. Maybe we could check the compat level of "master" specifically and fall back to the user database compat level only if the user can't connect to master. I think master will always have the highest compat level for a given edition.

shueybubbles (changes requested):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
```diff
@@ -162,55 +183,111 @@ public ParseOptions ParseOptions
...
+            if (server.DatabaseEngineType == DatabaseEngineType.SqlAzureDatabase)
+            {
+                return TransactSqlVersion.Azure;
+            }
 
-            switch (versionMajor)
+            // Determine the language version to use - we can't just use VersionMajor directly because there are engine versions (such as MI)
+            // whose language version they support is higher than the actual server version. So we choose the highest compat level from
+            // between the server version and compat level
+            var compatLevel = Math.Max(server.VersionMajor * 10, (int)GetServerCompatabilityLevel(server));
```
shueybubbles, line 234: Compatibility

Charles-Gagnon: 🤦‍♂️ This is what happens when I go to an IDE that doesn't have a spellchecker...

shueybubbles (approved):

alanrenmsft (approved):

PR #1212, opened 2021-05-18
Filtering Schema Compare error from warnings

ssreerama:
This changes is a fix for the ADS schema compare extension test failure, where the warning messages had been sent as error messages which displays message even though the comparison succeed. This change will filter out the warning messages and only sent the fatal error messages to the ADS.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
@@ -123,7 +123,10 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                 // Appending the set of errors that are stopping the schema compare to the ErrorMessage
                 // GetErrors return all type of warnings, and error messages. Only filtering the error type messages here
                 var errorsList = ComparisonResult.GetErrors().Where(x => x.MessageType.Equals(Microsoft.SqlServer.Dac.DacMessageType.Error)).Select(e => e.Message).Distinct().ToList();
-                ErrorMessage = string.Join("\n", errorsList);
+                if (errorsList.Count > 0)
+                {
+                    ErrorMessage = string.Join("\n", errorsList);
+                }
             }
             catch (Exception e)
             {
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
@@ -120,8 +120,9 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                     }
                 }
 
-                // Appending the set of errors that are stopping the schema compare to the ErrorMessage 
-                var errorsList = ComparisonResult.GetErrors().Select(e => e.Message).Distinct().ToList();
+                // Appending the set of errors that are stopping the schema compare to the ErrorMessage
+                // GetErrors return all type of warnings, and error messages. Only filtering the error type messages here
+                var errorsList = ComparisonResult.GetErrors().Where(x => x.MessageType.Equals(Microsoft.SqlServer.Dac.DacMessageType.Error)).Select(e => e.Message).Distinct().ToList();
                 ErrorMessage = string.Join("\n", errorsList);
             }
             catch (Exception e)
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -120,8 +120,9 @@ public void Execute(TaskExecutionMode mode)
                     }
                 }
 
-                // Appending the set of errors that are stopping the schema compare to the ErrorMessage 
-                var errorsList = ComparisonResult.GetErrors().Select(e => e.Message).Distinct().ToList();
+                // Appending the set of errors that are stopping the schema compare to the ErrorMessage
+                // GetErrors return all type of warnings, and error messages. Only filtering the error type messages here
+                var errorsList = ComparisonResult.GetErrors().Where(x => x.MessageType.Equals(Microsoft.SqlServer.Dac.DacMessageType.Error)).Select(e => e.Message).Distinct().ToList();
```
kisantia, line 125: can you add a check in the schema compare tests to check that the expected value of the error message?

udeeshagautam: + A test to see if we are getting errors only when that result.success is false or otherwise. 

Please try to update the dlls in ADS and run the integration tests for ADS also. Let us know for any help needed there.

kisantia (approved):

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -919,6 +919,54 @@ public async Task SchemaCompareIncludeExcludeWithDependencies()
...
+                SchemaCompareOperation schemaCompareOperation = new SchemaCompareOperation(schemaCompareParams, result.ConnectionInfo, result.ConnectionInfo);
+                schemaCompareOperation.Execute(TaskExecutionMode.Execute);
+
+                // Expected 'data loss could occur' warning messages while comparing 
+                var warnings = schemaCompareOperation.ComparisonResult.GetErrors().Where(x => x.MessageType.Equals(Microsoft.SqlServer.Dac.DacMessageType.Warning)).Select(e => e.Message).Distinct().ToList();
+                var errors = schemaCompareOperation.ComparisonResult.GetErrors().Where(x => x.MessageType.Equals(Microsoft.SqlServer.Dac.DacMessageType.Error)).Select(e => e.Message).Distinct().ToList();
+
+                // Assertions
+                Assert.AreEqual(2, schemaCompareOperation.ComparisonResult.Differences.Count());
+                Assert.AreEqual(2, warnings.Count);
```
kisantia, line 958: what are the 2 warnings? Is there another one in addition to the data loss warning?

ssreerama: The target database have two tables being dropped, these two warnings are for those two tables.

kisantia: can you add a comment with that? thanks!

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -120,8 +120,9 @@ public void Execute(TaskExecutionMode mode)
...
                 ErrorMessage = string.Join("\n", errorsList);
```
udeeshagautam, line 126: just to triple confirm - the ErrorMessage doesnt get passed to ADS if the result is true (passe) - right?

udeeshagautam (approved):

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
ssreerama: done!

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
ssreerama: True, if the comparison is success, the ErrorMessage will be empty!

udeeshagautam (approved):
looks good to me

kisantia (approved):

PR #1215, opened 2021-06-02
Update Assessment Nuget to Enable Logging

cssuh:
Updated the Assessment Nuget to version 1.0.20210528.591 
This enables logging. Logs will be put in the `\azuredatastudio\extensions\mssql\sqltoolsservice\Windows\3.0.0-release.105` file path
You can change the file path by setting the `SqlAssessmentConfiguration.AssessmentReportAndLogsRootFolderPath` variable


Charles-Gagnon (commented):
So by default this just logs to the same folder as the DLL? What I'd suggest is to put these in the same place as the other STS logs - that is `Logger.LogFileFullPath` - so that we aren't ending up with logs in lots of random places.

aasimkhan30 (approved):

PR #1216, opened 2021-06-10
Localization added to sqlservicetools

smartguest:
This PR adds localization to projects that have strings to be localized along with newly translated string resources,

It also adds in a test for SqlTools.Credentials localization by modifying the CommandOptions to change the internal SR culture as with the Kusto/SqlTools ServiceLayer (This is required for testing, otherwise the string resources will be null during testing, though this does not affect the translation in actual usage). A similar localization test has been added for Kusto Service Layer (both being based on the SrTest for the SqlTools ServiceLayer)

Additionally,  the same modification to CommandOptions has been added to ResourceProvider to accommodate future localization tests if strings are ever added to the strings file. 

The above modified CommandOptions also change any dependent projects' SR culture for future testing as well since they do not have CommandOptions themselves and are accessed by the four projects that do use them.


src/Microsoft.SqlTools.Hosting/Properties/AssemblyInfo.cs
```diff
@@ -49,3 +49,4 @@
 [assembly: InternalsVisibleTo("MicrosoftSqlToolsServiceLayer")]
 [assembly: InternalsVisibleTo("MicrosoftKustoServiceLayer")]
 [assembly: InternalsVisibleTo("MicrosoftSqlToolsCredentials")]
+[assembly: InternalsVisibleTo("SqlToolsResourceProviderService")]
```
Charles-Gagnon, line 52: What is this needed for? Generally we should try to avoid this - if something needs to be accessed by other assemblies just make it public.

src/Microsoft.SqlTools.ResourceProvider/ResourceCommandOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            try
+            {
+                LocaleSetter(locale);
+
+                // Setting our internal SR culture to our global culture
+                SR.Culture = CultureInfo.CurrentCulture;
+
+                // Setting Core SR culture to our global culture
+                Microsoft.SqlTools.ResourceProvider.Core.SR.Culture = CultureInfo.CurrentCulture;
```
Charles-Gagnon, line 28: Is this really the only way to do this? It seems odd to be manually setting the culture like this ourself - typically C# programs get strings based on the current culture from satellite DLLs that are used based on the culture of the current thread. Are you following some pre-existing code here with doing it this way?

src/Microsoft.SqlTools.Hosting/Properties/AssemblyInfo.cs
smartguest: This is meant to access the functions for CommandOptions, such as LocaleSetter, which throws up an access error if the internals are not visible. 

See:
https://github.com/microsoft/sqltoolsservice/blob/alex/localizationbuild/src/Microsoft.SqlTools.ServiceLayer/Utility/ServiceLayerCommandOptions.cs

and 

https://github.com/microsoft/sqltoolsservice/blob/alex/localizationbuild/src/Microsoft.Kusto.ServiceLayer/Utility/ServiceLayerCommandOptions.cs

For examples

src/Microsoft.SqlTools.ResourceProvider/ResourceCommandOptions.cs
smartguest: This is meant for testing only, it works fine when using sqltoolsservice in azuredatastudio without it, but the tests do not work without it, as string resources are null instead of being properly translated when the SR.culture = culture line is removed.

This is following:
https://github.com/microsoft/sqltoolsservice/blob/alex/localizationbuild/src/Microsoft.Kusto.ServiceLayer/Utility/ServiceLayerCommandOptions.cs

and 

https://github.com/microsoft/sqltoolsservice/blob/alex/localizationbuild/src/Microsoft.SqlTools.ServiceLayer/Utility/ServiceLayerCommandOptions.cs

src/Microsoft.SqlTools.Hosting/Properties/AssemblyInfo.cs
Charles-Gagnon: Where is LocaleSetter defined? If we could just make that public I'd much prefer that - InternalsVisibleTo isn't really a great pattern to get in the habit of using since it makes determining dependencies much more difficult.

smartguest: It is defined here: https://github.com/microsoft/sqltoolsservice/blob/alex/localizationbuild/src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs However, turning the function from internal to public doesn't work as the ResourceCommandOptions has no context for LocaleSetter without the InternalsVisible line.

smartguest: I set LocaleSetter to public and also removed internalsvisible for credentials and kusto, (ran the unit tests on them without issue, but I can restore them if necessary, ServiceLayer still needs it though).

src/Microsoft.SqlTools.Credentials/Localization/sr.cs
```diff
@@ -69,6 +69,14 @@ public static string CredentialServiceWin32CredentialDisposed
             }
         }
 
+        public static string TestLocalizationConstant
```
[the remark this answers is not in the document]
smartguest: I did initially, but I thought others might change the string content, but if they aren't going to, I will use the invalid critical handle string.

src/Microsoft.SqlTools.Credentials/CredentialsCommandOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public override void SetLocale(string locale)
+        {
+            try
+            {
+                LocaleSetter(locale);
+
+                // Setting our internal SR culture to our global culture
+                SR.Culture = CultureInfo.CurrentCulture;
```
Charles-Gagnon, line 25: This still seems a little odd to me - does the generated class not handle looking up the culture when fetching strings already?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.Credentials/Localization/sr.pt-BR.resx
```diff
@@ -0,0 +1,135 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
alanrenmsft, line 1: BR here is upper case, is it not going to cause issues?

alanrenmsft: I just spotted a few other places where the BR in file name is in upper case.

Charles-Gagnon: That shouldn't cause issues with this stuff since we tend to just glob all resx files, but yeah being consistent is good.

src/Microsoft.SqlTools.Credentials/CredentialsCommandOptions.cs
smartguest: Apparently, the reason this is done is because in the tests, it attempts to access the Credential SR culture's name which is null of course if its not initialized. It does not change the localization however when it happens, I just tested. Therefore if we remove the culture name test, it should work without creating a new command class.

alanrenmsft (approved):

Issue #1217, opened 2021-06-24
Update SRGen to support Net5.0+


aasimkhan30:
Currently SRGen requires dotnet 3.1 to run, the tool should be updated to support newer versions as well so we can remove the need to install the older framework for building.

Charles-Gagnon:
There's only really one issue left so renaming this and updating the description. Here's the original other issues for historical purposes : 

~~1. Updating STS to dotnet 5.0~~ Updated

~~1. Solve all the code analysis warnings we see while building this project.~~ This isn't really feasible at this time given the huge amount of warnings, can make a separate issue if we decide to take this up in the future

~~1. Trying to use a new version of ESRP that uses dotnet 5.0( if present). Current one uses dotnet 2.2~~  Updated

~~1. Update risky npm packages~~ All issues fixed and monitoring set up for new alerts

shueybubbles:
@Charles-Gagnon @smartguest I think you can just pass `-f Net5.0` to `dotnet run` here.
https://github.com/microsoft/sqltoolsservice/blob/693bb0480aacca5bdfb1c021526a3c5cb0574812/build.cake#L646

shueybubbles:
oh I should have remembered that, it's what the targets file that ships with srgen 3.2 does

```
Microsoft.Data.Tools.StringResourceTool.targets:    <SrGenFileExe Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' OR '$(TargetFrameworkIdentifier)' == '.NETStandard'">dotnet --roll-forward LatestMajor "$(MSBuildThisFileDirectory)/../lib/netcoreapp3.1/srgen.dll"</SrGenFileExe>
```

PR #1218, opened 2021-06-24
Updating to dotnet 5.0

aasimkhan30:
Updating STS to dotnet 5.0. This will be one of the first steps to remove the several build warnings we see during our pipelines runs for using older versions of dotnet that are nearing EOL. 

General Notes: 

 1. Unfortunately, due to SRGEN not available to use with dotnet 5.0, I have added a pipeline step that installs dotnet 3.1.
 2. Will follow up with a separate PR that fixes all the code analysis warnings. 
 3. This will also require a bigger change in https://github.com/microsoft/azuredatastudio/blob/main/extensions/mssql/config.json as the archive names are getting changed.


test/Microsoft.Kusto.ServiceLayer.UnitTests/Microsoft.Kusto.ServiceLayer.UnitTests.csproj
```diff
@@ -1,8 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>netcoreapp3.1</TargetFramework>
-
+        <TargetFramework>$(TestProjectsTargetFramework)</TargetFramework>
```
aasimkhan30, line 4: Changing target framework of Kusto from a hardcoded value to the same as rest of the project.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -64,7 +64,7 @@ public void CanCancelConnectRequest()
 
             // Given a connection that times out and responds to cancellation
             var mockConnection = new Mock<DbConnection> { CallBase = true };
-            CancellationToken token;
+            CancellationToken token = new CancellationToken();
```
aasimkhan30, line 67: This was throwing a compilation error

aasimkhan30:
Mentioning this issue https://github.com/microsoft/sqltoolsservice/issues/1217

azure-pipelines/build.yml
```diff
@@ -43,6 +43,11 @@ steps:
     projects: '$(Build.SourcesDirectory)/src/Microsoft.Kusto.ServiceLayer'
     arguments: '--configuration Release'
 
+- task: UseDotNet@2
```
shueybubbles, line 46: maybe we should rewrite srgen in go so it's not framework-dependent anymore. Volunteers? :-)

shueybubbles (approved):

aasimkhan30:
@Charles-Gagnon I had tried a locally built one, but I will try one from the azure pipelines

aasimkhan30:
@Charles-Gagnon, tried the win-x64 zip from the adhoc build. Everything seems to work fine.

Charles-Gagnon:
Thanks, could you also test on Mac and Linux?

udeeshagautam:
@aasimkhan30 as mentioned in chat this can affect SQL project build since it used the dotnet dlls at runtime. There can be two impacts (I am not sure if there certainly will be or not)
1) we might need users to depend on dotnet 5 for build (customer impact)
2) we might also need to update dacfx which would be a bigger issue. 

Please wait for for a validation from our side.

kisantia (approved):
I verified that the projects and dacpac extensions are still working with these changes.

JustinMDotNet (approved):
Looks good. Tested...

Connect to Kusto Cluster
Execute Query
Generate Take 10 Script
Generate Function Script
Execute Function
Connect to Notebook
Execute query in Notebook
Hover in Query
Hover in Notebook
Autocomplete in Query
Autocomplete in Notebook
Dashboard
Navigating object explorer

aasimkhan30:
Verified the build on mac, linux and ubuntu and everything seems to work. Pushing this change in.

PR #1219, opened 2021-07-07
Enable Multiple Assessments

cssuh:
- Added an additional parameter to handle running assessments for multiple databases
- Build multiple connection strings for running assessments for each database


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -140,10 +140,18 @@ public void InitializeService(ServiceHost serviceHost)
                     Version = Version.Parse(serverInfo.ServerVersion),
                     Platform = hostInfo.Platform
                 };
-
                 var db = SqlAssessmentService.GetDatabaseLocator(server, connection.Database);
-                var connectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
-                var results = await GetAssessmentItems(server, connectionString);
+                var connectionStrings = new List<string>();
+                if (parameters.Databases != null)
```
Charles-Gagnon, line 145: This is changing behavior - before it would pass in the connection string for the original db and use that. Is the intention that the callers are now required to pass in the list of dbs to use? What happens if they pass in an empty array?

Also am I missing something or is the db object above not actually used anywhere?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -140,10 +140,18 @@ public void InitializeService(ServiceHost serviceHost)
...
+                var connectionStrings = new List<string>();
+                if (parameters.Databases != null) 
+                {
+                    foreach (string database in parameters.Databases)
+                    {
+                        connInfo.ConnectionDetails.DatabaseName = database;
+                        connectionStrings.Add(ConnectionService.BuildConnectionString(connInfo.ConnectionDetails));
+                    }
+                }
+                string[] assessmentConnectionStrings = connectionStrings.ToArray();
```
brian-harris, line 153: should this entire section be skipped parameters.Databases == null

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
cssuh: Yes, we now have the user select the DBs that will be assessed, so those DBs will get passed in through this parameter.  I use the parameters.Databases in line 147 to loop through each DB and build a connection string for each DB.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
cssuh: are you saying I should remove the `if (parameters.Databases != null)`?

brian-harris: should lines 154 - 157 be executed if the database count is 0, i.e. assessmentConnectionStrings is empty
if no, can they go inside the if check.

aasimkhan30:
Looks good overall. Have you checked if we are not running assessment over extra databases?

brian-harris (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -140,13 +140,22 @@ public void InitializeService(ServiceHost serviceHost)
...
-
                 var db = SqlAssessmentService.GetDatabaseLocator(server, connection.Database);
-                var connectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails);
-                var results = await GetAssessmentItems(server, connectionString);
-                var result = new MigrationAssessmentResult();
-                result.Items.AddRange(results);
-                await requestContext.SendResult(result);
+                var connectionStrings = new List<string>();
+                if (parameters.Databases != null) 
+                {
```
Charles-Gagnon, line 146: Do you really need to check for Databases not being null here? If the contract says they shouldn't be null then typically it should be fine to assume it's not null

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -140,13 +140,22 @@ public void InitializeService(ServiceHost serviceHost)
                     Version = Version.Parse(serverInfo.ServerVersion),
                     Platform = hostInfo.Platform
                 };
-
                 var db = SqlAssessmentService.GetDatabaseLocator(server, connection.Database);
```
Charles-Gagnon, line 143: Is this being used anywhere?

Charles-Gagnon (approved):

PR #1220, opened 2021-07-10
Adding cpu count and ram size to server info

aasimkhan30:
These properties are going to be used in Telemetry for sql migration


alanrenmsft:
you should provide some context, where will this be used.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -753,6 +758,55 @@ public static ServerHostInfo GetServerHostInfo(IDbConnection connection)
...
+                try
+                {
+                    ExecuteReader(
+                        connection,
+                        SqlConnectionHelperScripts.GetHostSysInfoBeforeVersion11,
+                        reader =>
+                        {
+                            reader.Read();
+                            sysInfo.CpuCount = Int32.Parse(reader[0].ToString(), CultureInfo.InvariantCulture);
+                            sysInfo.PhysicalMemoryInKb = long.Parse(reader[1].ToString(), CultureInfo.InvariantCulture)/1024;
```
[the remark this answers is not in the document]
shueybubbles: Try to use SMO because it also knows about the special case for SQL Managed Instance, which has data in sys.server_resource_stats instead of dm_os_host_info

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/SqlConnectionHelperScripts.cs
```diff
@@ -66,5 +66,7 @@ END CATCH
 ";
         public const string GetHostInfo = @"SELECT [host_platform], [host_distribution], [host_release], [host_service_pack_level], [host_sku], [os_language_version] FROM sys.dm_os_host_info";
         public const string GetHostWindowsVersion = @"SELECT windows_release FROM sys.dm_os_windows_info";
+        public const string GetHostSysInfoBeforeVersion11 = @"SELECT cpu_count, physical_memory_in_bytes from sys.dm_os_sys_info";
```
alanrenmsft, line 69: you don't need these anymore, right?

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -827,6 +856,11 @@ public static ServerInfo GetServerVersion(IDbConnection connection)
                 //  otherwise - Windows Server 2019 Standard 10.0
                 serverInfo.OsVersion = hostInfo.Distribution != null ? string.Format("{0} {1}", hostInfo.Distribution, hostInfo.Release) : string.Format("{0} {1}", hostInfo.Platform, hostInfo.Release);
 
+                var osInfo = GetServerSysInfo(connection);
```
Charles-Gagnon, line 859: I would name this just sysInfo - it's more than just information about the OS

Charles-Gagnon (approved):
Few things to clean up but looks good otherwise. I'm slightly worried about making a ton of SMO server objects everywhere unnecessarily but I suppose we can investigate that later on if it becomes an issue.

alanrenmsft (approved):
don't forget to remove the sql scripts you added that are not needed now.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -753,6 +759,38 @@ public static ServerHostInfo GetServerHostInfo(IDbConnection connection)
...
+                if (reliableSqlConnection != null)
+                {
+                    conn = reliableSqlConnection.GetUnderlyingConnection();
+                }
+                else if (sqlConnection != null)
+                {
+                    conn = sqlConnection;
+                }
+                var server = new Server(new ServerConnection(conn));
+                server.SetDefaultInitFields(server.GetType(), new String[] { "Processor", "PhysicalMemory" });
```
shueybubbles, line 783: use nameof(server.Processor), nameof(server.PhysicalMemory) instead of string constants

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -753,6 +759,38 @@ public static ServerHostInfo GetServerHostInfo(IDbConnection connection)
...
+                sysInfo.CpuCount = server.Processors;
+                sysInfo.PhysicalMemoryInKb = server.PhysicalMemory;
+            }
+            catch (Exception ex)
+            {
+                Logger.Write(TraceEventType.Error, ex.ToString());
```
shueybubbles, line 789: shouldn't you put a value in the structure to tell the client you couldn't get the right data? The user experience on an error is just going to be seeing zeroes, correct?

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -676,6 +676,12 @@ public class ServerHostInfo
             public string ServicePackLevel;
         }
 
+        public class ServerSysInfo
```
kburtram, line 679: nit: unless you're copying the naming from somewhere else, I typically prefer to spell out the whole word (as recommended in .Net naming conventions guide).  So `ServerSystemInfo` in this case.

kburtram (approved):
LGTM

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
Charles-Gagnon: Typically we would want to throw an exception so that the caller will get an error on their side - I agree that silently just sending 0's is not preferable.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -753,6 +759,38 @@ public static ServerHostInfo GetServerHostInfo(IDbConnection connection)
...
+        /// <param name="connection">The connection</param>
+        public static ServerSysInfo GetServerSysInfo(IDbConnection connection)
+        {
+            var sysInfo = new ServerSysInfo();
+            try
+            {
+               SqlConnection conn = null;
+               ReliableSqlConnection reliableSqlConnection = connection as ReliableSqlConnection;
+               SqlConnection sqlConnection = connection as SqlConnection;
+                if (reliableSqlConnection != null)
```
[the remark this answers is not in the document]
aasimkhan30: I just got this logic from 
https://github.com/microsoft/sqltoolsservice/blob/839acf67cda8b41c1fdd4a241ebfecbae51f0f17/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/TaskServices/TaskServiceTests.cs#L142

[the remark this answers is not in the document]
aasimkhan30: There is a helper function for this. I am replacing my code with that.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ServerInfo.cs
```diff
@@ -75,7 +75,7 @@ public class ServerInfo
         /// <summary>
         /// The physical memory of the host running the server in KBs.
```
Charles-Gagnon, line 76: Update this as well

PR #1222, opened 2021-07-16
Adding server metadata to SQL Migration  Assessment API

aasimkhan30:
All of this additional info will be used in migration telemetry

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
@@ -23,15 +23,15 @@ namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
         /// <summary>
         /// Result of the assessment
         /// </summary>
-        public ServerProperties Result { get; set; }
+        public ServerAssessmentProperties Result { get; set; }
         /// <summary>
         /// Start time of the assessment
         /// </summary>
-        public string StartedOn { get; set; }
+        public string StartTime { get; set; }
         /// <summary>
         /// End time of the assessment
         /// </summary>
-        public string EndedOn { get; set; }
+        public string EndedTime { get; set; }
     }
 
     /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
@@ -4,6 +4,7 @@
 //
 
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlServer.Migration.Assessment.Common.Contracts.Models;
 
 namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
 {
@@ -30,6 +31,10 @@ namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
         /// End time of the assessment
         /// </summary>
         public string EndedTime { get; set; }
+        /// <summary>
+        /// Contains the raw assessment response 
+        /// </summary>
+        public ISqlMigrationAssessmentModel RawAssessmentResult { get; set; }
     }
 
     /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
@@ -3,9 +3,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-using System.Collections.Generic;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
-using Microsoft.SqlServer.Migration.Assessment.Common.Contracts.Models;
 
 namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
 {
@@ -23,7 +21,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
         /// <summary>
         /// Result of the assessment
         /// </summary>
-        public ServerAssessmentProperties Result { get; set; }
+        public ServerAssessmentProperties AssessmentResult { get; set; }
         /// <summary>
         /// Start time of the assessment
         /// </summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
@@ -17,25 +17,25 @@ namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
     public class MigrationAssessmentResult
     {
         /// <summary>
-        /// Gets the collection of assessment results.
+        /// Errors that happen while running the assessment
         /// </summary>
-        public List<MigrationAssessmentInfo> Items { get; } = new List<MigrationAssessmentInfo>();
+        public ErrorModel[] Errors { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating
-        /// if assessment operation was successful.
-        /// </summary>
-        public bool Success { get; set; }
-
-        /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -186,203 +183,134 @@ public bool TryGetData(string column, out object value)
             }
         }
 
-        internal async Task<List<MigrationAssessmentInfo>> GetAssessmentItems(SqlObjectLocator target, string connectionString)
+        internal async Task<MigrationAssessmentResult> GetAssessmentItems(string connectionString)
         {
```
brian-harris, line 187: be sure to coordinate with the following changes 
https://github.com/microsoft/sqltoolsservice/pull/1219

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
```diff
@@ -5,6 +5,7 @@
 
 using System.Collections.Generic;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlServer.Migration.Assessment.Common.Contracts.Models;
```
kburtram, line 8: you should try to keep the using statements sorted, with the system namespaces listed first (per project convention)

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
```diff
@@ -16,20 +17,21 @@ public class MigrationAssessmentsParams
...
-        /// if assessment operation was successful.
+        /// Result of the assessment
         /// </summary>
-        public bool Success { get; set; }
-
+        public ServerProperties Result { get; set; }
+        /// <summary>
+        /// Start time of the assessment
+        /// </summary>
+        public string StartedOn { get; set; }
```
kburtram, line 30: nit: I would probably use names StartTime and EndTime, but that is mostly just preference

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationObjectProperties.cs
```diff
@@ -0,0 +1,129 @@
+using System.Collections.Generic;
```
kburtram, line 1: add missing copy write header

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -186,203 +183,134 @@ public bool TryGetData(string column, out object value)
...
-                        impactedObjects[i] = impactedObject;
-                    }
-                    item.ImpactedObjects = impactedObjects;
-                }
-
-                result.Add(item);
+                assessmentResultLookup.Add(CreateAssessmentResultKey(r as ISqlMigrationAssessmentResult), r as ISqlMigrationAssessmentResult);
             }
-            return result;
+            ISqlMigrationAssessmentModel assessmentResultBig = await engine.GetTargetAssessmentResultsList(System.Threading.CancellationToken.None);
```
kburtram, line 198: I not sure what "Big" means in the name.  Perhaps a comment would help, or simply name assessmentResult?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -186,203 +183,134 @@ public bool TryGetData(string column, out object value)
...
+                    Tags = check.Tags.ToArray(),
+                    RulesetName = Engine.Configuration.DefaultRuleset.Name,
+                    RulesetVersion = Engine.Configuration.DefaultRuleset.Version.ToString(),
+                    RuleId = r.FeatureId.ToString(),
+                    Message = r.Message,
+                    AppliesToMigrationTargetPlatform = r.AppliesToMigrationTargetPlatform.ToString(),
+                    IssueCategory = r.IssueCategory.ToString(),
+                    ImpactedObjects = ParseImpactedObjects(r.ImpactedObjects).ToArray()
+                };
+            }).ToList();
```
kburtram, line 285: it looks we do a ToList here then immediately do a ToArray on the return value.  It maybe better to use ToArray here directly to avoid so much conversation & data copying.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -186,203 +183,134 @@ public bool TryGetData(string column, out object value)
...
             {
-                await requestContext.SendResult(true);
-            }
+                return new ImpactedObjectInfo()
+                {
+                    Name = i.Name,
+                    ImpactDetail = i.ImpactDetail,
+                    ObjectType = i.ObjectType
+                };
+            }).ToList();
```
kburtram, line 297: same comment as above here regarding ToList

kburtram (approved):
LGTM

brian-harris (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationObjectProperties.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <summary>
+        /// Returns the assessment status
+        /// </summary>
+        public int SqlAssessmentStatus { get; set; }
+        /// <summary>
+        /// Count of Dbs assessed
+        /// </summary>
+        public long AssessedDatabaseCount{get; set;}
+        /// <summary>
+        /// Give database statistics for the server
```
Charles-Gagnon, line 53: Is this the correct description?

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationAssessmentsRequest.cs
```diff
@@ -16,20 +17,21 @@ public class MigrationAssessmentsParams
...
-
+        public ErrorModel[] Errors { get; set; }
         /// <summary>
-        /// Gets or sets a value indicating
-        /// if assessment operation was successful.
+        /// Result of the assessment
         /// </summary>
-        public bool Success { get; set; }
-
+        public ServerProperties Result { get; set; }
```
Charles-Gagnon, line 26: It seems a little odd to have the "result" of an assessment be a set of server properties. It seems like it'd be better to have the result be a boolean and then have a separate property that contains the server properties

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/MigrationObjectProperties.cs
```diff
@@ -0,0 +1,129 @@
+using System.Collections.Generic;
+using Microsoft.SqlServer.Migration.Assessment.Common.Contracts.TargetAssessment.Models;
+
+namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
+{
+    public class ServerProperties
```
Charles-Gagnon, line 6: ServerProperties is a bit generic - even if it's in the Migration namespace I'd suggest naming it something more clear like `ServerAssessmentProperties`. (especially since it contains migration/assessment specific properties - not just properties of the server itself)

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -186,203 +183,134 @@ public bool TryGetData(string column, out object value)
...
-                }
-
-                result.Add(item);
+                assessmentResultLookup.Add(CreateAssessmentResultKey(r as ISqlMigrationAssessmentResult), r as ISqlMigrationAssessmentResult);
             }
-            return result;
+            ISqlMigrationAssessmentModel assessmentResultBig = await engine.GetTargetAssessmentResultsList(System.Threading.CancellationToken.None);
+            return new MigrationAssessmentResult()
+            {
+                Result = ParseServerAssessmentInfo(assessmentResultBig.Servers[0], assessmentResultLookup),
```
Charles-Gagnon, line 201: Just to verify - you're only allowing a single server per assessment currently? I'm assuming based on this the API allows passing in more than one but we're just not using that functionality.

aasimkhan30: Yes, in the frontend we only are connected to one server at a time. Therefore, we are never using multi-server capabilities of this api.

PR #1227, opened 2021-08-04
Add changeConnectionUri action for query

smartguest:
This adds a new action that reassigns the query to a different owneruri in case a query has been saved to a different file name. (Can also work for the same owneruri as well)

This is to address this bug in ADS: https://github.com/microsoft/azuredatastudio/issues/1876 where to fix this issue, a new file query editor input containing the current query input's contents will be returned. However the URI for the query has not been updated on the sqltoolsservice side, and thus whenever sqltoolsservice tries to run a query, it fails to do so since it can no longer find the ownerUri containing the file. 

This PR will add the action that is called during a save, where instead of disposing the existing query, rather it assigns it to the new URI and changes the query editorConnection URI to match. (The companion PR for ADS will be created at a later time once any issues or bugs have been fixed for it). This PR may cause side effects when working with an actual query, so this is currently a draft for review.


Charles-Gagnon:
So is the current thought for how this is going to work be : 

1. User saves untitled file
2. STS receives document message for the "new" file - which then wouldn't have the connection for the untitled doc associated with it
3. The code you're adding in ADS will then send this message telling STS to move the connection from the old URI over to the new one

Did you look at all into whether the language server stuff has any messages already for dealing with updating a document that has been saved? A quick glance shows https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_didSave that would be worth looking in to - although I'm not currently seeing anything indicating a "new" or "old" URI so correlating the documents might not be possible. But if we can use the existing stuff that would be greatly preferred.

smartguest:
@Charles-Gagnon We do not currently have DdSaveTextDocument notifications implemented in the TextDocument.cs Contract for SqlTools ServiceLayer, but it can be added in, just not sure how to use it, whether it is automatically done when a file is saved or needs to be manually called.

Charles-Gagnon:
That would be handled by the VS Code language service stuff - so you should just be able to add in the handler for that message.

smartguest:
queryRename works now in conjunction with changes to my ADS branch and dataprotocol, it disconnects the original untitled URI (and deleting the map in the process). When ADS connects to STS with the new file uri, there won't be an entry for the untitled uri in the map anymore so ADS can create a new untitled query on the same database again.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -351,6 +355,45 @@ public void InitializeService(ServiceHost serviceHost)
...
+                if (!ActiveQueries.TryGetValue(OriginalOwnerUri, out query))
+                {
+                    await requestContext.SendError(SR.QueryServiceRequestsNoQuery);
+                    return;
+                }
+                else
+                {
+                    ActiveQueries.TryRemove(OriginalOwnerUri, out query);
+                    // Delete the map from ConnectionService uri connection map, reconnect with new uri later.
+                    ConnectionService.Disconnect(new DisconnectParams()
```
Charles-Gagnon, line 380: Why are you disconnecting the connection? A connection is independent of the editor that opened it - if we rename a file we should update the URI but the connection should stay open as it was.

smartguest:
The code before created a new connection info object when ADS connected with the new file uri, the code now reuses the one previously associated with the untitled uri.

Charles-Gagnon:
@kburtram @alanrenmsft What are your thoughts on the naming of this message (`queryRename`)?

The main issue I have is that one of the things we've always talked about being not great behavior is the fact that we tie connections to the file URIs as it is. And this ties it even closer to that.

We'll still always need some way to tell STS to update a connection URI, but I wonder if we should just make this be `updateConnectionUri` or something instead.

The main thing thinking maybe it's fine to keep this is if we could find a use for this event besides updating the connection URI. But I can't really think of a good scenario for that.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -172,6 +172,13 @@ public void RemoveAllConnections()
                 DbConnection connection;
                 ConnectionTypeToConnectionMap.TryRemove(type, out connection);
             }
+        }
+
+        /// <summary>
+        /// Changes the OwnerUri in case of a query rename.
+        /// </summary>
+        public void changeOwnerUri(String newOwnerUri){
```
Charles-Gagnon, line 180: C# methods should be CapitalCased by convention

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -732,6 +735,30 @@ public async Task HandleDidChangeTextDocumentNotification(ScriptFile[] changedFi
...
+        /// <param name="eventContext"></param>
+        /// <returns></returns>
+        public Task HandleDidSaveTextDocumentNotification(
+            string uri,
+            ScriptFile scriptFile,
+            EventContext eventContext)
+        {
+            try
+            {
+                // TODO: need to do something here.
```
Charles-Gagnon, line 752: Is this needed anymore?

Also did you ever look into using this event? Was there some reason you couldn't use it?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -385,6 +385,23 @@ public Task<ExecutionPlan> GetExecutionPlan(int batchIndex, int resultSetIndex)
             Batches[saveParams.BatchIndex].SaveAs(saveParams, fileFactory, successHandler, failureHandler);
         }
 
+        /// <summary>
+        /// Changes the editorConnection's OwnerURI in case queryRename has happened.
+        /// </summary>
+        /// <param name="newOwnerUri">Name of the new OwnerURI</param>
+        public void changeConnectionOwnerURI(String newOwnerUri)
```
Charles-Gagnon, line 392: I'd suggest just making this a property getter/setter

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryRenameRequest.cs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /// <summary>
+    /// Parameters to return as the result of a query dispose request
+    /// </summary>
+    public class QueryRenameResult
+    {
+        /// <summary>
+        /// Any error messages that occurred during disposing the result set. Optional, can be set
+        /// to null if there were no errors.
```
Charles-Gagnon, line 26: 1. We should just throw an error if that happens I think - having it complete "successfully" but then have errors would be confusing I think
2. If we did do that you should generally also include a Successful property so it's easy to tell whether the request succeeded or not

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
smartguest: It's generally not needed, I just did not know what to do with this event on the ADS side as its not implemented in any project that I was involved with.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryRenameRequest.cs
smartguest: Error handling is handled in the handleRenameRequest task in QueryExecutionService, I don't see any other examples of QueryActionRequest that throw an error, if there is an example, can you please let me know?

Charles-Gagnon: I'm saying that errors shouldn't be sent back in a result. You either call SendError or let the exception bubble up - both result in the caller on the ADS side getting an exception from the call with the error message from the exception.

Results should only be used for indicating success. And in this case since you don't really have any additional metadata I would say it would even be cleaner just to have the "response" be void instead of a whole object like this (that should be possible IIRC...)

smartguest: I'm following what is done in simpleExecuteRequest

smartguest: Alright I'll leave the renameresult empty, I'm following what is done in dispose which also returns an empty QueryDisposeResult object upon completion. If it is possible to remove the requestContext sendresult message from the code without any side effects, I'll do it, but there needs to be a QueryRenameResult object for the requestContext.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryChangeConnectionUriNotification.cs
```diff
@@ -0,0 +1,24 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
+{
+    /// <summary>
+    /// Parameters for the query cancellation request
```
Charles-Gagnon, line 11: This isn't correct

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/QueryChangeConnectionUriNotification.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public class QueryChangeConnectionUriParams
+    {
+        public string NewOwnerUri { get; set; }
+        public string OriginalOwnerUri { get; set; }
+    }
+    public class QueryChangeConnectionUriNotification
```
Charles-Gagnon, line 18: The name of this class and file should be changed.

But also - I was thinking more about how we name notifications vs requests and I think that we should rename the notification to `query/connectionUriChanged` to more clearly indicate that this is a one-way notification for clients telling them what happened.

Charles-Gagnon (approved):

PR #1228, opened 2021-08-04
Get Azure functions operation

kisantia:
This adds the GetAzureFunctions operation which returns the names of all the Azure functions in a given file. This is needed for the add sql binding in vscode scenario so that the user can choose the specific Azure function to add the binding to. 

Unfortunately, the Azure functions extension in vscode doesn't expose an api to get all the Azure functions in an AF project and the way they do it for their extension only looks at the generated files. Since we need to know the source file an Azure function is in to add the binding there, we have to add this ourselves.

I also noticed I accidentally created the test files as .ts, so I updated those to be .cs (inserting sql bindings using this will only be supported for C# files).


src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
```diff
@@ -0,0 +1,71 @@
...
+                SyntaxTree tree = CSharpSyntaxTree.ParseText(text);
+                CompilationUnitSyntax root = tree.GetCompilationUnitRoot();
+
+                // look for Azure Functions in the file
+                IEnumerable<AttributeArgumentSyntax> functionNameAttributes = from methodDeclaration in root.DescendantNodes().OfType<MethodDeclarationSyntax>()
+                                          where methodDeclaration.AttributeLists.Count > 0
+                                          where methodDeclaration.AttributeLists.Where(a => a.Attributes.Where(attr => attr.Name.ToString().Contains(functionAttributeText)).Count() == 1).Count() == 1
+                                          select methodDeclaration.AttributeLists.Select(a => a.Attributes.Where(attr => attr.Name.ToString().Contains(functionAttributeText)).First().ArgumentList.Arguments.First()).First();
+
+                // remove quotes from around the names
```
Charles-Gagnon, line 52: Will they always be quoted? Might be better to be safe and only remove them if they exist.

And are quotes valid in the name? If so you might need to unescape those in the name as well

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// <summary>
+        /// Gets the names of all the azure functions in a file
+        /// </summary>
+        /// <returns>the result of trying to get the names of all the Azure functions in a file</returns>
+        public GetAzureFunctionsResult GetAzureFunctions()
+        {
+            try
+            {
+                string text = File.ReadAllText(Parameters.filePath);
```
Charles-Gagnon, line 41: Is this something that you could contribute to their extension? They might find it useful - and having all that kind of code in one place would be handy I think.

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                SyntaxTree tree = CSharpSyntaxTree.ParseText(text);
+                CompilationUnitSyntax root = tree.GetCompilationUnitRoot();
+
+                // look for Azure Functions in the file
+                IEnumerable<AttributeArgumentSyntax> functionNameAttributes = from methodDeclaration in root.DescendantNodes().OfType<MethodDeclarationSyntax>()
+                                          where methodDeclaration.AttributeLists.Count > 0
+                                          where methodDeclaration.AttributeLists.Where(a => a.Attributes.Where(attr => attr.Name.ToString().Contains(functionAttributeText)).Count() == 1).Count() == 1
+                                          select methodDeclaration.AttributeLists.Select(a => a.Attributes.Where(attr => attr.Name.ToString().Contains(functionAttributeText)).First().ArgumentList.Arguments.First()).First();
```
udeeshagautam, line 50: Question - so any function with this attribute will be considered as an Az func - yeah? does user have to ad this attribute manually or autorest does that automatically?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/AzureFunctions/AzureFunctionsServiceTests.cs
```diff
@@ -21,8 +21,8 @@ class AzureFunctionsServiceTests
         public void AddSqlInputBinding()
         {
             // copy the original file because the input binding will be inserted into the file
-            string originalFile = Path.Join(testAzureFunctionsFolder, "AzureFunctionsNoBindings.ts");
-            string testFile = Path.Join(Path.GetTempPath(), string.Format("InsertSqlInputBinding-{0}.ts", DateTime.Now.ToString("yyyy-dd-M--HH-mm-ss")));
+            string originalFile = Path.Join(testAzureFunctionsFolder, "AzureFunctionsNoBindings.cs");
```
udeeshagautam, line 24: oh the content on these was cs before also right? I missed that :)

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsService.cs
```diff
@@ -33,6 +33,7 @@ public static AzureFunctionsService Instance
         public void InitializeService(ServiceHost serviceHost)
         {
             serviceHost.SetRequestHandler(AddSqlBindingRequest.Type, this.HandleAddSqlBindingRequest);
+            serviceHost.SetRequestHandler(GetAzureFunctionsRequest.Type, this.HandleGetAzureFunctionsRequest);
```
udeeshagautam, line 36: Do I remember correctly that you mentioned in ADS the option shows only for .cs files - I am assuming we are checking file extns on ADS side itself?

udeeshagautam (approved):
just some questions, other than that looks good!

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsService.cs
kisantia: yup the command only shows in the command palette when `editorLangId == csharp`

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
kisantia: Yup that's correct. If the user creates a new function using the "Create Function" command in vscode or creates an AF project using autorest, this attribute gets added automatically.

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            try
+            {
+                string text = File.ReadAllText(Parameters.filePath);
+
+                SyntaxTree tree = CSharpSyntaxTree.ParseText(text);
+                CompilationUnitSyntax root = tree.GetCompilationUnitRoot();
+
+                // look for Azure Functions in the file
+                IEnumerable<AttributeArgumentSyntax> functionNameAttributes = from methodDeclaration in root.DescendantNodes().OfType<MethodDeclarationSyntax>()
```
Benjin, line 47: Can you switch this to code-style linq instead of query-style linq so it's easier to work through with a debugger?

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
```diff
@@ -0,0 +1,71 @@
...
+                                          where methodDeclaration.AttributeLists.Count > 0
+                                          where methodDeclaration.AttributeLists.Where(a => a.Attributes.Where(attr => attr.Name.ToString().Contains(functionAttributeText)).Count() == 1).Count() == 1
```
Benjin, line 49: Anything you can do to take this complex linq query and make it more readable?  Maybe break it up into a couple steps with comments or extract the `Where()` clauses into helper functions?

`First().ArgumentList.Arguments.First()).First()` is a particularly fun time, lol

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
Benjin: `.Trim("\'")` (I think you can pass in characters to trim) might be another option that would handle both quoted and unquoted cases.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/AzureFunctions/AzureFunctionsServiceTests.cs
```diff
@@ -21,8 +21,8 @@ class AzureFunctionsServiceTests
...
+            string testFile = Path.Join(Path.GetTempPath(), string.Format("InsertSqlInputBinding-{0}.cs", DateTime.Now.ToString("yyyy-dd-M--HH-mm-ss")));
```
Benjin, line 25: String interpolation is glorious!

Instead of
`string.Format("InsertSqlInputBinding-{0}.cs", DateTime.Now.ToString("yyyy-dd-M--HH-mm-ss"))`
just
`$"InsertSqlInputBinding-{DateTime.Now.ToString("yyyy-dd-M--HH-mm-ss")}.cs"`

Benjin: Also, not that it matters, but you probably want MM so that it spits out two-digit month numbers.

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
kisantia: Added the trim. The names will always be quoted and quotes in the name are invalid, otherwise building the AF project will fail .

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
kisantia: I can look into contributing this to their extension later and can remove this if that gets added. The current exposed [api](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/vscode-azurefunctions.api.d.ts) is pretty limited and it would take more time to get it added there, so I'll stick with this approach for now to get this out sooner.

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/GetAzureFunctionsOperation.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                return new GetAzureFunctionsResult()
+                {
+                    Success = true,
+                    azureFunctions = aFNames
+                };
+            }
+            catch (Exception e)
+            {
+                return new GetAzureFunctionsResult()
```
Charles-Gagnon, line 70: Do you want to return a result like this? If an exception is thrown then it'll return an error result already so this shouldn't be necessary.

Charles-Gagnon (approved):

PR #1231, opened 2021-08-09
Fixing GetServerSystemInfo for Azure DBs

aasimkhan30:
This piece of code throws when running on an azure-based SQL connection thereby breaking Object explorer and dashboards for azure databases. 

I have changed the behavior to log the errors instead and continue with the workflow.

This PR fixes:
https://github.com/microsoft/azuredatastudio/issues/16626
https://github.com/microsoft/azuredatastudio/issues/16643

### DIFF
--- a/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
@@ -776,8 +776,10 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
             }
             catch (Exception ex)
             {
-                 // we don't want to fail when an error occurs while fetching these properties.
-                 // just logging them here and moving on with the workflow. 
+                // We don't want to fail the normal flow if any unexpected thing happens
+                // since these properties are not available for types of sql servers and users 
+                // and it is not essential to always include them
+                // just logging the errors here and moving on with the workflow. 
                 Logger.Write(TraceEventType.Error, ex.ToString());
             }
             return sysInfo;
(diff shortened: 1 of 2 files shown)


kburtram (approved):
Looks ok.  But it's not ideal to not let the caller know that an error has occurred and some of the return values may be incorrect.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -776,8 +776,9 @@ public static ServerSystemInfo GetServerSystemInfo(IDbConnection connection)
             }
             catch (Exception ex)
             {
+                 // we don't want to fail when an error occurs while fetching these properties.
```
Charles-Gagnon, line 779: Would be good to log WHY we don't want to fail here. (something along the lines of "these properties may not be available for all server types or users and since these properties aren't critical it's ok to not always include them). 

You should also make these properties optional in the typings file on the ADS side so callers know to handle them not being there. (this is assuming that these values default to null, which I think should be how they're handled so callers know whether they weren't retrieved or actually came back as 0 for some reason)

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -763,7 +763,7 @@ public static ServerHostInfo GetServerHostInfo(IDbConnection connection)
         /// Gets the server host cpu count and memory from sys.dm_os_sys_info view
         /// </summary>
         /// <param name="connection">The connection</param>
-        public static ServerSystemInfo GetServerSystemInfo(IDbConnection connection)
+        public static ServerSystemInfo GetServerCpuAndMemoryInfo(IDbConnection connection)
         {
             var sysInfo = new ServerSystemInfo();
```
Charles-Gagnon, line 768: It looks like this just has these values be ints - would suggest making them nullable so that they get sent over as undefined if an error occurs fetching them.

Charles-Gagnon (approved):

PR #1232, opened 2021-08-09
AzureMonitor Fix Dashboard Tables

JustinMDotNet:
Changed results for AzureMonitor GetChildObjects for Dashboard

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoQueryUtils.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/KustoQueryUtils.cs
@@ -91,16 +91,22 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource
                 dictionary[key] = new Dictionary<string, T> {{metadata.Name, metadata}};
             }
         }
-        
-        public static void SafeAdd(this Dictionary<string, List<DataSourceObjectMetadata>> dictionary, string key, DataSourceObjectMetadata node)
+
+        public static void SafeAdd<T>(this Dictionary<string, SortedDictionary<string, DataSourceObjectMetadata>> dictionary, string key,
+            T node) where T : DataSourceObjectMetadata
         {
             if (dictionary.ContainsKey(key))
             {
-                dictionary[key].Add(node);
+                if (dictionary[key].ContainsKey(node.PrettyName))
+                {
+                    return;
+                }
+
+                dictionary[key].Add(node.PrettyName, node);
             }
             else
             {
-                dictionary[key] = new List<DataSourceObjectMetadata> {node};
+                dictionary[key] = new SortedDictionary<string, DataSourceObjectMetadata> {{node.PrettyName, node}};
             }
         }
         
--- a/src/Microsoft.Kusto.ServiceLayer/DataSource/Monitor/MonitorDataSource.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/DataSource/Monitor/MonitorDataSource.cs
@@ -108,8 +108,31 @@ namespace Microsoft.Kusto.ServiceLayer.DataSource.Monitor
 
             if (parentMetadata.MetadataType == DataSourceMetadataType.Cluster && includeSizeDetails)
             {
-                var child = _nodes[parentMetadata.Urn].FirstOrDefault();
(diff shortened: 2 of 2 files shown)


src/Microsoft.Kusto.ServiceLayer/DataSource/Monitor/MonitorDataSource.cs
```diff
@@ -108,8 +108,31 @@ public override IEnumerable<DataSourceObjectMetadata> GetChildObjects(DataSource
 
             if (parentMetadata.MetadataType == DataSourceMetadataType.Cluster && includeSizeDetails)
             {
-                var child = _nodes[parentMetadata.Urn].FirstOrDefault();
-                return child == null ? Enumerable.Empty<DataSourceObjectMetadata>() : _nodes[child.Urn];
+                var tableGroups = _nodes[parentMetadata.Urn];
```
qifahs, line 111: This code does not look specific to the dashboard. When we get children of the cluster in the Object explorer, shouldn't we return the table/function groups, rather than list of tables.

qifahs (changes requested):

src/Microsoft.Kusto.ServiceLayer/DataSource/Monitor/MonitorDataSource.cs
JustinMDotNet: The `includeSizeDetails` is only used for the Dashboard. It doesn't do a great job of describing that. I can rename it if you want.

qifahs: Can't figure that out by reading the code. Can you please add a comment above the if statement about the case this code covers.

src/Microsoft.Kusto.ServiceLayer/DataSource/Monitor/MonitorDataSource.cs
```diff
@@ -108,8 +108,31 @@ public override IEnumerable<DataSourceObjectMetadata> GetChildObjects(DataSource
...
+
+                if (tableGroups == null)
+                {
+                    return Enumerable.Empty<DataSourceObjectMetadata>();
+                }
+
+                var results = new SortedList<string, DataSourceObjectMetadata>(StringComparer.OrdinalIgnoreCase);
+                foreach (var tableGroup in tableGroups)
```
qifahs, line 119: You could use some Linq statements here to make this more concise. Something to do for the future.

src/Microsoft.Kusto.ServiceLayer/DataSource/Monitor/MonitorDataSource.cs
```diff
@@ -108,8 +108,31 @@ public override IEnumerable<DataSourceObjectMetadata> GetChildObjects(DataSource
...
+                {
+                    var tables = _nodes[tableGroup.Urn];
+                    foreach (var table in tables)
+                    {
+                        if (results.ContainsKey(table.PrettyName))
```
qifahs, line 124: Probably better to do this with a Dictionary. ContainsKey is a O(Log n) operation for SortedList, and approaches  O(1) for  Dictionary. Dictionary will be much faster for a large number of tables.

qifahs (approved):

src/Microsoft.Kusto.ServiceLayer/DataSource/KustoQueryUtils.cs
```diff
@@ -91,16 +91,22 @@ public static bool  IsClusterLevelQuery(string query)
...
-        
-        public static void SafeAdd(this Dictionary<string, List<DataSourceObjectMetadata>> dictionary, string key, DataSourceObjectMetadata node)
+
+        public static void SafeAdd<T>(this Dictionary<string, SortedDictionary<string, DataSourceObjectMetadata>> dictionary, string key,
+            T node) where T : DataSourceObjectMetadata
         {
             if (dictionary.ContainsKey(key))
             {
-                dictionary[key].Add(node);
+                if (dictionary[key].ContainsKey(node.PrettyName))
```
qifahs, line 100: Minor, but the return statement can be avoided with a negative check:
if(!dictionary[key]....)
{
    dictionary[key].Add(....);
}

qifahs (approved):

PR #1236, opened 2021-08-23
Added check if metadata is empty to throw error exception.

JustinMDotNet:


JustinMDotNet:
This code change fixes this issue:
https://github.com/microsoft/azuredatastudio/issues/16816

PR #1237, opened 2021-08-25
Enable scripting for subobjects such as triggers

kburtram:
Part of the fix for https://github.com/microsoft/azuredatastudio/issues/2093

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/ObjectMetadata.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/ObjectMetadata.cs
@@ -30,6 +30,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Metadata.Contracts
         public string Schema { get; set; }
 
         public string Name { get; set; }
+
+        public string ParentName { get; set; }
         
         public string Urn { get; set; }
     }
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
@@ -105,9 +105,9 @@ namespace Microsoft.SqlTools.ServiceLayer.Scripting
             urnBuilder.AppendFormat("Server[@Name='{0}']/", server.ToUpper());
             urnBuilder.AppendFormat("Database[@Name='{0}']/", Urn.EscapeString(database));
 
-            if (!string.IsNullOrWhiteSpace(scriptingObject.Table))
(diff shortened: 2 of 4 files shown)


kburtram:
...looking at test failure.

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -100,15 +101,31 @@ public static Urn ToUrn(this ScriptingObject scriptingObject, string server, str
...
+                if (!string.IsNullOrWhiteSpace(scriptingObject.Schema))
+                {
+                    urnBuilder.AppendFormat(" and @Schema = '{0}'", Urn.EscapeString(scriptingObject.Schema));
+                }
+                urnBuilder.Append("]/");
+            }
+
+            urnBuilder.AppendFormat("{0}[@Name='{1}'", scriptingObject.Type, Urn.EscapeString(scriptingObject.Name));
+
+            // add schema to object only if there is no parent object specified
```
Charles-Gagnon, line 120: Would be good to explain why we wouldn't add the schema in this case

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingExtensionMethods.cs
```diff
@@ -100,15 +101,31 @@ public static Urn ToUrn(this ScriptingObject scriptingObject, string server, str
...
-                scriptingObject.Schema != null ? string.Format("and @Schema = '{0}'", Urn.EscapeString(scriptingObject.Schema)) : string.Empty);
-
-            return new Urn(urn);
+            StringBuilder urnBuilder = new StringBuilder();
+            urnBuilder.AppendFormat("Server[@Name='{0}']/", server.ToUpper());
+            urnBuilder.AppendFormat("Database[@Name='{0}']/", Urn.EscapeString(database));
+
+            if (!string.IsNullOrWhiteSpace(scriptingObject.ParentName))
+            {
+                urnBuilder.AppendFormat("Table[@Name='{0}'", Urn.EscapeString(scriptingObject.ParentName));
```
Charles-Gagnon, line 110: So parent objects are always going to be tables?

src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/ObjectMetadata.cs
```diff
@@ -30,6 +30,8 @@ public class ObjectMetadata
         public string Schema { get; set; }
```
Charles-Gagnon, line 30: What is the MetadataType set to for things like triggers that aren't in the MetadataType enum?

PR #1243, opened 2021-09-09
migrating SSMS settings

### LINKED ISSUE #392: sqltoolservice is broken on Ubuntu 17.04
The service as used by vscode-mssql and mssql-scripter is broken on Ubuntu 17.04. This also includes the alpha build that uses .NET Core 2.0 preview as used in the alpha builds of vscode-mssql.

cavonac:
This fixes #392 and raises the default amount of characters returned per column in a grid to 2MB

SSMS and STS were out of sync with the chars to store and present to the grid. Instead of 65K, it should be 2MB. Will need to make sure this is performant.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
@@ -17,6 +17,14 @@ namespace Microsoft.SqlTools.ServiceLayer.SqlContext
     {
         #region Constants
 
+        #region WellKnownNumbers
+        internal const int int64K = 65535;
+        internal const int int1MB = 1048576;
+        internal const int int2MB = 2097152;
+        internal const int int5MB = 5242880;
+        internal const int int2GB = 2147483647;
+        #endregion
+
         /// <summary>
         /// Default value for batch separator (de facto standard as per SSMS)
         /// </summary>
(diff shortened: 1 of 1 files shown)


cavonac:
I opted to simply increase the value because it has the intended effect, the settings were out of sync, and there were very little changes to make to solve the issue. There may be more work to be done with ADS to pass along and set different configuration values everywhere. Now, STS will at least match what SSMS has.

Charles-Gagnon:
Whatever happened to https://github.com/microsoft/azuredatastudio/pull/16955?

cavonac:
> Whatever happened to [microsoft/azuredatastudio#16955](https://github.com/microsoft/azuredatastudio/pull/16955)?

I cancelled it and withdrew the PR as it was not what I was looking for. Removed the branch to clean it up. It was a draft anyways

src/Microsoft.SqlTools.ServiceLayer/SqlContext/QueryExecutionSettings.cs
```diff
@@ -25,13 +33,13 @@ public class QueryExecutionSettings : GeneralRequestDetails
         /// <summary>
         /// Default number of chars to store for long text fields (de facto standard as per SSMS)
         /// </summary>
-        private const int DefaultMaxCharsToStore = 65535; // 64 KB - QE default
+        private const int DefaultMaxCharsToStore = int2MB; // #392: Raised from 64K to 2MB as found in SSMS
```
kburtram, line 36: we probably don't need the comment regarding the history of the value.  anyone interested where the value came from can always check the source control history.

kburtram (changes requested):
Do we also need a corresponding change to only send a certain number of characters to the front-end?  i.e. are we now going to be sending 2MB to frontend instead of 64KB, and if so won't that negatively impact perf in the UI?

kburtram:
The other thing to test while fixing this bug is to confirm the user setting for this value is respected.  So if the user increases or decreases the setting then that has the expected impact on the size of the exported data.

Microsoft-code (approved):

kburtram:
Closing older stale PRs.  Thanks!

PR #1245, opened 2021-09-16
Kusto bubble up Auth exceptions

JustinMDotNet:
Changed RefreshAuthToken in ConnectionService and KustoClient to follow TryParse pattern.

Auth exceptions will now show in the error dialog instead of a null reference exception.

image

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
@@ -213,9 +213,13 @@ namespace Microsoft.Kusto.ServiceLayer.Connection
             return completeParams;
         }
 
-        internal string RefreshAuthToken(string ownerUri)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -213,9 +213,13 @@ public async Task<ConnectionCompleteParams> Connect(ConnectParams connectionPara
             return completeParams;
         }
 
-        internal string RefreshAuthToken(string ownerUri)
+        internal bool TryRefreshAuthToken(string ownerUri, out string token)
         {
-            TryFindConnection(ownerUri, out ConnectionInfo connection);
+            token = string.Empty;
+            if (!TryFindConnection(ownerUri, out ConnectionInfo connection))
```
JustinMDotNet, line 219: If there is no existing connection setup then we return false.

qifahs (approved):

Issue #1244, opened 2021-09-16
Calling Cancel on a running query causes execution of handlecancelrequest to fail


smartguest:
When you click cancel from ADS, the task HandleCancelRequest is called, which then tries to call Cancel on the Query object. This then attempts to call Cancel on the cancellationtokensource. However after this, the rest of the code fails to execute and we do not get a response back from the task, causing ADS query run to hang. This needs to be investigated as to why this is not working.

kburtram:
@smartguest I've assigned this issue to you since I assume this is part of what you're looking at as part of https://github.com/microsoft/azuredatastudio/issues/3231.  Please let me know if this isn't related to the Cancelation bug you're looking at in ADS.

smartguest:
@kburtram This is related to it. Also it seems that the design of SqlClient currently does not allow for queries to be cancelled immediately if it goes into a infinite loop or long running query loops: caused by a deadlock of some sort. 

This is not the case with Sql Server apparently, which uses a synchronous approach. This means we have to rewrite the code of SqlServerTools to use the approach used by Sql Server Management Studio, though this might cause issues.

Should I go ahead with rewriting the code of STS to use the SSMS approach? The SqlClient used by STS does not currently allow for infinite or very long running queries to be cancelled at the moment.

Here are two examples of code

https://github.com/dotnet/SqlClient/issues/44

https://stackoverflow.com/questions/48461567/canceling-query-with-while-loop-hangs-forever?noredirect=1#comment83955305_48461567

kburtram:
@smartguest SSMS also uses SqlClient and is able to cancel queries correctly.  I'm not sure what rewriting STS to use SSMS approach entails.  Maybe better to discuss on a call.  But at a high-level we should correct the STS to properly support cancelation.

Also, please note the category of ADS cancelation bugs are not primarily related to infinite loop scenarios.  So if that one case is particularly complicated for some reason you may also want to understand why ADS can't properly cancel other types of queries as well.

kburtram:
...I see some notes about synchronous code as being difference between SSMS/STS.  SSMS uses many threads and I don't believe the query execution is occurring on the UI thread, though I'd have to double check to be sure.  Perhaps let's look through what code you're referring to get better context on that comment.

smartguest:
The code where execution happens is in Batch.cs, this one uses an async task and readers to execute the command. According to the links I posted, the async functions do not work properly for infinite loop or long running queries from my understanding.

[...]

PR #1246, opened 2021-09-17
Fix for cancel query by adding synchronous readers.

smartguest:
Due to an ongoing known issue that affects the asynchronous reader in SqlParser, it cannot cancel any long running query that isn't expected to end shortly or is an infinite loop scenario. 

Since this is a common issue among many users and does not reproduce on SQL Server Management Studio due to it not using the asynchronous reader. I have replaced the problematic async reader with the synchronous version. 

Queries still execute the same so far without any major changes, but infinite running queries are now cancelled properly when run from the ADS side. I have tested this with a select top query and an infinite loop query scenario so far.

This addresses: https://github.com/microsoft/azuredatastudio/issues/3231 and https://github.com/microsoft/sqltoolsservice/issues/1244

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -350,7 +350,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
             {
                 try
                 {
-                    await ExecuteOnce(conn, cancellationToken);
+                    await SyncExecuteOnce(conn, cancellationToken);
                 }
                 catch (DbException dbe)
                 {
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -350,7 +350,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
             {
                 try
                 {
-                    await SyncExecuteOnce(conn, cancellationToken);
+                    await ExecuteOnce(conn, cancellationToken);
                 }
                 catch (DbException dbe)
                 {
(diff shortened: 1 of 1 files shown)


alanrenmsft:
I am not sure whether this is the correct fix. async/cancellation token is the right approach. instead of changing to sync data reader, have you tested why is the cancellation not working previously?

smartguest:
@alanrenmsft @Charles-Gagnon Yes, I have tested the issue extensively, debugging the code, the execution always stops at that line to call cancel on the dbcommand without any warning or error whatsoever, and there is no error returned, it seems the query on the server, continues to execute while the call to cancel is ignored entirely or hangs forever.  

According to the reports on this bug:

https://github.com/dotnet/SqlClient/issues/44#issuecomment-362757646
https://github.com/dotnet/SqlClient/issues/44#issuecomment-494950240

SqlClient async watch is unable to handle any infinite loop scenarios due to a race condition occuring between the cancellation and closing in SqlCommand.

as @saurabh500 says:

"In this case the query is executed on the server, but it will never complete as it is an infinite loop.
In the client, after sending a query for execution, the client waits for an acknowledgement from the server by locking the TdsParserStateObject and then waiting for a network packet.
How cancellation works is, the client is responsible for making sure that it completes the existing requests and then sends a cancellation request to the server. Since the existing request is never going to complete, the cancellation will not go through either.
The hang happens in case of queries like above which are never expected to complete, or if you try to cancel an operations while streaming endless data from SqlServer (like XEvents) and try to close the reader before cancelling.

This is essentially a race condition between Reader.Close and Reader.Cancel"

@Wraith2 says that:

At the moment a lock is taken in EndExecuteNonQuery and then in TdsParserStateObject.Cancel the lock is attempted and cancellation isn't allowed if it can't be acquired.

The only workaround so far as suggested by @cheenamalhotra 
https://github.com/dotnet/SqlClient/issues/44#issuecomment-791853924

Is to either use Timeout or use Sync APIs to directly cancel.

Charles-Gagnon (approved):
It looks like a fix in sqlclient is merged in now (https://github.com/dotnet/SqlClient/pull/956) but given our experiences with taking brand new sqlclient releases doing this fix for now seems reasonable. 

This shouldn't impact performance as the queries are spun on their own thread anyways and aren't awaited from what I can see.

smartguest:
@Charles-Gagnon I'm happy to hear that SqlClient has fixed this issue. Until we put the latest one into our code, I think using sync readers is fine for now.

alanrenmsft:
how do we cancel the query execution now? just ignore it? does it mean we have a thread that never stops execution?(or the execution timeout will be used in the case?)

also, we have to make sure when a query is canceled it won't send anything back to ADS.

Wraith2:
> given our experiences with taking brand new sqlclient releases

That sounds like bad experiences, what happened and was it preventable from the SqlClient side?

smartguest:
We can still call cancel on the sync reader @alanrenmsft without any issue when the cancellationtoken is cancelled as there is a small callback in the function that calls cancel on the DbCommand,, and it seems to work as expected with the query cancelling. It's just that the async reader does not respond at all to this command to cancel at all.

Charles-Gagnon:
@alanrenmsft What do you mean? The cancellation token will work as expected now since we're using the sync functions - that's the fix here (the cancellation was being ignored previously as the bugs that Alex linked to said)

If a cancel isn't requested then yes, the query can execute infinitely. It should generally be up to the caller to handle that and add an appropriate timeout if needed though I would say.

@Wraith2 - There was a fix earlier (https://github.com/dotnet/SqlClient/pull/796) that we updated sqlclient for. But the initial release with this fix ended up having another bug in it (https://github.com/dotnet/efcore/issues/19842 I believe). This issue was fixed quickly but we just happened to pick it up before that issue was noticed/fixed and so ended up releasing with that version and immediately had a bunch of customers hit the issue. So something like that is hard to make completely preventable - you can always add more tests (and ones were) but given that it's hard to test what you don't know is broken I think this was just a case of bad timing and the perfect storm of events.

PR #1251, opened 2021-09-28
Kusto Format Fix

JustinMDotNet:
Added Format function to DataSourceFactory. Removed unsued code / commented code in TSqlFormatterService.

This change resolves this issue:
https://github.com/microsoft/azuredatastudio/issues/17159

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/Formatter/TSqlFormatterService.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/Formatter/TSqlFormatterService.cs
@@ -6,8 +6,8 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.IO;
 using System.Threading.Tasks;
+using Microsoft.Kusto.ServiceLayer.DataSource;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Kusto.ServiceLayer/Formatter/TSqlFormatterService.cs
```diff
@@ -18,6 +18,7 @@
 using Microsoft.Kusto.ServiceLayer.Workspace;
 using Microsoft.Kusto.ServiceLayer.Workspace.Contracts;
 using Microsoft.SqlTools.Utility;
+using FormattingOptions = Microsoft.Kusto.ServiceLayer.Formatter.Contracts.FormattingOptions;
 using Range = Microsoft.Kusto.ServiceLayer.Workspace.Contracts.Range;
 
 namespace Microsoft.Kusto.ServiceLayer.Formatter
```
Charles-Gagnon, line 24: You should probably rename this while you're here since this isn't T-SQL...

gupta1 (approved):

PR #1252, opened 2021-09-28
added preserve user number format and date/time format along with dotnet bump.

smartguest:
When the locale is changed for ADS, it overrides the user set date/time and number format with whatever is used by the language in ADS, this means user preferences are lost unless the user is using a language pack that has the same region as the user's system, this PR preserves these preferences while also letting the language to be changed (these are the two properties that can be changed on a CultureInfo object)

This addresses:

https://github.com/microsoft/azuredatastudio/issues/16617

and 

https://github.com/microsoft/azuredatastudio/issues/16438


global.json
```diff
@@ -1,5 +1,5 @@
 {
     "sdk": {
-        "version": "5.0.301"
+        "version": "5.0.401"
```
Charles-Gagnon, line 3: Why did you update this?

external/Microsoft.SqlTools.Hosting.v2/Utility/CommandOptions.cs
```diff
@@ -144,12 +144,22 @@ public virtual void SetLocale(string locale)
         /// culture-specific messages
         /// </summary>
         /// <param name="locale"></param>
-        internal void LocaleSetter(string locale)
+        internal void LocaleSetter(string locale, Boolean useUserSettings = false)
```
Charles-Gagnon, line 147: What is the default for? Is there a scenario where the default would be used?

external/Microsoft.SqlTools.Hosting.v2/Utility/CommandOptions.cs
```diff
@@ -144,12 +144,22 @@ public virtual void SetLocale(string locale)
...
         {
             // Creating cultureInfo from our given locale
             CultureInfo language = new CultureInfo(locale);
             Locale = locale;
 
+            if (useUserSettings)
+            {
+                // Allow the system set Number Format and Date Format to be preserved when changing the locale.
+                NumberFormatInfo NumberFormat = CultureInfo.CurrentCulture.NumberFormat;
```
Charles-Gagnon, line 156: This is going to override the format even for languages that we support, right? @kburtram is that what we want? I guess it seems potentially reasonable to say that the languages in ADS are purely for localization purposes and globalization settings are always going to be taken from the system but I'm not sure that's the more intuitive.

@smartguest What does VS Code do for numbers displayed?

global.json
smartguest: I encountered an issue where I had a missing dotnet installation when I attempted to run dotnet restore, I tried to find 5.0.301 but it wasn't available for download anymore.

external/Microsoft.SqlTools.Hosting.v2/Utility/CommandOptions.cs
smartguest: Likely not, but I put this in just in case if it won't break anything, I'll remove it

external/Microsoft.SqlTools.Hosting.v2/Utility/CommandOptions.cs
smartguest: Yes this will override the format for languages that we support, but again users seem to prefer it.

global.json
Charles-Gagnon: 5.0.7 from https://dotnet.microsoft.com/download/dotnet/5.0 seems to be the one.

Updating this is something we should probably do, but please don't include unrelated changes like this into PRs. Create a new PR for stuff like this.

external/Microsoft.SqlTools.Hosting.v2/Utility/CommandOptions.cs
Charles-Gagnon: Where are you getting your "users seem to prefer it" statement from? I'm slightly worried about making this change and having people get upset because it changed what they were used to (if they had a different setting in their system than ADS).

Maybe not a huge issue, but something to consider when making a change that's potentially going to affect a lot of our users.

smartguest: Karl told me that we should respect the user's region settings and that the current behavior is incorrect in regards to user expectations.

Charles-Gagnon: Alright, if we get complaints we can look into adding a setting then.

Charles-Gagnon (approved):

PR #1253, opened 2021-09-29
Kusto - Added code coverage to Build

JustinMDotNet:
Modified build.yml to run code coverage and modified logfilename to nunit.trx

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -213,7 +213,13 @@ steps:
     allowPackageConflicts: true
 
 - task: PublishTestResults@2
-  displayName: 'Publish Test Results **/xunit.trx'
+  displayName: 'Publish Test Results **/SqlTools.ServiceLayer.UnitTests.trx'
   inputs:
     testResultsFormat: VSTest
-    testResultsFiles: '**/xunit.trx'
+    testResultsFiles: '**/SqlTools.ServiceLayer.UnitTests.trx'
+
+- task: PublishTestResults@2
+  displayName: 'Publish Test Results **/Kusto.ServiceLayer.UnitTests.trx'
+  inputs:
+    testResultsFormat: VSTest
+    testResultsFiles: '**/Kusto.ServiceLayer.UnitTests.trx'
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -72,14 +72,14 @@ steps:
   inputs:
     command: test
     projects: test/Microsoft.SqlTools.ServiceLayer.UnitTests
-    arguments: '--logger "trx;LogFileName=xunit.trx"'
+    arguments: '--logger "trx;LogFileName=SqlTools.ServiceLayer.UnitTests.trx" --collect "Code coverage"'
 
 - task: DotNetCoreCLI@2
   displayName: 'dotnet test test/Microsoft.Kusto.ServiceLayer.UnitTests'
   inputs:
     command: test
     projects: test/Microsoft.Kusto.ServiceLayer.UnitTests
-    arguments: '--logger "trx;LogFileName=nunit.trx" --collect "Code coverage"'
+    arguments: '--logger "trx;LogFileName=Kusto.ServiceLayer.UnitTests.trx" --collect "Code coverage"'
 
 - task: Npm@1
   displayName: 'npm install -g gulp-cli'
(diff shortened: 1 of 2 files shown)


azure-pipelines/build.yml
```diff
@@ -79,7 +79,7 @@ steps:
   inputs:
     command: test
     projects: test/Microsoft.Kusto.ServiceLayer.UnitTests
-    arguments: '--logger "trx;LogFileName=xunit.trx"'
+    arguments: '--logger "trx;LogFileName=nunit.trx" --collect "Code coverage"'
```
Charles-Gagnon, line 82: What was the issue here? Is this something that should also be done for SqlTools.ServiceLayer.UnitTests?

JustinMDotNet: It could be done for the SqlTools Service. This generates the code coverage for unit tests for the project. I changed the LogFileName so it wasn't the same as SqlTools and also because Kusto uses nunit not xunit.

Charles-Gagnon: Would suggest naming it `Kusto.ServiceLayer.UnitTests.trx` or something then to make it that much clearer.

Would you mind adding this to the Sql Tools service layer as well (and updating the name similarly there as well)? Seems like a good thing to have.

JustinMDotNet: Here's a link to this build. It now contains a "Code Coverage" tab and has a Coverage stat with the test pass stat.

image


https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=121958&view=results

Charles-Gagnon (approved):
Thanks for adding the sql tools stuff!

azure-pipelines/build.yml
```diff
@@ -210,10 +210,4 @@ steps:
     command: push
     packagesToPush: '$(Build.SourcesDirectory)/artifacts/nugetPackages/**/*.nupkg;!$(Build.ArtifactStagingDirectory)/**/*.symbols.nupkg'
     publishVstsFeed: '2191dd5f-4aec-491b-ac50-568bbc331c8a'
-    allowPackageConflicts: true
-
-- task: PublishTestResults@2
```
Charles-Gagnon, line 215: Why are you removing this?

Charles-Gagnon (changes requested):

azure-pipelines/build.yml
JustinMDotNet: I'm looking through this doc on Azure Pipelines and it seems like it's not necessary because it's being done by the Microsoft.NET.Test.SDK reference in the test projects. I linked the doc below. Atm I'm just pushing stuff and testing builds to ensure everything still works.

https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/dotnet-core?view=azure-devops&tabs=dotnetfive#collect-code-coverage

JustinMDotNet: Atm I'm running into an issue with the code coverage for the Kusto.ServiceLayer and Sql.ServiceLayer being combined into a single file

Charles-Gagnon: I see. Could you convert this into a draft PR until it's ready to review then?

azure-pipelines/build.yml
```diff
@@ -72,14 +72,16 @@ steps:
...
+    testRunTitle: SqlTools.ServiceLayer.UnitTests
+    arguments: '--configuration $(buildConfiguration) --collect "Code coverage"'
 
 - task: DotNetCoreCLI@2
   displayName: 'dotnet test test/Microsoft.Kusto.ServiceLayer.UnitTests'
   inputs:
     command: test
     projects: test/Microsoft.Kusto.ServiceLayer.UnitTests
-    arguments: '--logger "trx;LogFileName=xunit.trx"'
+    testRunTitle: Kusto.ServiceLayer.UnitTests
```
JustinMDotNet, line 83: testRunTitle sets the name of the test

image

azure-pipelines/build.yml
JustinMDotNet: The reason I started looking into this further is because I was seeing this warning during the builds

image

Here's a link to that build.
https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=122194&view=results

Charles-Gagnon (approved):

PR #1255, opened 2021-10-02
set issupportedVersion flag

alanrenmsft:
the ADS side change is: https://github.com/microsoft/azuredatastudio/pull/17219

This PR implements the supported version check in STS.

I am working with the PM and docs team to create a page that documents the SQL Servers that ADS supports, once it is available I will come back and update the message property.

this is what i am getting when trying to connect to a sql 2012 server 
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionCompleteNotification.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionCompleteNotification.cs
@@ -52,6 +52,16 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
         /// The type of connection that this notification is for
         /// </summary>
         public string Type { get; set; } = ConnectionType.Default;
+
+        /// <summary>
+        /// Gets or sets a boolean value indicates whether the current server version is supported by the service.
+        /// </summary>
+        public bool IsSupportedVersion { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionCompleteNotification.cs
```diff
@@ -52,6 +52,16 @@ public class ConnectionCompleteParams
...
+
+        /// <summary>
+        /// Gets or sets a boolean value indicates whether the current server version is supported by the service.
+        /// </summary>
+        public bool IsSupportedVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets the additional warning message about the unsupported server version.
+        /// </summary>
+        public string UnsupportedVersionMessage { get; set; }
```
Charles-Gagnon, line 64: Not sure why you added this if you're not using it here (and it's pretty unlikely to ever be used with how things are now...)

alanrenmsft: as I mentioned in the description, I am working with Alan Yu on this, hopefully next week sometime we will have the web page.

kburtram (approved):
LGTM, but don't you need to set the message string somewhere?

Charles-Gagnon (approved):

alanrenmsft:
the common message is set in the ADS, extensions can provide additional messages.

PR #1258, opened 2021-10-05
Kusto connection map to ConnectionManager

JustinMDotNet:
Created ConnectionManager and moved connectionMap. Changed references from ConnectionService to ConnectionManager.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionManager.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionManager.cs
@@ -1,0 +1,39 @@
+using System.Collections.Concurrent;
+using System.Composition;
+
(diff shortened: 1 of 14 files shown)


src/Microsoft.Kusto.ServiceLayer/Connection/ConnectionManager.cs
```diff
@@ -0,0 +1,39 @@
+using System.Collections.Concurrent;
+using System.Composition;
+
+namespace Microsoft.Kusto.ServiceLayer.Connection
+{
+    [Export(typeof(IConnectionManager))]
+    public class ConnectionManager : IConnectionManager
```
JustinMDotNet, line 7: New class to hold all connections separate from the Connection Service. Separating the responsibilities of communicating for connections (`ConnectionService`) and maintaining connections (`ConnectionManager`)

PR #1259, opened 2021-10-05
Add error for sql bindings when .net 5

kisantia:
The attribute used to specify Azure Functions was changed in .NET 5 to `Function` instead of `FunctionName` in .NET 3.1. Since sql bindings aren't currently supported for .NET 5, this will show the appropriate error instead of "No Azure Functions in the current file" as reported in https://github.com/microsoft/azuredatastudio/issues/17252.

image


chlafreniere:
Out of curiosity, is there a way to just hide the command when the project is .net 5?

Charles-Gagnon (approved):
Since we aren't planning on keeping this around forever this seems fine as a short-term solution. Hiding the command itself would probably be pretty difficult since the only way I can think of doing that is to do some parsing and set a context key based on that (so probably not impossible, but questionably worth it)

PR #1272, opened 2021-10-17
handle add column request

alanrenmsft:
1. handle add column request
2. handle the deserialization of the property information


src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -126,8 +126,14 @@ public void Stop()
                     TParams typedParams = default(TParams);
                     if (requestMessage.Contents != null)
                     {
-                        // TODO: Catch parse errors!
-                        typedParams = requestMessage.Contents.ToObject<TParams>();
+                        try
```
alanrenmsft, line 129: previously the STS will crash if the deserialization of request parameter fails, I am catching the error here to stop it from crashing and also send the error back to ADS.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Components/TableProperties.cs
```diff
@@ -23,11 +32,42 @@ public class TableProperties<T> : ComponentPropertiesBase
         /// <summary>
         /// All properties of the object.
         /// </summary>
-        public DesignerDataPropertyInfo[] ItemProperties { get; set; }
+        public List<DesignerDataPropertyInfo> ItemProperties { get; set; }
```
Charles-Gagnon, line 35: I believe you should also be able to just do 

`public List<DesignerDataPropertyInfo> ItemProperties { get; set; } = new List<DesignerDataPropertyInfo>();` 

as of a recent-ish C# update.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/DataModel/ObjectDataModelBase.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// The base class for data model object.
+    /// </summary>
+    public abstract class ObjectDataModelBase
+    {
+        public ObjectDataModelBase()
+        {
+            this.Name = new InputBoxProperties();
+        }
+
+        public InputBoxProperties Name { get; set; }
```
Charles-Gagnon, line 18: `public InputBoxProperties Name { get; set; } = new InputBoxProperties();`

Although it seems a bit odd to have something as generic as "Name" actually be an input box. `NameInputBox` would be clearer...(or even have both since I would have expected name would just be a string you could access regardless...)

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/DataModel/TableColumnDataModel.cs
```diff
@@ -3,14 +3,23 @@
...
     /// <summary>
     /// The data model of a table column object
     /// </summary>
-    public class TableColumn
+    public class TableColumnDataModel : ObjectDataModelBase
     {
-        public InputBoxProperties Name { get; set; }
+        public TableColumnDataModel()
+        {
+            this.Type = new DropdownProperties();
```
Charles-Gagnon, line 17: See other comments about auto-property initializers. 

Although I also wonder why you even have the property getters/setters defined if they're both public. Why not just something like

`public DropdownProperties Type = new DropdownProperties();`

Same comment about the names and/or types here though, `TypeDropdown` seems better for example and/or having Type be just a straight string.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/DataModel/TableDataModel.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class TableColumnCollection : TableProperties<TableColumnDataModel>
+    {
+        [JsonIgnore]
+        protected override string NewObjectNamePrefix { get { return "column"; } }
+
+        protected override TableColumnDataModel CreateNew(string name)
+        {
+            //TODO: Add the default values
+            var column = new TableColumnDataModel();
+            column.Name.Value = this.GetDefaultNewObjectName();
```
Charles-Gagnon, line 40: The fact that this is tying the model to the specific type of UI to use doesn't seem like something we normally want. Why isn't the model just the current values for the object that gets updated when a user selects a new one?

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -82,6 +83,16 @@ private async Task HandleProcessTableDesignerEditRequest(ProcessTableDesignerEdi
                            try
                            {
                                // TODO
+                               switch (requestParams.TableChangeInfo.Type)
+                               {
+                                   case DesignerEditType.Add:
+                                       this.HandleAddItemRequest(requestParams);
+                                       break;
+                                   case DesignerEditType.Remove:
```
Charles-Gagnon, line 91: I'm assuming the TODO is about the remove (and maybe others?)

Adding some context there would be helpful

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -111,6 +122,37 @@ private async Task HandleSaveTableChangesRequest(SaveTableChangesRequestParams r
                        });
         }
 
+        private void HandleAddItemRequest(ProcessTableDesignerEditRequestParams requestParams)
+        {
+            var property = requestParams.TableChangeInfo.Property;
+            if (property.GetType() == typeof(string))
```
Charles-Gagnon, line 128: No error or warning if it's not a string?

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -111,6 +122,37 @@ private async Task HandleSaveTableChangesRequest(SaveTableChangesRequestParams r
...
+            {
+                string propertyName = property as string;
```
Charles-Gagnon, line 130: You can also just do 

``` csharp
string propertyName = property as string;
if(propertyName != null)
{
...
```

That's typically how I see these types of checks being done.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -111,6 +122,37 @@ private async Task HandleSaveTableChangesRequest(SaveTableChangesRequestParams r
...
+                        break;
+                    default:
+                        break;
+                }
+            }
+        }
+
+        private List<string> GetSupportedColumnTypes(TableInfo tableInfo)
+        {
+            //TODO:
```
Charles-Gagnon, line 144: Please add a comment explaining what needs to be done

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -111,6 +122,37 @@ private async Task HandleSaveTableChangesRequest(SaveTableChangesRequestParams r
...
+            var columnTypes = new List<string>();
+            return columnTypes;
+        }
+
+        private List<string> GetSchemas(TableInfo tableInfo)
+        {
+            //TODO:
```
Charles-Gagnon, line 151: Comment

test/Microsoft.SqlTools.ServiceLayer.UnitTests/TableDesigner/TableDesignerChangeInfoTest.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+using Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts;
+using NUnit.Framework;
+using Newtonsoft.Json;
+
+namespace Microsoft.SqlTools.ServiceLayer.UnitTests.TableDesigner
+{
+    public class TableDesignerChangeInfoTest
+    {
+        [Test]
```
Charles-Gagnon, line 15: Glad to see a test, but having a few more cases here would be grand.

(plus tests for the other parts of your code)

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/DataModel/ObjectDataModelBase.cs
alanrenmsft: not quite following this: ```Why isn't the model just the current values for the object that gets updated when a user selects a new one?```

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
alanrenmsft: I'll be more specific about what to do, all the TODOs here are essential and will be implemented in a few weeks.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/TableDesigner/TableDesignerChangeInfoTest.cs
alanrenmsft: so far majority of the the code I added are just entities to match the ADS side. I guess the only one other logic that is not test covered is adding new column, I'll have a test for that.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/TableChangeInfo.cs
```diff
@@ -27,8 +28,41 @@ public class TableDesignerChangeInfo
...
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return true;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            object property;
+            if (reader.TokenType == JsonToken.StartObject)
```
Charles-Gagnon, line 50: Make sure you're handling parse failures gracefully

Charles-Gagnon (approved):

PR #1274, opened 2021-10-20
load the schema information

alanrenmsft:
load the available schemas. SSDT doesn't support this information, schema cannot be selected in SSDT table designer but available in SSMS, so I look into the code and get the query from there.

ADS
image

SSMS
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.Data.SqlClient;
 using Microsoft.SqlTools.Hosting.Protocol;
+using Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
 using Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -170,10 +174,57 @@ private List<string> GetSupportedColumnTypes(TableInfo tableInfo)
...
+            string schema;
+            if (tableInfo.IsNewTable)
+            {
+                schema = schemas.Contains("dbo") ? "dbo" : schemas[0];
+            }
+            else
+            {
+                schema = tableInfo.Schema;
+            }
+            tableViewModel.Schema.Value = schema;
```
Charles-Gagnon, line 190: Could also just remove the schema var and set this directly in the if/else

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -170,10 +174,57 @@ private List<string> GetSupportedColumnTypes(TableInfo tableInfo)
...
+            // TODO: set the view information
+            var view = new TableDesignerView();
+            return view;
+        }
+
         private List<string> GetSchemas(TableInfo tableInfo)
         {
-            //TODO: get the schemas.
             var schemas = new List<string>();
+            using (var connection = new SqlConnection(tableInfo.ConnectionString))
```
Charles-Gagnon, line 212: We should try to avoid having boilerplate code like this in each function that needs to make a query. I believe the ExecuteReader method in the ReliableConnectionHelper can be useful here - or if you can't find something similar that does what you need then I'd suggest making one that does and handles all the creating connection/opening/etc

PR #1275, opened 2021-10-23
Skip unused build steps

cssuh:
Skipping the x86, arm x86 and arm x64 builds
I was not able to find any size changes in the files:
Screen Shot 2021-10-22 at 5 28 34 PM
Screen Shot 2021-10-22 at 5 27 59 PM

according to [this documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#capabilities-and-limitations) we will have to use self-hosted agents if we want to increase the disk-space.

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -179,19 +179,19 @@ steps:
     includeRootFolder: false
     archiveFile: '$(Build.SourcesDirectory)/artifacts/package/Microsoft.SqlTools.ServiceLayer-win-x64-net5.0.zip'
 
-- task: ArchiveFiles@1
-  displayName: 'Archive windows 32 bit build'
-  inputs:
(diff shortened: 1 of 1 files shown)


alanrenmsft (approved):

kburtram (approved):
LGTM but can we skip all the architectures we don't use (i.e. x86, win arm x86, and win arm x64. IIRC).

Charles-Gagnon (commented):
Not sure what you were trying to say with your screenshot - that's just saying what the size of the zipped archive is (175358184 bytes ~= 168MB). But there's a whole bunch of other stuff that happens before it zips it up so even if the zip didn't change size (which it appears it didn't) then it just may be a previous step generating a bunch more stuff. If we wanted to figure out what it was (which wouldn't be a bad idea...) I'd suggest finding a tool to use that can scan and report the disk size and then doing an ad-hoc build with that to see what folder is taking up the most space.

In general though I don't think skipping the zip step is going to help much, we should instead be looking to skip building those arch's completely in the first place.

kburtram (approved):

PR #1277, opened 2021-11-02
STS version updates to support DacFx_160

ssreerama:
DacFx now has SqlTasks project moved from SSDT to its repo and have done plenty of changes along with vbump to 160, to support the changes, some of the STS dlls version have to be updated and these changes are targeting for upcoming release.

All integration and unit tests are passed in STS and ADS, and manual testing is also done. (no regressions found).

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -20,12 +20,12 @@
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
 		<PackageReference Update="Microsoft.Data.SqlClient" Version="3.0.0"/>
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="161.46367.54" />
-		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="160.5317.5-preview" GeneratePathProperty="true" />
+		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="160.5323.3-preview" GeneratePathProperty="true" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Data" Version="9.0.4" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Language" Version="9.0.4"/>
 		<PackageReference Update="Microsoft.SqlServer.Assessment"  Version="[1.0.305]" />
 		<PackageReference Update="Microsoft.SqlServer.Migration.Assessment"  Version="1.0.20210902.7" />
-		<PackageReference Update="Microsoft.SqlServer.Management.SqlParser" Version="160.21267.55" />
+		<PackageReference Update="Microsoft.SqlServer.Management.SqlParser" Version="160.21292.55" />
 		<PackageReference Update="Microsoft.Azure.OperationalInsights" Version="1.0.0" />
 		<PackageReference Update="Microsoft.CodeAnalysis.CSharp" Version="3.10.0" />
 		<PackageReference Update="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.10.0" />
--- a/Packages.props
+++ b/Packages.props
@@ -18,9 +18,9 @@
 		<PackageReference Update="Microsoft.Azure.Management.Sql" Version="1.41.0-preview" />
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.1.3"/>
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="3.0.0"/>
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="161.46367.54" />
-		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="150.5290.2-preview" GeneratePathProperty="true" />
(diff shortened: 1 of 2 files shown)


Packages.props
```diff
@@ -18,9 +18,9 @@
 		<PackageReference Update="Microsoft.Azure.Management.Sql" Version="1.41.0-preview" />
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.1.3"/>
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="3.0.0"/>
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="161.46367.54" />
-		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="150.5290.2-preview" GeneratePathProperty="true" />
+		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="160.5317.5-preview" GeneratePathProperty="true" />
```
llali, line 23: is this nuget already pushed to nuget.org? if not then we need to do that then you can use it here.

llali (changes requested):

Packages.props
Charles-Gagnon: Please make sure this process is documented somewhere (DacFX repo seems like a good place...) if it isn't already 😄

Packages.props
```diff
@@ -18,9 +18,9 @@
 		<PackageReference Update="Microsoft.Azure.Management.Sql" Version="1.41.0-preview" />
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="2.1.3"/>
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="3.0.0"/>
```
Charles-Gagnon, line 21: Do you know what changes there were between v2 and v3? Major version bumps like this should be taken with much more caution so it would be best to know what differences to expect and whether they'll affect us.

Have you done any testing of connections/queries on the various platforms as well?

Packages.props
llali: we already have a doc for dacfx nuget. Sai please take a look and if it's not clear let me know

Packages.props
ssreerama: Yes, I have done connections/querying along with testing on Windows and I'll take Kim/Leila's help for Mac testing.

I see these 3 are the [breaking changes](https://github.com/dotnet/SqlClient/blob/main/release-notes/3.0/3.0.0.md) with MDS3, #1 and #2 looks ok to me, for the #3 one I debugged the sqlDataReader code lines just to observe the functionality and seems good too.
1. The minimum supported .NET Framework version has been increased to v4.6.1. .NET Framework v4.6.0 is no longer supported.
2. User Id connection property now requires Client Id instead of Object Id for User-Assigned Managed Identity 
3. SqlDataReader now returns a DBNull value instead of an empty byte[]. Legacy behavior can be enabled by setting AppContext switch Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehavior

alanrenmsft: # sign has a special meaning, it will automatically be linked to PR or issues in this repo. you might want to edit the description so that people won't get confused :)

alanrenmsft: image

alanrenmsft: I've updated your comments and removed the links

ssreerama: Thanks Alan!

alanrenmsft:
@ssreerama have you released the new dacfx nuget package? if so please rerun the builds here.

Packages.props
```diff
@@ -18,9 +18,9 @@
...
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="161.46367.54" />
-		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="150.5290.2-preview" GeneratePathProperty="true" />
+		<PackageReference Update="Microsoft.SqlServer.DACFx" Version="160.5317.5-preview" GeneratePathProperty="true" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Data" Version="9.0.4" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Language" Version="9.0.4"/>
 		<PackageReference Update="Microsoft.SqlServer.Assessment"  Version="[1.0.305]" />
```
ssreerama, line 26: @aasimkhan30 , I see the version is mismatched with two version for <PackageReference Update="Microsoft.SqlServer.Management.SqlParser" Version="160.21267.55" />, can you confirm if I can update it to either 160.21267.51 or 160.21292.55. I don't see v160.21267.55 in [nuget.org](https://www.nuget.org/packages/Microsoft.SqlServer.Management.SqlParser/). please correct me if i'm missing any info here, Thanks :)

aasimkhan30: @ssreerama version 160.21292.55 should be the version. Can you please make the change here?

ssreerama:
DacFX 160 nuget will be ready by EOD today hopefully, Drew and Leila are finalizing the release notes and can release it quickly! Once it is done, will rerun the build and plan to merge! Also, Kim is going to test the STS/ADS in Man before the merge! Thanks :)

kisantia (approved):

alanrenmsft (approved):

Charles-Gagnon (approved):

PR #1280, opened 2021-11-03
Porting showplan code from ssms

aasimkhan30:
I have added a simple test case that just checks if the showplan graph is returned properly. I will try to find existing tests for this feature and port them here in a separate PR. 

The ssms code was using a graphing library "Microsoft.AnalysisServices.Graphing" to create and maintain graphs.
I have replaced that with a simple graph implementation that should server our purpose.


Charles-Gagnon:
Is this something we could put in a separate package?

(and what would be truly awesome is if SSMS & STS could both be updated to use that same package...not sure how heavily you had to modify the code to work here)

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Comparison/UniquePatternService.cs
```diff
@@ -0,0 +1,75 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Drawing;
```
kburtram, line 6: should we remove this namespace?  I don't think we should have any UI related code in this package.

kburtram:
@Charles-Gagnon the code is being pretty heavily modified as part of the port to STS, as the SSMS code couples logic & UI pretty heavily.  Any effort to unify this code and SSMS would need to be scoped separately, and it would be along the lines of "update SSMS to use STS (or shared package)".  Similar to what we're doing with Table Designer, or plan is to keep potentially sharable code isolated within a directory structure such that it will be easy to move somewhere else in the future, but not to go through the process of creating or integrating into a separate framework/package until we have a concrete use case.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
```diff
@@ -3061,1525 +3061,8976 @@ public static string SqlBindingsNet5NotSupported
             }
         }
 
-        public static string ConnectionServiceListDbErrorNotConnected(string uri)
+        public static string Statement
```
kburtram, line 3064: there are a ton of string updates here.  I assume you verified we need all these.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ActualPlanDataReaderNodeBuilder.cs
```diff
@@ -0,0 +1,87 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
```
kburtram, line 6: I'd recommend to try to split the "ShowPlanService" and the showplan graph build into separate components, and put the generic showplan code in a subfolder and the STS specific code in a code level folder.  This would allow the more generic code to be moved somewhere else in the future if we want to go that route.

kburtram (approved):
For the most part this looks good to me.  I still seem somethings about node coloring and other "UI" related content, so I think there will need to be more clean-ups/deletes.  But I think we can do that incrementally, maybe using code coverage to remove code blocks we don't need.

For this initial check-in, I think once you get a clean CI build you can merge so we can start integrating with the UI.  Then we can do future cleanups separately.  I  would verify we aren't adding any new framework dependencies if we don't need them, such as System.Drawing, etc.

abist (approved):

PR #1282, opened 2021-11-04
Applying Changes to SQL Database Projects and Using Projects in Comparison

Benjin:
Restarting this PR for fresh eyes, fresh comments.  Time to complete the feature! @nouryehia

---

This PR includes added logic to apply differences in a schema comparison to a sql database project (when the project is the target endpoint). Specifically, this includes:
- new SchemaComparePublishProjectChangesOperation
- new route to call this operation in the SchemaCompareService
- new parameters class that includes the information needed to apply the changes
- new fields added to the SchemaCompareEndpoint info class, which includes project-specific fields
- support for the project endpoint in the SchemaCompareOpenScmpOperation


global.json
```diff
@@ -1,5 +1,5 @@
 {
     "sdk": {
-        "version": "5.0.301"
+        "version": "5.0.402"
```
Charles-Gagnon, line 3: What is this needed for?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishChangesOperation.cs
```diff
@@ -1,80 +1,36 @@
-﻿//
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-//
+﻿using System.Threading;
```
Charles-Gagnon, line 1: copyright header

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishChangesOperation.cs
```diff
@@ -1,80 +1,36 @@
...
-            }
+        public SchemaComparePublishChangesOperation(SchemaComparisonResult comparisonResult)
+        {
+            Validate.IsNotNull(nameof(comparisonResult), comparisonResult);
+            ComparisonResult = comparisonResult;
         }
+        
+        public abstract void Execute(TaskExecutionMode mode);
 
         // The schema compare public api doesn't currently take a cancellation token so the operation can't be cancelled
```
Charles-Gagnon, line 30: I'm confused by this comment - it looks like you're passing around the cancellation token in other places so is this actually still accurate?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishDatabaseChangesOperation.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+            try
+            {
+                PublishResult = ComparisonResult.PublishChangesToDatabase(CancellationToken);
+                if (!PublishResult.Success)
+                {
+                    // Sending only errors and warnings - because overall message might be too big for task view
+                    ErrorMessage = string.Join(Environment.NewLine, this.PublishResult.Errors.Where(x => x.MessageType == SqlServer.Dac.DacMessageType.Error || x.MessageType == SqlServer.Dac.DacMessageType.Warning));
+                    throw new Exception(ErrorMessage);
```
Charles-Gagnon, line 46: It would be good to use a more specific type of exception, such as InvalidOperationException if there isn't a better one to use

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishProjectChangesOperation.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            try
+            {
+                PublishResult = ComparisonResult.PublishChangesToProject(Parameters.TargetProjectPath, Parameters.TargetFolderStructure);
+                
+                if (!PublishResult.Success)
+                {
+                    // Sending only errors and warnings - because overall message might be too big for task view
+                    ErrorMessage = PublishResult.ErrorMessage;
+                    throw new Exception(ErrorMessage);
```
Charles-Gagnon, line 46: Same here, InvalidOperationException or something more specific here

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishProjectChangesOperation.cs
```diff
@@ -0,0 +1,57 @@
...
+                throw new OperationCanceledException(CancellationToken);
+            }
+
+            try
+            {
+                PublishResult = ComparisonResult.PublishChangesToProject(Parameters.TargetProjectPath, Parameters.TargetFolderStructure);
+                
+                if (!PublishResult.Success)
+                {
+                    // Sending only errors and warnings - because overall message might be too big for task view
```
Charles-Gagnon, line 44: Is this comment accurate?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
```diff
@@ -14,20 +14,29 @@
...
+            FileAttributes attr = File.GetAttributes(path);
+
+            // If path is a directory, delete it and all subdirs/subfiles
+            if ((attr & FileAttributes.Directory) == FileAttributes.Directory)
+            {
+                new DirectoryInfo(path).Delete(true);
+            }
+
+            // If path is a file, delete it
+            else
```
Charles-Gagnon, line 37: Would prefer not to split up if/else like this. Putting the comments within the blocks is typically how I approach that.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
```diff
@@ -52,6 +61,34 @@ internal static string CreateDacpac(SqlTestDb testdb)
             return extractParams.PackageFilePath;
         }
 
+        internal static string CreateProject(SqlTestDb testdb, string projectName)
+        {
+            var result = GetLiveAutoCompleteTestObjects();
+            string folderPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "SchemaCompareTest", projectName);
```
Charles-Gagnon, line 67: It would be preferred to save this in a temp location instead

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -370,6 +565,80 @@ public async Task SchemaComparePublishChangesDacpacToDatabase()
...
+                Assert.True(schemaCompareOperation.ComparisonResult.IsValid);
+                Assert.True(schemaCompareOperation.ComparisonResult.IsEqual);
+                Assert.That(schemaCompareOperation.ComparisonResult.Differences, Is.Empty);
+
+                // cleanup
+                SchemaCompareTestUtils.VerifyAndCleanup(sourceProjectPath);
+            }
+            finally
+            {
+                sourceDb.Cleanup();
```
Charles-Gagnon, line 637: You should always be cleaning up any other artifacts created - such as the project files

Charles-Gagnon (approved):
Didn't spend too much time on the logic since that isn't really something I'm familiar with, but left some general comments to consider

llali (approved):

kisantia (approved):

global.json
Benjin: Not actually necessary, but is there any reason _not_ to casually update to the latest minor version, with the latest patches?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishChangesOperation.cs
Benjin: Same; removing.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishProjectChangesOperation.cs
Benjin: Looks like copypasta; removed

global.json
Charles-Gagnon: Nope, although I'd generally prefer updates like that be done in a separate PR so stuff isn't "snuck" in.

But this should be fine this time. Just try to avoid doing that in the future unless it's actually a part of the PR.

Benjin: Eh, nah you're right - I'll revert so that no one gets whiplash if either needs to be reverted.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaComparePublishChangesOperation.cs
```diff
@@ -1,80 +1,36 @@
-﻿//
-// Copyright (c) Microsoft. All rights reserved.
```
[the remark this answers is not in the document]
Benjin: Must've just replaced the entire file

PR #1287, opened 2021-11-08
Update to XElite

smartguest:
This PR adds in XELite based calls to profiler, which uses a stream based structure instead of poll based bursts, it appears to work with start, pause, and stop, but the status for multiple profiler sessions is unclear so far (the feed appears to stop for the initial session, and resume on the new session if its on the same database/server from one of my tests, but I'm not sure if this is a regression or not) and we're currently having issues with multiple sessions in profiler anyway, so that will need to be addressed first before this PR can go in. This PR is just for getting feedback on the code and to get known bugs that I may have overlooked in testing.


smartguest:
Identified an issue where upon stopping the session from another ADS client, the main ADS client does not display a message saying the session stopped unexpectedly, this means an XEventException is not being fired, will need to investigate what happens to the stream when a session stops.

smartguest:
Added exception handling task to handle in case session was stopped.

smartguest:
I will implement any feedback made on the current branch as it is right now. So far, the profiler stream appears to be functional, but when another ADS client tries to run on the same session, events are lost on the primary client (though this is by design with filtering old profiler events), though this might be how XELite works rather than anything on the SQL Server side.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1104,16 +1104,16 @@ public static bool IsDedicatedAdminConnection(ConnectionDetails connectionDetail
         /// Build a connection string from a connection details instance
         /// </summary>
         /// <param name="connectionDetails"></param>
-        public static string BuildConnectionString(ConnectionDetails connectionDetails)
+        public static string BuildConnectionString(ConnectionDetails connectionDetails, Boolean isXELite = false)
```
kburtram, line 1107: we should use C# primitives rather than .Net BCL classes when we have a choice.  i.e. please use `bool` here per .Net Framework coding guidelines.

Also, not sure the parameter name is ideal.  We're introducing a coupling between profiler and connectionservice that's inverted.  i.e., profilerservice can know about connectionservice but connectionservice shouldn't know about profilerservice.  Is there a different parameter name that more generically describes the customizations you're trying to make to the connection string for use is XELite?  And could you explain a bit why these customizations are needed?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -120,6 +121,39 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                    if (connInfo != null)
+                    {
+                        CancellationTokenSource threadCancellationToken = new CancellationTokenSource();
+                        // create a new XEvent session and Profiler session
+                        var xeSession = this.XEventSessionFactory.GetXEventSession(parameters.SessionName, connInfo);
+                        // //Create XELiveEvsentStream here, using ConnectionInfo Strings.
+                        var connectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails, true);
```
kburtram, line 143: don't see this being used anywhere.

src/Microsoft.SqlTools.ServiceLayer/Profiler/IProfilerSessionMonitor.cs
```diff
@@ -22,7 +22,7 @@ public interface IProfilerSessionMonitor
         /// <summary>
         /// Starts monitoring a profiler session
         /// </summary>
-        bool StartMonitoringSession(string viewerId, IXEventSession session);
+        bool StartMonitoringStream(string viewerId, IXEventSession session, ConnectionInfo connInfo);
```
kburtram, line 25: we pass the connection to `XEventSessionFactory.GetXEventSession` to create the session.  It seems strange to also pass it here.  I think the `GetXEventSession` should do everything we need with connection and once we start the session we're already connected.  Also, I don't see a good reason to rename the method.  XEvents uses the session terminology and so I think it makes sense to stick with that, unless there is a good reason to change the name.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -120,6 +121,39 @@ public void InitializeService(ServiceHost serviceHost)
...
+                        // //need to start session otherwise XLiveEventStreamer won't work.
```
kburtram, line 144: looks like there are double comment indicators `//`.  nit: not really sure saying we need to start the session or else streaming won't work.  That seems self-explanatory, IMO.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -120,6 +121,39 @@ public void InitializeService(ServiceHost serviceHost)
...
+            {
+                try
+                {
+                    ConnectionInfo connInfo;
+                    ConnectionServiceInstance.TryFindConnection(
+                        parameters.OwnerUri,
+                        out connInfo);
+                    if (connInfo != null)
+                    {
+                        CancellationTokenSource threadCancellationToken = new CancellationTokenSource();
```
kburtram, line 139: unused variable.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -164,8 +198,10 @@ internal async Task HandleCreateXEventSessionRequest(CreateXEventSessionParams p
                             xeSession = this.XEventSessionFactory.CreateXEventSession(parameters.Template.CreateStatement, parameters.SessionName, connInfo);
                         }
 
+                        var connectionString = ConnectionService.BuildConnectionString(connInfo.ConnectionDetails, true);
                         // start monitoring the profiler session
-                        monitor.StartMonitoringSession(parameters.OwnerUri, xeSession);
+                        //monitor.StartMonitoringSession(parameters.OwnerUri, xeSession);
```
kburtram, line 203: dead-code and unused variable above.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -20,6 +20,7 @@ public class ProfilerSession
         private static readonly TimeSpan DefaultPollingDelay = TimeSpan.FromSeconds(1);
         private object pollingLock = new object();
         private bool isPolling = false;
+        public bool isStreaming = false;
```
kburtram, line 23: this should be private and use property functions if it really needs to be public.  if we're adding this variable do we still need `isPolling` and polling variables below?  the new variable seems to be replacing the existing one?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -56,6 +54,9 @@ public Viewer(string Id, bool active, int xeId)
         // XEvent Session Id's matched to their Profiler Sessions
         private Dictionary<int, ProfilerSession> monitoredSessions = new Dictionary<int, ProfilerSession>();
 
+        // XEvent Session Id's matched to their Profiler Sessions
```
kburtram, line 57: comment isn't clear.  seems to mapping IDs to cancellationTokens, not sessions.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -72,23 +73,21 @@ public void AddSessionListener(IProfilerSessionListener listener)
...
                 {
-                    this.processorThread = Task.Factory.StartNew(ProcessSessions);
+                    this.processorThread = Task.Factory.StartNew(ProcessStreams);
                 }
 
                 // create new profiling session if needed
                 if (!this.monitoredSessions.ContainsKey(session.Id))
                 {
                     var profilerSession = new ProfilerSession();
+                    profilerSession.ConnectionInfo = connInfo;
```
kburtram, line 90: if we really need the connection I would recommend storing it in the xeventsession interface, since the session is supposed to be connected already.

kburtram (changes requested):
There are several things that should be cleaned up.  I've left a few comments.  I'll look through a bit more, but here are some initial comments.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
smartguest: XELite does not accept certain options in the connection string and won't function (IE, no data will be streamed) when a string containing them is passed as a parameter to the XELiveEventStreamer. This boolean flag is to used to omit those problematic options from the connection string when it is being generated so that it works properly. I'll name it to something else and change the Boolean to bool.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
smartguest: That is a debug string, will remove.

src/Microsoft.SqlTools.ServiceLayer/Profiler/IProfilerSessionMonitor.cs
smartguest: It is passed there because we need to get the connectionString from the connectionInfo for XElite, and the ProfilerSession object does not include that by default. Regardless of how its done there must be a way to get the proper connection string in there.

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -120,6 +121,39 @@ public void InitializeService(ServiceHost serviceHost)
             this.SessionMonitor.AddSessionListener(this);
         }
 
+        /// <summary>
+        /// Handle request to start a profiling session
+        /// </summary>
+        internal async Task HandleStartProfilingRequest(StartProfilingParams parameters, RequestContext<StartProfilingResult> requestContext)
+        {
+            await Task.Run(async () =>
```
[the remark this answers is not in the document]
smartguest: This was how the code was originally found in the profiler extension, I kept it as is to avoid disrupting functionality, let me know if this is needed or not. @alanrenmsft Do we still need to keep this async?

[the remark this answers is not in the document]
alanrenmsft: this issue is not just with profiler handlers, it is a much wider problem that I am looking into. so it sounds good to me to keep things as is. I am planning to come up with something that will address all the issues at once.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +185,96 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
+            return connDetails;
+        }
+
+        /// <summary>
+        /// Creates a stream for new XEvents if the session hasn't streamed yet.
+        /// </summary>
+        private void ProcessStream(int id, ProfilerSession session)
+        {
+            CancellationTokenSource threadCancellationToken = new CancellationTokenSource();
+            ConnectionDetails trimmedDetails = CreateXEliteConnectionDetails((session.XEventSession as XEventSession).ConnInfo.ConnectionDetails);
```
Charles-Gagnon, line 260: Why does the XEventSession not have a connection details that already can be used for xelite connections? It seems like we should be doing this as soon as the session is created.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +185,96 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
+        /// </summary>
+        private ConnectionDetails CreateXEliteConnectionDetails(ConnectionDetails connDetails){
+            connDetails.ConnectRetryCount = null;
+            connDetails.ConnectRetryInterval = null;
+            connDetails.MultiSubnetFailover = null;
+            return connDetails;
+        }
+
+        /// <summary>
+        /// Creates a stream for new XEvents if the session hasn't streamed yet.
```
Charles-Gagnon, line 255: I'm not seeing anywhere that you're checking if the session "hasn't streamed yet". Why do you need that check anyways?

Charles-Gagnon (commented):
It's looking mostly fine (although I haven't gone through all the logic super in depth yet...)

@kburtram @smartguest What's the plan for releasing this update though? Are we just going to switch over to this completely and not try to keep the old polling functionality around for users to switch back to if there's issues?

If that's the case what release are we targeting this for?

smartguest:
@Charles-Gagnon So far it seems we are planning on doing a complete switchover. Having SMO alongside XElite would require bringing back old lines of code and adding logic and dialogue to choose which type of profiler stream the user wants.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
smartguest: This code is only run when the session hasn't started streaming yet, this code should only be run once otherwise we would have multiple constantly running streams coming from the same session and it would lead to tons of duplicate events or possibly worse. That check is done outside of this code where ProcessStream is called.

Charles-Gagnon: How/why would we get multiple streams attempting to be ran? It definitely seems like that if that was ever to happen we should be throwing an error, right?

smartguest: The ProcessStreams task/function constantly cycles through the available monitored sessions (this is a holdover when SMO was polling each session constantly), and checks each session whether they are already streaming or not before starting. I'm just saying that without the session stream check, it would try to create another stream using the same session.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
smartguest: The ConnectionInfo class currently does not allow for properties such as connection details to be altered after they are created, the set function for it is private. Unless we want to change the class to allow for that, you have to copy the ConnectionDetails and clean it separately.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -1535,6 +1535,10 @@
     <value>Cannot access the specified path on the server: {0}</value>
     <comment></comment>
   </data>
+  <data name="ConnInfoIsNull" xml:space="preserve">
+    <value>Connection Info is null</value>
```
Charles-Gagnon, line 1539: Let's avoid using terms like "null" in user-facing messages. Something more general like "Unable to find connection info for URI {0}" would be better I think.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -120,6 +121,39 @@ public void InitializeService(ServiceHost serviceHost)
             this.SessionMonitor.AddSessionListener(this);
         }
 
+        /// <summary>
+        /// Handle request to start a profiling session
+        /// </summary>
+        internal async Task HandleStartProfilingRequest(StartProfilingParams parameters, RequestContext<StartProfilingResult> requestContext)
```
Charles-Gagnon, line 127: While you're refactoring this anyways, I would strongly suggest taking the time to add in debug logging for key events (start stream, stop stream, errors, received stream data, etc.) Basically anything that would be helpful in debugging an issue a customer has when you don't have access to their machine.

I'd probably suggest doing this as a separate follow up PR just to make it easier to review and not add more churn to this one, but getting started by adding comments in your local environment would be a good idea.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -120,6 +121,39 @@ public void InitializeService(ServiceHost serviceHost)
...
+                        await requestContext.SendResult(result);
+                    }
+                    else 
+                    {
+                        throw new Exception(SR.ConnInfoIsNull);
+                    }
+                }
+                catch (Exception e)
+                {
+                    await requestContext.SendError(new Exception(SR.CreateSessionFailed(e.Message)));
```
Charles-Gagnon, line 152: You should add in the original exception as well since it contains all the useful information like callstack

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -40,73 +34,33 @@ public class ProfilerSession
...
-        /// The delay between session polls
+        /// Functions for marking and retrieving the status of the ProfilerSession having an active XELite stream running. 
         /// </summary>
-        public TimeSpan PollingDelay
-        {
-            get
-            {
-                return pollingDelay;
-            }
+        public void toggleStreamLock(){
```
Charles-Gagnon, line 50: Do you really want this to be a toggle instead of a setter? It seems like the scenarios this is being used for specifically want to turn streaming on/off and so toggling could result in unexpected state.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +185,99 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
-            }
-        }
-
         /// <summary>
-        /// The core queue processing method
+        /// The core queue processing method, cycles through monitored sessions and creates a stream for them if not already.
         /// </summary>
-        /// <param name="state"></param>
-        private void ProcessSessions()
+        private void ProcessStreams()
```
Charles-Gagnon, line 191: Given that everything else is still referring to these as "sessions" I'd say just keep this as `ProcessSessions`

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
Charles-Gagnon: Yes, you'd create a new ConnectionDetails at some point. I'm saying that when the `ProfilerSession` is created then the connection info it creates there calls `CreateXEliteConnectionDetails` so that this object already has the expected details and no further processing is needed.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -148,7 +144,8 @@ public async Task TestStopProfilingRequest()
         /// Test pausing then resuming a session
         /// </summary>
         /// <returns></returns>
-        [Test]
+        //[Test]
```
Charles-Gagnon, line 147: Please fix these tests before merging in

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
smartguest: The ConnectionInfo class for profiler is retrieved by calling ConnectionServiceInstance.tryfindconnection which uses a trygetvalue on a map, if we want to alter the creation of the connectioninfo's connectiondetails as its being created, it would require alterations to the connection service, so it's not practical there, if we want to store the connection details in XEventSessions, that might be a better option.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -40,73 +34,36 @@ public class ProfilerSession
...
-        /// The delay between session polls
+        /// Functions for marking and retrieving the status of the ProfilerSession having an active XELite stream running. 
         /// </summary>
-        public TimeSpan PollingDelay
-        {
-            get
-            {
-                return pollingDelay;
-            }
+        public void enableStreamLock(){
```
Charles-Gagnon, line 50: I would just make the property public at this point. 

Also I'm not sure that using the term "lock" here is appropriate. It's not actually being used as a lock in the standard C# sense (see sessionsLock) and so only adds confusion.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +185,88 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
-            finally
+        }
+
+        /// <summary>
+        /// Function that creates a brand new stream from a session, this is called from ProcessSessions when a session doesn't have a stream running currently.
+        /// </summary>
+        private void StartStream(int id, ProfilerSession session)
+        {
+            CancellationTokenSource threadCancellationToken = new CancellationTokenSource();
+            var connectionString = ConnectionService.BuildConnectionString((session.XEventSession as XEventSession).ConnDetails);
```
Charles-Gagnon, line 252: Is the cast necessary here? IXEventSession seems to have ConnDetails already

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +185,88 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
+            var eventStreamer = new XELiveEventStreamer(connectionString, (session.XEventSession as XEventSession).Session?.Name);
```
Charles-Gagnon, line 253: What's the scenario for the session being null here? You don't check that in the line above

src/Microsoft.SqlTools.ServiceLayer/Profiler/IXEventSession.cs
```diff
@@ -15,6 +16,8 @@ public interface IXEventSession
         /// </summary>
         int Id { get; }
 
+        ConnectionDetails ConnDetails { get; set; }
```
alanrenmsft, line 19: why not use ConnectionDetails as the property name?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
alanrenmsft: yes, let's just do one sendError call.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -383,6 +393,7 @@ public IXEventSession GetXEventSession(string sessionName, ConnectionInfo connIn
             var sqlConnection = ConnectionService.OpenSqlConnection(connInfo);
             SqlStoreConnection connection = new SqlStoreConnection(sqlConnection);
             BaseXEStore store = CreateXEventStore(connInfo, connection);
+            ConnectionDetails newConnectionDetails = CreateXEliteConnectionDetails(connInfo.ConnectionDetails);
```
alanrenmsft, line 396: no need to create a variable for this, you can do it inline directly in L413

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -72,9 +65,6 @@ public void AddSessionListener(IProfilerSessionListener listener)
             }
         }
 
-        /// <summary>
-        /// Start monitoring the provided session
-        /// </summary>
         public bool StartMonitoringSession(string viewerId, IXEventSession session)
```
alanrenmsft, line 68: It seems this method will always return true, you can change its return type to be void.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +185,88 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
+
+        /// <summary>
+        /// Function that creates a brand new stream from a session, this is called from ProcessSessions when a session doesn't have a stream running currently.
+        /// </summary>
+        private void StartStream(int id, ProfilerSession session)
+        {
+            CancellationTokenSource threadCancellationToken = new CancellationTokenSource();
+            var connectionString = ConnectionService.BuildConnectionString((session.XEventSession as XEventSession).ConnDetails);
+            var eventStreamer = new XELiveEventStreamer(connectionString, (session.XEventSession as XEventSession).Session?.Name);
+            //Start streaming task here, will run until cancellation or error with the feed.
```
alanrenmsft, line 254: nit, for all the code comments, there should be a space between // and the comment.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -229,7 +229,7 @@ public async Task TestPauseProfilingRequest()
         /// <summary>
         /// Test notifications for stopped sessions
         /// </summary>
-        //[Test]
+        [Test]
         // TODO: Fix test to work with XElite.
```
Charles-Gagnon, line 233: Remove

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -254,21 +254,14 @@ public async Task TestStoppedSessionNotification()
...
-            pollingTimer.Interval = 10000;
-            pollingTimer.Start();
-            bool timeout = false;
-            pollingTimer.Elapsed += new System.Timers.ElapsedEventHandler((s_, e_) => { timeout = true; });
-            while (sessionStopped == false && !timeout)
-            {
-                Thread.Sleep(250);
-            }
-            pollingTimer.Stop();
+            // stop session with fake stop error.
```
Charles-Gagnon, line 263: I'm not sure I understand how this is a "fake stop error". There's no error involved at all. And if you're simulating an error it would be better to mock the stream and then use the mock to send an error that way instead.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -259,16 +259,23 @@ private void StartStream(int id, ProfilerSession session)
...
             }, TaskContinuationOptions.OnlyOnFaulted);
 
             this.monitoredCancellationTokenSources.Add(id, threadCancellationToken);
             session.enableStreamLock();
         }
 
+        /// <summary>
+        /// Helper function for firing error and stopping session in case the session is stopped on the server. This is public for tests.  
+        /// </summary>
+        public void stopSessionError(int Id){
```
Charles-Gagnon, line 278: C# convention is capitalized method names (`StopSessionError`)

A couple suggestions : 

1. Just call this StopSession or something similar
2. Remove the mention about it being "public for tests". That isn't really important for people to know. Just describe what it's actually accomplishing (which in reality is "sending notification to listeners and then removing the session from our lists")

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
Charles-Gagnon: @smartguest I missed seeing this comment from Karl. So you can put the getter/setter back in if that makes sense to you (you can just use the auto-getter and setter `public bool IsStreaming { get; set; }` to make that easy)

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -101,12 +44,16 @@ public bool EventsLost
             }
         }
 
+        public bool isStreamActive(){
```
Charles-Gagnon, line 47: Do we need this if we're exposing IsStreaming directly?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -95,6 +90,7 @@ public async Task TestStartProfilingRequest()
         /// Test stopping a session and receiving event callback
         /// </summary>
         /// <returns></returns>
+        // TODO: Fix test to work with XElite.
```
Charles-Gagnon, line 93: Remove

src/Microsoft.Kusto.ServiceLayer/Localization/sr.xlf
```diff
@@ -2024,6 +2018,12 @@
         <note>.
  Parameters: 0 - filePath (string) </note>
       </trans-unit>
+      <trans-unit id="CreateSessionFailed">
+        <source>Failed to create session: {0}</source>
+        <target state="new">Failed to create session: {0}</target>
+        <note>.
```
Charles-Gagnon, line 2024: Do you know why there's the extra newline in these? Can you remove them?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -120,6 +121,39 @@ public void InitializeService(ServiceHost serviceHost)
...
+                        await requestContext.SendResult(result);
+                    }
+                    else 
+                    {
+                        throw new Exception(SR.ProfilerConnectionNotFound);
+                    }
+                }
+                catch (Exception e)
+                {
+                    await requestContext.SendError(new Exception (SR.CreateSessionFailed(e.Message), e));
```
Charles-Gagnon, line 152: Let's use different errors for these so we know what actually failed - this one should be `StartProfilingFailed`

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -160,8 +148,17 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
             lock (this.sessionsLock)
             {
-                if (this.monitoredSessions.Remove(sessionId, out session))
+                //cancel running XEventStream for session.
+                CancellationTokenSource targetToken;
+                if (monitoredCancellationTokenSources.Remove(sessionId, out targetToken))
+                {
+                    targetToken.Cancel();
+                }
+                if (this.monitoredSessions.Remove(sessionId, out session) && session.IsStreaming)
```
Charles-Gagnon, line 157: We should remove all viewers even if the session wasn't streaming, right? Not sure if that could ever actually happen but that's what the old logic was doing. So in this case I think you can just remove the `session.IsStreaming` check completely.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +183,102 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
+                        ProfilerSession session;
+                        this.monitoredSessions.TryGetValue(id, out session);
+                        if (!session.IsStreaming)
                         {
-                            List<string> viewers = this.sessionViewers[session.XEventSession.Id];
-                            if (viewers.Any(v => allViewers[v].active))
-                            {
-                                ProcessSession(session);
-                            }
+                            StartStream(id, session);
```
Charles-Gagnon, line 201: Why did you remove the check to make sure that at least one viewer is active?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +183,102 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
-            }
-            catch (Exception ex)
-            {
-                Logger.Write(TraceEventType.Warning, "Failed to poll session. error: " + ex.Message);
-            }
-            finally
-            {
-                session.IsPolling = false;
+                RemoveSession(id, out tempSession);
+                throw new Exception(SR.SessionMissingDetails(id));
```
Charles-Gagnon, line 271: I think a better error message here would be something along the lines of `Unable to start streaming session {0} due to missing session details`. 

Although I'm curious when you expect this to ever happen in the first place.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -149,11 +146,12 @@ public async Task TestStopProfilingRequest()
         /// </summary>
         /// <returns></returns>
         [Test]
+        // TODO: Add more in-depth testing for pause effects on events received.
         public async Task TestPauseProfilingRequest()
         {
             bool success = false;
             string testUri = "test_session";
-            bool recievedEvents = false;
+            // bool recievedEvents = false; // Removed due to incompatability with XElite stream.
```
Charles-Gagnon, line 154: Just delete this and any other unused code

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -164,74 +162,71 @@ public async Task TestPauseProfilingRequest()
...
+            // }
+            // pollingTimer.Stop();
 
             // confirm that polling works
-            Assert.True(recievedEvents);
+            //Assert.True(recievedEvents);
 
             // pause viewer
             await profilerService.HandlePauseProfilingRequest(requestParams, requestContext.Object);
             Assert.True(success);
```
Charles-Gagnon, line 208: Where is success being set to true in this test?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
smartguest: It shouldn't as long as the profilersession object is created properly in previous steps

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/ProfilerServiceTests.cs
smartguest: It is set true on a callback on L160 when HandlePauseProfilingRequest is called

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -1571,6 +1571,16 @@
     <value>An XEvent session named {0} already exists</value>
     <comment>.
  Parameters: 0 - sessionName (String) </comment>
+  </data>
+  <data name="SessionMissingDetails" xml:space="preserve">
+    <value>Unable to start streaming session {0} due to missing session details.</value>
+    <comment>.
```
Charles-Gagnon, line 1577: Do you know why the comments are on a separate line like this? Seems like that's existing behavior but it looks pretty odd...is that something we can fix?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
Charles-Gagnon: @smartguest Just to check, are you planning on doing this? It'd be hugely valuable and this is something I feel we should get in before pushing this to stable so that if customers start having issues we can quickly and easily track down where issues may be occurring. 

@kburtram FYI

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -374,6 +375,16 @@ private static BaseXEStore CreateXEventStore(ConnectionInfo connInfo, SqlStoreCo
...
         }
 
+        /// <summary>
+        /// Nulls out properties in ConnectionDetails that aren't compatible with XElite.
+        /// </summary>
+        private static ConnectionDetails CreateXEliteConnectionDetails(ConnectionDetails connDetails){
+            connDetails.ConnectRetryCount = null;
+            connDetails.ConnectRetryInterval = null;
+            connDetails.MultiSubnetFailover = null;
+            return connDetails;
```
Charles-Gagnon, line 385: Given that we're modifying the existing object this doesn't really need to return it. Maybe just call this `RemoveIncompatibleConnectionProperties` and have the return type be void.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -160,8 +148,17 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
             {
+                //cancel running XEventStream for session.
+                CancellationTokenSource targetToken;
+                if (monitoredCancellationTokenSources.Remove(sessionId, out targetToken))
+                {
+                    targetToken.Cancel();
+                }
                 if (this.monitoredSessions.Remove(sessionId, out session))
                 {
+                    // Toggle isStreaming status of removed session to not streaming.
```
Charles-Gagnon, line 159: This isn't really toggling (toggle implies switching to whatever it currently isn't set to)

But I don't think this needs a comment, it's clear enough from the name what's going on.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +183,105 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
-            }
-            lock (this.pollingLock)
-            {
-                Monitor.Pulse(pollingLock);
-            }
-        }
-
         /// <summary>
-        /// The core queue processing method
+        /// The core queue processing method, cycles through monitored sessions and creates a stream for them if not already.
```
Charles-Gagnon, line 187: One thing to think about, why do we even need to do this? Why don't we start the session right away when we add it to the dictionary? It's fine to leave it with the existing behavior for now, but a good follow-up task would be to see if this is actually necessary (and if it is then make sure it's documented in the code why we do this)

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -186,113 +183,105 @@ private bool RemoveSession(int sessionId, out ProfilerSession session)
...
                 }
             }
         }
 
         /// <summary>
-        /// Process a session for new XEvents if it meets the polling criteria
+        /// Helper function used to process the XEvent feed from a session's stream.
         /// </summary>
-        private void ProcessSession(ProfilerSession session)
+        internal async Task HandleXEvent(IXEvent xEvent, ProfilerSession session)
```
Charles-Gagnon, line 214: Does this need to be internal? Seems like private should be fine

Charles-Gagnon (approved):
This looks fine, although I didn't spend a lot of time going over the exact logic here of the streaming since that generally seems like better done through actual testing.

But let's make sure that we're getting this in at the right time. Is December really when we want to be doing this? This is a pretty major change and there's likely to be some kinks to work out, so maybe this could wait until after the holidays to get in?

@kburtram ?

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
smartguest: This is auto generated Charles, I'm not sure why this is the case, but I'll change it manually for this PR for now.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
smartguest: Likely it is not necessary, but I will investigate in another follow up task.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
smartguest: I plan on getting this into insiders by December at the very least.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
smartguest: It's generated by the String Resource Tool included, so I suppose this is the expected result.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
Charles-Gagnon: So keep in mind that if you check this in and don't want it in stable then you're essentially blocking anyone else from making changes in STS that they want to get into stable. So if you don't want to ship this in stable then we'd need to discuss when this should actually be merged in.

PR #1291, opened 2021-11-10
Package published projects

Charles-Gagnon:
1. Add some missing projects that were ignored because of .gitignore
2. Fix up some path issues
3. Add pack and sign to build

https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=128050&view=results

Note that I'm not sure this is the best/cleanest way to do this. A couple things that I'm not sure about : 

1. The runtime packages are ALL downloaded since they are just dependencies of the root package. We should ideally only be downloading the files needed for the host runtime
2. I'm not sure the package structure is correct - we put all the files under the runtimes folder (even the non-native stuff) 
3. For the target RIDs it seems like we're being more specific than is necessary - for example why are we specifying win7-x64 instead of just win-x64?

https://natemcmaster.com/blog/2017/11/11/build-tools-in-nuget/ has some good suggestions about how we could refactor this to fit the purpose of this package better. 

Anyways this will continue to be a work in progress - to start with I'm just getting this to match what the .NET Interactive folks need and we can make improvements down the road as we see fit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,6 @@ msbuild.wrn
 # code coverage artifacts
 coverage.xml
 node_modules
-packages
 reports
 opencovertests.xml
 outputCobertura.xml
@@ -173,7 +172,6 @@ publish/
 *.nuget.props
 *.nuget.targets
 src/**/*.nupkg
-**/packages/*
 
 # Windows Azure Build Output
 csx/
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -151,6 +151,34 @@ steps:
     MaxConcurrency: 5
   condition: and(succeeded(), eq(variables['signed'], true))
 
+- task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
+  displayName: 'ESRP Code Signing - Nuget Package'
+  inputs:
+    ConnectedServiceName: 'Code Signing'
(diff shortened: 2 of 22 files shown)


chlafreniere (approved):

chlafreniere:
@Charles-Gagnon did we get clarity on win7-x86/win10-arm64 support (and whether that's needed for any existing scenarios)?

Charles-Gagnon:
We can `build what we need`, so I guess it's fine?

Given that we don't support those at the moment then I think releasing a package for those would not be the best idea anyways. If we did end up needing to support them for some reason then we'll have to come back and evaluate whether we can do that as-is or whether other changes need to happen. 

The main reason for this PR being needed is for the Linux distros since currently we only publish a package for RHEL but not any of the others (which then also means we can't run in codespaces since that runs on non-RHEL linux)

chlafreniere:
> We can "build what we need", so I guess it's fine?
> 
> Given that we don't support those at the moment then I think releasing a package for those would not be the best idea anyways. If we did end up needing to support them for some reason then we'll have to come back and evaluate whether we can do that as-is or whether other changes need to happen.
> 
> The main reason for this PR being needed is for the Linux distros since currently we only publish a package for RHEL but not any of the others (which then also means we can't run in codespaces since that runs on non-RHEL linux)

Thanks for clarifying. Makes sense.

PR #1292, opened 2021-11-10
use path to identify edit target

alanrenmsft:
1. changes to reflect the property => path change in ADS: [17638](https://github.com/microsoft/azuredatastudio/pull/17638)
2. add a few column properties


src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
```diff
@@ -56,7 +56,7 @@ public virtual Task SendError(string errorMessage, int errorCode = 0)
         public virtual Task SendError(Exception e)
         {
             // Overload to use the parameterized error handler
-            return SendError(e.Message, e.HResult);
+            return SendError(string.Format("{0}{1}{2}", e.Message, Environment.NewLine, e.StackTrace), e.HResult);
```
alanrenmsft, line 59: while working on the new feature, I ran into a couple errors that only shows me the error message without stacktrace.
it would be helpful for us to debug the issue especially when users ran into issues.

Charles-Gagnon: I'm not sure I like doing this for ALL errors - because we often will end up displaying these to the user. I'd suggest adding an extra StackTrace field to the Error object we're writing back instead.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/DesignerPathUtils.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections;
+using System.Linq;
+using Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
+{
+    public static class DesignerPathUtils
+    {
+        ///<summary>
+        /// Parse the path in the table designer change information.
```
Charles-Gagnon, line 16: Please given information about what the format of this path is supposed to be like.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/DesignerPathUtils.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (!isValid)
+            {
+                throw new ArgumentException(SR.InvalidTableEditPathException(path, editType.ToString()));
+            }
+            return list;
+        }
+
+        private static string[] Split(string path)
+        {
+            return path.Split('/', StringSplitOptions.RemoveEmptyEntries);
```
Charles-Gagnon, line 78: Can this path potentially contain /'s in the segment names (such as the name of objects)?

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -57,103 +57,89 @@ public void InitializeService(ServiceHost serviceHost)
...
+                try
+                {
+                    await action();
+                }
+                catch (Exception e)
+                {
+                    await requestContext.SendError(e);
+                }
+            });
+            return Task.CompletedTask;
```
Charles-Gagnon, line 73: I assume you're purposely not blocking on the task completing here? I'd make it clear if that's the case - and add why you're doing that. 

And to make sure I'm getting this correctly - this will make it so it immediately continues to process other messages but the request context itself is what sends the results/errors and so on the ADS side it'll still be waiting until we send one of those, correct?

Any concern about STS processing multiple designer events simultaneously and causing some race conditions?

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -57,103 +57,89 @@ public void InitializeService(ServiceHost serviceHost)
             this.ServiceHost.SetRequestHandler(SaveTableChangesRequest.Type, HandleSaveTableChangesRequest);
             this.ServiceHost.SetRequestHandler(DisposeTableDesignerRequest.Type, HandleDisposeTableDesignerRequest);
         }
+        private Task HandleRequest<T>(RequestContext<T> requestContext, Func<Task> action)
```
Charles-Gagnon, line 60: This seems like something that could be put in a more general location so others can use it too...

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/RequestContext.cs
alanrenmsft: makes sense, I'll revert the change here and have a separate PR.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/DesignerPathUtils.cs
alanrenmsft: I've added comment about the `/` character as a reserved character in ADS typings file.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/DesignerPathUtils.cs
```diff
@@ -0,0 +1,93 @@
...
+		///     b. 'propertyName1/index-1/propertyName2'. Example: add a column mapping to the first foreign key: 'foreignKeys/0/mappings'.
+		/// 2. 'Update' scenario
+		///     a. 'propertyName1'. Example: update the name of the table: 'name'.
+		///     b. 'propertyName1/index-1/propertyName2'. Example: update the name of a column: 'columns/0/name'.
+		///     c. 'propertyName1/index-1/propertyName2/index-2/propertyName3'. Example: update the source column of an entry in a foreign key's column mapping table: 'foreignKeys/0/mappings/0/source'.
+		/// 3. 'Remove' scenario
+        ///     a. 'propertyName1/index-1'. Example: remove a column from the columns property: 'columns/0'.
+		///     b. 'propertyName1/index-1/propertyName2/index-2': Example: remove a column mapping from a foreign key's column mapping table: 'foreignKeys/0/mappings/0'.
+        ///<summary>
+        public static ArrayList Parse(string path, DesignerEditType editType)
```
Charles-Gagnon, line 30: Any reason you're using a string for this instead of just passing an array directly anyways? Seems like that would save a good amount of effort here.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/DesignerPathUtils.cs
```diff
@@ -0,0 +1,80 @@
...
+                    if (!isValid)
+                    {
+                        break;
+                    }
+                }
+            }
+
+            if (!isValid)
+            {
+                throw new ArgumentException(SR.InvalidTableEditPathException(string.Join(',', path), editType.ToString()));
```
Charles-Gagnon, line 76: Would suggest adding a message about why this was invalid too

Charles-Gagnon (approved):

PR #1296, opened 2021-11-11
Applying Changes to SQL Database Projects and Using Projects in Comparison

Benjin:
Time to complete the feature! @nouryehia

... but without putting Schema Compare in Azure Data Studio in a broken state while those PRs go through this time 😄 

---

This PR includes added logic to apply differences in a schema comparison to a sql database project (when the project is the target endpoint). Specifically, this includes:
- new SchemaComparePublishProjectChangesOperation
- new route to call this operation in the SchemaCompareService
- new parameters class that includes the information needed to apply the changes
- new fields added to the SchemaCompareEndpoint info class, which includes project-specific fields
- support for the project endpoint in the SchemaCompareOpenScmpOperation


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
```diff
@@ -198,13 +200,48 @@ await requestContext.SendResult(new ResultStatus()
...
                 SchemaComparisonResult compareResult = schemaCompareResults.Value[parameters.OperationId];
-                operation = new SchemaComparePublishChangesOperation(parameters, compareResult);
+                operation = new SchemaComparePublishDatabaseChangesOperation(parameters, compareResult);
+                SqlTask sqlTask = null;
+                TaskMetadata metadata = new TaskMetadata();
+                metadata.TaskOperation = operation;
+                metadata.ServerName = parameters.TargetServerName;
+                metadata.DatabaseName = parameters.TargetDatabaseName;
+                metadata.Name = SR.PublishChangesTaskName;
+                sqlTask = SqlTaskManagerInstance.CreateAndRun<SqlTask>(metadata);
```
kisantia, line 216: should be able to declare sqlTask here instead of initially being set to null above

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareService.cs
```diff
@@ -231,6 +268,51 @@ await requestContext.SendResult(new ResultStatus()
...
+        {
+            SchemaComparePublishProjectChangesOperation operation = null;
+            try
+            {
+                SchemaComparisonResult compareResult = schemaCompareResults.Value[parameters.OperationId];
+                operation = new SchemaComparePublishProjectChangesOperation(parameters, compareResult);
+
+                TaskMetadata metadata = new()
+                {
+                    TaskOperation = operation,
```
kisantia, line 285: can you also update the PublishDatabases one to initialize the metadata like this?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -437,6 +706,260 @@ public async Task SchemaComparePublishChangesDatabaseToDatabase()
...
+                {
+                    OperationId = schemaCompareOperation.OperationId,
+                    TargetProjectPath = targetProjectPath,
+                    TargetFolderStructure = SqlServer.Dac.DacExtractTarget.Flat,
+                };
+
+                SchemaComparePublishProjectChangesOperation publishChangesOperation = new(publishChangesParams, schemaCompareOperation.ComparisonResult);
+                publishChangesOperation.Execute(TaskExecutionMode.Execute);
+                Assert.True(publishChangesOperation.PublishResult.Success);
+                Assert.AreEqual(publishChangesOperation.PublishResult.ErrorMessage, "");
```
kisantia, line 853: string.Empty instead of ""

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -437,6 +706,260 @@ public async Task SchemaComparePublishChangesDatabaseToDatabase()
...
+                string sourceProjectPath = SchemaCompareTestUtils.CreateProject(sourceDb, "SourceProject");
+                string[] sourceScripts = SchemaCompareTestUtils.GetProjectScripts(sourceProjectPath);
+                string targetProjectPath = SchemaCompareTestUtils.CreateProject(targetDb, "TargetProject");
+                string[] targetScripts = SchemaCompareTestUtils.GetProjectScripts(targetProjectPath);
+
+                SchemaCompareEndpointInfo sourceInfo = new();
+                sourceInfo.EndpointType = SchemaCompareEndpointType.Project;
+                sourceInfo.ProjectFilePath = Path.Combine(sourceProjectPath, "SourceProject.sqlproj");
+                sourceInfo.TargetScripts = sourceScripts;
+                sourceInfo.DataSchemaProvider = "150";
```
kisantia, line 898: it would be nice if there was a little helper function for creating source/target project endpoint infos since it looks like the same thing getting created for so many tests

kisantia (approved):

guillaume86:
@Benjin is it relying on an internal build of Microsoft.SqlServer.DacFx ?

I tried to open an scmp with a project target with `SchemaComparison` and I still hit this error:
> The specified file could not be opened as a schema comparison file.  The file might be corrupt or it might refer to a target or source endpoint that is not of type database or dacpac. Only database endpoints and dacpac endpoints are supported. (Parameter 'scmpFilePath')

I could substitute the target endpoint with a built dacpac but that defeats the purpose a bit :).

PR #1300, opened 2021-11-12
Sending show plan graph to ADS on Result Set updated event

aasimkhan30:
This PR does 2 following things 
1. Sending show plan graph on result set updated event to the frontend.
1. I have converted the showplan graph to simpler object to avoid circular references.


src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -2213,6 +2213,16 @@ SizeInMegaBytesFormat = {0} MB
 SizeInGigaBytesFormat = {0} GB
 ;Size in TeraBytes format
 SizeInTeraBytesFormat = {0} TB
+; display string for the operator cost property - 0.###### - is the float number format specifier
+OperatorDisplayCost(double cost, int percentage) = {0:0.#######} ({1}%)
+;This node name will be visible in Active Accesibility
+NodeDisplayPropertiesName2(string title, long nodeId) = {0}[{1}]
+;This node name will be visible in Active Accesibility
+NodeDisplayPropertiesName1(long nodeId) = [{0}]
```
Charles-Gagnon, line 2221: These don't seem like things that actually need to be localized - there's no text to localize here

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -0,0 +1,103 @@
+using System.Collections.Generic;
```
Charles-Gagnon, line 1: copyright

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Node.cs
```diff
@@ -55,6 +57,186 @@ public int GroupIndex
...
+                }
+
+                return secondLine == null || secondLine.Length == 0
+                    ? firstLine
+                    : String.Format(CultureInfo.CurrentCulture, "{0}\n{1}", firstLine, secondLine);
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets node name (visible through AA)
```
Charles-Gagnon, line 128: What's AA? And `Name2` isn't very descriptive, is there not something better you could use?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Node.cs
```diff
@@ -55,6 +57,186 @@ public int GroupIndex
...
+        /// </summary>
+        public string Name2
+        {
+            get
+            {
+                // Make AA names compatible with the previos implementation
```
Charles-Gagnon, line 134: What does this mean? What previous implementation? Remember people reading this code have no context for where this came from originally so explaining why this stuff is actually necessary is better.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Node.cs
```diff
@@ -55,6 +57,186 @@ public int GroupIndex
...
+                }
+                else
+                {
+                    return SR.NodeDisplayPropertiesName2(name, this.ID);
+                }
+            }
+
+            set
+            {
+                ;
```
Charles-Gagnon, line 150: Is this needed?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Node.cs
```diff
@@ -55,6 +57,186 @@ public int GroupIndex
...
+        /// Gets the value that indicates whether the Node has critical warnings.
+        /// </summary>
+        private bool HasCriticalWarnings
+        {
+            get
+            {
+                if (this["Warnings"] != null)
+                {
+                    ExpandableObjectWrapper wrapper = this["Warnings"] as ExpandableObjectWrapper;
+                    if (wrapper["NoJoinPredicate"] != null)
```
Charles-Gagnon, line 198: Why this specific property?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ShowPlan/ShowPlanTests.cs
```diff
@@ -16,17 +17,15 @@ namespace Microsoft.SqlTools.ServiceLayer.UnitTests.ShowPlan
...
             StreamReader reader = new StreamReader(scriptStream);
             string text = reader.ReadToEnd();
-            var showPlanGraphs = ShowPlanGraph.ParseShowPlanXML(text, ShowPlanType.Actual);
-            Assert.AreEqual(1, showPlanGraphs.Length, "Single show plan graph not generated from the test xml file");
-            var testShowPlanGraph = showPlanGraphs[0];
-            Assert.NotNull(testShowPlanGraph, "graph should not be null");
-            Assert.NotNull(testShowPlanGraph.Root, "graph should have a root");
+            var showPlanGraphs = ShowPlanGraphUtils.CreateShowPlanGraph(text);
+            Assert.NotNull(showPlanGraphs, "graph should not be null");
+            Assert.NotNull(showPlanGraphs.Root, "graph should have a root");
```
Charles-Gagnon, line 28: Why did you remove these other checks? Are there no other checks you can do?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -893,13 +894,18 @@ public async Task<ResultSetSubset> InterServiceResultSubset(SubsetParams subsetP
...
+                {
+                    var xmlString = r.GetRow(0)[0].DisplayValue;
+                    graph = ShowPlanGraphUtils.CreateShowPlanGraph(xmlString);
+                }
                 ResultSetUpdatedEventParams eventParams = new ResultSetUpdatedEventParams
                 {
                     ResultSetSummary = r.Summary,
-                    OwnerUri = ownerUri
+                    OwnerUri = ownerUri,
+                    ShowPlanGraph = graph
```
Charles-Gagnon, line 907: I'm concerned about sending over a huge showplan graph - have you tested if there's any perf issues with those?

Charles-Gagnon (commented):
There seems to be a lot more than just sending the showplan graph, in the future please either make sure the title/description is accurate or better yet split into multiple PRs.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
aasimkhan30: I got these strings from SSMS. Should I move them somewhere else? I see that they have added some basic formated texts in .strings file.

Charles-Gagnon: I would just embed them directly into the code using string interpolation or string.Format

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Node.cs
aasimkhan30: I got these AA comments from SSMS. Will remove this prop as it is not used currently.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Node.cs
aasimkhan30: Got this piece of code from SSMS. I haven't really investigated the logic of this property

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
aasimkhan30: Since we are generating the show plan graph only for the last result set update, I don't think it will be a perf issue for the users. Also, most of the show plans graph are not excessively big.

Charles-Gagnon: When it's generated doesn't really matter if the size is too large.

How are you coming to your statement "most of the show plan graph are not that excessively big"? I'm not sure that's something we can take for granted. I would suggest talking to the QP experts in that area to get a sense for whether this is something we should be more careful of. And at the very least you should be doing testing to see if there's a limit for when perf starts to become a noticeable issue.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -14,6 +14,7 @@
 using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts.ExecuteRequests;
 using Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage;
 using Microsoft.SqlTools.ServiceLayer.SqlContext;
+using Microsoft.SqlTools.ServiceLayer.ShowPlan;
```
kburtram, line 17: sort order of usings looks incorrect.  Also, I see one out of order below.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -893,13 +894,18 @@ public async Task<ResultSetSubset> InterServiceResultSubset(SubsetParams subsetP
             // Setup the ResultSet updated callback
             ResultSet.ResultSetAsyncEventHandler resultUpdatedCallback = async r =>
             {
+                ExecutionPlanGraph graph = null;
+                if(r.Summary.Complete && r.Summary.SpecialAction.ExpectYukonXMLShowPlan == true)
```
kburtram, line 898: nit: add space between 'f' and '('.  Also comparing `ExpectYukonXMLShowPlan` with `true` is somewhat redundant since that is default test in if condition.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -893,13 +894,18 @@ public async Task<ResultSetSubset> InterServiceResultSubset(SubsetParams subsetP
...
+                {
+                    var xmlString = r.GetRow(0)[0].DisplayValue;
```
kburtram, line 900: I guess we always know that we can safely reference this element with `(0)[0]`.  i.e. no needs to bounds-check?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
kburtram: we should look at the perf of this scenario to understand it.  but we need to transmit the graph data to the client to render it at some point.  I guess maybe the XML is more compact than the processed JSON format. Though I suspect compared to the size of other things we pass around (e.g. query resultsets).  I suspect the delta between the XML format and JSON format isn't that impactful.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    /// <summary>
+    /// Execution plan graph object that is sent over JSON RPC
+    /// </summary>
+    public class ExecutionPlanGraph
+    {
+        /// <summary>
+        /// Root of the execution plan tree
+        /// </summary>
+        public ExecutionPlanNode Root {get; set;}
```
kburtram, line 13: inconsistent spacing around the '{' and 'g' and ';' and '}' compared to other class.

kburtram (approved):
Looks reasonable for a work in progress.  Several items that can be cleaned up.  But mostly fine for this PR.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
kburtram: do need to push more changes to resolve this?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        public ExecutionPlanNode Root { get; set; }
+        /// <summary>
+        /// Underlying query for the execution plan graph
+        /// </summary>
+        public string Query { get; set; }
+        /// <summary>
+        ///  Execution plan graph error
+        /// </summary>
+        public string error { get; set; }
```
kburtram, line 21: `error` should be capitalized

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Node.cs
```diff
@@ -55,6 +57,159 @@ public int GroupIndex
...
+                // The first line is the Physical name and the physical kind in parenthesis
+                // The second line should contains either Object value or LogicalOp name.
+                // The second line should not show the same content as the first line.
+
+                string firstLine = this["PhysicalOp"] as string;
+                if (firstLine == null)
+                {
+                    if (this.Operation == null)
+                    {
+                        return String.Empty;
```
kburtram, line 82: nit: `string.Empty` is preferred to `String.Empty` per .Net Framework Coding Guidelines.  Basically primitive types should be preferred to BCL classes when both are available.  Probably not worth rewriting the code, more FYI for next time.

kburtram (approved):
Looks mostly fine for hooking up this initial event.  I think there is a missing copyright and some other pending comments.  Though overall once major issues are addressed, I push this and take remaining issues for follow-up PRs as we iterate on the feature.

PR #1302, opened 2021-11-12
Fix version string

Charles-Gagnon:


Charles-Gagnon:
Not planning on it - this is just for the nuget packages I'm going to publish for the .NET Interactive folks

PR #1303, opened 2021-11-12
Updating ManagedBatchParser project target framework to netstandard2.1 as the…

ssreerama:
… DacFx is changed to 2.1 with 160 vbump.

This change is to fix warnings getting while restoring packages and building the projects. DacFx has updated to netstandard2.1 with the 160 vbump and the BatchParser project is using its 160_Nuget, which causes these warning with mismatched frameworks.
image

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,7 +4,7 @@
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
   <PropertyGroup>
-	<TargetFramework>netstandard2.1</TargetFramework>
+    <ManagedBatchParserTargetFramework>netstandard2.1</ManagedBatchParserTargetFramework>
     <CredentialsTargetFramework>net5.0</CredentialsTargetFramework>
     <ServiceLayerTargetFramework>net5.0</ServiceLayerTargetFramework>
     <ResourceProviderTargetFramework>net5.0</ResourceProviderTargetFramework>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,6 +4,7 @@
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
   <PropertyGroup>
+	<TargetFramework>netstandard2.1</TargetFramework>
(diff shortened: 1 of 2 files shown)


ssreerama:
Can someone please tag the owner of this project, would like to know if they are okay with the changes! Thanks :)

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFramework>netstandard2.0</TargetFramework>
+		<TargetFramework>netstandard2.1</TargetFramework>
```
Charles-Gagnon, line 3: Could you move this into https://github.com/microsoft/sqltoolsservice/blob/main/Directory.Build.props like the others? Might as well try to keep them together.

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
```diff
@@ -1,6 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFramework>netstandard2.0</TargetFramework>
```
ssreerama, line 3: Deleted the property only for ManagedBatchParser! remaining projects still targeting to what they have! let me know if you want to apply it for all other projects.

Directory.Build.props
```diff
@@ -4,6 +4,7 @@
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
   <PropertyGroup>
+	<TargetFramework>netstandard2.1</TargetFramework>
```
Charles-Gagnon, line 7: This should be`ManagedBatchParserTargetFramework` (also fix the spacing)

Charles-Gagnon (approved):

kisantia (approved):

PR #1305, opened 2021-11-15
Switch to using XLF for localization

smartguest:
Currently the build process uses the localized XLF file to generate the resx files, overwriting any changes made to the localized resx files. Manually copying strings from the resx updates into the XLF files is not feasible for localization, therefore I have decided to change the localization file to use XLF instead of resx, to ensure easier localization.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/LocProject.json
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/LocProject.json
@@ -4,8 +4,8 @@
       "LanguageSet": "AzureKatal_Languages",
       "LocItems": [
         {
-          "SourceFile": "src\\Microsoft.SqlTools.ServiceLayer\\Localization\\sr.resx",
-          "LclFile": "src\\Microsoft.SqlTools.ServiceLayer\\Localization\\LCL\\{Lang}\\sr.resx.lcl",
+          "SourceFile": "src\\Microsoft.SqlTools.ServiceLayer\\Localization\\sr.xlf",
+          "LclFile": "src\\Microsoft.SqlTools.ServiceLayer\\Localization\\LCL\\{Lang}\\sr.xlf.lcl",
           "CopyOption": "LangIDOnName",
           "OutputPath": "src\\Microsoft.SqlTools.ServiceLayer\\Localization"
         }


kburtram (changes requested):
I don't understand the need for this change.  Do we still add strings in the strings files and use srgen?  How does this tie into XLF vs. resx?  If it's not simple to explain as a PR comment perhaps this can be a topic for our weekly design sync meeting?

smartguest:
@kburtram @Charles-Gagnon We still add strings using sr.strings. What happens is that when I run build.cmd, the srgen task uses the strings found in sr.language.xlf in transxliff to generate sr.language.resx files using the strings found in the localized XLF files.

This means that if we copy over localized resx files into the STS directory from the CI build drop, they will be overwritten with outdated strings in the XLF files whenever we run the complete build task, this is very annoying to deal with.

Charles-Gagnon:
Could you remind me what the process is for updating the strings currently? I'm a little confused on where the copying of stuff from the CI pipeline comes into play here.

(did you ever get this documented anywhere? That would be very useful to have for future reference)

smartguest:
To update the strings in STS, here are the following steps:

- Update sr.strings with the strings you want to add, then run build.cmd srgen task, this will create newly generated sr.cs, sr.resx, and sr.xlf files with the new strings.

- Push the changes to STS, wait for LCL files to come back from localization team.

- Merge the LCL files into STS, wait for a new build from this pipeline: https://dev.azure.com/ms/sqltoolsservice/_build?definitionId=437

- Download the drop from the latest run in the pipeline.

The next step I want to change with this PR: 

Currently: 

- Copy over the localized resx files (inside the drop) into the STS localized language folder, and push it into STS.

This means localized XLF files won't be updated and srgen build task will overwrite the updated resx strings with outdated strings from the localized XLF file when its run each time (it uses them as input)

Planned:

- Copy over the localized XLF files into transXliff folder.

- Then run srgen task and it will update the resx files appropriately with the latest strings.

- Push the updated files into STS.

smartguest:
This process is not documented anywhere, it's similar to that of ADS, but I just realized this when running SRGEN after the LCL files were checked in that STS uses localized XLFs for input and updates the localized resx files instead of leaving the localized resx files alone. Therefore this PR is designed to reflect how the SRGEN task works.

Charles-Gagnon (approved):
So by updating this the loc pipeline you linked is going to output XLF files instead? 

Anyways I still don't have a clear understanding of this process (how do we get these initial LCL files) but if you've verified that this flow works then this seems fine.

Please document these steps (and be extremely specific in each steps about where the input files are coming from and what needs to be done with them) in the docs repo.

smartguest:
@Charles-Gagnon Yes precisely. We get the LCL files from a PR bot as far as I know, that's what is used to generate these resx and now XLF files in the CI pipeline.

PR #1306, opened 2021-11-16
Target win-x64 instead of win7-x64

Charles-Gagnon:
We were already bundling and using STS as if it was targeting win-x64 (all versions of windows used the same version) - and given that we aren't using any native APIs then this shouldn't have any actual effect on the service functionality. Tested locally and didn't see any immediate issues.

(the reason for making this change now is because we need to be targeting the correct runtime for the .NET Interactive packages, since we were targeting win7-x64 then the generated packages weren't working for any non-win7 Windows machines)

Also : 

1. Updating the package descriptions so we hopefully deter external people from using the package 
2. Fixing a path issue with one of the linux runtime packages

### DIFF
--- a/RefreshDllsForTestRun.cmd
+++ b/RefreshDllsForTestRun.cmd
@@ -5,13 +5,13 @@ IF [%_BuildConfiguration%] NEQ []  GOTO Start
 SET _BuildConfiguration=Debug
 
 :Start
-SET _PerfTestSourceLocation="%WORKINGDIR%\test\Microsoft.SqlTools.ServiceLayer.PerfTests\bin\%_BuildConfiguration%\net5.0\win7-x64\publish"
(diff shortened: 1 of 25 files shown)


chlafreniere (approved):

alanrenmsft (approved):

kburtram (approved):
LGTM.  I'm not sure the exact difference between these two targets, but I assume all the same Windows platforms are supported.

Charles-Gagnon:
win-x64 is a superset of win7-x64

https://docs.microsoft.com/en-us/dotnet/core/rid-catalog

So basically we're saying that instead of supporting specifically win7 we can run on any Windows platform (7 and up at least). There could potentially be problems with this down the line (dependent packages being framework dependent), but given that right now ADS is using this for all windows platforms and we haven't had any problems yet I'm not expecting any issues.

PR #1310, opened 2021-11-17
Always sign artifacts

Charles-Gagnon:
Previously signing was controlled by the `signed` var for binaries. To avoid the possibility of releasing unsigned bits I'm changing it to just always sign even things, even if we aren't releasing it.

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -124,7 +124,6 @@ steps:
      ]
     SessionTimeout: 600
     MaxConcurrency: 5
-  condition: and(succeeded(), eq(variables['signed'], true))
 
 - task: BatchScript@1
   displayName: 'Package published projects'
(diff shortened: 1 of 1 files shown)


kburtram (approved):
Is signing causing some type of problem?  Do we force this variable to true when the release flag is true?  Otherwise, I'm concerned we may end up shipping unsigned bits at some point.

Charles-Gagnon:
Not specifically, but the signing of the binaries was controlled by the signing var so this is more about being consistent.

But your suggestion of forcing signed to true when release is true is a good one. I'll do that as well.

kburtram:
Thanks, that makes sense.  If we can ensure all the packages we push to github are signed I think we're good.  Otherwise, the risk of shipping unsigned bits at some point in the future seems unnecessarily high (...and we should likely just sign everything, in that case).

alanrenmsft (approved):

Charles-Gagnon:
Alright, updated it to just always sign things regardless

alanrenmsft (approved):

PR #1316, opened 2021-11-18
add foreign keys and constraints

alanrenmsft:
1. add foreign keys and constraints to STS
2. make changes to reflect the ADS interface changes

note: the code that actually populate the foreign key and constraints data are not included in this PR, I am waiting for a new dacfx package.


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Constants.cs
```diff
@@ -11,6 +11,7 @@ public static class TablePropertyNames
         public const string Schema = "schema";
         public const string Description = "description";
         public const string Columns = "columns";
+        public const string ForeignKeys = "";
```
caohai, line 14: any reason to use empty string?

alanrenmsft: good catch. didn't notice this because we haven't enabled editing. :)

caohai (approved):

PR #1326, opened 2021-12-01
handle large decimal

alanrenmsft:
this PR fixes https://github.com/microsoft/azuredatastudio/issues/275
I was working on https://github.com/microsoft/azuredatastudio/issues/17578 and realized it is a dupe of https://github.com/microsoft/azuredatastudio/issues/275

the root cause is .NET's max value for decimal is 10^28 and the sql server's decimal max value is 10^38, when a value exceeds the max value for .NET decimal, conversion overflow exception will throw. 

for us, we only need to display the value, so we can use the sqldecimal value directly and do toString on it. dataReader.getValues() will convert the values to .NET type and dataReader.getSqlValues will return sqlDecimal type directly.

after making the change to use getSqlValues I noticed the current handlers simply doesn't work (it is not currently being used and was never used probably) since you can't cast a sql data type directly to its corresponding type in .NET. we need to get its underlying value by calling the sqldatatype.Value. I tested with the sql types that I have modified in this PR.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamWriter.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamWriter.cs
@@ -92,19 +92,19 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
                 {typeof(byte[]), val => WriteBytes((byte[]) val)},
                 {typeof(Guid), val => WriteGuid((Guid) val)},
 
-                {typeof(SqlString), val => WriteNullable((SqlString) val, obj => WriteString((string) obj))},
-                {typeof(SqlInt16), val => WriteNullable((SqlInt16) val, obj => WriteInt16((short) obj))},
-                {typeof(SqlInt32), val => WriteNullable((SqlInt32) val, obj => WriteInt32((int) obj))},
-                {typeof(SqlInt64), val => WriteNullable((SqlInt64) val, obj => WriteInt64((long) obj)) },
-                {typeof(SqlByte), val => WriteNullable((SqlByte) val, obj => WriteByte((byte) obj)) },
-                {typeof(SqlBoolean), val => WriteNullable((SqlBoolean) val, obj => WriteBoolean((bool) obj)) },
-                {typeof(SqlDouble), val => WriteNullable((SqlDouble) val, obj => WriteDouble((double) obj)) },
-                {typeof(SqlSingle), val => WriteNullable((SqlSingle) val, obj => WriteSingle((float) obj)) },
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/StorageDataReader.cs
```diff
@@ -98,7 +98,7 @@ public Task<bool> ReadAsync(CancellationToken cancellationToken)
         /// <returns>The value of the given column</returns>
         public object GetValue(int i)
         {
-            return sqlDataReader == null ? DbDataReader.GetValue(i) : sqlDataReader.GetValue(i);
+            return sqlDataReader == null ? DbDataReader.GetValue(i) : sqlDataReader.GetSqlValue(i);
```
alanrenmsft, line 101: I check that both in product and test code the data reader is already SqlDataReader, @kburtram  do you remember why we need to check whether is a SqlDataReader?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/StorageDataReader.cs
kburtram: no, I don't recall the details on why we're checking this here.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamWriter.cs
```diff
@@ -92,19 +92,19 @@ public ServiceBufferFileStreamWriter(Stream stream, QueryExecutionSettings setti
...
 
-                {typeof(SqlString), val => WriteNullable((SqlString) val, obj => WriteString((string) obj))},
-                {typeof(SqlInt16), val => WriteNullable((SqlInt16) val, obj => WriteInt16((short) obj))},
-                {typeof(SqlInt32), val => WriteNullable((SqlInt32) val, obj => WriteInt32((int) obj))},
-                {typeof(SqlInt64), val => WriteNullable((SqlInt64) val, obj => WriteInt64((long) obj)) },
-                {typeof(SqlByte), val => WriteNullable((SqlByte) val, obj => WriteByte((byte) obj)) },
-                {typeof(SqlBoolean), val => WriteNullable((SqlBoolean) val, obj => WriteBoolean((bool) obj)) },
-                {typeof(SqlDouble), val => WriteNullable((SqlDouble) val, obj => WriteDouble((double) obj)) },
-                {typeof(SqlSingle), val => WriteNullable((SqlSingle) val, obj => WriteSingle((float) obj)) },
+                {typeof(SqlString), val => WriteNullable((SqlString) val, obj => WriteString(((SqlString) obj).Value))},
```
kburtram, line 95: is this the same approach we use in SSMS to avoid this bug?

kburtram (approved):
LGTM

PR #1333, opened 2021-12-08
handle sql variable type

alanrenmsft:
for: https://github.com/microsoft/azuredatastudio/issues/17825

in this pr: https://github.com/microsoft/sqltoolsservice/pull/1326, I changed the code to use getSqlValues and as a result, the underlying types for sql variant are also sql types now, but the code is not able to resolve the type correctly from the string, the fix is to make the string fully qualified by appending the assembly's name.

with the fix:

highlighted columns in screenshot below are sql_variant columns.

image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
@@ -142,7 +142,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
                     colType = Type.GetType(sqlVariantType);
                     if (colType == null)
                     {
-                        throw new ArgumentException(SR.QueryServiceUnsupportedSqlVariantType(column.ColumnName, sqlVariantType));
+                        throw new ArgumentException(SR.QueryServiceUnsupportedSqlVariantType(sqlVariantType, column.ColumnName));
                     }
                 }
                 else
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -134,14 +134,15 @@ public IList<DbCellValue> ReadRow(long fileOffset, long rowId, IEnumerable<DbCol
                         continue;
                     }
 
-                    // The typename is stored in the string
+                    // We need to specify the assembly name for SQL types in order to resolve the type correctly.
+                    if (sqlVariantType.StartsWith("System.Data.SqlTypes."))
+                    {
+                        sqlVariantType = sqlVariantType + ", System.Data.Common";
```
Charles-Gagnon, line 140: Is this code from SSMS or something? how are they handling it otherwise?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
alanrenmsft: similar to SSMS with modification, but this specific line, SSMS doesn't have to do this.

alanrenmsft: as per the microsoft doc:
image

it should be full qualified name, not sure why SSMS doesn't need it, I can try to see whether .net 4 and .net 5 makes a difference.

alanrenmsft: .net 4.7 and .net 5 all behave the same, null is returned is only the first part is provided, I don't have a SSMS debug environment, I can ask Matteo to debug and share the findings.

PR #1349, opened 2021-12-22
Remove XElite for now

smartguest:
This reverts commit 196364e81ee16a70768ecd566b3f82ab772b0a2e.

This will be done until any major bugs with XElite are addressed

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
@@ -1572,16 +1572,6 @@
     <comment>.
  Parameters: 0 - sessionName (String) </comment>
   </data>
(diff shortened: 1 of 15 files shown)


src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -1571,16 +1571,6 @@
     <value>An XEvent session named {0} already exists</value>
     <comment>.
  Parameters: 0 - sessionName (String) </comment>
-  </data>
```
alanrenmsft, line 1574: we should also remove these from the xlf file, right?

smartguest: that is correct

alanrenmsft (approved):

PR #1351, opened 2021-12-23
Parsing nested properties in show plan nodes.

aasimkhan30:
Adding nested properties in show plan nodes. This code will recursively parse a property that has a collection of properties as its value.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -47,7 +47,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// <summary>
         /// Node properties to be shown in the tooltip
         /// </summary>
-        public List<ExecutionPlanGraphElementProperties> Properties { get; set; }
+        public List<ExecutionPlanGraphPropertyBase> Properties { get; set; }
         /// <summary>
         /// Display name for the node
         /// </summary>
@@ -64,16 +64,12 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         public List<ExecutionPlanEdges> Edges { get; set; }
     }
 
-    public class ExecutionPlanGraphElementProperties
+    public class ExecutionPlanGraphPropertyBase
     {
         /// <summary>
         /// Name of the property
         /// </summary>
         public string Name { get; set; }
-        /// <summary>
-        /// Formatted value for the property. In case of nested properties, it can an array of ExecutionPlanGraphElementProperties
-        /// </summary>
-        public object Value { get; set; }
         /// <summary>
         /// Flag to show/hide props in tooltip
         /// </summary>
@@ -86,6 +82,22 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// Flag to indicate if the property has a longer value so that it will be shown at the bottom of the tooltip
         /// </summary>
         public bool IsLongString { get; set; }
+    }
+
+    public class NestedExecutionPlanGraphProperty: ExecutionPlanGraphPropertyBase
+    {
+         /// <summary>
+        /// Formatted value for the property. In case of nested properties, it can an array of ExecutionPlanGraphElementProperties
+        /// </summary>
+        public List<ExecutionPlanGraphPropertyBase> Value { get; set; }
+    }
+
+    public class ExecutionPlanGraphProperty : ExecutionPlanGraphPropertyBase
+    {
+        /// <summary>
+        /// Formatted value for the property. In case of nested properties, it can an array of ExecutionPlanGraphElementProperties
+        /// </summary>
+        public string Value { get; set; }
     }
 
     public class ExecutionPlanEdges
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -71,9 +71,9 @@ public class ExecutionPlanGraphElementProperties
         /// </summary>
```
Charles-Gagnon, line 71: The name `ExecutionPlanGraphElementProperties` doesn't seem like an accurate name given how it seems to be used below - shouldn't it be something like `ExecutionPlanGraphProperty` instead?

Charles-Gagnon (commented):
Please adds tests.

And is this something that needs to be fixed in the SSMS code as well?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
aasimkhan30: This property belongs to graph nodes and edges and not the graph itself. Will the new name still make sense?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -71,9 +71,9 @@ public class ExecutionPlanGraphElementProperties
...
         public string Name { get; set; }
         /// <summary>
-        /// Formatted value for the property
+        /// Formatted value for the property. In case of nested properties, it can an array of ExecutionPlanGraphElementProperties
         /// </summary>
-        public string FormattedValue { get; set; }
+        public object Value { get; set; }
```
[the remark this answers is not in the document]
aasimkhan30: In the ads implementation, we are converting the SSMS graph object into a generic graph that will consumed by the front-end. This issue therefore does not exist in SSMS.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
Charles-Gagnon: It's still a part of an ExecutionPlanGraph though so that still seems to make sense to me. But feel free to make it more generic if you want, I was mostly commenting on the `ElementProperties` part of the name

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -64,17 +64,13 @@ public class ExecutionPlanNode
         public List<ExecutionPlanEdges> Edges { get; set; }
     }
 
-    public class ExecutionPlanGraphElementProperties
+    public class ExecutionPlanGraphPropertyBase
```
Charles-Gagnon, line 67: I would just simplify this into a single class that takes a generic as the value `ExecutionPlanGraphProperty<T>` - that way you don't need to create the two separate implementations (which don't actually do anything different)

Charles-Gagnon: (and if you wanted to have a strong type for the nested values then you should be able to just do this 

`public class NestedExecutionPlanGraphProperty extends ExecutionPlanGraphProperty<IList<ExecutionPlanGraphProperty>> { }`

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraphUtils.cs
```diff
@@ -50,19 +50,37 @@ private static ExecutionPlanEdges ConvertShowPlanEdgeToExecutionPlanEdge(Edge ed
...
 
-        private static List<ExecutionPlanGraphElementProperties> GetProperties(PropertyDescriptorCollection props)
+        private static List<ExecutionPlanGraphPropertyBase> GetProperties(PropertyDescriptorCollection props)
         {
-            List<ExecutionPlanGraphElementProperties> propsList = new List<ExecutionPlanGraphElementProperties>();
+            List<ExecutionPlanGraphPropertyBase> propsList = new List<ExecutionPlanGraphPropertyBase>();
             foreach (PropertyValue prop in props)
             {
-                propsList.Add(new ExecutionPlanGraphElementProperties()
+                var complexProperty = prop.Value as Microsoft.SqlTools.ServiceLayer.ShowPlan.ShowPlanGraph.ExpandableObjectWrapper;
```
lewis-sanchez, line 58: Can simplify type name to just `ExpandableObjectWrapper` because the namespace is already being declared on line 9.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
lewis-sanchez: I agree with Charles on this, but I also think this class can be made an interface since the intention seems to be to facilitate creating the props list declared in `showPlanGraphUtils.cs` to contain both subclasses: 
image

In that case, I would rename and make `ExecutionPlanGraphPropertyBase` an interface since it's not defining any behavior and it's also not being instantiated directly and keep the 2 subclasses.

kburtram (approved):

PR #1354, opened 2022-01-06
Corrects duplicate files in publish output error.

lewis-sanchez:
This PR resolves a new error generated by .NET SDK (NETSDK1152) in situations where files from different source paths would be copied to the same file path in the publish output.


Charles-Gagnon:
What error is this? I'm not sure hiding the error is the best action here - or is this some expected behavior from dotnet? Why are we even copying a file from different places into the same destination? If they're from different source paths then that more seems like a publishing error that should be fixed.

lewis-sanchez:
The error is a new one that was introduced with .NET SDK 6.0.100. I came across the error when I ran "dotnet publish -r win-x64" in the terminal to locally publish the sql tools service after building it.

lewis-sanchez:
Hiding the error isn't sitting well with me. I'll close this PR until a better approach is found.

PR #1357, opened 2022-01-10
Addin new Dac Deployment Option and getting description from DacFx API

ssreerama:
This PR contains:
1. Newly added or missed Dac deployment options to the dacfx service, now dacpac/SC/SQL DB Proj extensions will have these available.
2. Description for the options are now pulling from theDacFx API and added to the options model, now we don't need to duplicate the huge list of descriptions in all extensions.
3. Modified and fixed the tests cases accordingly.


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -3,13 +3,22 @@
...
 using Microsoft.SqlServer.Dac;
 
 namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 {
+    /// <summary>
+    /// Class to define deployment option default value and the description
+    /// </summary>
+    public class DeploymentOptionProps
+    {
+        public object value { get; set; }
```
Charles-Gagnon, line 19: Would suggest making this a generic class so the value has an actual type

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -3,13 +3,22 @@
...
+        public string description { get; set; } = string.Empty;
```
Charles-Gagnon, line 20: C# properties should be capitalized

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -234,11 +270,25 @@ public DeploymentOptions()
             foreach (var deployOptionsProp in deploymentOptionsProperties)
             {
                 var prop = options.GetType().GetProperty(deployOptionsProp.Name);
+                var attribute = prop.GetCustomAttributes(typeof(DescriptionAttribute), true)[0];
```
Charles-Gagnon, line 273: 1. Is this safe to do? It would be best to assume this property won't always have the attribute on it and handle that correctly.
2. Would suggest using the generics version of this method (`GetCustomAttributes<DescriptionAttribute>(true)`)
3. Generally you should do `FirstOrDefault` instead of directly accessing the first element, especially when you can't be sure that the list isn't empty. 
4. You shouldn't need the cast to DescriptionAttribute with the proper typing on the GetCustomAttributes call

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -234,11 +270,25 @@ public DeploymentOptions()
...
+                var description = (DescriptionAttribute)attribute;
 
                 // Note that we set excluded object types here since dacfx has this value as null;
                 if (prop != null && deployOptionsProp.Name != "ExcludeObjectTypes")
                 {
-                    deployOptionsProp.SetValue(this, prop.GetValue(options));
+                    if (deployOptionsProp.Name == "DoNotDropObjectTypes")
```
Charles-Gagnon, line 279: Please add a comment about why you're ignoring this one. 

And generally we should avoid using magic strings like this - please make a const in a common location somewhere for these special names.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -234,11 +270,25 @@ public DeploymentOptions()
...
+                    {
+                        deployOptionsProp.SetValue(this, prop.GetValue(options));
+                    }
+                    else
+                    {
+                        DeploymentOptionProps setProp = new DeploymentOptionProps()
+                        {
+                            value = prop.GetValue(options),
+                            description = description.Description
```
Charles-Gagnon, line 288: Why is this one special and have its value be a nested DeploymentOptionProps value? If that's expected shouldn't prop.GetValue be returning the correct type of object to begin with?

And the description seems like something you should have all the properties be adding as long as you're doing the proper error checking as I mentioned above.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -296,10 +346,25 @@ public void SetOptions(DacDeployOptions options)
             foreach (var deployOptionsProp in deploymentOptionsProperties)
             {
                 var prop = options.GetType().GetProperty(deployOptionsProp.Name);
+                var attribute = prop.GetCustomAttributes(typeof(DescriptionAttribute), true)[0];
```
Charles-Gagnon, line 349: Duplicated code, please refactor into helper method

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -32,7 +32,24 @@ internal static DacDeployOptions CreateSchemaCompareOptions(DeploymentOptions de
                 var prop = dacOptions.GetType().GetProperty(deployOptionsProp.Name);
                 if (prop != null)
                 {
-                    prop.SetValue(dacOptions, deployOptionsProp.GetValue(deploymentOptions));
+                    var optionVal = deployOptionsProp.GetValue(deploymentOptions);
+                    if (deployOptionsProp.Name == "DoNotDropObjectTypes" || deployOptionsProp.Name == "ExcludeObjectTypes")
```
Charles-Gagnon, line 36: Could you not check if the type of the value is DeploymentOptionProps? That would make this future proof so you don't have to keep coming back and adding more props to the list here and elsewhere in this code if/when more are added you want to exclude.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -32,7 +32,24 @@ internal static DacDeployOptions CreateSchemaCompareOptions(DeploymentOptions de
...
+                    {
+                        prop.SetValue(dacOptions, optionVal);
+                    }
+                    else
+                    {
+                        // Get value of the DeploymentOptionsProps option value property
+                        var value = ((DeploymentOptionProps)optionVal).value;
+                        
+                        // Convert Int64 to Int32
```
Charles-Gagnon, line 45: Why? Comments should strive to explain why something is being done, we can already see the code is already converting from Int64 to Int32. 

Is overflow not something you're worried about?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1827,12 +1827,34 @@ private void CreateAndValidateScmpFile(SchemaCompareEndpointInfo sourceInfo, Sch
                 DeploymentOptions = new DeploymentOptions()
                 {
                     // change some random ones explicitly
-                    AllowDropBlockingAssemblies = true,
-                    DropConstraintsNotInSource = true,
-                    IgnoreAnsiNulls = true,
-                    NoAlterStatementsToChangeClrTypes = false,
-                    PopulateFilesOnFileGroups = false,
-                    VerifyDeployment = false,
+                    AllowDropBlockingAssemblies = new DeploymentOptionProps
```
Charles-Gagnon, line 1830: It would be good to add some tests verifying that the logic handles loading the nested props types and types without a description properly.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: prop.GetValue(options) gives the value of the property coming from the DacFx which is a single bool/int/string value. where in STS I have defined a new class to include the description attribute text value along with the above value to pass it to ADS, which made me to create a new instance with value/description to set the values to these props.

Is there a way we can directly access the description attribute text defined in the c#(in DacFx) in typeScript(ADS)? Or am i missing anything! Thanks.

Charles-Gagnon: So that part is fine, I was just confused why on 281 you were setting the value directly while here on 290 you were creating this new DeploymentOptionProps object and assigning that instead. But looking at it I see that's because those types are just simple arrays instead of needing to be this props object.

Is there any reason that those aren't "DeploymentOptionProps" as well? It seems like they could be and just have the array of enum values be the value. That would simplify this logic a bit which would be nice.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -3,13 +3,22 @@
...
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.SqlServer.Dac;
 
 namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 {
+    /// <summary>
+    /// Class to define deployment option default value and the description
+    /// </summary>
+    public class DeploymentOptionProps
```
Charles-Gagnon, line 17: `DeploymentOptionProp` seems like a better name since each instance is just for a single property (and consider `DeploymentOptionProperty`. Generally unless you really need to it's best to avoid shortening names to avoid possible confusion)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Sure, Since these two props are not coming directly from DacFx and have no description, I just didn't updated them, But keeping them in as DeploymentOptionProperty would be good, have made changes and working on test cases! will add the Iteration. Thanks

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -296,14 +336,66 @@ public void SetOptions(DacDeployOptions options)
...
+        /// </summary>
+        /// <param name="prop"></param>
+        /// <param name="val"></param>
+        /// <param name="deployOptionsProp"></param>
+        public void SetGenericDeployOptionProps(PropertyInfo prop, DacDeployOptions options, PropertyInfo deployOptionsProp)
+        {
+            var val = prop.GetValue(options);
+            var attribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
+
+            // DeploymentOptionProperty is a generic type, which requires value type to intitialize
```
Charles-Gagnon, line 358: Couldn't you just use Object as the type instead since you don't really care about the specific type when setting it?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -296,14 +336,66 @@ public void SetOptions(DacDeployOptions options)
...
+            {
+                DeploymentOptionProperty<bool> setProp = new DeploymentOptionProperty<bool>()
+                {
+                    Value = (bool)val,
+                    Description = attribute.Description
+                };
+
+                deployOptionsProp.SetValue(this, setProp);
+            }
+            else if (val != null && val.GetType() == typeof(System.Object))
```
Charles-Gagnon, line 387: If the value is `ObjectType[]` then I don't think this is ever going to be true (and assuming that any object is that specific type array doesn't seem correct.

(but with my comment above I don't think you even need all this logic so this is mostly as an FYI)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Yes, we really don't care what type of value the property has, we mostly focusing on bool and other types which are not null! In such case I am going to revert the generic type to just an object, since setting object type to boolean or something would be an exception case unless typecasting. or is there any other approach to leave as generic and achieve the same with object type!

Charles-Gagnon: You should leave the class generic since that's useful to have when accessing the properties on the object. I'm just saying here when you're creating it you just create it with the type of object so that it can take anything.

ssreerama: Looks like without typecasting work is not going to be done for at least ObjectTypes[]! Since the DeploymentOptionProperty<object> has to set to the Dac Options of <ObjectType[]>.

DeploymentOptionProperty<object> is going to set to DeploymentOptionProperty<bool/string/ObjectType[]> not the reverse way! so I am going to do all the properties as object type except the ObjectType[] which needs a typecasting by verifying with the property type.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -24,15 +25,24 @@ internal static class SchemaCompareUtils
...
             foreach (var deployOptionsProp in deploymentOptionsProperties)
             {
                 var prop = dacOptions.GetType().GetProperty(deployOptionsProp.Name);
                 if (prop != null)
                 {
-                    prop.SetValue(dacOptions, deployOptionsProp.GetValue(deploymentOptions));
+                    var val = deployOptionsProp.GetValue(deploymentOptions);
+                    var selectedVal = val.GetType().GetProperty("Value").GetValue(val);
+
+                    // The integer value coming from the DeploymentOptions is of Int64 which overflows to set to DacDeployOptions.
```
Charles-Gagnon, line 39: What does this mean? You should be explaining why this conversion is necessary. And I'd also explain why it isn't a concern to be converting an int64 into an int32 - the convert function is going to throw if it's given a value that's too large 

`Run-time exception (line 10): Value was either too large or too small for an Int32.`

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -296,14 +472,47 @@ public void SetOptions(DacDeployOptions options)
...
+        /// </summary>
+        /// <param name="prop"></param>
+        /// <param name="val"></param>
+        /// <param name="deployOptionsProp"></param>
+        public void SetGenericDeployOptionProps(PropertyInfo prop, DacDeployOptions options, PropertyInfo deployOptionsProp)
+        {
+            var val = prop.GetValue(options);
+            var attribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
+
+            if (prop != null && prop.PropertyType == typeof(ObjectType[]))
```
Charles-Gagnon, line 494: You access prop above so the null check isn't necessary here (and you could shortcut it anyways with `prop?.PropertyType == typeof(ObjectType[])`)

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
ssreerama: Here the DacDeployOptions(DacFx model) has the int type of 32bin signed integer. when the DeploymentOptionProperty model coming back from ADS, these int values are as Int64 (really not sure what made this, any idea). when preparing the DacDeployOptions model with these return values, it is throwing an error **{"Object of type 'System.Int64' cannot be converted to type 'System.Int32'."}**. 

Since the default value is coming from DacFx as Int32 and is not going to change in ADS (as these fields are not available to edit in UI). I believe the exception won't be an issue here! Thanks

Charles-Gagnon: I'm guessing it's because the JSON RPC layer just transfers numbers and not anything about them so it defaults to Int64 on the C# side. But in general leaving stuff up to a guess and hoping things work out isn't a great practice so I'd suggest : 

1. Spending a bit of time digging into it and finding out whether my guess above was correct or whether it's an issue elsewhere
2. Whatever it ends up being explain in the comment exactly why this needs to be done and why it's safe to assume this will work.

ssreerama: https://stackoverflow.com/questions/8297541/how-do-i-change-the-default-type-for-numeric-deserialization

https://web.archive.org/web/20160606030722/http://json.codeplex.com/workitem/20832

With some time spent on researching the reason, it came that "JSON by default reads integer values as Int64 because there is no way to know whether the value should be Int32 or Int64, and Int64 is less likely to overflow. For a typed property the deserializer knows to convert the Int64 to a Int32" but here the type is object, so it gives as 64bit which needs a conversion. Thanks, will dosome more research if there is any other limitations and update comment more clear.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -296,14 +472,58 @@ public void SetOptions(DacDeployOptions options)
...
+        /// </summary>
+        /// <param name="prop"></param>
+        /// <param name="val"></param>
+        /// <param name="deployOptionsProp"></param>
+        public void SetGenericDeployOptionProps(PropertyInfo prop, DacDeployOptions options, PropertyInfo deployOptionsProp)
+        {
+            var val = prop.GetValue(options);
+            var attribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
+
+            if (prop.PropertyType == typeof(ObjectType[]))
```
Charles-Gagnon, line 494: My PR solves needing to have these if/else statements which greatly reduces the duplication and also has the benefit of not hardcoding the "supported" types so adding a new type later on is easier. I'd highly recommend looking at taking in my solution there instead of reverting back to this...

ssreerama: Oh No! I missed reviewing the total changes of your PR, I barely saw few comments and thought those are all changes! let me work on the suggessions! Thanks

Charles-Gagnon: And note that the purpose of that PR was that you could merge it into your branch and thus not require actually making the changes yourself. So if things look good just click "squash and merge" and then go pull down the latest from your branch.

ssreerama: Actually, I have copied the changes and made all updates and later saw the comment about merging. since already added these changes can we close the other PR or is there a way to merge that into it. need help here! Thanks :)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -3,192 +3,215 @@
...
+        {
+            this.Value = value;
+            this.Description = description;
+            this.DisplayName = displayName;
+        }
+
+        public T Value { get; set; }
+        public string Description { get; set; } = string.Empty;
+
+        // To Diaply the options in ADS extensions UI in SchemaCompare/SQL-DB-Project/Dacpac extensions
```
Charles-Gagnon, line 31: 1. Comments should be XML doc (`///`) format
2. You spelled display incorrectly

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -3,192 +3,215 @@
...
+        public DeploymentOptionProperty<bool> IgnoreColumnOrder { get; set; }
 
-        public bool IgnoreTablePartitionOptions { get; set; } // DW Specific
+        public DeploymentOptionProperty<bool> IgnoreTablePartitionOptions { get; set; } // DW Specific
 
-        public string AdditionalDeploymentContributorPaths { get; set; } = string.Empty;
+        public DeploymentOptionProperty<string> AdditionalDeploymentContributorPaths { get; set; }
 
-        public ObjectType[] DoNotDropObjectTypes { get; set; } = null;
+        public DeploymentOptionProperty<ObjectType[]> DoNotDropObjectTypes { get; set; } = new DeploymentOptionProperty<ObjectType[]>(null);
```
Charles-Gagnon, line 210: Why null? Seems like you should just not define or have an empty array if there aren't any ObjectTypes you want to include here

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -210,15 +233,172 @@ public class DeploymentOptions
...
+        /// Adding new properties here would give easy handling of new option to all extensions
+        /// </summary>
+        private void SetDisplayNameForOption()
+        {
+            #region Display Name and Dac Options Mapping
+            DacDeployOptions d = new DacDeployOptions();
+            _displayNameMapDict = new Dictionary<string, string>();
+
+            // Ex: displayNameMapDict["DacFx Option Name"] = "ADS UI Display Name"
+            _displayNameMapDict[nameof(d.AdditionalDeploymentContributorArguments)] = "Additional Deployment Contributor Arguments";
```
Charles-Gagnon, line 283: Is this stuff something you could put into DacFX instead? Having to define it here still isn't great since that means every consumer of this functionality has to redefine these strings. @kisantia FYI

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -210,15 +233,172 @@ public class DeploymentOptions
...
         #endregion
 
+        /// <summary>
+        /// Mapping the DisplayName to the dac deploy option
+        /// Adding new properties here would give easy handling of new option to all extensions
+        /// </summary>
+        private void SetDisplayNameForOption()
+        {
+            #region Display Name and Dac Options Mapping
+            DacDeployOptions d = new DacDeployOptions();
```
Charles-Gagnon, line 279: Seems kind of weird to have this in its own function that creates a temporary object soley to get some name values. You should be able to just create the dictionary where you define it above and use `nameof(DacDeployOptions.<PropName>)` for the keys

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: @udeeshagautam this is what we discussed to get the display names from the DacFx like Description that we are already getting, which can be reusable at all extensions. I'll create a new task for this one and will be working on right after this PR.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -2,192 +2,217 @@
...
+using System.Collections.Generic;
 
 namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 {
     /// <summary>
-    /// Class to define deployment options. 
+    /// Class to define deployment option default value and the description
+    /// Value: Default and selected value of the deployment options
+    /// Description: Description of the deployment options
+    /// DisplayName: To Diaply the options in ADS extensions UI in SchemaCompare/SQL-DB-Project/Dacpac extensions
```
kisantia, line 20: typo - "To _display_ the options"

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -2,192 +2,217 @@
...
+    /// </summary>
+    public class DeploymentOptionProperty<T>
+    {
+        public DeploymentOptionProperty(T value, string description = "", string displayName = "")
```
kisantia, line 24: Why do description and displayName need to have default values of empty string?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -2,192 +2,217 @@
...
+        public DeploymentOptionProperty<bool> CreateNewDatabase { get; set; }
 
-        public bool CompareUsingTargetCollation { get; set; }
+        public DeploymentOptionProperty<bool> CompareUsingTargetCollation { get; set; }
 
-        public bool CommentOutSetVarDeclarations { get; set; }
+        public DeploymentOptionProperty<bool> CommentOutSetVarDeclarations { get; set; }
 
-        public int CommandTimeout { get; set; } = 120;
+        public DeploymentOptionProperty<int> CommandTimeout { get; set; } = new DeploymentOptionProperty<int>(120);
```
kisantia, line 141: please add comments for what these timeout values mean, like how setting LongRunningCommandTimeout to 0 means to wait infinitely

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -209,15 +234,172 @@ public class DeploymentOptions
...
+            _displayNameMapDict[nameof(d.PopulateFilesOnFileGroups)] = "Populate Files On File Groups";
+            _displayNameMapDict[nameof(d.PreserveIdentityLastValues)] = "Preserve Identity Last Values";
+            _displayNameMapDict[nameof(d.RebuildIndexesOfflineForDataPhase)] = "Rebuild Indexes Offline For Data Phase";
+            _displayNameMapDict[nameof(d.RegisterDataTierApplication)] = "Register Data Tier Application";
+            _displayNameMapDict[nameof(d.RestoreSequenceCurrentValue)] = "Restore Sequence Current Value";
+            _displayNameMapDict[nameof(d.RunDeploymentPlanExecutors)] = "Run Deployment Plan Executors";
+            _displayNameMapDict[nameof(d.ScriptDatabaseCollation)] = "Script Database Collation";
+            _displayNameMapDict[nameof(d.ScriptDatabaseCompatibility)] = "Script Database Compatibility";
+            _displayNameMapDict[nameof(d.ScriptDatabaseOptions)] = "Script Database Options";
+            _displayNameMapDict[nameof(d.ScriptDeployStateChecks)] = "Script DeployS tate Checks";
```
kisantia, line 384: typo - Deploy State

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -209,15 +234,172 @@ public class DeploymentOptions
...
+            _displayNameMapDict[nameof(d.DoNotAlterChangeDataCaptureObjects)] = "Do Not Alter Change Data Capture Objects";
+            _displayNameMapDict[nameof(d.DoNotAlterReplicatedObjects)] = "Do Not Alter Replicated Objects";
+            _displayNameMapDict[nameof(d.DoNotDropDatabaseWorkloadGroups)] = "Do Not Drop Database Workload Groups";
+            _displayNameMapDict[nameof(d.DoNotDropObjectTypes)] = "Do Not Drop Object Types";
+            _displayNameMapDict[nameof(d.DoNotDropWorkloadClassifiers)] = "Do Not Drop Workload Classifiers";
+            _displayNameMapDict[nameof(d.DoNotEvaluateSqlCmdVariables)] = "Do Not Evaluate Sql Cmd Variables";
+            _displayNameMapDict[nameof(d.DropConstraintsNotInSource)] = "Drop Constraints Not In Source";
+            _displayNameMapDict[nameof(d.DropDmlTriggersNotInSource)] = "Drop Dml Triggers Not In Source";
+            _displayNameMapDict[nameof(d.DropExtendedPropertiesNotInSource)] = "Drop Extended Properties Not In Source";
+            _displayNameMapDict[nameof(d.DropIndexesNotInSource)] = "Drop Indexes NotInSource";
```
kisantia, line 320: Not In Source should be 3 words here and the line below

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -209,15 +234,172 @@ public class DeploymentOptions
...
+            _displayNameMapDict[nameof(d.IgnoreDatabaseWorkloadGroups)] = "Ignore Database Workload Groups";
+            _displayNameMapDict[nameof(d.IgnoreDdlTriggerOrder)] = "Ignore Ddl Trigger Order";
+            _displayNameMapDict[nameof(d.IgnoreDdlTriggerState)] = "Ignore Ddl Trigger State";
+            _displayNameMapDict[nameof(d.IgnoreDefaultSchema)] = "Ignore Default Schema";
+            _displayNameMapDict[nameof(d.IgnoreDmlTriggerOrder)] = "Ignore Dml Trigger Order";
+            _displayNameMapDict[nameof(d.IgnoreDmlTriggerState)] = "Ignore Dml Trigger State";
+            _displayNameMapDict[nameof(d.IgnoreExtendedProperties)] = "Ignore Extended Properties";
+            _displayNameMapDict[nameof(d.IgnoreFileAndLogFilePath)] = "Ignore File And Log File Path";
+            _displayNameMapDict[nameof(d.IgnoreFileSize)] = "Ignore File Size";
+            _displayNameMapDict[nameof(d.IgnoreFilegroupPlacement)] = "Ignore File group Placement";
```
kisantia, line 345: capitalize "G" in group for consistency

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Some DeploymentOptions like CommandTimeout/LongRunnigCommandTimeout are giving with a default value. Instead of passing a null values to the description and displayNames during that initialization for all other options, have added a default null value at here!

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -2,192 +2,220 @@
...
+using System.Collections.Generic;
 
 namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 {
     /// <summary>
-    /// Class to define deployment options. 
+    /// Class to define deployment option default value and the description
+    /// Value: Default and selected value of the deployment options
+    /// Description: Description of the deployment options
+    /// DisplayName: To display the options in ADS extensions UI in SchemaCompare/SQL-DB-Project/Dacpac extensions
```
kisantia, line 20: move the comments for these properties to be over the actual properties below

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -2,192 +2,220 @@
...
+
+        public T Value { get; set; }
+        public string Description { get; set; } = string.Empty;
+        public string DisplayName { get; set; }
+    }
+
+    /// <summary>
+    /// Class to define deployment options.
     /// Keeping the order and defaults same as DacFx
     /// The default values here should also match the default values in ADS UX
```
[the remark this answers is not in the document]
ssreerama: Added

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -24,15 +25,26 @@ internal static class SchemaCompareUtils
...
                 var prop = dacOptions.GetType().GetProperty(deployOptionsProp.Name);
                 if (prop != null)
                 {
-                    prop.SetValue(dacOptions, deployOptionsProp.GetValue(deploymentOptions));
+                    var val = deployOptionsProp.GetValue(deploymentOptions);
+                    var selectedVal = val.GetType().GetProperty("Value").GetValue(val);
+
+                    // JSON.NET by default reads Number type as Int64, deserializing an object type to dacOptions of Int32 type required to convert into Int32 from Int64.
+                    // If not converted setting value(Int64) to dacOption(Int32) will throw {"Object of type 'System.Int64' cannot be converted to type 'System.Int32'."}.
+                    // As these integet type options are non-editable and are not availbale in ADS to update, integer overflow exception will not be happening here.
```
Benjin, line 41: typo: integer

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -843,15 +845,19 @@ private bool ValidateOptions(DeploymentOptions expected, DeploymentOptions actua
...
+                var actualPValue = actualP.GetType().GetProperty("Value").GetValue(actualP);
 
                 if (v.Name == "ExcludeObjectTypes")
                 {
                     Assert.True((defaultP as ObjectType[])?.Length == (actualP as ObjectType[])?.Length, "Number of excluded objects is different not equal");
                 }
                 else
                 {
-                    Assert.True((defaultP == null && actualP == null) || (defaultP == null && (actualP as string) == string.Empty) || defaultP.Equals(actualP), $"Actual Property from Service is not equal to default property for {v.Name}, Actual value: {actualP} and Default value: {defaultP}");
+                    Assert.True((defaultP == null && actualP == null) || (defaultP == null && (actualP as string) == string.Empty) || defaultP.Equals(actualP)
```
Benjin, line 858: Can you break this up into multiple lines and/or multiple asserts and add comments?

Also, `String.IsNullOrEmpty(arg)` may be helpful in making this more readable.

Benjin, line 858: Can you break this up into multiple lines and/or multiple asserts and add comments?

Also, `String.IsNullOrEmpty(arg)` may be helpful in making this more readable.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
```diff
@@ -135,11 +135,17 @@ internal static void CompareOptions(DeploymentOptions deploymentOptions, DacDepl
...
-                var dacValue = dacProp.GetValue(dacDeployOptions);
+                var changedDacValue = deployOptionsValue != null ? deployOptionsValue.GetType().GetProperty("Value").GetValue(deployOptionsValue) : deployOptionsValue;
+                var dafaultDacValue = dacProp.GetValue(dacDeployOptions);
 
                 if (deployOptionsProp.Name != "ExcludeObjectTypes") // do not compare for ExcludeObjectTypes because it will be different
                 {
-                    Assert.True((deployOptionsValue == null && dacValue == null) || deployOptionsValue.Equals(dacValue), $"DacFx DacDeploy property not equal to Tools Service DeploymentOptions for { deployOptionsProp.Name}, SchemaCompareOptions value: {deployOptionsValue} and DacDeployOptions value: {dacValue} ");
+                    Assert.True((deployOptionsValue == null && dafaultDacValue == null) 
+                        || deployOptionsValue.Equals(dafaultDacValue)
+                        || changedDacValue == null && (dafaultDacValue as string) == string.Empty
```
Benjin, line 145: `String.IsNullOrEmpty()`

Benjin, line 145: `String.IsNullOrEmpty()`

Benjin (approved):

PR #1359, opened 2022-01-12
Use Activator.CreateInstance to create generic type at runtime

Charles-Gagnon:
Had to play around with the types a bit but I think this should work and allow you to keep the strong typings on the properties.

I didn't actually try it at runtime, if you could do that and things look good then merge this in and update the rest of the properties to have the correct type.

(you could also use reflection to set the properties, but making a constructor seemed easiest)

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
@@ -18,6 +18,14 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     /// </summary>
     public class DeploymentOptionProperty<T>
     {
+
+        public DeploymentOptionProperty(T value, string description = "", string displayName = "")
+        {
+            this.Value = value;
+            this.Description = description;
+            this.DisplayName = displayName;
+        }
+
         public T Value { get; set; }
         public string Description { get; set; } = string.Empty;
 
@@ -33,7 +41,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     {
         #region Properties
 
-        public DeploymentOptionProperty<object> IgnoreTableOptions { get; set; }
+        public DeploymentOptionProperty<bool> IgnoreTableOptions { get; set; }
 
         public DeploymentOptionProperty<object> IgnoreSemicolonBetweenStatements { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -197,13 +205,13 @@ public class DeploymentOptions
 
         public DeploymentOptionProperty<object> IgnoreTablePartitionOptions { get; set; } // DW Specific
 
-        public DeploymentOptionProperty<object> AdditionalDeploymentContributorPaths { get; set; } = new DeploymentOptionProperty<object>();
+        public DeploymentOptionProperty<object> AdditionalDeploymentContributorPaths { get; set; } = new DeploymentOptionProperty<object>(null);
```
Charles-Gagnon, line 208: I don't know what an appropriate values for these would be. Do you even want to initialize them here vs just leaving the property null to begin with?

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -33,7 +41,7 @@ public class DeploymentOptions
     {
         #region Properties
 
-        public DeploymentOptionProperty<object> IgnoreTableOptions { get; set; }
+        public DeploymentOptionProperty<bool> IgnoreTableOptions { get; set; }
```
Charles-Gagnon, line 44: After you merge in you should change the rest of the properties to their actual types.

Charles-Gagnon:
Ping @ssreerama - this is meant to be merged into your branch. Let me know if you have any questions otherwise you can handle doing the merge yourself when you're ready.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: In one of my previous iterations these properties are changed to bool/string type, In runtime the property Intialization with values required type casting which ends with 'if else if' statements. I'll update them now as my final iteration(hopefully :)). thanks.

Charles-Gagnon: I'm a bit confused here - what exactly are you saying is the issue? Were you able to find a way to have strong typing without the `Activator.CreateInstance` solution I have in this PR?

We shouldn't keep these types as `object` - that kind of defeats the purpose of doing generics in the first place.

PR #1373, opened 2022-01-27
Adding recommendations to query plan 

aasimkhan30:
Adding index recommendations to show plan xml
image
The formatted query text when the user clicks on recommendations.
image

SSMS for comparison: (Note: SSMS does not support displaying multiple recommendations)
image
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -144,11 +144,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
     public class ExecutionPlanGraphFile 
     {
         /// <summary>
-        /// file contents
+        /// File contents
         /// </summary>
         public string GraphFileContent { get; set; }
         /// <summary>
-        /// file type for execution plan. This will be the file type of the editor when the user opens the graph file
+        /// File type for execution plan. This will be the file type of the editor when the user opens the graph file
         /// </summary>
         public string GraphFileType { get; set; }
     }
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -23,7 +23,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// <summary>
         /// Underlying xml string used for generating execution plan graph
         /// </summary>
-        public string XmlString { get; set; }
+        public string RawGraph { get; set; }
+        /// <summary>
+        /// Index recommendations given by show plan to improve query performance
+        /// </summary>
+        public List<ExecutionPlanRecommendation> Recommendations;
     }
 
     public class ExecutionPlanNode
@@ -113,12 +117,27 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// <summary>
         /// Size of the rows returned by the subtree of the edge.
         /// </summary>
-        /// <value></value>
         public double RowSize { get; set; }
         /// <summary>
         /// Edge properties to be shown in the tooltip.
         /// </summary>
-        /// <value></value>
         public List<ExecutionPlanGraphPropertyBase> Properties { get; set; }
+    }
+
+
+    public class ExecutionPlanRecommendation
+    {
+        /// <summary>
+        /// Text displayed in the show plan graph control
+        /// </summary>
+        public string DisplayString { get; set; }
+        /// <summary>
+        /// Raw query that is recommended to the user
+        /// </summary>
+        public string QueryText { get; set; }
+        /// <summary>
+        /// Query that will be opened in a new file once the user click on the recommendation
+        /// </summary>
+        public string FormattedQueryText { get; set; }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -134,11 +134,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// <summary>
         /// Raw query that is recommended to the user
         /// </summary>
-        public string QueryText { get; set; }
+        public string Query { get; set; }
         /// <summary>
         /// Query that will be opened in a new file once the user click on the recommendation
         /// </summary>
-        public string FormattedQueryText { get; set; }
+        public string QueryWithDescription { get; set; }
     }
 
     public class RawExecutionPlanGraph 
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -23,11 +23,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// <summary>
         /// Underlying xml string used for generating execution plan graph
         /// </summary>
-        public string RawGraph { get; set; }
+        public RawExecutionPlanGraph RawGraph { get; set; }
         /// <summary>
         /// Index recommendations given by show plan to improve query performance
         /// </summary>
-        public List<ExecutionPlanRecommendation> Recommendations;
+        public List<ExecutionPlanRecommendation> Recommendations { get; set; }
     }
 
     public class ExecutionPlanNode
@@ -139,5 +139,17 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// Query that will be opened in a new file once the user click on the recommendation
         /// </summary>
         public string FormattedQueryText { get; set; }
+    }
+
+    public class RawExecutionPlanGraph 
+    {
+        /// <summary>
+        /// Raw execution plan graph from which the query plan graph is generated
+        /// </summary>
+        public string RawGraphString { get; set; }
+        /// <summary>
+        /// file type for raw execution plan. This will be the file type of the editor when the user opens the raw graph
+        /// </summary>
+        public string RawGraphFileType { get; set; }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -21,9 +21,9 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// </summary>
         public string Query { get; set; }
         /// <summary>
-        /// Underlying xml string used for generating execution plan graph
+        /// Graph file that used to generate ExecutionPlanGraph
         /// </summary>
-        public RawExecutionPlanGraph RawGraph { get; set; }
+        public ExecutionPlanGraphFile GraphFile { get; set; }
         /// <summary>
         /// Index recommendations given by show plan to improve query performance
         /// </summary>
@@ -141,15 +141,15 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         public string QueryWithDescription { get; set; }
     }
 
-    public class RawExecutionPlanGraph 
+    public class ExecutionPlanGraphFile 
     {
         /// <summary>
-        /// Raw execution plan graph from which the query plan graph is generated
+        /// file contents
         /// </summary>
-        public string RawGraphString { get; set; }
+        public string GraphFileContent { get; set; }
         /// <summary>
-        /// file type for raw execution plan. This will be the file type of the editor when the user opens the raw graph
+        /// file type for execution plan. This will be the file type of the editor when the user opens the graph file
         /// </summary>
-        public string RawGraphFileType { get; set; }
+        public string GraphFileType { get; set; }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Description.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Description.cs
@@ -1,9 +1,9 @@
-using System;
 //
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
 using System.Collections.Generic;
 
 namespace Microsoft.SqlTools.ServiceLayer.ShowPlan.ShowPlanGraph
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Description.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Description.cs
@@ -3,6 +3,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System.Collections.Generic;
+
 namespace Microsoft.SqlTools.ServiceLayer.ShowPlan.ShowPlanGraph
 {
     public class Description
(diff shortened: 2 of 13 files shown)


src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -23,7 +23,11 @@ public class ExecutionPlanGraph
         /// <summary>
         /// Underlying xml string used for generating execution plan graph
         /// </summary>
-        public string XmlString { get; set; }
+        public string RawGraph { get; set; }
+        /// <summary>
+        /// Index recommendations given by show plan to improve query performance
+        /// </summary>
+        public List<ExecutionPlanRecommendation> Recommendations;
```
lewis-sanchez, line 30: Is there a reason why we aren't using `{get; set;}` for `Recommendations`? All of the other list properties in the other classes defined in this file have them.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,100 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
...
+            stmtXmlDocument.LoadXml(XmlDocument);
+            var nsmgr = new XmlNamespaceManager(stmtXmlDocument.NameTable);
+            nsmgr.AddNamespace("shp", "http://schemas.microsoft.com/sqlserver/2004/07/showplan");
+            XmlNode rootNode = stmtXmlDocument.DocumentElement;
+
+            XmlNode missingIndexes = rootNode.SelectSingleNode("descendant::shp:MissingIndexes", nsmgr);
+            //Returning null if no missing indexes are found in the xml file
+            if (missingIndexes == null)
+            {
+                return null;
```
Charles-Gagnon, line 394: Why are you returning null if there's no indices? `Description` seems like a much more general thing - so is there a reason not to send it even if there's no indices? (you already have `HasMissingIndex` anyways)

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,100 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
...
+            }
+            XmlNodeList indexGroups = missingIndexes.SelectNodes("descendant::shp:MissingIndexGroup", nsmgr);
+
+            bool memoryOptimzed = false;
+            XmlNode scan = rootNode.SelectSingleNode("descendant::shp:IndexScan", nsmgr);
```
Charles-Gagnon, line 399: Where is all this logic from? Some explanation of what is going on would be helpful...this is really hard to understand currently without expertise on how showplan is set up.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,100 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
...
+            }
+            if (scan != null && scan.Attributes["Storage"] != null)
+            {
+                if (0 == string.Compare(scan.Attributes["Storage"].Value, "MemoryOptimized", StringComparison.Ordinal))
+                {
+                    memoryOptimzed = true;
+                }
+            }
+
+            const string createIndexTemplate = "CREATE NONCLUSTERED INDEX [<Name of Missing Index, sysname,>]\r\nON {0}.{1} ({2})\r\n";
```
Charles-Gagnon, line 412: quote/escape schema/table

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraphUtils.cs
```diff
@@ -12,14 +13,15 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
 {
     public class ShowPlanGraphUtils
     {
-        public static List<ExecutionPlanGraph> CreateShowPlanGraph(string xml)
+        public static List<ExecutionPlanGraph> CreateShowPlanGraph(string xml, string fileName = "")
```
Charles-Gagnon, line 16: When would fileName ever not be specified?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraphUtils.cs
```diff
@@ -85,5 +87,32 @@ private static List<ExecutionPlanGraphPropertyBase> GetProperties(PropertyDescri
...
+                {
+                    DisplayString = mi.MissingIndexCaption,
+                    QueryText = mi.MissingIndexQueryText,
+                    FormattedQueryText = ParseMissingIndexQueryText(fileName, mi.MissingIndexImpact, mi.MissingIndexDatabase, mi.MissingIndexQueryText)
+                }).ToList();
+            }
+            return null;
+        }
+
+        private static string ParseMissingIndexQueryText(string doctitle, string impact, string database, string query)
```
Charles-Gagnon, line 108: XML doc for these. 

And maybe update the `doctitle` param - in all the previous methods you call it `fileName` so that seems reasonable to use as the name here too.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,100 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
...
+        private Description ParseDescription(ShowPlanGraph graph, string XmlDocument, int index)
+        {
+
+            Description description = new Description();
+            description.HasMissingIndex = true;
+            description.QueryText = graph.Statement;
+
+            XmlDocument stmtXmlDocument = new XmlDocument();
+            stmtXmlDocument.LoadXml(XmlDocument);
+            var nsmgr = new XmlNamespaceManager(stmtXmlDocument.NameTable);
```
lewis-sanchez, line 386: Nitpick: rename to nsMgr

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,100 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
...
+                {
+                    foreach (XmlNode column in columnGroup.ChildNodes)
+                    {
+                        string columnName = column.Attributes["Name"].Value;
+                        if (0 != string.Compare(columnGroup.Attributes["Usage"].Value, "INCLUDE", StringComparison.Ordinal))
+                        {
+                            if (indexColumns == string.Empty)
+                                indexColumns = columnName;
+                            else
+                                indexColumns = string.Format("{0},{1}", indexColumns, columnName);
```
lewis-sanchez, line 438: Can also use string interpolation: `indexColumns = $"{indexColumns},{columnName}"`, slightly more compact way of doing the same thing.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
aasimkhan30: I got this code from ssms. It seems that names are already escaped in the query plan. 
image

Charles-Gagnon: Oh huh, that's interesting. I would definitely recommend adding comments indicating that.

Could you check with a database/schema/table that have ] in their names to verify that things are being escaped correctly?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -23,7 +23,11 @@ public class ExecutionPlanGraph
         /// <summary>
         /// Underlying xml string used for generating execution plan graph
         /// </summary>
-        public string XmlString { get; set; }
+        public string RawGraph { get; set; }
```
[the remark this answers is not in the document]
aasimkhan30: I am trying to think of a generalized name as it is going to be used across different providers. XML string makes it very sql server specific.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
aasimkhan30: image
It seems those names are properly escaped. The original table name for the query was [master].[dbo].[insert\\]l]

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Description.cs
```diff
@@ -44,45 +47,69 @@ public bool IsClusteredMode
...
+                this.missingIndexQueryText = value;
+            }
         }
 
         public string MissingIndexImpact
         {
             get { return this.missingIndexImpact; }
+            set
+            {
+                this.missingIndexImpact = value;
```
Charles-Gagnon, line 97: For these you can just remove the implementations and use the auto setter-getter functionality when there isn't any additional logic you're doing

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Description.cs
```diff
@@ -1,8 +1,11 @@
+using System;
```
Charles-Gagnon, line 1: This should go below copyright

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,100 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
...
+            XmlDocument stmtXmlDocument = new XmlDocument();
+            stmtXmlDocument.LoadXml(graph.XmlDocument);
+            var nsMgr = new XmlNamespaceManager(stmtXmlDocument.NameTable);
+            nsMgr.AddNamespace("shp", "http://schemas.microsoft.com/sqlserver/2004/07/showplan");
+            XmlNode rootNode = stmtXmlDocument.DocumentElement;
+
+            XmlNode missingIndexes = rootNode.SelectSingleNode("descendant::shp:MissingIndexes", nsMgr);
+
+            List<MissingIndex> parsedIndexes = new List<MissingIndex>();
+            //Returning null if no missing indexes are found in the xml file
```
Charles-Gagnon, line 388: This doesn't seem true anymore - you still return the Description object

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraphUtils.cs
```diff
@@ -85,5 +90,40 @@ private static List<ExecutionPlanGraphPropertyBase> GetProperties(PropertyDescri
             }
             return propsList;
         }
+
+        private static List<ExecutionPlanRecommendation> ParseRecommendations(ShowPlanGraph.ShowPlanGraph g, string fileName)
+        {
+            if (g.Description != null && g.Description.MissingIndices != null)
```
Charles-Gagnon, line 96: `g.Description?.MissingIndices != null`

But also - can this ever be null currently? We're always setting a list, it just might be empty.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraphUtils.cs
```diff
@@ -85,5 +90,40 @@ private static List<ExecutionPlanGraphPropertyBase> GetProperties(PropertyDescri
...
+        /// Creates query file text for the recommendations. It has the missing index query along with some lines of description.
+        /// </summary>
+        /// <param name="fileName">query file name that has generated the plan</param>
+        /// <param name="impact">impact of the missing query on performance</param>
+        /// <param name="database">database name to create the missing index in</param>
+        /// <param name="query">actual query that will be used to create the missing index</param>
+        /// <returns></returns>
+        private static string ParseMissingIndexQueryText(string fileName, string impact, string database, string query)
+        {
+            return $@"{string.Format(SR.MissingIndexDetailsTitle, fileName, impact)}
```
Charles-Gagnon, line 118: This should also be parameterized like you did for MissingIndexFormat

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Microsoft.SqlTools.ServiceLayer.UnitTests.csproj
```diff
@@ -33,4 +33,8 @@
 	    <Content Remove=".\ShowPlan\TestExecution.xml" />
     	<EmbeddedResource Include=".\ShowPlan\TestExecutionPlan.xml" />
 	</ItemGroup>
+	<ItemGroup>
+	    <Content Remove=".\ShowPlan\TestExecutionPlanRecommendations.xml" />
+    	<EmbeddedResource Include=".\ShowPlan\TestExecutionPlanRecommendations.xml" />
```
Charles-Gagnon, line 38: Align this correctly

aasimkhan30: This seems to be aligned correctly on my end
image

I wonder why it looks weird over here

Charles-Gagnon: Turn on whitespace visualization - it looks like there's a mix of tabs and spaces throwing things off. 

image

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -23,7 +23,11 @@ public class ExecutionPlanGraph
         /// <summary>
         /// Underlying xml string used for generating execution plan graph
         /// </summary>
-        public string XmlString { get; set; }
+        public RawExecutionPlanGraph RawGraph { get; set; }
```
Charles-Gagnon, line 26: 1. The comment isn't correct anymore
2. Can this just be ExecutionPlanGraph? "Raw" isn't really accurate here since it's been parsed and whatnot already...

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -113,12 +117,39 @@ public class ExecutionPlanEdges
...
+        /// </summary>
+        public string QueryWithDescription { get; set; }
+    }
+
+    public class RawExecutionPlanGraph 
+    {
+        /// <summary>
+        /// Raw execution plan graph from which the query plan graph is generated
+        /// </summary>
+        public string RawGraphString { get; set; }
```
Charles-Gagnon, line 149: This is the XML right? Maybe just `GraphXml` or something

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -113,12 +117,39 @@ public class ExecutionPlanEdges
...
+        /// <summary>
+        /// file type for raw execution plan. This will be the file type of the editor when the user opens the raw graph
+        /// </summary>
+        public string RawGraphFileType { get; set; }
```
Charles-Gagnon, line 153: This can just be `GraphFileType` I would think

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,99 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
             }
         }
 
+        private Description ParseDescription(ShowPlanGraph graph, int index)
+        {
+            XmlDocument stmtXmlDocument = new XmlDocument();
```
Charles-Gagnon, line 379: I would at least indicate where you got this logic from - and it would be great if you could spend some time understanding what it's doing and comment it appropriately. It's a bit dense and hard to understand at a glance.

Charles-Gagnon (approved):
Didn't spend too much time going over the specific parsing logic since that's coming from SSMS - I would definitely suggest adding some more tests to cover the various cases (with/without memory optimized, with/without included columns, etc.) to ensure all code paths are handled correctly.

lewis-sanchez (approved):

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
aasimkhan30: I believe these names should be generic as GraphXml won't make sense for postgres, etc..

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
kburtram: I would remove `Raw` from all the names as it isn't really necessary, IMO.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -45,16 +44,20 @@ public ShowPlanGraph[] Execute(object dataSource)
...
+            int i = 0;
             foreach (BaseStmtInfoType statement in EnumStatements(plan))
             {
                 // Reset currentNodeId (used through Context) and create new context
                 this.currentNodeId = 0;
                 NodeBuilderContext context = new NodeBuilderContext(new ShowPlanGraph(), this.showPlanType, this);
                 // Parse the statement block
                 XmlPlanParser.Parse(statement, null, null, context);
+                // Get the statement XML for the graph.
+                context.Graph.XmlDocument = GetSingleStatementXml(dataSource, i++);
```
kburtram, line 58: incrementing a loop counter in the middle of a loop body is a little unusual code structure.  I would consider (1) putting the increment on its own line to make it more obvious and (2) consider giving it a more descriptive name (e.g. `statementIndex`) that makes it more intuitive why it's being incremented at that point in the code.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/XmlPlanNodeBuilder.cs
```diff
@@ -371,6 +374,99 @@ private IEnumerable<BaseStmtInfoType> EnumStatements(StmtBlockType statementBloc
...
+                                if (includeColumns == string.Empty)
+                                    includeColumns = columnName;
+                                else
+                                    includeColumns = $"{indexColumns},{columnName}";
+                            }
+                        }
+                    }
+
+                    string queryText = string.Format((memoryOptimzed) ? addIndexTemplate : createIndexTemplate, schemaName, tableName, indexColumns);
+                    if (!String.IsNullOrEmpty(includeColumns))
```
kburtram, line 444: nit: prefer `string` to `String` here.

kburtram (approved):
A few comments, but looks good overall to me

PR #1374, opened 2022-01-27
Use old dotnet embedded resource convention

### LINKED ISSUE #1372: Could not find the resource "Microsoft.SqlTools.ManagedBatchParser.Localization.SR.resources" among the resources "Microsoft.SqlTools.ManagedBatchParser.SR.resources" embedded in the assembly "Microsoft.SqlTools.ManagedBatchParser"
https://github.com/microsoft/azuredatastudio/issues/17967

https://github.com/microsoft/vscode-mssql/issues/17221

These two issues show that for some reason managedbatchparser is not finding the string resources under the right name. I searched for ManagedBatchParser.SR, but have yet to find any results so far. 

The embedded resource resx file appears to have been included in the project file, but it's still not showing up.

smartguest:
The changes to dotnet caused the embedded localization resources file to not be recognized properly. 

In order to fix this, for every part of sqltoolsservice that has embedded localization strings, I set the EmbeddedResourceUseDependentUponConvention flag to false to go back to the old behavior pre bump.

See here for the issue: https://docs.microsoft.com/en-us/dotnet/core/compatibility/msbuild

This fixes #1372

and its associated issues:

https://github.com/microsoft/azuredatastudio/issues/17967

https://github.com/microsoft/vscode-mssql/issues/17221

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Microsoft.SqlTools.Hosting.csproj
+++ b/src/Microsoft.SqlTools.Hosting/Microsoft.SqlTools.Hosting.csproj
@@ -4,6 +4,7 @@
 		<EnableDefaultItems>false</EnableDefaultItems>
 		<EnableDefaultCompileItems>false</EnableDefaultCompileItems>
 		<EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
+		<EmbeddedResourceUseDependentUponConvention>false</EmbeddedResourceUseDependentUponConvention>
 		<EnableDefaultNoneItems>false</EnableDefaultNoneItems>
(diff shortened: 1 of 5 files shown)


Charles-Gagnon:
I can't seem to repro this error myself - do you know why that is?

smartguest:
@Charles-Gagnon I'm not sure, but on a regular installation of ADS, if you try to run this query: 

SELECT * FROM sys.databases WHERE name = 'master

The missing resources error message shows up, instead of the correct malformed string error, I'll post a screenshot:

Charles-Gagnon (approved):
Ah there we go, I had tried one of the other ones and didn't seem to repro it but that does indeed break for me too.

PR #1377, opened 2022-01-31
Integration test for Get SKU Recommendation.

neetusingh-ms:
Adding Integration test for 'Get SKU Recommendation' call.
Adding dummy perf data files required for integration tests.


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
```diff
@@ -34,7 +38,34 @@ public async Task TestHandleMigrationAssessmentRequest()
...
+                DatabaseAllowList = new List<string> { "test", "test1" }
+            };
+
+            var requestContext = RequestContextMocks.Create<GetSkuRecommendationsResult>(r => result = r).AddErrorHandling(null);
+
+            MigrationService service = new MigrationService();
+            await service.HandleGetSkuRecommendationsRequest(requestParams, requestContext.Object);
+            Assert.IsNotNull(result, "Get SKU Recommendation result is null");
+            Assert.IsNotNull(result.SqlMiRecommendationResults, "Get MI SKU Recommendation result is null");
+            Assert.AreEqual(result.SqlMiRecommendationResults.First().PositiveJustifications.Count, 6, "Invalid number of positive justifications for MI SKU Recommendation result");
```
raymondtruong, line 68: I would say change this to:
1) use PositiveJustifications concatenated with NegativeJustifications
2) change it so that the combined length is maybe >0 or >1 or >2 or something, instead of equal to exactly 6 (since the number of justifications returned can change depending on the characteristics of the specific source instance, even if we don't change the recommendation logic)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
```diff
@@ -34,7 +38,34 @@ public async Task TestHandleMigrationAssessmentRequest()
...
+        }
+
+        [Test]
+        public async Task TestHandleMigrationGetSkuRecommendationsRequest()
+        {
+            GetSkuRecommendationsResult result = null;
+
+            var requestParams = new GetSkuRecommendationsParams()
+            {
+                DataFolder = Path.Combine("..", "..", "..", "Migration"),
```
raymondtruong, line 51: nit: maybe place the test data inside a dedicated folder within the 'Migration' folder called 'Data' or something? Just so we don't clutter the main folder with a mix of test classes and sample data

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
neetusingh-ms: My understanding is if we get a negative justification, we will exclude that SKU from final recommended list.

raymondtruong: That's a good point. In practice you're right, if the SKU recommendation result actually makes it back to the end user then it won't have any negative justifications. Although it's possible that in the future (in cases where we recommend multiple recommendations at a time, for example) we may start to start using it. So I figured I'd mention it now to include it just in case, but it shouldn't matter for now

raymondtruong (approved):

PR #1378, opened 2022-02-01
Expose SqlInstanceRequirements in SKU recommendation output

raymondtruong:
In this PR, we add a new field to `GetSkuRecommendationsResult`. Alongside the existing list of actual SKU recommendation results for SQL DB, MI, and VM, we expose a field of type `SqlInstanceRequirements` (coming from the SQL assessment NuGet) which contains essentially an aggregated view of the source instance's performance requirements. With this object, we can output the actual CPU/memory/IO/etc. requirement of the source instance, derived from the collected performance data. This object is essentially the input to the SKU recommendation engine. This way, when displaying recommendation results, we can also display the characteristics of the source instance alongside the actual recommended SKUs.

Also:
- updated tests
- fixed a typo in test class name

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
@@ -19,7 +19,7 @@ using NUnit.Framework;
 
 namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.Migration
 {
-    public class MigrationgentServiceTests
+    public class MigrationServiceTests
     {
         [Test]
         public async Task TestHandleMigrationAssessmentRequest()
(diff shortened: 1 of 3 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
```diff
@@ -19,7 +19,7 @@
 
 namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.Migration
 {
-    public class MigrationgentServiceTests
+    public class MigrationServiceTests
```
raymondtruong, line 22: @neetusingh-ms, I made sure that the `TestHandleMigrationGetSkuRecommendationsRequest` test works as intended, which it does, but when running the other one in this class `TestHandleMigrationAssessmentRequest`, I get this error:
image

Am I missing something?

neetusingh-ms: Did you do one time set up for running integration tests?

neetusingh-ms (approved):

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
raymondtruong: Ah yes, that would explain it :) I did not, thanks Neetu!

PR #1380, opened 2022-02-01
Optional application name parameter for connection string

VasuBhog:
When building the connection string for a server we should give an option to not include the application name. It is always set to true (within the vscode-mssql extension API) and only when set to false will we not set it for the connection string.

### DIFF
--- a/docs/guide/jsonrpc_protocol.md
+++ b/docs/guide/jsonrpc_protocol.md
@@ -533,12 +533,12 @@ Get a connection string for the provided connection.
         public string OwnerUri { get; set; }
 
         /// <summary>
-        /// Indicates whether the password should be return in the connection string
+        /// Indicates whether the password should be return in the connection string. Default is false.
         /// </summary>
         public bool IncludePassword { get; set; }
 
         /// <summary>
-        /// Indicates whether the application name should be return in the connection string. Default is false.
+        /// Indicates whether the application name should be return in the connection string. Default is true.
         /// </summary>
         public bool? IncludeApplicationName { get; set;}
     }
--- a/docs/guide/jsonrpc_protocol.md
+++ b/docs/guide/jsonrpc_protocol.md
@@ -538,7 +538,7 @@ Get a connection string for the provided connection.
         public bool IncludePassword { get; set; }
 
         /// <summary>
-        /// Indicates whether the application name should be return in the connection string
+        /// Indicates whether the application name should be return in the connection string. Default is false.
         /// </summary>
         public bool? IncludeApplicationName { get; set;}
     }
--- a/docs/guide/jsonrpc_protocol.md
+++ b/docs/guide/jsonrpc_protocol.md
@@ -536,6 +536,11 @@ Get a connection string for the provided connection.
         /// Indicates whether the password should be return in the connection string
         /// </summary>
         public bool IncludePassword { get; set; }
+
+        /// <summary>
+        /// Indicates whether the application name should be return in the connection string
+        /// </summary>
+        public bool? IncludeApplicationName { get; set;}
     }
 ```
 #### Response
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -1327,10 +1327,14 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                         {
                             connStringBuilder.Password = ConnectionService.PasswordPlaceholder;
                         }
-
-                        if (connStringParams.IncludeApplicationName)
+                        // default connection string application name to always be included unless set to false
+                        if (connStringParams.IncludeApplicationName != false)
                         {
-                           connStringBuilder.ApplicationName = "sqlops-connection-string";
+                            connStringParams.IncludeApplicationName = true;
+                        }
+                        if (connStringParams.IncludeApplicationName.HasTrue())
+                        {
+                            connStringBuilder.ApplicationName = "sqlops-connection-string";
                         }
                         connectionString = connStringBuilder.ConnectionString;
                     }
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -1328,8 +1328,10 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                             connStringBuilder.Password = ConnectionService.PasswordPlaceholder;
                         }
 
-                        connStringBuilder.ApplicationName = "sqlops-connection-string";
-
+                        if (connStringParams.IncludeApplicationName)
+                        {
+                           connStringBuilder.ApplicationName = "sqlops-connection-string";
+                        }
                         connectionString = connStringBuilder.ConnectionString;
                     }
                     catch (Exception e)
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -1328,11 +1328,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                             connStringBuilder.Password = ConnectionService.PasswordPlaceholder;
                         }
                         // default connection string application name to always be included unless set to false
-                        if (connStringParams.IncludeApplicationName != false)
-                        {
-                            connStringParams.IncludeApplicationName = true;
-                        }
-                        if (connStringParams.IncludeApplicationName.HasTrue())
+                        if (!connStringParams.IncludeApplicationName.HasValue || connStringParams.IncludeApplicationName.Value == true)
                         {
                             connStringBuilder.ApplicationName = "sqlops-connection-string";
                         }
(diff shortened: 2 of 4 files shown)


chlafreniere:
@VasuBhog please search the codebase for `GetConnectionStringParams`. There are other places that we should be updating here, from tests to docs.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
```diff
@@ -19,5 +19,10 @@ public class GetConnectionStringParams
         /// Indicates whether the password should be return in the connection string
         /// </summary>
         public bool IncludePassword { get; set; }
+
+        /// <summary>
+        /// Indicates whether the application name should be return in the connection string
+        /// </summary>
+        public bool IncludeApplicationName{ get; set; }
```
chlafreniere, line 26: I'm not at all familiar with STS code, so just asking questions 😄.

Are all of these params optional, or is this now a required field? Do we need a `?`?

Also nit: spacing (`IncludeApplicationName {` instead of `IncludeApplicationName{`)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1328,8 +1328,10 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
                             connStringBuilder.Password = ConnectionService.PasswordPlaceholder;
                         }
 
-                        connStringBuilder.ApplicationName = "sqlops-connection-string";
-
+                        if (connStringParams.IncludeApplicationName)
```
Charles-Gagnon, line 1331: This should be defaulted to true in STS so we don't have to update every client that uses it

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
VasuBhog: This is defaulted to true in mssql-vscode. This won't update every client since it already defaulted and only would change when someone sets this parameter to false.

Charles-Gagnon: More than just VS Code uses STS 😄 

The default here should be that if it isn't specified the behavior stays the same as previously.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
VasuBhog: I followed the same pattern as above since I am not that familiar as well. All these parameters are optional, we make ownerURI required in the many different places we utilize getConnectionStringRequest as that is what we use to set build the connectionString in STS.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
VasuBhog: Makes sense to set this to true then.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
Charles-Gagnon: I imagine for this you're going to want to make it optional. I haven't looked too closely at the code that turns the JSON params into objects but my guess is that it just uses the default for the type (false in this case) if it isn't specified when deserializing into the object. Hopefully whatever does that would then leave that value as null instead if it wasn't specified.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1327,9 +1327,15 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
                         {
```
chlafreniere, line 1327: @VasuBhog please search the codebase for GetConnectionStringParams. There are other places that we should be updating here, from tests to docs.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1327,9 +1327,15 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
                         {
                             connStringBuilder.Password = ConnectionService.PasswordPlaceholder;
                         }
-
-                        connStringBuilder.ApplicationName = "sqlops-connection-string";
-
+                        // default connection string application name to always be included unless set to false
+                        if (connStringParams.IncludeApplicationName != false)
```
Charles-Gagnon, line 1331: Just simplify this to 

```
if (!connStringParams.IncludeApplicationName.HasValue() || connStringParams.IncludeApplicationName.Value == true)
{
    connStringBuilder.ApplicationName = "sqlops-connection-string";
}
```

(you could also make an extension method HasFalse like we've done for HasTrue. Could be useful elsewhere)

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
```diff
@@ -19,5 +19,10 @@ public class GetConnectionStringParams
         /// Indicates whether the password should be return in the connection string
         /// </summary>
         public bool IncludePassword { get; set; }
+
+        /// <summary>
+        /// Indicates whether the application name should be return in the connection string
```
Charles-Gagnon, line 24: This should indicate what the default is

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
```diff
@@ -124,7 +124,8 @@ public async Task GetCurrentConnectionStringTest()
             var requestParams = new GetConnectionStringParams()
             {
                 OwnerUri = result.ConnectionInfo.OwnerUri,
-                IncludePassword = false
+                IncludePassword = false,
+                IncludeApplicationName = true
```
Charles-Gagnon, line 128: Seems like a good opportunity to add a test for not having the application name as well 😄

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
```diff
@@ -139,5 +140,38 @@ public async Task GetCurrentConnectionStringTest()
...
+            {
+                OwnerUri = result.ConnectionInfo.OwnerUri,
+                IncludePassword = false,
+                IncludeApplicationName = false
+            };
+
+            await service.HandleGetConnectionStringRequest(requestParams, requestContext.Object);
+            requestContext.VerifyAll();
+
+            requestContext.Setup(x => x.SendResult(It.Is<string>((connectionString) => !connectionString.Contains(resultApplicationName))))
```
Charles-Gagnon, line 170: Why do you have this second check? Couldn't you just check the full `Application Name={resultApplicationName}` the first time?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
```diff
@@ -139,5 +140,38 @@ public async Task GetCurrentConnectionStringTest()
             await service.HandleGetConnectionStringRequest(requestParams, requestContext.Object);
             requestContext.VerifyAll();
         }
+
+        /// <summary>
+        /// Test HandleGetConnectionStringRequest
+        /// When Application Name is set to false, the connection string should not contain the application name
```
Charles-Gagnon, line 146: `When IncludeApplicationName is set to false...`

docs/guide/jsonrpc_protocol.md
```diff
@@ -536,6 +536,11 @@ Get a connection string for the provided connection.
         /// Indicates whether the password should be return in the connection string
         /// </summary>
         public bool IncludePassword { get; set; }
+
+        /// <summary>
+        /// Indicates whether the application name should be return in the connection string
```
Charles-Gagnon, line 541: This should also be updated to indicate the default value.

Charles-Gagnon (approved):

docs/guide/jsonrpc_protocol.md
```diff
@@ -536,6 +536,11 @@ Get a connection string for the provided connection.
         /// Indicates whether the password should be return in the connection string
         /// </summary>
         public bool IncludePassword { get; set; }
+
+        /// <summary>
+        /// Indicates whether the application name should be return in the connection string. Default is false.
```
Charles-Gagnon, line 541: `Default is true`? False isn't the correct default...

Charles-Gagnon (changes requested):

docs/guide/jsonrpc_protocol.md
VasuBhog: Yeah mean to put this for IncludePassword.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
```diff
@@ -139,5 +140,34 @@ public async Task GetCurrentConnectionStringTest()
...
+            {
+                OwnerUri = result.ConnectionInfo.OwnerUri,
+                IncludePassword = false,
+                IncludeApplicationName = false
+            };
+
+            await service.HandleGetConnectionStringRequest(requestParams, requestContext.Object);
+            requestContext.VerifyAll();
+
+            await service.HandleGetConnectionStringRequest(requestParams, requestContext.Object);
```
Charles-Gagnon, line 169: Why do you have these lines duplicated?

Charles-Gagnon (approved):
Thank you for adding the default to the other comments as well!

PR #1381, opened 2022-02-02
table designer - support second level entities

alanrenmsft:
1. vbump dacfx to bring in the new changes
2. add support for second level entities: columns mapping for indexes and foreign keys.
3. post processing for publish changes operation
4. add confirmation for column deletion
5. refactoring

@kisantia  please confirm the new dacfx works with dacpac and sqlproj extensions


caohai (approved):

kisantia (approved):
I did some basic testing of dacpac, schema compare, and sql projects extension and everything still seems to be working as expected

PR #1386, opened 2022-02-03
Move code coverage packages to devDependencies

Charles-Gagnon:
For component governance - dev dependencies won't trigger alerts.


Charles-Gagnon:
Merging this in to get alerting set up properly - let me know if there's any issues with this approach.

PR #1390, opened 2022-02-04
Disable intellisense if connection needs refresh

cssuh:
Disabling intellisense if a connection needs refresh, first step in fixing: https://github.com/microsoft/azuredatastudio/issues/17179

(next I will initiate a call to azurecore prompting the user to refresh connection)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1212,6 +1212,28 @@ private bool RequiresReparse(ScriptParseInfo info, ScriptFile scriptFile)
                 || !string.Equals(prevSqlText, currentSqlText);
         }
 
+        private bool RefreshNeeded(string uri)
+        {
+            ConnectionInfo connInfo;
+            System.Data.Common.DbConnection connection;
+            connectionService.TryFindConnection(uri, out connInfo);
```
alanrenmsft, line 1219: we do support intellisense for disconnected scenario, your change will break it, you should handle the scenario when the return value is false (disconnected scenario) and return false.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1212,6 +1212,28 @@ private bool RequiresReparse(ScriptParseInfo info, ScriptFile scriptFile)
                 || !string.Equals(prevSqlText, currentSqlText);
         }
 
+        private bool RefreshNeeded(string uri)
```
alanrenmsft, line 1215: please add comment

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1212,6 +1212,28 @@ private bool RequiresReparse(ScriptParseInfo info, ScriptFile scriptFile)
...
+            ConnectionInfo connInfo;
+            System.Data.Common.DbConnection connection;
+            connectionService.TryFindConnection(uri, out connInfo);
+            if (connInfo.ConnectionDetails.AuthenticationType != "AzureMFA") 
+            {
+                return false;
+            } 
+            else 
+            {
+                if (connInfo.TryGetConnection(ConnectionType.Default, out connection))
```
Charles-Gagnon, line 1226: How does this happen - shouldn't we have a connection already (after all the user had to connect the query in the first place)? Does it get cleared out if an error happens or something?

cssuh: https://github.com/microsoft/sqltoolsservice/blob/1b5ac4c017129b1f33dad10ce68b62f49d2d2800/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs#L763 This line seems to indicate that I can use TryFindConnection but I'll do some more digging just to make sure

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
alanrenmsft: this comment was resolved without code changes or follow up comments. reactivating it.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
Charles-Gagnon: Are you actually testing this with a connection that's timed out and needs to be refreshed?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
cssuh: I changed it to false

cssuh: I will update it because that doesn't seem right either

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
cssuh: @Charles-Gagnon It seems like this will still return a connection even if the token has expired, so I'll need another way to check this.

PR #1393, opened 2022-02-08
improve error handling, add RefreshPerfDataCollection 

raymondtruong:
Changes:
- improve error handling
- add new RefreshPerfDataCollection request, meant to be called periodically, returning the latest messages and errors back to the ADS frontend (since the perf data collector is essentially the console app without a console, the messages need to go somewhere).  As far as I know there's no easier way to send data from the STS back to the ADS frontend without making an explicit request, doing a periodic refresh to fetch the latest messages is the most straightforward approach. That way, any backend data collection errors will be surfaced back to the UI with a ~30 second delay at most.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
@@ -28,6 +28,7 @@ using Microsoft.SqlServer.Migration.SkuRecommendation.Billing;
 using Microsoft.SqlServer.Migration.SkuRecommendation.Contracts;
 using Microsoft.SqlServer.Migration.SkuRecommendation.Contracts.Models.Sku;
 using Microsoft.SqlServer.Migration.SkuRecommendation.Contracts.Models;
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -456,6 +571,38 @@ internal static List<AzureSqlSkuCategory> GetEligibleSkuCategories(string target
                                                     AzureSqlPaaSServiceTier.GeneralPurpose,
                                                     ComputeTier.Provisioned,
                                                     AzureSqlPaaSHardwareType.Gen5));
+                    //if (includePreviewSkus)
```
neetusingh-ms, line 574: why commented?

neetusingh-ms (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
raymondtruong: The version of the NuGet I was using up until this point was the old one without the new SKUs so none of this actually worked. Now that we've released it I'll update it here and uncomment all this out

PR #1394, opened 2022-02-08
Modify Get Connection String Request to use connection details

VasuBhog:
This will allow for both vscode-mssql and sql-database-projects to get the connection string via connection details without having to connect to the server. 

Issue currently: `Exception thrown: 'System.ArgumentNullException' in System.Private.CoreLib.dll` when checking OwnerURI == null.


Charles-Gagnon:
Not sure I understand the issue - you're saying when you try to use this you're getting that error? Where is that being thrown?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1317,13 +1317,22 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
                 string connectionString = string.Empty;
                 ConnectionInfo info;
-                if (TryFindConnection(connStringParams.OwnerUri, out info))
+                SqlConnectionStringBuilder connStringBuilder;
+                if (connStringParams.OwnerUri == null)
+                {
+                    // Use connection info to create connection string
+                    if (connStringParams.ConnectionDetails != null)
+                    {
+                        connStringBuilder = CreateConnectionStringBuilder(connStringParams.ConnectionDetails);
```
Charles-Gagnon, line 1326: You still want this to use the same logic as below for checking includePassword and includeApplicationName. That's all common to either way so just first get the connStringBuilder based on what type of connection info we're given and then run the common logic.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
```diff
@@ -15,11 +15,16 @@ public class GetConnectionStringParams
         /// </summary>
         public string OwnerUri { get; set; }
 
+        /// <summary>
+        /// Connection information of the connection
+        /// </summary>
+        public ConnectionDetails ConnectionDetails { get; set; }
```
Charles-Gagnon, line 21: Remember to update the docs too

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ConnectionServiceTests.cs
```diff
@@ -124,6 +124,7 @@ public async Task GetCurrentConnectionStringTest()
             var requestParams = new GetConnectionStringParams()
             {
                 OwnerUri = result.ConnectionInfo.OwnerUri,
+                ConnectionDetails = result.ConnectionInfo.ConnectionDetails,
```
Charles-Gagnon, line 127: Make sure you're adding/updating tests to test this new functionality

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
Charles-Gagnon, line 21: @alanrenmsft There isn't any common pattern we use for getting union types over to the C# land right? On the TS side we can just do `connectionContext: string | IConnectionInfo` where the user can pass in either the URI or the connection info object, but the only way I can think of here is to just have it be a generic object - but I don't think that'll work with the deserializing that needs to happen on this end. Am I missing anything you can think of?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1317,13 +1317,22 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
+                }
+                else if (TryFindConnection(connStringParams.OwnerUri, out info))
                 {
                     try
                     {
-                        SqlConnectionStringBuilder connStringBuilder = CreateConnectionStringBuilder(info.ConnectionDetails);
+                        connStringBuilder = CreateConnectionStringBuilder(info.ConnectionDetails);
 
-                        if (!connStringParams.IncludePassword)
+                        if (!connStringParams.IncludePassword.HasValue)
```
alanrenmsft, line 1335: HasValue doesn't necessarily means the value is true. you should check both.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
```diff
@@ -15,11 +15,16 @@ public class GetConnectionStringParams
...
+
         /// <summary>
         /// Indicates whether the password should be return in the connection string
         /// default is set to false
         /// </summary>
-        public bool IncludePassword { get; set; }
+        public bool? IncludePassword { get; set; }
```
alanrenmsft, line 27: I am wondering why we need to make this nullable?

Charles-Gagnon: Yeah, why did you make this change @VasuBhog? As far as I can tell the ADS side always specifies this when sending the message so I'm not sure where we'd expect this to ever be undefined.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
alanrenmsft: this is the place in Table Designer where we are not sure about the type of the property: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/TableChangeInfo.cs#L33

but I guess it is different because the possible data types are all primitive types. for your case, I think you will need a JsonConverter: https://stackoverflow.com/questions/20432166/how-to-deserialize-a-json-property-that-can-be-two-different-data-types-using-js

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
VasuBhog: Switched it back.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringParams.cs
Charles-Gagnon: Doing that would require doing some refactoring of the base message parsing stuff, don't think that's worth it at this time. Having separate fields isn't the prettiest but this isn't really user facing so I think it's fine to just go for the separate fields for now.

docs/guide/jsonrpc_protocol.md
```diff
@@ -532,7 +532,11 @@ Get a connection string for the provided connection.
         /// </summary>
         public string OwnerUri { get; set; }
 
+        /// Connection info of the connection
+        /// </summary>
+        public ConnectionDetails ConnectionInfo { get; set; }
```
Charles-Gagnon, line 537: This should be `ConnectionDetails`

Charles-Gagnon (approved):

PR #1399, opened 2022-02-09
[Feature] SKU recommendations in SQL migration extension

raymondtruong:
**For the corresponding frontend changes, see https://github.com/microsoft/azuredatastudio/pull/18252**

-----

This PR implements changes to enable the new SKU recommendation experience in the SQL migration extension.

High level changes include:
- updated Microsoft.SqlServer.Migration.Assessment NuGet version from 1.0.20210902.7 to 1.0.20220208.23
  - physically replaced nupkg file and updated package reference in Packages.props
  - updated assessment rulesets/metadata. Also removed a duplicate set of assessment rulesets/metadata that seemed to be incorrectly placed one folder level too deep
  - added SKU recommendation metadata that likewise will need to be updated with every new version of the assessment NuGet
- implemented new `migration/getskurecommendations` request
  - this calls into the SQL assessment NuGet and generates SKU recommendations for specified target platforms. This SKU recommendation uses the newest logic from the latest assessment NuGet, including support for new VM recommendation logic, as well as new optional preview SKUs for MI (premium series and premium series memory optimized) and VM (Eb series)
- implemented new `migration/startperfdatacollection` request
  - this involves implementing a new `DataCollectionController` for performance data collection, which is a stripped-down version of the console app shipped with DMA 
  - this controller periodically queries SQL DMVs for performance data, aggregates the data, and persists it locally on the user's machine in a specified folder as csv files. Since the data aggregation and persistence implementation is identical to DMA, this enables scenarios where users collecting data from one tool (say, DMA) can import and view the recommendation results in another tool (like ADS)
- implement new `migration/stopperfdatacollection` request
  - this stops any existing data collection process
- implement new `migration/refreshperfdatacollection` request
  - since this `DataCollectionController` "console app" has no console to actually write messages to, this call fetches the status and latest messages and errors generated by the data collection process and returns it to the UI. This will be called periodically on a timer so that the user will be notified of any errors during data collection

Testing:
- existing assessment functionality in ADS still works as intended
- new SKU recommendation functionality in ADS is being tested in private feature branch: https://github.com/microsoft/azuredatastudio/tree/MigrationSkuRecommendations
- added unit tests


neetusingh-ms (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/SqlDataQueryController.cs
```diff
@@ -0,0 +1,374 @@
...
+        /// <summary>
+        /// Log exceptions to file.
+        /// </summary>
+        /// <param name="ex">Exception to log</param>
+        private void Logging(Exception ex)
+        {
+            this.errors.Add(new KeyValuePair<string, DateTime>(ex.Message, DateTime.UtcNow));
+
+            var error = new UnhandledSqlExceptionErrorModel(ex, ErrorScope.General);
+            _logger.Log(error, ErrorLevel.Error, TelemetryScope.PerfCollection);
```
neetusingh-ms, line 214: logger is always null... should we remove 214,215 for now

raymondtruong:
> As part of updating the dependencies in Packages.props we require that any PRs opened also verify that they've done the following checks.
> 
> **Please respond to this comment verifying that you've done the appropriate validation (or explain why it's not necessary) before merging in the PR**
> 
> * [ ]  Built and tested the change locally to validate that the update doesn't cause any regressions and fixes the issues intended
> * [ ]  Tested changes on all major platforms (Windows/Mac/Linux)
> * [ ]  Check the source repo for any open issues with the release being updated to (if available)

Yes, changes were built and tested locally. Only APIs changed were sql-migration specific. Existing assessment functionality was not modified (aside from updating assessment metadata as part of the NuGet upgrade, which was already tested on the NuGet side) and still works as intended.

No, changes were tested on Windows but not on Mac/Linux. However, SKU recommendation is a v1 preview experience (assessment is also in preview to begin with) and non-Windows platforms are not explicitly supported. There shouldn't be anything that would cause assessment functionality to not work on Mac/Linux, especially as we had updated the NuGet in the past to remove Windows-specific functionality, but it has not been thoroughly tested enough yet.

Yes, no relevant currently open issues.

src/Microsoft.SqlTools.ServiceLayer/Migration/SqlDataQueryController.cs
```diff
@@ -0,0 +1,382 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace Microsoft.SqlTools.ServiceLayer.Migration
+{
+    /// <summary>
+    /// Controller to manage the collection, aggregation, and persistence of SQL performance and static data for SKU recommendation.
+    /// </summary>
+    public class SqlDataQueryController : IDisposable
```
raymondtruong, line 22: FYI, this implementation is pulled from DMA

brian-harris (approved):

PR #1401, opened 2022-02-09
Fixing faulty showInTooltip logic and adding prop display value for Query Plan 

aasimkhan30:
Previously, I was using the wrong flag for determining whether the property should be shown in tooltip or not. This PR corrects that logic and also adds a method for getting a display string for complex properties.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -87,7 +87,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// </summary>
         public int DisplayOrder { get; set; }
         /// <summary>
-        /// Flag to indicate if the property has a longer value so that it will be shown at the bottom of the tooltip
+        /// Flag to show property at the bottom of tooltip. Generally done for for properties with longer value.
         /// </summary>
         public bool PositionAtBottom { get; set; }
         /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
@@ -90,6 +90,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         /// Flag to indicate if the property has a longer value so that it will be shown at the bottom of the tooltip
         /// </summary>
         public bool IsLongString { get; set; }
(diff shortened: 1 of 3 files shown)


Charles-Gagnon:
It would be helpful to explain what you're fixing - it's not obvious what the "faulty" logic is

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraphUtils.cs
```diff
@@ -122,5 +124,26 @@ private static string ParseMissingIndexQueryText(string fileName, string impact,
 */
 ";
         }
+
+        private static string GetPropertyText(PropertyDescriptor property)
+        {
+            try
+            {
+                // Get the property value.
+                object propertyValue = property.GetValue(property);
```
Charles-Gagnon, line 133: Is this correct? Normally you pass in the object that has the property you want the value of. This is getting the value of the descriptor

aasimkhan30: Thanks for pointing it out. Fixed the issue.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -89,7 +89,12 @@ public class ExecutionPlanGraphPropertyBase
         /// <summary>
         /// Flag to indicate if the property has a longer value so that it will be shown at the bottom of the tooltip
```
alanrenmsft, line 90: comments need to be updated

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/PropertyValue.cs
```diff
@@ -49,6 +49,14 @@ public int DisplayOrder
             }
         }
 
+        public bool ShowInToolTip
```
alanrenmsft, line 52: isn't tooltip a word?

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraphUtils.cs
```diff
@@ -122,5 +124,26 @@ private static string ParseMissingIndexQueryText(string fileName, string impact,
 */
 ";
         }
+
+        private static string GetPropertyDisplayValue(PropertyValue property)
+        {
```
alanrenmsft, line 129: I feel like this method can be simplified in to just a couple lines.

alanrenmsft (approved):

Charles-Gagnon (approved):

PR #1413, opened 2022-02-19
Export headers in empty file

benrr101:
**Description**: A friend of mine (@engineerchange) recently mentioned an issue to me. The issue was that if you export an empty table as a CSV with header export set to true, it will generate an empty file. The behavior he expected was that a file would be generated with just the headers, but no data. After some investigation, I determined that 1) I wrote most of this code and 2) the issue is caused by headers being output only if `WriteRow` is called. In the case of an empty table, `WriteRow` is never called, thereby no headers are ever emitted. 

The proposed solution in this PR is to emit the headers as part of the constructor for the CSV SaveAs writer. Because this requires knowledge of the columns during construction, the reader/writer factories had to be refactored to get the columns when constructing the writer. Along with the following changes, it turns out there's little difference between the SaveAs reader/writer factories, so I extracted an abstract  These changes, along with unit tests are the main part of the PR.

Secondarily, I noticed that the service buffer writers and factory basically shares no functionality with the other reader/writer factories. So, I refactored the service buffer reader/writer to have it's own interfaces. This allows the code to better follow SOLID principles (by removing implementations that always throw).

There's a few smaller changes going on in this PR, too.

**Testing**:
* Existing unit tests pass
* New unit tests also pass


kburtram:
@lewis-sanchez could you please code review this PR when you have a chance?  Thanks!

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/SaveResults/ServiceIntegrationTests.cs
```diff
@@ -389,11 +419,15 @@ public async Task SaveResultsExcelNonExistentQuery()
...
         {
-            // Given: 
+            // Given:
             // ... A working query and workspace service
             WorkspaceService<SqlToolsSettings> ws = Common.GetPrimedWorkspaceService(Constants.StandardQuery);
-            ConcurrentDictionary<string, byte[]> storage;
-            QueryExecutionService qes = Common.GetPrimedExecutionService(Common.StandardTestDataSet, true, false, false, ws, out storage);
+            QueryExecutionService qes = Common.GetPrimedExecutionService(
+                Common.StandardTestDataSet,
+                true, false,
```
lewis-sanchez, line 427: Put `false` on it's own line for consistency.

benrr101:
Closing this for now in order to split it into a couple smaller PRs.

PR #1415, opened 2022-03-03
Defines Query Plan Comparison Request Handler

lewis-sanchez:
This PR features a new service dedicated to compare show plans along with the necessary request handlers and associated parameter and result data transfer objects.

Additional tests have also been added to confirm that compare logic was ported over correctly.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GraphComparisonRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GraphComparisonRequest.cs
@@ -5,12 +5,30 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan.Contracts
 {
     public class GetGraphComparisonParams
     {
+        /// <summary>
+        /// First query plan's XML file text for comparison.
+        /// </summary>
+        public string FirstQueryPlanXmlText { get; set; }
 
+        /// <summary>
+        /// Second query plan's XML file text for comparison.
+        /// </summary>
+        public string SecondQueryPlanXmlText { get; set; }
+
+        /// <summary>
+        /// Flag to indicate if the database name should be ignored
+        /// during comparisons.
+        /// </summary>
+        public bool IgnoreDatabaseName { get; set; }
     }
 
     public class GetGraphComparisonResult
     {
-
+        /// <summary>
+        /// Flag indicating the compared graphs are the same when
+        /// true and different when false.
+        /// </summary>
+        public bool IsEquivalent { get; set; }
     }
 
     public class GraphComparisonRequest
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GraphComparisonRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GraphComparisonRequest.cs
@@ -1,0 +1,22 @@
+﻿using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.ShowPlan.Contracts
+{
+    public class GetGraphComparisonParams
+    {
+
+    }
+
+    public class GetGraphComparisonResult
+    {
+
+    }
+
+    public class GraphComparisonRequest
+    {
+        public static readonly
+            RequestType<GetGraphComparisonParams, GetGraphComparisonResult> Type =
+                RequestType<GetGraphComparisonParams, GetGraphComparisonResult>.Create("showplan/compareshowplans");
+    }
+}
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Comparison/ShowPlanComparisonService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Comparison/ShowPlanComparisonService.cs
@@ -1,0 +1,61 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.Hosting.Protocol;
+using Microsoft.SqlTools.ServiceLayer.Hosting;
+using System;
+
+namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
+{
+    /// <summary>
+    /// Main class for Show Plan Comparison functionality.
+    /// </summary>
+    public sealed class ShowPlanComparisonService : IDisposable
+    {
+        private static readonly Lazy<ShowPlanComparisonService> instance = new(() => new ShowPlanComparisonService());
+
+        private bool disposed;
+
+        /// <summary>
+        /// Constructs a new ShowPlanComparison instance with default parameters.
+        /// </summary>
+        public ShowPlanComparisonService() { }
+
+        /// <summary>
+        /// Gets singleton instance object.
+        /// </summary>
+        public static ShowPlanComparisonService Instance => instance.Value;
+
+        /// <summary>
+        /// Service host object for sending/receiving requests/events.
+        /// Internal for testing purposes.
+        /// </summary>
+        internal IProtocolEndpoint ServiceHost
+        {
+            get;
+            set;
+        }
+        
+        /// <summary>
+        /// Initializes the ShowPlanComparison service.
+        /// </summary>
+        /// <param name="serviceHost"></param>
+        public void InitializeService(ServiceHost serviceHost)
+        {
+            ServiceHost = serviceHost;
+        }
+
+        /// <summary>
+        /// Disposes the ShowPlanComparison service.
+        /// </summary>
+        public void Dispose()
+        {
+            if (!disposed)
+            {
+                disposed = true;
+            }
+        }
+    }
+}
--- a/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Comparison/ShowPlanComparisonService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ShowPlanGraph/Comparison/ShowPlanComparisonService.cs
@@ -5,7 +5,11 @@
 
 using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
+using Microsoft.SqlTools.ServiceLayer.ShowPlan.Contracts;
+using Microsoft.SqlTools.ServiceLayer.ShowPlan.ShowPlanGraph;
+using Microsoft.SqlTools.ServiceLayer.ShowPlan.ShowPlanGraph.Comparison;
 using System;
+using System.Threading.Tasks;
 
 namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
 {
@@ -45,6 +49,38 @@ namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
         public void InitializeService(ServiceHost serviceHost)
         {
             ServiceHost = serviceHost;
+            ServiceHost.SetRequestHandler(GraphComparisonRequest.Type, HandleGraphComparisonRequest);
+        }
+
+        internal async Task HandleGraphComparisonRequest(
+            GetGraphComparisonParams parameter,
+            RequestContext<GetGraphComparisonResult> requestContext)
+        {
+            try
+            {
+                var firstGraphSet = ShowPlanGraph.ShowPlanGraph.ParseShowPlanXML(parameter.FirstQueryPlanXmlText, ShowPlanType.Unknown);
+                var firstRootNode = firstGraphSet?[0]?.Root;
+
+                var secondGraphSet = ShowPlanGraph.ShowPlanGraph.ParseShowPlanXML(parameter.SecondQueryPlanXmlText, ShowPlanType.Unknown);
+                var secondRootNode = secondGraphSet?[0]?.Root;
+
+                var skeletonManager = new SkeletonManager();
+                var firstSkeletonNode = skeletonManager.CreateSkeleton(firstRootNode);
+                var secondSkeletonNode = skeletonManager.CreateSkeleton(secondRootNode);
+
+                var isEquivalent = skeletonManager.AreSkeletonsEquivalent(firstSkeletonNode, secondSkeletonNode, parameter.IgnoreDatabaseName);
+
+                var result = new GetGraphComparisonResult()
+                {
+                    IsEquivalent = isEquivalent
+                };
+
+                await requestContext.SendResult(result);
+            }
+            catch (Exception e)
+            {
+                await requestContext.SendError(e.ToString());
+            }
         }
 
         /// <summary>
--- a/test/Microsoft.SqlTools.ServiceLayer.UnitTests/ShowPlan/ShowPlanTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.UnitTests/ShowPlan/ShowPlanTests.cs
@@ -38,13 +38,45 @@ namespace Microsoft.SqlTools.ServiceLayer.UnitTests.ShowPlan
             var skeletonManager = new SkeletonManager();
             var skeletonNode = skeletonManager.CreateSkeleton(rootNode);
             var skeletonNode2 = skeletonManager.CreateSkeleton(rootNode);
+
             var skeletonCompareResult = skeletonManager.AreSkeletonsEquivalent(skeletonNode, skeletonNode2, ignoreDatabaseName: true);
 
             Assert.AreEqual(true, skeletonCompareResult);
         }
 
         [Test]
-        public void CompareShowPlan_DifferentSkeletons()
+        public void CompareShowPlan_TwoNullSkeletons()
+        {
+            Node firstRoot = null;
+            Node secondRoot = null;
+
+            var skeletonManager = new SkeletonManager();
+            var firstSkeletonNode = skeletonManager.CreateSkeleton(firstRoot);
+            var secondSkeletonNode = skeletonManager.CreateSkeleton(secondRoot);
+
+            var skeletonCompareResult = skeletonManager.AreSkeletonsEquivalent(firstSkeletonNode, secondSkeletonNode, ignoreDatabaseName: true);
+
+            Assert.AreEqual(true, skeletonCompareResult);
+        }
+
+        [Test]
+        public void CompareShowPlan_ComparingSkeletonAgainstNull()
+        {
+            ReadFile(".ShowPlan.TestExecutionPlan.xml");
+            ShowPlanGraph[] graphs = ShowPlanGraph.ParseShowPlanXML(queryPlanFileText, ShowPlanType.Unknown);
+            var rootNode = graphs[0].Root;
+
+            var skeletonManager = new SkeletonManager();
+            var skeletonNode = skeletonManager.CreateSkeleton(rootNode);
+            SkeletonNode skeletonNode2 = null;
+
+            var skeletonCompareResult = skeletonManager.AreSkeletonsEquivalent(skeletonNode, skeletonNode2, ignoreDatabaseName: true);
+
+            Assert.AreEqual(false, skeletonCompareResult);
+        }
+
+        [Test]
+        public void CompareShowPlan_DifferentChildCount()
         {
             ReadFile(".ShowPlan.TestExecutionPlan.xml");
             ShowPlanGraph[] graphs = ShowPlanGraph.ParseShowPlanXML(queryPlanFileText, ShowPlanType.Unknown);
@@ -54,6 +86,7 @@ namespace Microsoft.SqlTools.ServiceLayer.UnitTests.ShowPlan
             var skeletonNode = skeletonManager.CreateSkeleton(rootNode);
             var skeletonNode2 = skeletonManager.CreateSkeleton(rootNode);
             skeletonNode2.Children.RemoveAt(skeletonNode2.Children.Count - 1);
+
             var skeletonCompareResult = skeletonManager.AreSkeletonsEquivalent(skeletonNode, skeletonNode2, ignoreDatabaseName: true);
 
             Assert.AreEqual(false, skeletonCompareResult);
--- a/test/Microsoft.SqlTools.ServiceLayer.UnitTests/ShowPlan/ShowPlanTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.UnitTests/ShowPlan/ShowPlanTests.cs
@@ -29,7 +29,7 @@ namespace Microsoft.SqlTools.ServiceLayer.UnitTests.ShowPlan
         }
 
         [Test]
-        public void CompareDuplicateShowPlanSkeletons()
+        public void CompareShowPlan_DuplicateSkeletons()
         {
             ReadFile(".ShowPlan.TestExecutionPlan.xml");
             ShowPlanGraph[] graphs = ShowPlanGraph.ParseShowPlanXML(queryPlanFileText, ShowPlanType.Unknown);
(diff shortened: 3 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ColorMatchingSectionsRequest.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// Created Skeleton Node for a show plan
+        /// </summary>
+        public SkeletonNodeDTO SecondSkeletonNode { get; set; }
+    }
+
+    public class ColorMatchingSectionsRequest
+    {
+        public static readonly
+            RequestType<ColorMatchingSectionsParams, ColorMatchingSectionsResult> Type =
```
Charles-Gagnon, line 45: It's slightly easier to read if you camelCase the message names (e.g. `queryExecutionPlan/colorMatchingSections`)

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/EdgeDTO.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public double EstimatedDataSize { get; set; }
+        public double EstimatedRowCount { get; set; }
+        public List<ExecutionPlanGraphPropertyBase> Properties { get; set; }
+        public double RowCount { get; set; }
+        public double RowSize { get; set; }
+        public NodeDTO FromNode { get; set; }
+        public NodeDTO ToNode { get; set; }
+
+        public EdgeDTO()
```
Charles-Gagnon, line 20: You shouldn't need to specifically define this - parameterless constructors are generated automatically

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/DescriptionDTO.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public class DescriptionDTO
+    {
+        public string Title { get; set; }
+        public string QueryText { get; set; }
+        public string ClusteredMode { get; set; }
+        public bool IsClusteredMode { get; set; }
+        public List<MissingIndex> MissingIndices { get; set; }
+
+        public DescriptionDTO()
```
Charles-Gagnon, line 19: Do you need this?

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/GraphComparisonRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    public class GetGraphComparisonParams
+    {
+        /// <summary>
+        /// First query plan's XML file text for comparison.
+        /// </summary>
+        public string FirstQueryPlanXmlText { get; set; }
```
Charles-Gagnon, line 15: Are all execution plans going to be in XML format? If we're supporting generic plans then the contract should be made more generic too - @aasimkhan30 you did that for your stuff already right?

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/GraphComparisonRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// true and different when false.
+        /// </summary>
+        public bool IsEquivalent { get; set; }
+    }
+
+    public class GraphComparisonRequest
+    {
+        public static readonly
+            RequestType<GetGraphComparisonParams, GetGraphComparisonResult> Type =
+                RequestType<GetGraphComparisonParams, GetGraphComparisonResult>.Create("queryexecutionplan/compareshowplans");
```
Charles-Gagnon, line 42: `compareExecutionPlans`

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/GraphDTO.cs
```diff
@@ -0,0 +1,16 @@
...
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    public class GraphDTO
+    {
+        public NodeDTO Root { get; set; }
+        public DescriptionDTO Description { get; set; }
+
+        public GraphDTO()
```
Charles-Gagnon, line 13: Not needed

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/NodeDTO.cs
```diff
@@ -0,0 +1,36 @@
...
+        public string LogicalOpUnlocName { get; set; }
+        public OperationDTO Operation { get; set; }
+        public NodeDTO Parent { get; set; }
+        public string PhysicalOpUnlocName { get; set; }
+        public List<ExecutionPlanGraphPropertyBase> Properties { get; set; }
+        public double RelativeCost { get; set; }
+        public NodeDTO Root { get; set; }
+        public double SubtreeCost { get; set; }
+
+        public NodeDTO()
```
Charles-Gagnon, line 33: Not needed

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/SkeletonNodeDTO.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class SkeletonNodeDTO
+    {
+        public NodeDTO BaseNode { get; set; }
+        public List<SkeletonNodeDTO> Children { get; set; } = new List<SkeletonNodeDTO>();
+        public int GroupIndex { get; set; }
+        public bool HasMatch { get; set; }
+        public List<SkeletonNodeDTO> MatchingNodes { get; set; } = new List<SkeletonNodeDTO>();
+        public SkeletonNodeDTO ParentNode { get; set; }
+        
+        public SkeletonNodeDTO()
```
Charles-Gagnon, line 19: Not needed

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/ActualPlanDataReaderNodeBuilder.cs
```diff
@@ -3,7 +3,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
-namespace Microsoft.SqlTools.ServiceLayer.ShowPlan.ShowPlanGraph
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.ExecPlanGraph
```
Charles-Gagnon, line 6: We should avoid shortening names unless necessary - `ExecutionPlanGraph` would be clearer here

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/Comparison/SkeletonManager.cs
```diff
@@ -22,8 +22,11 @@ public class SkeletonManager
         /// </summary>
         /// <param name="root">Node to construct skeleton of</param>
         /// <returns>SkeletonNode with children representing logical descendants of the input node</returns>
-       public SkeletonNode CreateSkeleton(Node root)
+        public SkeletonNode CreateSkeleton(Node root)
```
Charles-Gagnon, line 25: Remove extra space

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/Comparison/SkeletonManager.cs
```diff
@@ -22,8 +22,11 @@ public class SkeletonManager
...
         {
+            if (root == null)
```
Charles-Gagnon, line 27: If's should always have parens for the body of the if.

But also - is root being null something you actually expect? Seems like letting it throw an error in that case is preferable (although maybe throw your own exception so it's clear what the issue is)

Charles-Gagnon: (same comment for all the other similar checks)

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/Comparison/SkeletonNode.cs
```diff
@@ -80,5 +82,51 @@ public Graph GetGraph()
...
 
+        public SkeletonNodeDTO ConvertToDTO()
+        {
+            var skeletonNodeDTO = new SkeletonNodeDTO();
+            LevelOrderCopy(this, skeletonNodeDTO);
+
+            return skeletonNodeDTO;
+        }
+
+        private static void LevelOrderCopy(SkeletonNode skeletonNode, SkeletonNodeDTO skeletonNodeDTO)
```
Charles-Gagnon, line 93: The name isn't really clear what it's doing. It seems to just be converting a SkeletonNode into a DTO? I'd just put this logic into the ConvertToDTO method - it doesn't really seem useful to have it split out like this.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/DataReaderNodeBuilder.cs
```diff
@@ -37,7 +37,7 @@ public DataReaderNodeBuilder()
         /// </summary>
         /// <param name="dataSource">Data Source.</param>
         /// <returns>An array of AnalysisServices Graph objects.</returns>
-        public ShowPlanGraph[] Execute(object dataSource)
+        public ExecutionPlanGraph[] Execute(object dataSource)
```
Charles-Gagnon, line 40: In the future I'd suggest that you do cleanup stuff like this in separate PRs - this is adding a ton of cruft to the PR and makes it harder to properly review.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/Node.cs
```diff
@@ -736,5 +738,83 @@ public void AddChild(Node child)
...
+        {
+            var nodeDTO = new NodeDTO();
+            LevelOrderCopy(this, nodeDTO);
+
+            return nodeDTO;
+        }
+
+        private static void LevelOrderCopy(Node node, NodeDTO rootDTO)
+        {
+            if (node == null)
```
Charles-Gagnon, line 751: Same here about just putting this in the ConvertToDTO method

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanService.cs
```diff
@@ -0,0 +1,219 @@
...
+using Microsoft.SqlTools.Hosting.Protocol;
+using Microsoft.SqlTools.ServiceLayer.Hosting;
+using Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts;
+using Microsoft.SqlTools.ServiceLayer.ExecutionPlan.ExecPlanGraph;
+using Microsoft.SqlTools.ServiceLayer.ExecutionPlan.ExecPlanGraph.Comparison;
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan
+{
+    /// <summary>
+    /// Main class for Migration Service functionality
```
Charles-Gagnon, line 17: Execution Plan Service

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanService.cs
```diff
@@ -0,0 +1,219 @@
...
+    /// </summary>
+    public sealed class ExecutionPlanService : IDisposable
+    {
+        private static readonly Lazy<ExecutionPlanService> instance = new Lazy<ExecutionPlanService>(() => new ExecutionPlanService());
+
+        private bool disposed;
+
+        /// <summary>
+        /// Construct a new MigrationService instance with default parameters
```
Charles-Gagnon, line 26: This should be private if you're using the singleton pattern (and fix the doc comment)

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/GraphComparisonRequest.cs
aasimkhan30: Yeah, @lewis-sanchez  there is a class ExecutionPlanGraphInfo that you can use.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/GraphDTO.cs
```diff
@@ -0,0 +1,16 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    public class GraphDTO
+    {
+        public NodeDTO Root { get; set; }
```
aasimkhan30, line 10: Please add comments to all the contract classes and properties.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/NodeDTO.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    public class NodeDTO
+    {
+        public List<NodeDTO> Children { get; set; } = new List<NodeDTO>();
```
aasimkhan30, line 12: Please add comments

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/OperationDTO.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.ServiceLayer.ExecutionPlan.ExecPlanGraph;
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    public class OperationDTO
+    {
+        public string Name { get; set; }
```
aasimkhan30, line 12: Comments

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/SkeletonNodeDTO.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    public class SkeletonNodeDTO
+    {
+        public NodeDTO BaseNode { get; set; }
```
aasimkhan30, line 12: Comments.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/executionplangraph.cs
```diff
@@ -6,12 +6,12 @@
...
 using System.Collections.Generic;
 
-namespace Microsoft.SqlTools.ServiceLayer.ShowPlan.ShowPlanGraph
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.ExecPlanGraph
 {
     /// <summary>
     /// Extension of graph with some handy included methods specific for ShowPlan use
     /// </summary>
-    public class ShowPlanGraph : Graph
+    public class ExecutionPlanGraph : Graph
```
aasimkhan30, line 14: Please avoid changing the names of the internal classes ported from SSMS. It will make future code porting and cross referencing a little difficult.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
```diff
@@ -59,7 +61,21 @@ private static ExecutionPlanEdges ConvertShowPlanEdgeToExecutionPlanEdge(Edge ed
             };
         }
 
-        private static List<ExecutionPlanGraphPropertyBase> GetProperties(PropertyDescriptorCollection props)
+        public static EdgeDTO ConvertShowPlanEdgeToDTO(Edge edge)
+        {
+            var edgeDTO = new EdgeDTO()
```
aasimkhan30, line 66: Why not use the existing ExecutionPlanEdges and ExecutionPlanNodes?

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/Comparison/SkeletonManager.cs
lewis-sanchez: There was a space missing, so when I formatted the document this extra space was added. It's hard to tell in the diff, but the "p" in public highlighted in red appears one character position too early 
image

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/Comparison/SkeletonManager.cs
lewis-sanchez: Null isn't something I would expect. I added that because I was writing a unit for this section of the ported `SkeletonManager`'s `AreSkeletonsEquivalent` method:
image

The setup for the aforementioned unit test looks like this:
image

As seen in the unit test, it starts by setting two `Node` variables to null and then uses those `Node` variables to create `SkeletonNodes`. The null check was then added to `CreateSkeleton` to prevent a null exception from being thrown when the method accessed a property on an object that is null.

Charles-Gagnon: Generally tests shouldn't be sending in invalid values. You should instead either create an actual Node object and pass it in or create a mock and use that.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecPlanGraph/ActualPlanDataReaderNodeBuilder.cs
lewis-sanchez: Using `ExectionPlanGraph` results in a build error because there's a class named `ExecutionPlanGraph` under the `/ExecutionPlan/Contracts/` directory that conflicts with it. Would renaming to something like `ExecutionGraph`, `PlanGraph`, or simply `Graph` be better since `ExecutionPlan` is already a part of the namespace?

lewis-sanchez: The namespace would then be this: `Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Graph`

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/Description.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System.Collections.Generic;
+using Microsoft.SqlTools.ServiceLayer.ExecutionPlan.ExecutionGraph;
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    /// <summary>
+    /// Holds all information that describes =the execution plan's graph.
+    /// </summary>
+    public class Description
```
aasimkhan30, line 14: This information already exists in executionPlanGraph class.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/SkeletonNode.cs
```diff
@@ -0,0 +1,40 @@
...
+//
+
+using System.Collections.Generic;
+
+namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
+{
+    /// <summary>
+    /// A simpler version of a node because of the reduced amount of properties.
+    /// </summary>
+    public class SkeletonNode
```
aasimkhan30, line 13: Can we rename this to ExecutionGraphComparisonResult or ExecutionGraphMatchingResult? 
The name SkeletonNode does not seem intuitive.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ColorMatchingSectionsRequest.cs
```diff
@@ -0,0 +1,48 @@
...
+        public ExecutionPlanGraphInfo SecondExecutionPlanGraphInfo { get; set; }
+
+        /// <summary>
+        /// Flag to indicate if the database name should be ignored
+        /// during comparisons.
+        /// </summary>
+        public bool IgnoreDatabaseName { get; set; }
+    }
+
+    public class ColorMatchingSectionsResult
```
aasimkhan30, line 29: Lets rename this class to GraphComparisonResult instead of ColorMatchingSectionResult

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ColorMatchingSectionsRequest.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// Created Skeleton Node for a show plan
+        /// </summary>
+        public SkeletonNode SecondSkeletonNode { get; set; }
+    }
+
+    public class ColorMatchingSectionsRequest
+    {
+        public static readonly
+            RequestType<ColorMatchingSectionsParams, ColorMatchingSectionsResult> Type =
+                RequestType<ColorMatchingSectionsParams, ColorMatchingSectionsResult>.Create("queryExecutionPlan/colorMatchingSections");
```
aasimkhan30, line 46: Let's rename the end point to executionPlan/compareExecutionPlanGraph

aasimkhan30 (approved):
Some renaming is required to make the classes more intuitive. Please try to use the existing contract objects to pass the information back. For example, the information in the description is already present in the ExecutionPlanGraph class.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ColorMatchingSectionsRequest.cs
Charles-Gagnon: I assume that means you'll want to update the other names in this file as well (including the file itself)

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/Description.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Indicates if clustered mode is enabled on the server
+        /// </summary>
+        public bool IsClusteredMode { get; set; }
+        /// <summary>
+        /// List of missing indices for the graph
+        /// </summary>
+        public List<MissingIndex> MissingIndices { get; set; }
+
+        /// <summary>
+        /// Instantiates a DescriptionDTO
```
Charles-Gagnon, line 38: You may not need this given Aasim's comment about, but fix these doc comments if you're keeping it

Charles-Gagnon (commented):
Could you take some time to go do the ShowPlan -> ExecutionPlan renames in another PR? That should be super quick to get in and would greatly reduce the clutter in this PR so it's easier to review properly. (and then just merge in main once that gets in to get rid of the changes here)

PR #1418, opened 2022-03-08
Change preview report type

abist:
ADS PR - https://github.com/microsoft/azuredatastudio/pull/18667

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
@@ -10,11 +10,23 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
     /// <summary>
     /// The service request to generate preview report describing the changes.
     /// </summary>
+    public class GeneratePreviewReportResult 
+    {
+        // The script generated in plaintext or markdown
+        public string Script;
+
+        // format of the string (plaintext or mardown)
+        public string Format;
+    }
+    
+    /// <summary>
+    /// The service request to generate preview report describing the changes.
+    /// </summary>
     public class GeneratePreviewReportRequest
     {
         /// <summary>
         /// Request definition
         /// </summary>
-        public static readonly RequestType<TableInfo, string> Type = RequestType<TableInfo, string>.Create("tabledesigner/generatepreviewreport");
+        public static readonly RequestType<TableInfo, GeneratePreviewReportResult> Type = RequestType<TableInfo, GeneratePreviewReportResult>.Create("tabledesigner/generatepreviewreport");
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
@@ -21,7 +21,7 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
         /// <summary>
         /// format (mimetype) of the string (plaintext or mardown)
         /// </summary>
-        public string Format;
+        public string MimeType;
     }
     
     /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
@@ -19,7 +19,7 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
         public string Report;
 
         /// <summary>
-        /// format (mimetype) of the string (plaintext or mardown)
+        /// format (mimetype) of the string
         /// </summary>
         public string MimeType;
     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
```diff
@@ -7,6 +7,18 @@
...
+    /// <summary>
+    /// The service request to generate preview report describing the changes.
+    /// </summary>
+    public class GeneratePreviewReportResult 
+    {
+        // The script generated in plaintext or markdown
+        public string Script;
+
+        // format of the string (plaintext or mardown)
+        public string Format;
```
Charles-Gagnon, line 19: I'd say this should be optional - and default to plaintext if none is specified.

I'd also suggest that this be `MimeType` instead - that's the standard way of controlling how content should be rendered. For markdown it'd be `text/markdown`

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
```diff
@@ -7,6 +7,18 @@
...
 {
+    /// <summary>
+    /// The service request to generate preview report describing the changes.
+    /// </summary>
+    public class GeneratePreviewReportResult 
+    {
+        // The script generated in plaintext or markdown
+        public string Script;
+
+        // format of the string (plaintext or mardown)
```
Charles-Gagnon, line 18: `markdown`

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
```diff
@@ -7,6 +7,18 @@
 
 namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
 {
+    /// <summary>
+    /// The service request to generate preview report describing the changes.
+    /// </summary>
+    public class GeneratePreviewReportResult 
+    {
+        // The script generated in plaintext or markdown
```
Charles-Gagnon, line 15: Use xml doc comments for these

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
```diff
@@ -7,6 +7,18 @@
...
+        public string Script;
```
alanrenmsft, line 16: this should be `Report` not script.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
```diff
@@ -7,6 +7,23 @@
...
 namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
 {
+    /// <summary>
+    /// The service request to generate preview report describing the changes.
+    /// </summary>
+
+    public class GeneratePreviewReportResult 
+    {
+        /// <summary>
+        /// The report generated in plaintext or markdown
```
Charles-Gagnon, line 17: I'd just remove the "plaintext or markdown" stuff in these comments. Just keeping this generic is fine - it's up to the callee to handle the various mime types appropriately but we shouldn't be specifying here what those types might be since we support sending anything.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -161,13 +161,17 @@ private Task HandleGenerateScriptRequest(TableInfo tableInfo, RequestContext<str
...
-            return this.HandleRequest<string>(requestContext, async () =>
+            return this.HandleRequest<GeneratePreviewReportResult>(requestContext, async () =>
             {
                 var table = this.GetTableDesigner(tableInfo);
                 var report = table.GenerateReport();
-                await requestContext.SendResult(report);
+                var generatePreviewReportResult = new GeneratePreviewReportResult();
+                // TODO - set report type by caohai
+                generatePreviewReportResult.Report = report;
+                generatePreviewReportResult.MimeType = "text/markdown";
```
Charles-Gagnon, line 173: Shouldn't this be `text/plain`?

abist: default is markdown
this will be changed by @caohai anyway

Charles-Gagnon: The report is being generated as markdown? I thought you had to hardcode the markdown string in your other PR because it wasn't being generated as markdown yet.

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
abist: Hai's working on doing that. He'll pick up from this PR.

Charles-Gagnon: It would be better to have this reflect the actual type that's being sent. He can update it as part of his PR then...or are you implying that he's going to be adding stuff to this PR?

alanrenmsft (approved):
please change the return type to text/plain, otherwise looks good.

PR #1419, opened 2022-03-09
Rename preview report type

abist:
Accompanying the type name change in this PR - https://github.com/microsoft/azuredatastudio/pull/18667

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
@@ -11,7 +11,7 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
     /// The service request to generate preview report describing the changes.
     /// </summary>
 
-    public class GeneratePreviewReportResult 
+    public class PreviewReportResult 
     {
         /// <summary>
         /// The report generated for publish preview
@@ -32,6 +32,6 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
         /// <summary>
         /// Request definition
         /// </summary>
-        public static readonly RequestType<TableInfo, GeneratePreviewReportResult> Type = RequestType<TableInfo, GeneratePreviewReportResult>.Create("tabledesigner/generatepreviewreport");
+        public static readonly RequestType<TableInfo, PreviewReportResult> Type = RequestType<TableInfo, PreviewReportResult>.Create("tabledesigner/generatepreviewreport");
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -161,17 +161,17 @@ private Task HandleGenerateScriptRequest(TableInfo tableInfo, RequestContext<str
             });
         }
 
-        private Task HandleGeneratePreviewReportRequest(TableInfo tableInfo, RequestContext<GeneratePreviewReportResult> requestContext)
+        private Task HandleGeneratePreviewReportRequest(TableInfo tableInfo, RequestContext<PreviewReportResult> requestContext)
```
Charles-Gagnon, line 164: Oh - so my comment in the other PR was just for the API stuff. This is fine as it was - having the result class be the name of the RPC method (`GeneratePreviewReport` in this case) + Result is the standard way of doing this.

So I don't think any changes here are necessary.

abist: Doesn't the API type name need to be the same in ADS and STS (I know it could work with different names because it's just json in the end), but different names would be an issue right? Also, we're the only ones consuming the Table Designer API, so doubt it would cause any confusion.

Charles-Gagnon: No, it doesn't have to be the same at all. Just the shape of the object has to map (so name and type of properties inside)

Given your comment in the other thread I'd say just revert that back to match the name here to make it clear they're related.

alanrenmsft: class/interface names do not matter when it comes to the communication between ADS and STS.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
```diff
@@ -11,7 +11,11 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
     /// The service request to generate preview report describing the changes.
     /// </summary>
 
+<<<<<<< HEAD
```
alanrenmsft, line 14: what is this? you pushed changes without resolving the conflicts?

abist: it didn't show the diff in vscode for some reason. Anyway closing this PR because it's not needed.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
abist: i said the same thing above if you read it. But we why have different names when we are the only ones using it. Anyway, changing it back to `GeneratePreviewReportResult`

PR #1420, opened 2022-03-09
new table designer features

alanrenmsft:
1. graph table
2. temporal table
3. memory-optimized table
4. add primary key support
5. some new validation rules


src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -4738,4 +4738,180 @@ The Query Processor estimates that implementing the following index could improv
...
+  <data name="TableDesignerAutoCreateHistoryTableDescription" xml:space="preserve">
+    <value>Specifies whether the engine should automatically create the history table.</value>
+    <comment></comment>
+  </data>
+  <data name="TableDesignerAutoCreateHistoryTableTitle" xml:space="preserve">
+    <value>Auto Create History Table</value>
+    <comment></comment>
+  </data>
+  <data name="TableDesignerNewHistoryTableDescription" xml:space="preserve">
+    <value>Specifies the name of the history name.</value>
```
abist, line 4902: Seems like this should be description and not name based on variable

PR #1422, opened 2022-03-10
Adding ExecutionPlanService to ADS

aasimkhan30:


src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -165,6 +165,9 @@ private static void InitializeRequestHandlersAndServices(ServiceHost serviceHost
             InitializeHostedServices(serviceProvider, serviceHost);
             serviceHost.ServiceProvider = serviceProvider;
 
+            ShowPlan.ExecutionPlanService.Instance.InitializeService(serviceHost);
```
kburtram, line 169: nit: "ShowPlan" is more of an internal SQL engine name, probably QueryExecutionPlan or ExecutionPlan is a better name.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GetExecutionPlanRequest.cs
```diff
@@ -0,0 +1,28 @@
...
+    public class GetExecutionPlanResult : ResultStatus
+    {
+        public List<ExecutionPlanGraph> Graphs { get; set; }
+    }
+
+    public class GetExecutionPlanRequest
+    {
+        public static readonly
+        RequestType<GetExecutionPlanParams, GetExecutionPlanResult> Type = 
+         RequestType<GetExecutionPlanParams, GetExecutionPlanResult>.Create("executionplan/getexecutionplan");
```
kburtram, line 26: for externally visible contracts I'd consider changing `executionplan` to `queryexecutionplan` though perhaps query is sorta implied.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GetExecutionPlanRequest.cs
```diff
@@ -0,0 +1,28 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections.Generic;
```
kburtram, line 6: these should generally be sorted, though there is also a convention to list System namespaces first, then sort the rest (which this appears to follow).  However the namespaces in the next file are sorted with System being in alphabetical order.  I would recommend to try to be consistent with one approach or the other (at least within a particular PR 😄).

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GetExecutionPlanRequest.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.ShowPlan
+{
+    public class GetExecutionPlanParams
+    {
+        public ExecutionPlanGraphFile GraphFile { get; set; }
```
kburtram, line 14: nit: not sure "File" is the right name here, i.e., does this content have to come from a file or could it come over the network, etc.  I think we've generally used the term "Info" to describe blobs of data, such as `ExecutionPlanGraphInfo`.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/ExecutionPlanService.cs
```diff
@@ -49,6 +50,23 @@ internal IProtocolEndpoint ServiceHost
...
             this.ServiceHost = serviceHost;
+            this.ServiceHost.SetRequestHandler(GetExecutionPlanRequest.Type, HandleGetExecutionPlan);
+        }
+
+        private async Task HandleGetExecutionPlan(GetExecutionPlanParams requestParams, RequestContext<GetExecutionPlanResult> requestContext)
+        {
+            try
+            {
+                var plans = ShowPlanGraphUtils.CreateShowPlanGraph(requestParams.GraphFile.GraphFileContent, "");
+                await requestContext.SendResult(new GetExecutionPlanResult
```
kburtram, line 61: do you need to set `ResultStatus.Success` to `true` (or not derive the return value from `ResultStatus`)?

kburtram (approved):
LGTM, a few nitpicks for consideration.

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
aasimkhan30: Will rename that in a separate PR. That will require a lot of file name changes.

src/Microsoft.SqlTools.ServiceLayer/ShowPlan/Contracts/GetExecutionPlanRequest.cs
aasimkhan30: Will rename that in a separate PR. That will require a lot of file name changes.

PR #1425, opened 2022-03-12
fix for issue that decimal number are exported to excel as strings

alanrenmsft:
In previous milestone, we fixed the issue that long but valid sql decimal can't be displayed in ADS (failed with conversion overflow), but we don't have special handler for those 2 types in the export to excel handler, as a result, the decimal values are being treated as strings. 

https://github.com/microsoft/azuredatastudio/issues/18615


Charles-Gagnon (approved):
Have you gone through all the other types and verified that they haven't changed how they're being exported either?

alanrenmsft:
> Have you gone through all the other types and verified that they haven't changed how they're being exported either?

yes

PR #1428, opened 2022-03-15
Backup/Restore Managed Instance

nemanja-milovancevic:
Added:

- Support for backup to URL and restore from URL
- RPC for creating blob container shared access signature


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/Contracts/CreateSasRequest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery.Contracts
+{
+    /// <summary>
+    /// Parameters passed for creating shared access signature
+    /// </summary>
+    public class CreateSasParams
+    {
+        /// <summary>
+        /// Sql server URI
```
Charles-Gagnon, line 16: `Connection URI` would be clearer I think

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -437,5 +442,34 @@ internal void ScriptBackup(BackupOperation backupOperation)
...
+           CreateSasParams optionsParams,
+           RequestContext<CreateSasResponse> requestContext)
+        {
+            try
+            {
+                ConnectionInfo connInfo;
+                bool supported = IsBackupRestoreOperationSupported(optionsParams.OwnerUri, out connInfo);
+                var response = new CreateSasResponse();
+
+                if (supported && connInfo != null)
```
Charles-Gagnon, line 456: Shouldn't you send an error if it's not supported or we can't get a connection?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -437,5 +442,34 @@ internal void ScriptBackup(BackupOperation backupOperation)
         {
             backupOperation.Execute(TaskExecutionMode.Script);
         }
+
+        internal async Task HandleCreateSasRequest(
```
Charles-Gagnon, line 446: Please make a new service for this - it doesn't really seem to belong to "disaster recovery"

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -195,8 +195,12 @@ public override Server Server
         /// Add a backup file to restore plan media list
         /// </summary>
         /// <param name="filePaths"></param>
-        public void AddFiles(string filePaths)
+        public void AddDevices(string filePaths, DeviceType deviceType)
         {
+            if (deviceType != DeviceType.File && deviceType != DeviceType.Url)
```
Charles-Gagnon, line 200: Please add a comment about why you're doing this. It seems like we should be throwing an exception if we get a device type we don't support instead of forcing it to be a file though...

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,116 @@
...
+            CredentialCollection credentials = sqlServer.Credentials;
+
+            Credential azureCredential = new Credential(sqlServer, credentialName);
+
+            // Container can have many SAS credentials coexisting, here we'll always drop existing one once customer choose to create new credential 
+            // since sql customer has no way to know its existency and even harder to retrive its secret string. 
+            if (credentials.Contains(credentialName))
+            {
+                Credential oldCredential = credentials[credentialName];
+                oldCredential.Drop();
```
Charles-Gagnon, line 57: Do we really want to drop the old credential without any sort of notification/prompt to the user?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+            sqlServer.Credentials.Refresh();
+
+            credentials = sqlServer.Credentials;
+
+            if (!credentials.Contains(credentialName))
```
Charles-Gagnon, line 63: Seems like we should throw an exception if the credential still exists or even just remove this check and try to create (and it'll fail if somehow one was created after the drop)

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,116 @@
...
+                {
+                    sasBuilder.Identifier = storedPolicyName;
+                }
+                Uri sasUri = containerClient.GenerateSasUri(sasBuilder);
+
+                return sasUri;
+            }
+            else
+            {
+                return null;
```
Charles-Gagnon, line 112: The return type should be `Uri?` if it can return null, but wouldn't you want to throw an exception if it can't generate the SAS URI?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,116 @@
...
+                                          string storedPolicyName = null)
+        {
+            // Check whether this BlobContainerClient object has been authorized with Shared Key.
+            if (containerClient.CanGenerateSasUri)
+            {
+                // Create a SAS token that's valid for one year.
+                BlobSasBuilder sasBuilder = new BlobSasBuilder()
+                {
+                    BlobContainerName = containerClient.Name,
+                    Resource = "c"
```
Charles-Gagnon, line 94: What's this resource from?

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
nemanja-milovancevic: This is how we handle backup operation too

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
nemanja-milovancevic: I removed this check.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
nemanja-milovancevic: Replaced `return null` with an exception throw

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
nemanja-milovancevic: I hope it's more clear now.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// <param name="container"></param>
+        /// <param name="policyName"></param>
+        /// <param name="selectedSaredAccessExpiryTime"></param>
+        public Uri GetServiceSasUriForContainer(BlobContainerClient containerClient,
+                                          string storedPolicyName = null)
+        {
+            // Check whether this BlobContainerClient object has been authorized with Shared Key.
+            if (containerClient.CanGenerateSasUri)
+            {
+                // Create a SAS token that's valid for one year.
```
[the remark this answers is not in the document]
nemanja-milovancevic: I agree.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// </summary>
+        public static BlobService Instance
+        {
+            get { return instance.Value; }
+        }
+
+        public void InitializeService(IProtocolEndpoint serviceHost)
+        {
+
+            // Create shared access secret
```
alanrenmsft, line 57: the comment doesn't provide any useful information, you can remove it together with the blank line above.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,123 @@
...
+            try
+            {
+                ConnectionInfo connInfo;
+                BlobService.ConnectionServiceInstance.TryFindConnection(
+                        ownerUri,
+                        out connInfo);
+
+                if (connInfo != null)
+                {
+                    using (sqlConn = ConnectionService.OpenSqlConnection(connInfo, "AzureBlob"))
```
alanrenmsft, line 102: I don't think you need to open the connection, you are not using it, you only checked the connInfo.IsCloud.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,123 @@
...
+                }
+                await requestContext.SendResult(response);
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex.ToString());
+            }
+        }
+
+        private bool IsBackupRestoreOperationSupported(string ownerUri, out ConnectionInfo connectionInfo)
```
alanrenmsft, line 90: please add comment how to determine whether backup restore is supported?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -195,8 +196,12 @@ public override Server Server
         /// Add a backup file to restore plan media list
         /// </summary>
         /// <param name="filePaths"></param>
-        public void AddFiles(string filePaths)
+        public void AddDevices(string filePaths, DeviceType deviceType)
         {
+            if (isSupportedDeviceType(deviceType))
```
udeeshagautam, line 201: Assuming that Url type is only getting supported for MI, where are we checking if the intended use is for MI or not?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/SharedAccessSignatureCreatorTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.SqlTools.ServiceLayer.AzureBlob;
+using Microsoft.SqlServer.Management.Smo;
+using Azure.Storage.Sas;
+
+namespace Microsoft.SqlTools.ServiceLayer.UnitTests.DisasterRecovery
+{
+    class SharedAccessSignatureCreatorTests
+    {
+
+        [Test]
```
udeeshagautam, line 18: are we planning to add more tests for other code paths in this PR?

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            try
+            {
+                ConnectionInfo connInfo;
+                bool supported = IsBackupRestoreOperationSupported(optionsParams.OwnerUri, out connInfo);
+                var response = new CreateSasResponse();
+
+                if (supported && connInfo != null)
+                {
+                    SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo, "Backup");
```
alanrenmsft, line 73: please put the disposable objects you are creating here into using statements.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -17,6 +17,8 @@
 using Microsoft.SqlTools.ServiceLayer.Management;
 using Microsoft.SqlTools.ServiceLayer.TaskServices;
 using Microsoft.SqlTools.Utility;
+using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlServer.Management.Common;
```
udeeshagautam, line 21: I don't see new usage for these in this file - am I missing something?

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/CreateSasRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// Blob container key
+        /// </summary>
+        public string BlobContainerKey { get; set; }
+        /// <summary>
+        /// Storage account name
+        /// </summary>
+        public string StorageAccountName { get; set; }
+        /// <summary>
+        /// Shared access signature expiration date
+        /// format: YYYY-MM-DD
```
alanrenmsft, line 33: is this UTC? are you sure it doesn't contain time part?

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/CreateSasRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// </summary>
+        public string ExpirationDate { get; set; }
+    }
+
+    /// <summary>
+    /// Response class for backup execution
```
alanrenmsft, line 39: copy/paste?

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            DateTimeOffset? expirationDate = null;
+            if (!String.IsNullOrEmpty(expirationDateString))
+            {
+                expirationDate = DateTimeOffset.Parse(expirationDateString);
+            }
+            var containerClient = new BlobContainerClient(new Uri(containerUri), new StorageSharedKeyCredential(accountName, accountKey));
+            Uri secretStringUri = GetServiceSasUriForContainer(containerClient, null, expirationDate);
+            string secretString = secretStringUri.ToString().Split('?')[1];
+            string identity = "Shared Access Signature";
```
alanrenmsft, line 37: do you need to make sure this id is not already been used?

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
```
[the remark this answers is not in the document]
nemanja-milovancevic: It's not happening to me

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
nemanja-milovancevic: OpenSqlConnection can return null

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
nemanja-milovancevic: Good point. I'm not sure if we want to enable URL backup/restore only for MI.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/SharedAccessSignatureCreator.cs
nemanja-milovancevic: This is not unique id. Sql Server can have stored multiple shared access signatures for blob containers (one per blob container). You can read more [here](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-credential-transact-sql?view=sql-server-ver15#:~:text=IDENTITY%20%3D%27identity_name%27%20Specifies%20the%20name%20of%20the%20account,signature%20%28SAS%29%2C%20the%20IDENTITY%20is%20SHARED%20ACCESS%20SIGNATURE.)

nemanja-milovancevic: This identity is not unique id. Sql Server can store more than one shared access signature (one per blob container). You can read more [here](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-credential-transact-sql?view=sql-server-ver15#:~:text=IDENTITY%20%3D%27identity_name%27%20Specifies%20the%20name%20of%20the%20account,signature%20%28SAS%29%2C%20the%20IDENTITY%20is%20SHARED%20ACCESS%20SIGNATURE.)

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/CreateSasRequest.cs
nemanja-milovancevic: @Charles-Gagnon @udeeshagautam I wanted to ask you guys about this. Should I enforce one date format in the RPC? Currently client can use [these formats](https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.parse?view=net-6.0).

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
nemanja-milovancevic: I agree about sqlConn. However, I'm not sure serverConnection and sqlServer are disposable objects.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/SharedAccessSignatureCreatorTests.cs
nemanja-milovancevic: I'm not sure we need more unit tests. However it would be nice to add some integration tests.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
nemanja-milovancevic: I removed the function. I don't think I need it since I'm already opening sqlConnection outside.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
nemanja-milovancevic: Is it better now?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/SharedAccessSignatureCreatorTests.cs
nemanja-milovancevic: I've managed to set up my environment for integration tests, and I added a backup integration test. We need to add an Environment variable for configuring azure blob connection parameters. I named the variable "AzureBlobSettingFileName", and it's a path to a configuration file. Config file has the following format:
```
{
	"BlobContainerUri" : "<test blob container uri>",
	"AccountKey" : "<test account key>",
	"AccountName" : "<test account name>"
}
```

I plan to add restore integration test too.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,77 @@
...
+                        ServerConnection serverConnection = new ServerConnection(sqlConn);
+                        Server sqlServer = new Server(serverConnection);
+
+                        SharedAccessSignatureCreator sharedAccessSignatureCreator = new SharedAccessSignatureCreator(sqlServer);
+                        string sharedAccessSignature = sharedAccessSignatureCreator.CreateSqlSASCredential(optionsParams.StorageAccountName, optionsParams.BlobContainerKey, optionsParams.BlobContainerUri, optionsParams.ExpirationDate);
+                        response.SharedAccessSignature = sharedAccessSignature;
+                    }
+                } else
+                {
+                    await requestContext.SendError("Create shared access signature is not supported opeation.");
```
bruzel, line 67: Typo

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/CreateSasRequest.cs
bruzel: It looks safer to me to split into three arguments - day, month and a year. You never know if client intent will be properly parsed on the other side having so many potential date formats...

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <param name="policyName"></param>
+        /// <param name="selectedSaredAccessExpiryTime"></param>
+        public Uri GetServiceSasUriForContainer(BlobContainerClient containerClient,
+                                          string storedPolicyName = null,
+                                          DateTimeOffset? expiringDate = null)
+        {
+            // Check whether this BlobContainerClient object has been authorized with Shared Key.
+            if (containerClient.CanGenerateSasUri)
+            {
+                // Create a SAS token that's valid for one year.
```
bruzel, line 88: not true if expiring date is explicitly given.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/CreateSasRequest.cs
Charles-Gagnon: We control the message being sent and how the date is formatted so that seems unnecessary. 

So you should make sure that this is formatted as expected (I agree using UTC time would probably be best) and then just ensure that it's formatted without regards to the current culture or anything. (you could also do something like use epoch time, but that shouldn't be necessary)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/DisasterRecovery/SharedAccessSignatureCreatorTests.cs
nemanja-milovancevic: I added restore integration test too and added ability to pass all Azure blob variables via environment variables.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/UnsupportedDeviceTypeException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
bruzel, line 1: Missing header information

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/CreateSasRequest.cs
```diff
@@ -0,0 +1,55 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
bruzel, line 3: Please update other headers to be consistent

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/UnsupportedDeviceTypeException.cs
nemanja-milovancevic: @Charles-Gagnon I'm using **dotnet build**

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,78 @@
...
+                        ServerConnection serverConnection = new ServerConnection(sqlConn);
+                        Server sqlServer = new Server(serverConnection);
+
+                        SharedAccessSignatureCreator sharedAccessSignatureCreator = new SharedAccessSignatureCreator(sqlServer);
+                        string sharedAccessSignature = sharedAccessSignatureCreator.CreateSqlSASCredential(optionsParams.StorageAccountName, optionsParams.BlobContainerKey, optionsParams.BlobContainerUri, optionsParams.ExpirationDate);
+                        response.SharedAccessSignature = sharedAccessSignature;
+                    }
+                } else
+                {
+                    await requestContext.SendError("Create shared access signature is not supported operation.");
```
Charles-Gagnon, line 68: 1. You should localize this string since we're likely going to want to show this to the user
2. I'd suggest making it clearer why it failed. There's two cases to consider, either we couldn't find the connection or it isn't a cloud DB. Ideally I'd suggest sending separate messages for both those cases
3. You should return after this, otherwise you'll also end up sending a result on line 70 which is invalid (you should also probably move the SendResult to line 65 since it doesn't make sense to send a result any other time. But it'd still be good to return early)

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,78 @@
...
+                }
+                await requestContext.SendResult(response);
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex.ToString());
```
Charles-Gagnon, line 74: You can just do `SendError(ex)`

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -0,0 +1,78 @@
...
+                       optionsParams.OwnerUri,
+                       out connInfo);
+                var response = new CreateSasResponse();
+
+                if (connInfo != null && !connInfo.IsCloud)
+                {
+                    using (
+                    SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo, "AzureBlob"))
+                    {
+                        // Connection gets discounnected when backup is done
```
Charles-Gagnon, line 58: disconnected

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/BlobSasResource.cs
```diff
@@ -0,0 +1,15 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.AzureBlob.Contracts
+{
+    class BlobSasResource
```
Charles-Gagnon, line 8: `public static class`

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/BlobSasResource.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public const string BLOB_CONTAINER = "c";
```
Charles-Gagnon, line 10: Please add a comment for these explaining what these values are

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/Contracts/CreateSasRequest.cs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /// <summary>
+    /// Request class for creating shared access signature
+    /// </summary>
+    public class CreateSasRequest
+    {
+        public static readonly
+            RequestType<CreateSasParams, CreateSasResponse> Type =
+                RequestType<CreateSasParams, CreateSasResponse>.Create("blob/createsas");
```
Charles-Gagnon, line 53: You'll need to make this `blob/createSas` with the changes I made - the method names are case sensitive

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,115 @@
...
+                oldCredential.Drop();
+            }
+
+            try
+            {
+                azureCredential.Create(identity, secretString);
+            }
+            catch (Exception ex)
+            {
+                throw new FailedOperationException("Create credential failed.", ex);
```
Charles-Gagnon, line 71: I'd consider having this be wrapped around the entire method since there's a lot of places that an exception can be thrown - and it'd be nice for all of those to know that the error was during the process of writing the SAS to the server. (you could also throw a separate exception for each place - either is fine)

Charles-Gagnon, line 71: Localize

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/SharedAccessSignatureCreator.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    sasBuilder.Identifier = storedPolicyName;
+                }
+                Uri sasUri = containerClient.GenerateSasUri(sasBuilder);
+
+                return sasUri;
+            }
+            else
+            {
+                throw new FailedOperationException("Cannot generate SAS URI for blob container");            }
```
Charles-Gagnon, line 112: Localize

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -211,12 +213,12 @@ public override void Execute()
                 {
                     string destName = Convert.ToString(this.backupInfo.BackupPathList[i], System.Globalization.CultureInfo.InvariantCulture);
                     int deviceType = (int)(this.backupInfo.BackupPathDevices[destName]);
+                    int backupDeviceType =
```
Charles-Gagnon, line 216: We're parsing the backupDeviceType here but this doesn't seem to be the one that's used in the below code - am I missing something here?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
```diff
@@ -229,6 +231,12 @@ public override void Execute()
                                 this.backup.Devices.AddDevice(destName, DeviceType.File);
                             }
                             break;
+                        case (int)DeviceType.Url:
+                            if (this.backupDeviceType == BackupDeviceType.Url)
```
Charles-Gagnon, line 235: What's the difference between the device type and the backupDeviceType? And shouldn't we be throwing or something if the backupDeviceType isn't URL here? Or is that an expected scenario?

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -195,8 +205,12 @@ public override Server Server
         /// Add a backup file to restore plan media list
         /// </summary>
         /// <param name="filePaths"></param>
-        public void AddFiles(string filePaths)
+        public void AddDevices(string filePaths, DeviceType deviceType)
```
Charles-Gagnon, line 208: Update xml doc

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -195,8 +205,12 @@ public override Server Server
...
         {
+            if (isSupportedDeviceType(this.Server.EngineEdition, deviceType))
```
Charles-Gagnon, line 210: I'd suggest just making a helper method `ThrowIfUnsupportedDeviceType`

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -223,6 +237,11 @@ public void AddFiles(string filePaths)
             }
         }
 
+        private bool isSupportedDeviceType(Edition engineEdition, DeviceType deviceType)
```
Charles-Gagnon, line 240: C# convention is `PascalCasedMethods`

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -77,6 +77,16 @@ public class RestoreDatabaseTaskDataObject : SmoScriptableOperationWithFullDbAcc
         private bool? isTailLogBackupWithNoRecoveryPossible = null;
         private string backupMediaList = string.Empty;
         private Server server;
+        private static DeviceType[] managedInstanceSupportedDeviceTypes = { DeviceType.Url };
+        private static DeviceType[] boxSupportedDeviceTypes = { DeviceType.File, DeviceType.Url };
+        private Dictionary<Edition, DeviceType[]> supportedDeviceTypes = new Dictionary<Edition, DeviceType[]>
```
Charles-Gagnon, line 82: `private static readonly`

(for all these)

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
```diff
@@ -223,6 +237,11 @@ public void AddFiles(string filePaths)
             }
         }
 
+        private bool isSupportedDeviceType(Edition engineEdition, DeviceType deviceType)
+        {
+            return supportedDeviceTypes.ContainsKey(engineEdition) && supportedDeviceTypes[engineEdition].Contains(deviceType);
```
Charles-Gagnon, line 242: I'd rather we not require that the edition be known since that would mean that if a new edition is added later on we'd have to remember to go update this to add it to the dictionary. 

Instead we should have a "DefaultSupportedDeviceTypes" - and then the dictionary would only contain "special" editions such as MI.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/UnsupportedDeviceTypeException.cs
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using Microsoft.SqlServer.Management.Smo;
+
+namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery.RestoreOperation
+{
+    class UnsupportedDeviceTypeException: Exception
```
Charles-Gagnon, line 11: `public class`

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/UnsupportedDeviceTypeException.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public UnsupportedDeviceTypeException(DeviceType deviceType) : base("Unsupported device type " + deviceType.ToString() + ".")
```
Charles-Gagnon, line 13: Localize this string

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestAzureBlobConnectionService.cs
```diff
@@ -0,0 +1,69 @@
...
+        private static string DefaultAzureBlobSettingFileName = Path.Combine(FileUtils.UserRootFolder, "azureBlobConnectionSettings.json");
+        private static TestAzureBlobConnectionService instance = new TestAzureBlobConnectionService();
+        private AzureBlobConnectionSetting settings;
+
+        private TestAzureBlobConnectionService()
+        {
+            LoadInstanceSettings();
+        }
+
+        public static TestAzureBlobConnectionService Intance
```
Charles-Gagnon, line 23: Instance

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestAzureBlobConnectionService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+using System.IO;
+using NUnit.Framework;
+
+namespace Microsoft.SqlTools.ServiceLayer.Test.Common
+{
+    public class TestAzureBlobConnectionService
+    {
+        private static string DefaultAzureBlobSettingFileName = Path.Combine(FileUtils.UserRootFolder, "azureBlobConnectionSettings.json");
```
Charles-Gagnon, line 14: This doesn't seem to be used anymore

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestAzureBlobConnectionService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        internal static AzureBlobConnectionSetting InitAzureBlobConnectionSetting()
+        {
+            try
+            {
+                AzureBlobConnectionSetting settings = new AzureBlobConnectionSetting();
+                settings.AccountKey = Environment.GetEnvironmentVariable(Constants.AzureStorageAccountKey);
+                settings.AccountName = Environment.GetEnvironmentVariable(Constants.AzureStorageAccountName);
+                settings.BlobContainerUri = Environment.GetEnvironmentVariable(Constants.AzureBlobContainerUri); 
+                Console.WriteLine("Azure Blob Connection Settings loaded successfully");
```
Charles-Gagnon, line 59: It would be a good idea to throw an error if any of these are null as well. But we only want to do that if we are actually trying to use this - so you should make the instance var 

`private static Lazy<TestAzureBlobConnectionService> instance = new Lazy<TestAzureBlobConnectionService>(() => new TestAzureBlobConnectionService());`

and then have instance return `instance.Value`

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/RestoreOperation/RestoreDatabaseTaskDataObject.cs
nemanja-milovancevic: Is this better?

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/BackupOperation/BackupOperation.cs
Charles-Gagnon: Hm, ok. This seems weird but since the code was already like this I guess it's fine to leave as is for now...would be nice to see if you can figure out why these checks would be necessary as you're working on this code though.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -742,6 +742,7 @@ RestoreBackupSetUserName = User Name
 RestoreBackupSetExpiration = Expiration
 TheLastBackupTaken = The last backup taken ({0})
 NoBackupsetsToRestore = No backupset selected to be restored
+UnsupportedDeviceType = Unsupported device type {0} for engine edition {1}.
```
Charles-Gagnon, line 745: This should take in parameters for these values

`UnsupportedDeviceType(string deviceType, string engineEdition) = Unsupported device type {0} for engine edition {1}.`

(I realize other strings in this file aren't doing it correctly, just ignore those for now and we can fix them later)

Charles-Gagnon (commented):
One last thing to fix, but otherwise looks good. 

Once you fix that I'll merge it in next week after we cut off the release branch.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -762,6 +763,11 @@ StopSessionFailed(String error) = Failed to stop session: {0}
 SessionNotFound = Cannot find requested XEvent session
 SessionAlreadyExists(String sessionName) = An XEvent session named {0} already exists
 
+############################################################################
+# Azure Blob Service
+NotSupportedCloudCreateSas = Create shared access signature is not supported for cloud instancess.
```
Charles-Gagnon, line 768: `instances`

Charles-Gagnon (approved):

alanrenmsft (approved):

Charles-Gagnon:
@nemanja-milovancevic Once the build with these changes comes out tomorrow then you can update https://github.com/microsoft/azuredatastudio/pull/18823 to include the STS version bump as well and we can get that in

PR #1429, opened 2022-03-16
add validation rule for identity setting

alanrenmsft:


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
```diff
@@ -459,4 +460,29 @@ public List<ValidationError> Run(TableViewModel table)
...
+            {
+                for (int i = 0; i < table.Columns.Items.Count; i++)
+                {
+                    var column = table.Columns.Items[i];
+                    if (column.IsIdentity && (column.IdentitySeed != 1 || column.IdentityIncrement != 1))
+                    {
+                        var propertyName = column.IdentitySeed != 1 ? TableColumnPropertyNames.IdentitySeed : TableColumnPropertyNames.IdentityIncrement;
+                        errors.Add(new ValidationError()
+                        {
+                            Message = "The use of seed and increment values other than 1 is not supported with memory optimized tables.",
```
caohai, line 479: nit "The use of **identity** seed and increment"?

caohai (approved):

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
alanrenmsft: I'll leave it like this, I copied the message from engine's error message.

PR #1433, opened 2022-03-18
[Table Designer] Change table designer preview report type to markdown

caohai:
We introduced the change to generate table designer preview report in markdown format via DacFx version bump in this commit https://github.com/microsoft/sqltoolsservice/commit/7a326b2487c208254a95541b84256ab331d0c6a5. This PR is to switch the file type.
Before this change:
image

After this change:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
@@ -169,9 +169,8 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
                 var table = this.GetTableDesigner(tableInfo);
                 var report = table.GenerateReport();
                 var generatePreviewReportResult = new GeneratePreviewReportResult();
-                // TODO - set report type by caohai
                 generatePreviewReportResult.Report = report;
-                generatePreviewReportResult.MimeType = "text/plain";
+                generatePreviewReportResult.MimeType = "text/markdown";
                 await requestContext.SendResult(generatePreviewReportResult);
             });
         }


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -169,9 +169,8 @@ private Task HandleGeneratePreviewReportRequest(TableInfo tableInfo, RequestCont
                 var table = this.GetTableDesigner(tableInfo);
                 var report = table.GenerateReport();
                 var generatePreviewReportResult = new GeneratePreviewReportResult();
-                // TODO - set report type by caohai
                 generatePreviewReportResult.Report = report;
-                generatePreviewReportResult.MimeType = "text/plain";
+                generatePreviewReportResult.MimeType = "text/markdown";
```
Charles-Gagnon, line 173: Does this mean the report from DacFX is always generated in markdown?

caohai: Yes, this report has been changed to use markdown format with DacFx version 160.6057.0-preview.

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
Charles-Gagnon: Hm, well I'd definitely suggest that you consider having this be an option that people calling GenerateReport pass in. Because not everyone is going to be able to support markdown and so having a plaintext option as well is still going to be something you want IMO.

@kisantia @udeeshagautam FYI

alanrenmsft: @Charles-Gagnon , this is not a dacfx wide thing, only table designer in dacfx is hardcoded to produce markdown report.

alanrenmsft: @caohai I don't see an immediate need, but I guess we can add an optional parameter in the tableDesigner.GenerateScript() method to specify the output format.

caohai: @Charles-Gagnon yeah as Alan mentioned, this is table designer specific and we are only using md for table designer preview purpose.

@alanrenmsft Sure I can add this later.

PR #1434, opened 2022-03-18
Export headers in an empty result set

benrr101:
**Description**: A friend of mine (@engineerchange) recently mentioned an issue to me. The issue was that if you export an empty table as a CSV with header export set to true, it will generate an empty file. The behavior he expected was that a file would be generated with just the headers, but no data. After some investigation, I determined that 1) I wrote most of this code and 2) the issue is caused by headers being output only if `WriteRow` is called. In the case of an empty result set, `WriteRow` is never called, therefore no headers are ever written. 

The proposed solution in this PR is to emit the headers as part of the constructor for the CSV SaveAs writer. Because this requires knowledge of the columns during construction, the reader/writer factories had to be refactored to get the columns when constructing the writer. Although I'd like to make more changes to improve the model for these classes, the changes would make this change a bit too big. Instead, we have some kinda dead functionality (ie, non-CSV SaveAs writers and factories take columns at construction, but do nothing with them).

I also made a smaller change to calculate selection start/end/length in the base class during construction. This just simplifies the code a bit, and I made sure to add unit tests to cover the changes

**Testing**:
* Existing unit tests pass
* New unit tests also pass


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -791,8 +791,8 @@ private async Task<long> AppendRowToBuffer(DbDataReader dbDataReader)
             {
                 throw new InvalidOperationException(SR.QueryServiceResultSetAddNoRows);
             }
-            
-            using (IFileStreamWriter writer = fileStreamFactory.GetWriter(outputFileName))
+
+            using (IFileStreamWriter writer = fileStreamFactory.GetWriter(outputFileName, null))
```
lewis-sanchez, line 795: Could specify null as the default argument, so you don't have to pass it in explicitly.

benrr101: Ok, change made. Though I still intend to do the refactor from the abandoned PR that will make the unused columns parameter go away.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamFactory.cs
```diff
@@ -40,18 +42,28 @@ public string CreateFile()
...
         /// SSMS formatted buffer file, file share is ReadWrite to allow concurrent reads/writes to the file.
         /// </summary>
         /// <param name="fileName">The file to write values to</param>
+        /// <param name="columns">
+        /// The entire list of columns for the result set. They will be filtered down as per the
+        /// request params.
+        /// </param>
         /// <returns>A <see cref="ServiceBufferFileStreamWriter"/></returns>
-        public IFileStreamWriter GetWriter(string fileName)
+        public IFileStreamWriter GetWriter(string fileName, IReadOnlyList<DbColumnWrapper> columns)
```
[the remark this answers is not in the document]
benrr101: Intentional. In order to output headers in the constructor for SaveAs CSV (the only one that was updated in this PR), we need columns in the constructor. In order to get columns to the constructor, we need columns in the `GetWriter` method. Since all the file writers (CSV, JSON, ServiceBuffer, etc) implement the same `IFileStreamFactory`, even the `ServiceBufferFileStreamFactory` has to take columns on `GetWriter`. It doesn't make any sense for it to since we don't know the columns until we've read the results, so we just pass in `null`. Avoiding this situation was what the bigger refactoring I did in the original PR was all about - split the factories into one that creates ServiceBuffer reader/writer, one that creates SaveAs writers and eliminates the methods that always throw when used on the wrong class. 

I've updated the comment to indicate this is argument ignored.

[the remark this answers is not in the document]
benrr101: Note, I still intend to submit the rest of that original PR since these changes really are hacky without it.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsWriterBase.cs
```diff
@@ -21,41 +22,54 @@ public abstract class SaveAsStreamWriter : IFileStreamWriter
         /// </summary>
         /// <param name="stream">The stream that will be written to</param>
         /// <param name="requestParams">The SaveAs request parameters</param>
-        protected SaveAsStreamWriter(Stream stream, SaveResultsRequestParams requestParams)
+        /// <param name="columns">
+        /// The entire list of columns for the result set. They will be filtered down as per the
```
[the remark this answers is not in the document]
benrr101: Updated to indicate at the base level, the columns are only used to determine the max columns to output.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsWriterBase.cs
```diff
@@ -73,7 +87,7 @@ public int WriteRow(StorageDataReader dataReader)
         /// </summary>
         /// <param name="row">The row of data to output</param>
         /// <param name="columns">The list of columns to output</param>
-        public abstract void WriteRow(IList<DbCellValue> row, IList<DbColumnWrapper> columns);
+        public abstract void WriteRow(IList<DbCellValue> row, IReadOnlyList<DbColumnWrapper> columns);
```
Charles-Gagnon, line 90: This should be `IReadOnlyList<DbColumnWrapper>? columns` given that you're passed in null as a possible value on line 374 of ResultSet. 

Although I have to ask why is null something you're passing in there anyways - is there a better way you can do that such that we don't have to handle a null set of columns?

Charles-Gagnon: And I assume you've made sure that the file stream writer is able to handle null columns correctly? It's not easy following the logic through a PR, what type of writer is that going to be anyways?

benrr101: > Should be `IReadOnlyList<DbColumnWrapper>?` 
Are we using nullable reference types in this project? I'm getting a warning that it has to be in a `#nullable` context if I'm going to make that change.

> Why is null something you're passing in there anyways
`SaveAsWriterBase` is only used for exporting files via SaveAs* commands. On line 374 of `ResultSet` we're generating a service buffer file, who's writer does not inherit from `SaveAsWriterBase`. I get that that's not clear from the name `fileStreamFactory` but that variable name was chosen prior to adding SaveAs support and I didn't want to go around changing anymore than necessary for this pared down PR.

The problem here is that for whatever reason (lost to time, though I vaguely remember working on this with Kevin Cunnane a long time ago) is the SaveAs writers and the service buffer writer share the same Reader/Writer factory interface despite being for significantly different purposes. The service buffer is for caching results as they come off the reader (everything is written to file, we don't know the columns at construction, writing results from a list of cell values is prohibited). The SaveAs writers are for exporting results (a selection of results can be selected, we do know columns at construction, writing results from a `DbDataReader` is prohibited). The only thing linking the two scenarios is that rows can be read from a service buffer file in both cases. The proposed refactor in the original PR was to give different reader/writer factory interfaces for SaveAs scenarios and for service buffer scenarios (and as I think on it more, I might just scrap the reader/writer factories altogether). But, that was apparently too big of a PR, so I've pared it back to just the changes necessary to get the SaveAs CSV writer to output headers on construction while not completely refactoring everything. I still intend to do that refactor as a later PR, so whatever issues come up as part of not doing those refactors will go away soon.

Charles-Gagnon: Gotcha, thanks for the explanation!

(I thought we had nullable enabled here but guess not!)

Charles-Gagnon (approved):

PR #1441, opened 2022-03-23
[SQL Migration] Show correct recommendation when viewing issues in assessment results

raymondtruong:
See corresponding frontend ADS PR: https://github.com/microsoft/azuredatastudio/pull/18801

---

In this PR, we fix the SQL migration extension's 'assessment results' dialog to show the correct recommendations provided by the assessment NuGet when viewing detected issues. Previously, the recommendation provided to the user was actually just the preview text for the 'more info' link duplicated. 

The issue was that the field meant to contain the recommendation message (`MigrationAssessmentInfo.Message`) contained the wrong text. By explicitly casting `ISqlMigrationAssessmentResult.Check` to the correct type `Microsoft.SqlServer.Management.Assessment.Checks.Check`, the correct text is now accessible. The value that was previously contained there is now remapped to `MigrationAssessmentInfo.DisplayName` so that it's still accessible, since the old `MigrationAssessmentInfo.DisplayName` was unused and was always null anyways.

See the [other PR](https://github.com/microsoft/azuredatastudio/pull/18801) for screenshots.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
@@ -504,21 +504,22 @@ namespace Microsoft.SqlTools.ServiceLayer.Migration
         {
             return assessmentResults.Select(r =>
             {
+                var check = (Microsoft.SqlServer.Management.Assessment.Checks.Check)r.Check;
                 return new MigrationAssessmentInfo()
                 {
-                    CheckId = r.Check.Id,
-                    Description = r.Check.Description,
-                    DisplayName = r.Check.DisplayName,
-                    HelpLink = r.Check.HelpLink,
-                    Level = r.Check.Level.ToString(),
+                    CheckId = check.Id,
+                    Description = check.Description,
+                    DisplayName = r.Message,
+                    HelpLink = check.HelpLink,
+                    Level = check.Level.ToString(),
                     TargetType = r.TargetType.ToString(),
                     DatabaseName = r.DatabaseName,
                     ServerName = r.ServerName,
-                    Tags = r.Check.Tags.ToArray(),
+                    Tags = check.Tags.ToArray(),
                     RulesetName = Engine.Configuration.DefaultRuleset.Name,
                     RulesetVersion = Engine.Configuration.DefaultRuleset.Version.ToString(),
                     RuleId = r.FeatureId.ToString(),
-                    Message = r.Message,
+                    Message = check.Message,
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -504,21 +504,22 @@ internal MigrationAssessmentInfo[] ParseAssessmentResult(IList<ISqlMigrationAsse
...
                     TargetType = r.TargetType.ToString(),
                     DatabaseName = r.DatabaseName,
                     ServerName = r.ServerName,
-                    Tags = r.Check.Tags.ToArray(),
+                    Tags = check.Tags.ToArray(),
                     RulesetName = Engine.Configuration.DefaultRuleset.Name,
                     RulesetVersion = Engine.Configuration.DefaultRuleset.Version.ToString(),
                     RuleId = r.FeatureId.ToString(),
-                    Message = r.Message,
+                    Message = check.Message,
```
raymondtruong, line 522: This is the main field that was incorrectly mapped

aasimkhan30 (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -504,21 +504,22 @@ internal MigrationAssessmentInfo[] ParseAssessmentResult(IList<ISqlMigrationAsse
...
+                    CheckId = check.Id,
+                    Description = check.Description,
+                    DisplayName = r.Message,
+                    HelpLink = check.HelpLink,
+                    Level = check.Level.ToString(),
                     TargetType = r.TargetType.ToString(),
                     DatabaseName = r.DatabaseName,
                     ServerName = r.ServerName,
-                    Tags = r.Check.Tags.ToArray(),
+                    Tags = check.Tags.ToArray(),
```
brian-harris, line 518: should there be null checks for check.Level, r.TargetType, or check.Tags, can they be null?

brian-harris (approved):

neetusingh-ms (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
raymondtruong: `check.Level` and `r.TargetType` cannot be null as they are fundamental to the migration assessment info. `check.Tags` can be empty but it won't be null. The NuGet provides a complete result regardless, this is just mapping the NuGet result to what ADS expects. Since most of these fields are unused anyways, prior to this change (where there never were any null checks to begin with) I hadn't seen any issues.

PR #1443, opened 2022-03-25
ADS Enhancement - Support Sending Query Results to a File

lewis-sanchez:
This PR is associated with this other PR that was created for Azure Data Studio (ADS): https://github.com/microsoft/azuredatastudio/pull/18822.

The purpose of this PR is to add a flag to indicate if a message contains a row count, which signals the end of a query. This helps ADS with merging query results and query messages into a file in the correct order.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ResultMessage.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ResultMessage.cs
@@ -37,7 +37,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
         /// <summary>
         /// Whether or not this message indicates the completion of a statement
         /// </summary>
-        public bool? IsStatementCompleted { get; set; }
+        public bool? HasRowCount { get; set; }
 
         /// <summary>
         /// Constructor with default "Now" time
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ResultMessage.cs
```diff
@@ -34,6 +34,11 @@ public class ResultMessage
         /// </summary>
         public string Message { get; set; }
 
+        /// <summary>
+        /// Whether or not this message indicates the completion of a statement
+        /// </summary>
+        public bool? HasRowCount { get; set; }
```
Charles-Gagnon, line 40: As mentioned in the other PR - I think it would be better to have this be `MessageType` instead corresponding to the messageType enum you made there. The row count message indicating the end of a batch is more of an implementation detail and not something our logic should need to care about.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ResultMessage.cs
```diff
@@ -45,6 +50,19 @@ public ResultMessage(string message, bool isError, int? batchId)
             Message = message;
         }
 
+        /// <summary>
+        /// Constructor with flag to indicate query statement completed
+        /// </summary>
+        public ResultMessage(string message, bool isError, int? batchId, bool? hasRowCount)
```
Charles-Gagnon, line 56: `hasRowCount` (or `messageType` if you update) should not be optional

Charles-Gagnon, line 56: You should just merge this in with the constructor above - the `hasRowCount`/`messageType` type param should always be included

Charles-Gagnon (changes requested):

kburtram:
@lewis-sanchez let's hold off on this and revisit in the future.  let's keep the branches published but we can close the PRs for now.  Thanks!

PR #1448, opened 2022-03-31
add table type to process and publish results

abist:
added `isNode`, `isEdge` and `isSystemVersioned` to edit and publish results for ADS to use

ADS change - https://github.com/microsoft/azuredatastudio/pull/18891

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
@@ -25,7 +25,7 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
         public string MimeType;
 
         /// <summary>
-        /// Metadata about the table to be captured
+        /// Metadata about the table
         /// </summary>
         public Dictionary<string, string> Metadata { get; set; }
     }
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System.Collections.Generic;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
 
 namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
@@ -22,6 +23,11 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts
         /// format (mimetype) of the string
         /// </summary>
         public string MimeType;
+
+        /// <summary>
+        /// Metadata about the table to be captured
+        /// </summary>
+        public Dictionary<string, string> Metadata { get; set; }
     }
     
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/ProcessTableDesignerEditRequest.cs
```diff
@@ -24,6 +24,12 @@ public class ProcessTableDesignerEditResponse
         public bool IsValid { get; set; }
 
         public TableDesignerIssue[] Issues { get; set; }
+
+        public bool IsEdge { get; set; }
```
Charles-Gagnon, line 28: As mentioned in the other PR, I don't think we want to be including these MSSQL-specific properties in a generic API like this. Would suggest instead having a property bag of key/value pairs that store metadata like this (similar to how OE nodes have the metadata field)

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/ProcessTableDesignerEditRequest.cs
```diff
@@ -24,6 +25,8 @@ public class ProcessTableDesignerEditResponse
         public bool IsValid { get; set; }
 
         public TableDesignerIssue[] Issues { get; set; }
+
+        public Dictionary<string, string> PropertyBag { get; set; }
```
Charles-Gagnon, line 29: `Metadata` seems like a better name to use (especially given that we use that same terminology elsewhere)

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -664,6 +666,27 @@ private bool GetBooleanValue(object value)
             return (bool)value;
         }
 
+        private bool GetIsNode(TableInfo tableInfo)
```
Charles-Gagnon, line 669: I'd suggest just making this an extension method on  the table designer (and put in a separate static class file)

```
public static bool IsNode(this TableDesigner tableDesigner)
{
   return tableDesigner.TableViewModel.IsEdge
}
```

and then in GetPropertyBag just do

```
var tableDesigner = this.GetTableDesigner(tableInfo);
var metadata = new Dictionary<string, string>()
{
            {"IsEdge", tableDesigner.IsEdge },
            // other ones here
}
return metadata;
```

It would also be nice to have Metadata be an extension function on TableInfo, but I don't know if you can do that and be able to get the TableDesigner from that...

abist: We get the table designer class from dacfx, so can't change that class.  I can change the name to metadata.

Charles-Gagnon: Extension methods are something you add in your own code, this isn't modifying the original class.

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/Contracts/Requests/GeneratePreviewReportRequest.cs
```diff
@@ -22,6 +23,11 @@ public class GeneratePreviewReportResult
         /// format (mimetype) of the string
         /// </summary>
         public string MimeType;
+
+        /// <summary>
+        /// Metadata about the table to be captured
```
alanrenmsft, line 28: what does `to be captured` mean?  please also update the comment in ads, i think you have the same comment there.

alanrenmsft (approved):

PR #1449, opened 2022-03-31
Adding decimal values to costs

aasimkhan30:
This PR fixes https://github.com/microsoft/azuredatastudio/issues/18884

For graphs with more than 20 nodes:
After:
image
Before:
image

For graphs with less than 20 nodes: 
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
@@ -33,7 +33,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan
         }
 
         public static ExecutionPlanNode ConvertShowPlanTreeToExecutionPlanTree(Node currentNode)
-        {
+        {   
             return new ExecutionPlanNode
             {
                 ID = currentNode.ID,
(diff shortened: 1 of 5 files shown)


kburtram (approved):
LGTM.  Is this the same precision that is used in SSMS?  Seems like a lot of decimal points to me, but if that is what users expects then is fine to me.

SSMS completely round up the digits so there is no decimal points. I will add Erin to this PR to get her feedback on it.

aasimkhan30:
SSMS completely round up the digits so there is no decimal points. I will add Erin to this PR to get her feedback on it.

erinstellato-ms:
@aasimkhan30 I reached out to the original requester, and a few other people, to get feedback on the number of decimal points.  First question would be whether it can be conditional.  In certain scenarios, show the whole numbers as we do now.  In other cases, go to decimals, with a max of either two (2) or three (3).

If you can make it conditional, I can provide suggestions for how to handle that.  It could either be based on number of operators, or number of operators and percentage numbers.

aasimkhan30:
@erinstellato-ms, updated screenshots. For now, graphs with less than 20 nodes will have rounded up relative costs whereas graphs with more than 20 nodes will have decimal relative costs.

erinstellato-ms (commented):
I think that this is a good operator count to start with for a threshold.  I assume that increasing it (if we get feedback that 20 is too low) is s somewhat simple (?) code change.

kburtram (approved):

PR #1451, opened 2022-04-01
Adding node overlays to execution plans

aasimkhan30:
It will look something like this on the frontend
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
@@ -171,7 +171,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         /// <summary>
         /// Type of the node overlay. This determines the icon that is displayed for it
         /// </summary>
-        public BadgeIconType Type { get; set; }
+        public BadgeType Type { get; set; }
 
         /// <summary>
         /// Text to display for the overlay tooltip
@@ -179,7 +179,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         public string Tooltip { get; set; }
     }
 
-    public enum BadgeIconType
+    public enum BadgeType
     {
         WARNING,
         CRITICALWARNING,
--- a/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
@@ -75,9 +75,9 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         public List<ExecutionPlanNode> Children { get; set; }
         public List<ExecutionPlanEdges> Edges { get; set; }
         /// <summary>
-        /// Add an overlay icon to the node like warnings and parallelism
+        /// Add badge icon to nodes like warnings and parallelism
         /// </summary>
-        public List<ExecutionPlanNodeOverlay> OverlayIcons { get; set; }
+        public List<Badge> Badges { get; set; }
     }
 
     public class ExecutionPlanGraphPropertyBase
@@ -166,16 +166,23 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         public string GraphFileType { get; set; }
     }
 
-    public class ExecutionPlanNodeOverlay
+    public class Badge
     {
         /// <summary>
         /// Type of the node overlay. This determines the icon that is displayed for it
         /// </summary>
-        public string Type { get; set; }
+        public BadgeIconType Type { get; set; }
 
         /// <summary>
         /// Text to display for the overlay tooltip
         /// </summary>
         public string Tooltip { get; set; }
+    }
+
+    public enum BadgeIconType
+    {
+        WARNING,
+        CRITICALWARNING,
+        PARALLELISM
     }
 }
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -161,4 +165,17 @@ public class ExecutionPlanGraphInfo
         /// </summary>
         public string GraphFileType { get; set; }
     }
+
+    public class ExecutionPlanNodeOverlay
+    {
+        /// <summary>
+        /// Type of the node overlay. This determines the icon that is displayed for it
+        /// </summary>
+        public string Type { get; set; }
```
Charles-Gagnon, line 174: This would probably be better as an enum

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -74,6 +74,10 @@ public class ExecutionPlanNode
         public string[] Subtext { get; set; }
         public List<ExecutionPlanNode> Children { get; set; }
         public List<ExecutionPlanEdges> Edges { get; set; }
+        /// <summary>
+        /// Add an overlay icon to the node like warnings and parallelism
+        /// </summary>
+        public List<ExecutionPlanNodeOverlay> OverlayIcons { get; set; }
```
Charles-Gagnon, line 80: Generally the names should match, so I'd expect either the property to be named "Overlays" or the class to be "ExecutionPlanNodeOverlayIcon"

But also - maybe "Badge" would be a better term here - that's usually what is used to indicate the status icons (like the connected status on servers)

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -4619,6 +4619,14 @@ The Query Processor estimates that implementing the following index could improv
     <comment>title of missing index details.
  Parameters: 0 - fileName (string), 1 - impact (string) </comment>
   </data>
+  <data name="WarningOverlayTooltip" xml:space="preserve">
+    <value>Warnings </value>
```
Charles-Gagnon, line 4623: Extra space

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
```diff
@@ -46,11 +46,34 @@ public static ExecutionPlanNode ConvertShowPlanTreeToExecutionPlanTree(Node curr
                 Properties = GetProperties(currentNode.Properties),
                 Children = currentNode.Children.Select(x => ConvertShowPlanTreeToExecutionPlanTree(x)).ToList(),
                 Edges = currentNode.Edges.Select(x => ConvertShowPlanEdgeToExecutionPlanEdge(x)).ToList(),
+                OverlayIcons = GenerateNodeOverlay(currentNode),
                 Name = currentNode.DisplayName,
                 ElapsedTimeInMs = currentNode.ElapsedTimeInMs
             };
         }
 
+        public static List<ExecutionPlanNodeOverlay> GenerateNodeOverlay(Node currentNode)
```
lewis-sanchez, line 55: You can make this an extension method for nodes.

aasimkhan30: I do not want to mix contract classes with showplan classes.

lewis-sanchez: In what way would they mix? The extension method would go in a separate static class of its own and would only be for the `Node` type. You wouldn't have to change the `Node` class itself.

aasimkhan30: I am a little confused by the suggestion. 

1. If the suggestion is to add it to ShowPlan/Node class

The badge feature seems more like an ExecutionPlan specific functionality and not a showplan functionality. In the past, I have tried to keep the showplan code as close to ssms as possible and just add my own logic to execution plan graph utils. It follows a similar pattern to that. Also, I do not see any additional benefit of adding badges method to the node class. 

2. If the suggestion is to add a setter method to ExecutionPlanNode class 
The parameters of the setter might be confusing then. What will be the input. Because providing a showplan/node as an input to the setter does not seem  intuitive

Charles-Gagnon: He's just saying make this the method (notice the "this" added)

`public static List<Badge> GenerateNodeOverlay(this Node node)`

and then you call it like this 

`node.GenerateNodeOverlay()` - although if you're doing it like this I would actually rename it to 

`public static List<Badge> GetBadges(this Node node)` 

It doesn't really change any of the logic. Extension methods are just nicer to use and read for situations like this.

(if you do this as he suggested normally extension methods go in their own separate static classes)

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -161,4 +165,24 @@ public class ExecutionPlanGraphInfo
...
+
+        /// <summary>
+        /// Text to display for the overlay tooltip
+        /// </summary>
+        public string Tooltip { get; set; }
+    }
+
+    public enum BadgeType
+    {
+        WARNING,
```
kburtram, line 184: nit: it's a good idea to have explicit values assigned to enum values, particularly if the enum will be duplicated in different codebase.  it prevents unexpected changes when adding values/changing order in the future.

kburtram (approved):

PR #1453, opened 2022-04-07
Fixing integration tests for sts 

aasimkhan30:
Test Build 
https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=148778&view=results

A bunch of tests are failing. We can have an offline discussion about that.


azure-pipelines/createBuildDirectories.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
+
+# This script creates the necessary directories that are required
```
Charles-Gagnon, line 3: Please include in the description why this is necessary with links to the relevant issues you've found about the problem.

azure-pipelines/integrationtests.yml
```diff
@@ -1,56 +1,67 @@
...
-  - task: CopyFiles@2
-    displayName: 'Copy testsettings file to bin'
+  - task: DownloadSecureFile@1
+    displayName: Download Test Environment Configuration
+    name: testEnvironmentPath
     inputs:
-      SourceFolder: '$(Agent.TempDirectory)'
-      Contents: '**\testsettings.json'
-      TargetFolder: '$(Build.SourcesDirectory)\test\Microsoft.SqlTools.ServiceLayer.IntegrationTests\bin\Debug\net6.0'
+      secureFile: 'SQLConnectionInstances.xml'
```
Charles-Gagnon, line 45: If we're using docker do we need this?

azure-pipelines/integrationtests.yml
```diff
@@ -1,56 +1,67 @@
...
+    displayName: Building Integration Tests
     inputs:
-      command: test
       projects: '**/Microsoft.SqlTools.ServiceLayer.IntegrationTests.csproj'
-      arguments: '--no-build'
-      testRunTitle: 'SqlToolsService Integration Tests'
-    enabled: false
+      arguments: '/p:CodeCoverageBuild=true'
 
-  - task: VSTest@2
```
Charles-Gagnon, line 45: Why are you removing the code coverage?

azure-pipelines/integrationtests.yml
```diff
@@ -1,56 +1,67 @@
...
+      releaseType: stable
 
-  - task: VisualStudioTestPlatformInstaller@1
-    displayName: 'Visual Studio Test Platform Installer'
-    inputs:
-      versionSelector: latestStable
+  - script: docker pull mcr.microsoft.com/mssql/server:2019-latest
+    displayName: Pull MSSQL Docker Image  
+
+  - script: 'docker run -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=$(sql-server-password)"
```
Charles-Gagnon, line 18: We should be generating the password in the pipeline, see the example pipelines I linked you

Charles-Gagnon (changes requested):

azure-pipelines/integrationtests.yml
aasimkhan30: That task is not supported on Linux

azure-pipelines/integrationtests.yml
aasimkhan30: image
I followed this guide

azure-pipelines/createBuildDirectories.sh
```diff
@@ -0,0 +1,99 @@
...
+    "it"
+    "it-IT"
+    "ja"
+    "ja-JP"
+    "ko"
+    "ko-KR"
+    "nl-NL"
+    "pl"
+    "pl-PL"
+    "pt-br"
```
Charles-Gagnon, line 62: We actually need to make all these folders? It seems odd that we need to create the folders for both cases, shouldn't just one be enough?

azure-pipelines/integrationtests.yml
```diff
@@ -1,56 +1,72 @@
...
+      releaseType: stable
 
-  - task: VisualStudioTestPlatformInstaller@1
-    displayName: 'Visual Studio Test Platform Installer'
-    inputs:
-      versionSelector: latestStable
+  - script: docker pull mcr.microsoft.com/mssql/server:2019-latest
+    displayName: Pull MSSQL Docker Image  
+
+  - bash: echo "##vso[task.setvariable variable=defaultSql2016_password;issecret=true]Test-$(Build.BuildNumber)-$(Get-Date -format yyyyMMdd-Hmmss)"
```
Charles-Gagnon, line 18: Since this is for a 2019 server we should be setting this to `defaultSql2019_password`

To start with it's fine to disable the tests for older versions of SQL, we can work to enable targeting multiple versions later.

test/Microsoft.SqlTools.ServiceLayer.TestEnvConfig/Program.cs
```diff
@@ -94,22 +94,24 @@ TestEnvConfig setting_file
...
+                var passwordEnvVariableValue = Environment.GetEnvironmentVariable((setting.Attribute("VersionKey").Value + "_password"));
+
+                settings.Add(new InstanceInfo(setting.Attribute("VersionKey").Value)
                 {
                     ServerName = setting.Element("DataSource").Value, // DataSource is required
                     ConnectTimeoutAsString = (string)setting.Element("ConnectTimeout"), //ConnectTimeout is optional
-                    User = (string)setting.Element("UserId"), // UserID is optional
-                    Password = (string)setting.Element("Password"),
+                     User = (string)setting.Element("UserId"), // UserID is optional
+                    Password = passwordEnvVariableValue != "" ? passwordEnvVariableValue : (string)setting.Element("Password"),
```
Charles-Gagnon, line 108: You need to use `string.IsNullOrEmpty` here - `GetEnvironmentVariable` returns null if the env var doesn't exist

https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-6.0

Charles-Gagnon (changes requested):

azure-pipelines/createBuildDirectories.sh
aasimkhan30: Removed a few directories that are unnecessary.

alanrenmsft: there are 2 issues:
1. dotnet build not able to create directories - we are adding a workaround in this PR
2. Linux is case sensitive, but we have mixed casing in multiple places: our own code and the nuget packages. in order to have just one folder per language, we have to make sure the loc folders in all nuget packages are following the same casing convention. 

that is what I was saying this PR is not ideal, but acceptable at least we can have the test running.

alanrenmsft (approved):

azure-pipelines/createBuildDirectories.sh
Charles-Gagnon: I'm still confused why we need multiple folders for the same loc - like `ru` and `ru-RU`. And then also still have some mixed case ones like `zh-hans` and `zh-HANS`

@aasimkhan30 Could you put a comment in here explaining why all these different ones are necessary? The comment you have explains why we need to create the folders in the first place, but it doesn't really explain why we have to make "duplicate" folders for the different languages.

azure-pipelines/integrationtests.yml
```diff
@@ -1,56 +1,72 @@
...
   - task: DotNetCoreCLI@2
-    displayName: 'Run integration tests'
+    displayName: Setting Up Test Environment
+    inputs:
+      command: run
+      projects: '**/Microsoft.SqlTools.ServiceLayer.TestEnvConfig.csproj'
+      arguments: $(testEnvironmentPath.secureFilePath)
+      testRunTitle: 'Setting tests'
+    env:
+      defaultSql2016_password: '$(defaultSql2019_password)'
```
Charles-Gagnon, line 58: This should be 2019 as well...

We may need to update some tests that are targeting 2016 to instead target 2019. But that's fine (and better IMO). If there are ones targeting both we can disable the 2016 tests for now I think.

azure-pipelines/createBuildDirectories.sh
aasimkhan30: I can add Alan's explanation to the description

Charles-Gagnon: Oh gotcha, so it's because of the nuget packages we consume. (because we can control the casing in our own repo so that shouldn't be a problem...)

Charles-Gagnon (approved):

PR #1455, opened 2022-04-08
Making star expansion test cross plat 

aasimkhan30:
Replacing the windows line breaks in test cases  with Environment New line.

Unfortunately, I cannot add Envrionment.Newline in the test case as it is generated during runtime


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -453,7 +453,7 @@ public async Task ExpandSqlStarExpressionsTest(string sqlStarQuery, int cursorLi
                 var starExpansionCompletionItem = await langService.GetCompletionItems(
                     textDocumentPosition, connectionInfoResult.ScriptFile, connectionInfoResult.ConnectionInfo);
 
-                Assert.AreEqual(expectedStarExpansionInsertText, starExpansionCompletionItem[0].InsertText, "Star expansion not found");
+                Assert.AreEqual(expectedStarExpansionInsertText.Replace("\r\n", Environment.NewLine).Replace("\n", Environment.NewLine), starExpansionCompletionItem[0].InsertText, "Star expansion not found");
```
Charles-Gagnon, line 456: You can use @ to add newlines in strings : 

```
@"[col1],
[col2],
[col3]"
```

Charles-Gagnon (approved):

PR #1456, opened 2022-04-08
A few table designer improvements

caohai:
1. vbump DacFx to 160.6093.0-preview
      - add full model validation prior to publishing
      - add constraint name uniqueness check
      - improve preview report
3. add description support to column, constraint and index (except for edge constraint, which doesn't support description).
4. add a unified field for column data type editing


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -463,6 +469,12 @@ private bool HandleUpdateItemRequest(ProcessTableDesignerEditRequestParams reque
                             case TableColumnPropertyNames.Type:
                                 column.DataType = GetStringValue(newValue);
                                 break;
+                            case TableColumnPropertyNames.AdvancedType:
```
alanrenmsft, line 472: you also need to add this field to the view information, right?

alanrenmsft (approved):

alanrenmsft:
@zijchen @kisantia we have done some basic validation for sql database proj, dacpac and schema compare and things look good.

PR #1470, opened 2022-04-21
Fix autoflush option not being used

Charles-Gagnon:
We have a command line option for setting autoflush on the log, but it was being overwritten in the Initialize call - because we didn't pass it in there it defaults to false and then sets it again there. This meant that we'd never be able to actually set and use the autoflush-log option. 

https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.Hosting/Utility/Logger.cs#L120


alanrenmsft:
we will be able to see the logs in the ADS's SqlToolsService channel with this change?

Charles-Gagnon:
No, this option isn't even used currently in ADS (and has nothing to do with the output channel even if it did)

I'll be following up with further improvements as I work through this, the log thing I want to discuss more.

alanrenmsft (approved):

PR #1471, opened 2022-04-21
Don't query for unsupported properties

Charles-Gagnon:
Was seeing this in the logs for Azure servers which don't support these properties - while it wasn't causing any problems because of the catch it's distracting and just adds unnecessary messages to the log.

### DIFF
--- a/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
@@ -14,6 +14,7 @@ using System.Security;
 using Microsoft.SqlTools.Utility;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlServer.Management.Smo;
+using System.Linq;
 
 namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
 {
--- a/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
@@ -795,9 +795,24 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection
             {
                 SqlConnection sqlConnection = GetAsSqlConnection(connection);
                 var server = new Server(new ServerConnection(sqlConnection));
-                server.SetDefaultInitFields(server.GetType(), new String[] { nameof(server.Processors), nameof(server.PhysicalMemory) });
-                sysInfo.CpuCount = server.Processors;
-                sysInfo.PhysicalMemoryInMB = server.PhysicalMemory;
+                var defaultFields = new List<string>();
+                var isProcessorsSupported = server.IsSupportedProperty(nameof(server.Processors));
+                if (isProcessorsSupported)
+                {
+                    defaultFields.Add(nameof(server.Processors));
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
@shueybubbles One thing I was thinking about - it seems like it might be useful to have a way for SMO to filter out unsupported properties itself. Maybe we even make SetDefaultInitFields ignore any unsupported properties? Or at least add a parameter to do that. Would save having to do this kind of check everywhere we set default fields.

aasimkhan30 (approved):

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableConnectionHelper.cs
```diff
@@ -795,9 +795,24 @@ public static ServerSystemInfo GetServerCpuAndMemoryInfo(IDbConnection connectio
             {
                 SqlConnection sqlConnection = GetAsSqlConnection(connection);
                 var server = new Server(new ServerConnection(sqlConnection));
-                server.SetDefaultInitFields(server.GetType(), new String[] { nameof(server.Processors), nameof(server.PhysicalMemory) });
-                sysInfo.CpuCount = server.Processors;
-                sysInfo.PhysicalMemoryInMB = server.PhysicalMemory;
+                var defaultFields = new List<string>();
+                var isProcessorsSupported = server.IsSupportedProperty(nameof(server.Processors));
```
alanrenmsft, line 799: is `isSupportedProperty` method a client side only operation?

Charles-Gagnon: Yes, it just checks whether the property is marked as supported in the definition XML's used to create the class objects

https://github.com/microsoft/sqlmanagementobjects/blob/main/src/Microsoft/SqlServer/Management/Smo/SqlSmoObject.cs#L4385

alanrenmsft (approved):

alanrenmsft:
PR validation failed. @Charles-Gagnon

shueybubbles:
@Charles-Gagnon  you aren't the first to ask

https://github.com/microsoft/sqlmanagementobjects/issues/84

shueybubbles (approved):
I am a bit surprised SetDefaultInitFields even matters for the Server class. 
I thought it only applied to populating objects in collections. 
If you have some analysis of how it improves things here, please add the information to the SMO wiki on github.

Charles-Gagnon:
Oh neat, I'll look into sending out a PR for this tomorrow since I'm in the area working on this stuff.

Charles-Gagnon:
> I am a bit surprised SetDefaultInitFields even matters for the Server class. I thought it only applied to populating objects in collections. If you have some analysis of how it improves things here, please add the information to the SMO wiki on github.

Interesting, I'll investigate this tomorrow. 

@aasimkhan30 Do you recall if you verified whether this was actually necessary?

aasimkhan30 (approved):

aasimkhan30:
@Charles-Gagnon , here is the link to that. https://github.com/microsoft/azuredatastudio/blob/d202fd703edc5935672bcc341b28b0f29407d174/extensions/sql-migration/src/models/stateMachine.ts#L718

aasimkhan30:
I have verified this and the migration extension uses the Assessment nuget package and not these properties to generate the required telemetry. It can be safely removed from sts.

Charles-Gagnon:
This was already promoted to the stable API so I'd rather just leave it in - it's not have any negative impact as far as I can tell outside the error message in the log.

aasimkhan30 (approved):

PR #1476, opened 2022-04-26
Refreshes token for intellisense

cssuh:
Current behavior: after the refresh token expires (~1 hour), intellisense will stop working for users writing connected azure database queries.

New behavior: the refresh token will get automatically refreshed behind the scenes so intellisense will continue to work indefinitely 

ADS PR: https://github.com/microsoft/azuredatastudio/pull/19214


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionInfo.cs
```diff
@@ -173,5 +173,14 @@ public void RemoveAllConnections()
                 ConnectionTypeToConnectionMap.TryRemove(type, out connection);
             }
         } 
+
+        /// <summary>
+        /// Updates the Auth Token and Expires On fields
```
alanrenmsft, line 178: missing closing tag

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -233,6 +233,43 @@ public ISqlConnectionFactory ConnectionFactory
         /// <param name="connectionParams">The params to validate</param>
```
alanrenmsft, line 233: the comments doesn't match the method

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -233,6 +233,43 @@ public ISqlConnectionFactory ConnectionFactory
...
+                Authority = connection.ConnectionDetails.GetOptionValue("azureTenantId", string.Empty),
+                Provider = "Azure",
+                Resource = "SQL",
+                Uri = ownerUri
+            };
+            try {
+                await this.ServiceHost.SendEvent(RefreshTokenNotification.Type, requestMessage);
+            }
+            catch (Exception e){
+                Console.WriteLine(e);
```
alanrenmsft, line 256: Do not use Console.WriteLine here for 2 reasons:
1. the STDIO are used for IPC communication in STS, writing unexpected stuff to it will break STS
2. writing to console is effectively the same as do nothing, we won't know what happened.
Please use logger here, you can ask @Charles-Gagnon for how to use the logger.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -233,6 +233,43 @@ public ISqlConnectionFactory ConnectionFactory
...
+                AccountId = connection.ConnectionDetails.GetOptionValue("azureAccount", string.Empty),
+                Authority = connection.ConnectionDetails.GetOptionValue("azureTenantId", string.Empty),
+                Provider = "Azure",
+                Resource = "SQL",
+                Uri = ownerUri
+            };
+            try {
+                await this.ServiceHost.SendEvent(RefreshTokenNotification.Type, requestMessage);
+            }
+            catch (Exception e){
```
alanrenmsft, line 255: in case of exception, shouldn't it return false?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -452,6 +453,9 @@ internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestCon
                 }
                 else
                 {
+                    if (RefreshNeeded(scriptFile.ClientUri)) {
```
alanrenmsft, line 456: As we chatted before, I think we should let ConnectionService handle the refresh automatically. for example when the connection is added, if it has expiresOn info, then we will automatically refresh it when it is about to expire. when the connection is removed we will delete the auto refresh handler.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: And add context to the message you write indicating what specifically failed

Charles-Gagnon (changes requested):

alanrenmsft (changes requested):
left some comments, especially, the token refresh should be handled by connection service so that the consumers of connection service don't have to know about it.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -233,6 +233,43 @@ public ISqlConnectionFactory ConnectionFactory
...
+            var requestMessage = new RefreshTokenNotificationParams
+            {
+                AccountId = connection.ConnectionDetails.GetOptionValue("azureAccount", string.Empty),
+                Authority = connection.ConnectionDetails.GetOptionValue("azureTenantId", string.Empty),
+                Provider = "Azure",
+                Resource = "SQL",
+                Uri = ownerUri
+            };
+            try {
+                await this.ServiceHost.SendEvent(RefreshTokenNotification.Type, requestMessage);
```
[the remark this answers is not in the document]
cssuh: yes, I have tested it a few times and it works.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
cssuh: We chatted about this in our team sync today, and it will have to be a two part plan - first we will get this PR in to get the customer unblocked, and then I will circle back and implement a more global solution for the token refresh.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -906,6 +910,18 @@ public async Task HandleDidChangeTextDocumentNotification(ScriptFile[] changedFi
             }
         }
 
+        internal async Task HandleRefreshTokenNotification(
+            RefreshTokenParams refreshTokenParams,
+            EventContext eventContext
+        )
+        {
+            await Task.Run(() =>
```
Charles-Gagnon, line 918: awaiting a Task.Run isn't really adding anything useful, you might as well get rid of it and just call `UpdateAuthToken` directly, then return `Task.CompletedTask;`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
alanrenmsft: this is still not resolved @cssuh

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
alanrenmsft: I am fine with it, please create an issue to track it.

alanrenmsft: this code block should be defined as a method and it seems you are not using the output of the method, what if the refresh fails?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -455,6 +455,7 @@ internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestCon
                 {
                     if (RefreshNeeded(scriptFile.ClientUri)) {
                         await connectionService.TryRefreshAuthToken(scriptFile.ClientUri);
+                        await requestContext.SendResult(null);
```
alanrenmsft, line 458: why sending null?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -69,6 +69,8 @@ public class ConnectionService
         /// </summary>
         internal Dictionary<string, ConnectionInfo> OwnerToConnectionMap { get; } = new Dictionary<string, ConnectionInfo>();
 
+        public Dictionary<string, Boolean> IsConnected = new Dictionary<string, bool>();
```
alanrenmsft, line 72: I don't think this is needed based on the code you have, you are adding/removing stuff to it but never uses it.

alanrenmsft:
@cssuh please reply to the comment indicating whether you are done resolving all the comments. I do see unresolved issues but seems you are waiting for review again.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
cssuh: I want to use offline intellisense if the auth token needs to be refreshed

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -227,12 +227,78 @@ public ISqlConnectionFactory ConnectionFactory
...
+                    {   
+                        // disable intellisense
+                    } 
+                    else 
+                    {
+                        return false;
+                    }
+                }
+            }
+            if (!this.TryFindConnection(ownerUri, out ConnectionInfo connection))
```
alanrenmsft, line 261: what do we have to do the check again? this seems can be the else block of the previous if statement.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -227,12 +227,78 @@ public ISqlConnectionFactory ConnectionFactory
         // Attempts to link a URI to an actively used connection for this URI
         public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connectionInfo) => this.OwnerToConnectionMap.TryGetValue(ownerUri, out connectionInfo);
 
+        /// <summary>
+        /// Refreshes the auth token of a given connection 
+        /// </summary>
+        /// <param name="ownerUri">The URI of the connection</param>
+        /// <returns>true upon successful refresh of the auth token,
+        /// false if the uri cannot be found</returns>
```
alanrenmsft, line 235: the description doesn't match the behavior. false can also mean no refresh needed

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -227,12 +227,78 @@ public ISqlConnectionFactory ConnectionFactory
...
+                {
+                    return false;
+                } 
+                else 
+                {
+                    // Check if token is expired
+                    var maxTolerance = 2 * 60; // two minutes
+                    if (connInfo.ConnectionDetails.ExpiresOn - DateTimeOffset.Now.ToUnixTimeSeconds() < maxTolerance)
+                    {   
+                        // disable intellisense
```
alanrenmsft, line 253: disable intellisense?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
alanrenmsft: and if refresh fails.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
alanrenmsft: we should try start the here, right?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -227,12 +227,78 @@ public ISqlConnectionFactory ConnectionFactory
...
+            return true;
+        }
+
         /// <summary>
         /// Validates the given ConnectParams object. 
         /// </summary>
         /// <param name="connectionParams">The params to validate</param>
         /// <returns>A ConnectionCompleteParams object upon validation error, 
         /// null upon validation success</returns>
+        internal bool UpdateAuthToken(RefreshTokenParams refreshToken)
```
alanrenmsft, line 291: description and the actual behavior don't match

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
alanrenmsft: what is the expected behavior here? I thought we want to wait for the refresh to complete and then have the correct intellisense send back to the UI.

Charles-Gagnon: @alanrenmsft We can't do that because of the synchronous processing issue that you have the PR out to start addressing. My understanding is that this is a temporary quick fix to unblock this until we can fully fix the synchronous processing issue - at which point this stuff should be removed and replaced with a Request call

cssuh: The disabling intellisense is part of the issue in the ICM, they're saying that the log files were growing very quickly once a token expired because intellisense kept trying to autofill but could not due to the expired token

The refresh token is able to be refreshed in the background, based on an access token that expires every couple months, so no user interaction should be necessary.

Charles-Gagnon: Refresh tokens aren't guaranteed to be valid - they can be revoked or time out for example. (I think there's also an AAD thing you can change to require user interaction each time - it would be good to verify this)

https://docs.microsoft.com/en-us/azure/active-directory/develop/refresh-tokens

So I think we should at least be considering the possibility of this happening. I would suggest looking at how other implementations in VS Code handles this.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,77 @@ public ISqlConnectionFactory ConnectionFactory
...
 
         /// <summary>
-        /// Validates the given ConnectParams object. 
+        /// Refreshes the auth token of a given connection 
         /// </summary>
-        /// <param name="connectionParams">The params to validate</param>
-        /// <returns>A ConnectionCompleteParams object upon validation error, 
-        /// null upon validation success</returns>
+        /// <param name="ownerUri">The URI of the connection</param>
+        /// <returns>true upon successful refresh of the auth token,
```
Charles-Gagnon, line 234: This is not true - the token is not actually refreshed when this method returns. It *requests* a refresh which will happen at a later point. So this should be `True if a refresh is needed, false if...`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,77 @@ public ISqlConnectionFactory ConnectionFactory
...
+            if (this.TryFindConnection(ownerUri, out connInfo))
+            {
+                // If not an azure connection, no need to refresh token
+                if (connInfo.ConnectionDetails.AuthenticationType != "AzureMFA") 
+                {
+                    return false;
+                } 
+                else 
+                {
+                    // Check if token is expired
```
Charles-Gagnon, line 250: `Check if token is expired or about to expire`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,77 @@ public ISqlConnectionFactory ConnectionFactory
...
+                    var maxTolerance = 2 * 60; // two minutes
+                    if (connInfo.ConnectionDetails.ExpiresOn - DateTimeOffset.Now.ToUnixTimeSeconds() < maxTolerance)
+                    {   
+                        var requestMessage = new RefreshTokenParams
+                        {
+                            AccountId = connInfo.ConnectionDetails.GetOptionValue("azureAccount", string.Empty),
+                            Authority = connInfo.ConnectionDetails.GetOptionValue("azureTenantId", string.Empty),
```
Charles-Gagnon, line 257: Why is "authority" being used to store the tenant ID? Authority != tenant

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,77 @@ public ISqlConnectionFactory ConnectionFactory
...
+                } 
+                else 
+                {
+                    // Check if token is expired
+                    var maxTolerance = 2 * 60; // two minutes
+                    if (connInfo.ConnectionDetails.ExpiresOn - DateTimeOffset.Now.ToUnixTimeSeconds() < maxTolerance)
+                    {   
+                        var requestMessage = new RefreshTokenParams
+                        {
+                            AccountId = connInfo.ConnectionDetails.GetOptionValue("azureAccount", string.Empty),
```
Charles-Gagnon, line 256: If we don't have an accountId or "authority" (tenant) then we shouldn't even try to request a refresh - since that's guaranteed to fail.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,77 @@ public ISqlConnectionFactory ConnectionFactory
...
+                            Authority = connInfo.ConnectionDetails.GetOptionValue("azureTenantId", string.Empty),
+                            Provider = "Azure",
+                            Resource = "SQL",
+                            Uri = ownerUri
+                        };
+                        try {
+                            await this.ServiceHost.SendEvent(RefreshTokenNotification.Type, requestMessage);
+                        }
+                        catch (Exception ex){
+                            Logger.Write(TraceEventType.Error, "Failed to refresh auth token " + ex.Message);
```
Charles-Gagnon, line 266: `Failed to send RefreshTokenNotification`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,77 @@ public ISqlConnectionFactory ConnectionFactory
...
+            else
+            {
+                return false;
+            }
+
+            
+        }
+
+        /// <summary>
+        /// Updates the azure auth token
```
Charles-Gagnon, line 286: These comments also need updating - we aren't actually updating it here. Just requesting an update.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/RefreshTokenNotification.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        public int ExpiresOn { get; set; }
+
+        /// <summmary>
+        /// Connection URI
+        /// </summary>
+        public string Uri { get; set; }
+    }
+
+    class RefreshToken
```
Charles-Gagnon, line 72: `TokenRefreshedNotification`

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -906,6 +911,15 @@ public async Task HandleDidChangeTextDocumentNotification(ScriptFile[] changedFi
             }
         }
 
+        internal Task HandleRefreshTokenNotification(
+            TokenRefreshedParams refreshTokenParams,
```
Charles-Gagnon, line 915: These names will also need to be updated - they should be `TokenRefreshed*`

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
cssuh: what do you mean by `TokenRefreshed*`? Isn't `TokenRefreshedParams` the right name?

Charles-Gagnon: `HandleTokenRefreshedNotification`
`tokenRefreshedParams`

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/RefreshTokenNotification.cs
```diff
@@ -0,0 +1,78 @@
...
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
+{
+    class RefreshTokenParams 
+    {
+        /// <summary>
+        /// Tenant for the token
```
Charles-Gagnon, line 13: The tenant isn't "for the token" - it's just a part of the scope of what the token is authenticating for. Just `ID of the tenant` is fine - that's just a standard part of any AD auth flow so shouldn't need a detailed explanation.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -452,6 +453,10 @@ internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestCon
                 }
                 else
                 {
+                    if (await connectionService.TryRefreshAuthToken(scriptFile.ClientUri))
+                    {
+                        await requestContext.SendResult(null);
```
Charles-Gagnon, line 458: Add comment here explaining why we're returning null in this situation

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
Charles-Gagnon: Ping - did you look into this at all? It's just going to keep trying to refresh constantly right? I'm not sure that's the behavior we want - it feels like we should cut it off at some point and just stop trying

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
```diff
@@ -41,6 +47,11 @@ class RequestSecurityTokenResponse
         /// Gets or sets the access token.
         /// </summary>
         public string Token { get; set; }
+
+        /// <summmary>
+        /// Gets or sets the token expiration
+        /// </summary>
+        public int ExpiresOn { get; set; }
```
Charles-Gagnon, line 54: Given that this is just an int I'd suggest being more explicit in the doc comment explaining what this value represents

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/RefreshTokenNotification.cs
```diff
@@ -0,0 +1,78 @@
...
+            EventType<RefreshTokenParams> Type =
+            EventType<RefreshTokenParams>.Create("account/refreshToken");
+    }
+
+    class TokenRefreshedParams
+    {
+        /// <summary>
+        /// Gets or sets the key that uniquely identifies a particular linked account.
+        /// </summary>
+        public string AccountKey { get; set; }
```
Charles-Gagnon, line 54: Is this being used anywhere?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,81 @@ public ISqlConnectionFactory ConnectionFactory
...
+            
+        }
+
+        /// <summary>
+        /// Requests an update of the azure auth token
+        /// </summary>
+        /// <param name="refreshToken">The token to update</param>
+        /// <returns>true upon successful update, false if it failed to find 
+        /// the connection</returns>
+        internal bool UpdateAuthToken(TokenRefreshedParams refreshToken)
```
Charles-Gagnon, line 295: This is returning a bool but I'm not seeing that being used

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,81 @@ public ISqlConnectionFactory ConnectionFactory
...
+            {
+                // If not an azure connection, no need to refresh token
+                if (connInfo.ConnectionDetails.AuthenticationType != "AzureMFA") 
+                {
+                    return false;
+                } 
+                else 
+                {
+                    // Check if token is expired or about to expire
+                    var maxTolerance = 2 * 60; // two minutes
```
Charles-Gagnon, line 251: This can just be a constant

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +228,81 @@ public ISqlConnectionFactory ConnectionFactory
         public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connectionInfo) => this.OwnerToConnectionMap.TryGetValue(ownerUri, out connectionInfo);
 
         /// <summary>
-        /// Validates the given ConnectParams object. 
+        /// Refreshes the auth token of a given connection
```
Charles-Gagnon, line 231: `of a given connection, if needed`

Charles-Gagnon:
You have build failures too FYI

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
cssuh: I have added a dictionary to keep track of the state - if the token is in the "Updating" state, I won't request another token

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -59,6 +60,7 @@ public class ConnectionService
         private readonly ConcurrentDictionary<CancelTokenKey, CancellationTokenSource> cancelTupleToCancellationTokenSourceMap =
                     new ConcurrentDictionary<CancelTokenKey, CancellationTokenSource>();
 
+        public Dictionary<string, TokenState> IsConnected = new Dictionary<string, TokenState>();
```
Charles-Gagnon, line 63: `IsConnected` isn't a very clear name, and I'm not sure why you'd need separate states anyways. Just a set of URIs currently being refreshed should be fine, you add it when a refresh is requested and then remove it when the token refreshed notification is received.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/RefreshTokenNotification.cs
Charles-Gagnon: @cssuh

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +230,93 @@ public ISqlConnectionFactory ConnectionFactory
...
+                {
+                    return false;
+                }
+                else
+                {
+                    // Check if token is expired or about to expire
+                    if (connInfo.ConnectionDetails.ExpiresOn - DateTimeOffset.Now.ToUnixTimeSeconds() < maxTolerance)
+                    {
+                        // Check if the token is updating already, in which case there is no need to request a new one
+                        if (this.IsConnected[ownerUri] == TokenState.Updating) {
```
Charles-Gagnon, line 256: Given that @alanrenmsft is currently working on enabling async message processing we're going to have to start being more careful about stuff like this. You should add a lock (or even better use a concurrent object like ConcurrentDictionary)

(I don't believe there's a concurrent set so despite that being preferred you can just use a dictionary and have the value be just true or something throwaway if you can't find a better solution...)

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/RefreshTokenNotification.cs
cssuh: It's not being used, I can remove it

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -59,6 +60,7 @@ public class ConnectionService
         private readonly ConcurrentDictionary<CancelTokenKey, CancellationTokenSource> cancelTupleToCancellationTokenSourceMap =
                     new ConcurrentDictionary<CancelTokenKey, CancellationTokenSource>();
 
+        public ConcurrentDictionary<string, Boolean> TokenUpdateUris = new ConcurrentDictionary<string, Boolean>();
```
Charles-Gagnon, line 63: `private readonly`

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +230,91 @@ public ISqlConnectionFactory ConnectionFactory
...
+                            return false;
+                        }
+                        try
+                        {
+                            this.TokenUpdateUris.TryAdd(ownerUri, true);
+                            await this.ServiceHost.SendEvent(RefreshTokenNotification.Type, requestMessage);
+                        }
+                        catch (Exception ex)
+                        {
+                            Logger.Write(TraceEventType.Error, "Failed to send RefreshTokenNotification " + ex.Message);
```
Charles-Gagnon, line 284: Just remove this try/catch. It would be pretty weird if it failed to send and would likely indicate other bigger problems - and regardless even if an error occurred since we need to refresh the token we wouldn't get intellisense anyways.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +230,91 @@ public ISqlConnectionFactory ConnectionFactory
...
+                    }
+                    else
+                    {
+                        return false;
+                    }
+                }
+            }
+            else
+            {
+                Logger.Write(TraceEventType.Error, "Failed to find connection when refreshing token");
```
Charles-Gagnon, line 297: Use the helper functions I added such as `Logger.Error`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +230,91 @@ public ISqlConnectionFactory ConnectionFactory
...
+        /// Requests an update of the azure auth token
+        /// </summary>
+        /// <param name="refreshToken">The token to update</param>
+        /// <returns>true upon successful update, false if it failed to find 
+        /// the connection</returns>
+        internal void UpdateAuthToken(TokenRefreshedParams tokenRefreshedParams)
+        {
+            if (!this.TryFindConnection(tokenRefreshedParams.Uri, out ConnectionInfo connection))
+            {
+                Logger.Write(TraceEventType.Error, "Failed to find connection when refreshing token");
```
Charles-Gagnon, line 312: `when updating refreshed token for URI {tokenRefreshedParams.Uri}` - log messages should be unique enough to tell where exactly they came from (currently this is duplicated in a different context on line 297) and contain information to help with debugging.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: And add a comment explaining what this is for

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
```diff
@@ -28,6 +28,12 @@ class RequestSecurityTokenParams
         /// Gets or sets the scope of the authentication request.
         /// </summary>
         public string Scope { get; set; }
+
+        /// <summary>
+        /// Gets or sets the account ID
+        /// </summary>
+        public string AccountId { get; set; }
```
Charles-Gagnon, line 35: Why are you changing these notifications? You don't appear to be modifying the handlers for these

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +230,91 @@ public ISqlConnectionFactory ConnectionFactory
...
+                        }
+                        var requestMessage = new RefreshTokenParams
+                        {
+                            AccountId = connInfo.ConnectionDetails.GetOptionValue("azureAccount", string.Empty),
+                            TenantId = connInfo.ConnectionDetails.GetOptionValue("azureTenantId", string.Empty),
+                            Provider = "Azure",
+                            Resource = "SQL",
+                            Uri = ownerUri
+                        };
+                        if (requestMessage.TenantId == null || requestMessage.TenantId == String.Empty)
```
Charles-Gagnon, line 267: `string.IsNullOrEmpty(requestMessage.TenantId)` (same for account)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -452,6 +453,11 @@ internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestCon
                 }
                 else
                 {
+                    if (await connectionService.TryRefreshAuthToken(scriptFile.ClientUri))
```
kburtram, line 456: it looks like you may be sending two SendResults per-request if the token needs to be refreshed since we don't return in the if or have the subsequent code in an else block.  This snippet may better look-up the default completion list since it would pass a null connectioninfo to GetCompletionItems method.

```C#

ConnectionInfo connInfo = null;
if (!await connectionService.TryRefreshAuthToken(scriptFile.ClientUri))
{
	// if token does not need to be refreshed then retrieve connection from connectionservice
	// get the current list of completion items and return to client
	ConnectionServiceInstance.TryFindConnection(
		scriptFile.ClientUri,
		out connInfo);
}
				
var completionItems = await GetCompletionItems(
	textDocumentPosition, scriptFile, connInfo);
await requestContext.SendResult(completionItems);
```

nit: the naming and semantics of `TryRefreshAuthToken` is not super-intuitive for me, since it succeeds when a refresh is needed, which seems to be more of the failure path in my mind.  Not a huge-deal if this makes more sense else in the code.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
cssuh: I think I did this on accident, I will change it back

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +234,84 @@ public ISqlConnectionFactory ConnectionFactory
...
         /// </summary>
-        /// <param name="connectionParams">The params to validate</param>
-        /// <returns>A ConnectionCompleteParams object upon validation error, 
-        /// null upon validation success</returns>
+        /// <param name="ownerUri">The URI of the connection</param>
+        /// <returns>true if a refresh is needed,
+        /// false if the uri cannot be found, refresh is not needed,
+        /// refresh token has already been requested, or if refresh fails </returns>
+
+        internal async Task<bool> TryRefreshAuthToken(string ownerUri)
```
Charles-Gagnon, line 244: Let's name this `TryRequestRefreshAuthToken` instead to make it clear this doesn't immediately refresh the token

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +234,84 @@ public ISqlConnectionFactory ConnectionFactory
...
 
         /// <summary>
-        /// Validates the given ConnectParams object. 
+        /// Refreshes the auth token of a given connection, if needed
         /// </summary>
-        /// <param name="connectionParams">The params to validate</param>
-        /// <returns>A ConnectionCompleteParams object upon validation error, 
-        /// null upon validation success</returns>
+        /// <param name="ownerUri">The URI of the connection</param>
+        /// <returns>true if a refresh is needed,
```
Charles-Gagnon, line 240: `True if a refreshed was needed and requested, false otherwise`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +234,84 @@ public ISqlConnectionFactory ConnectionFactory
...
+        /// Requests an update of the azure auth token
+        /// </summary>
+        /// <param name="refreshToken">The token to update</param>
+        /// <returns>true upon successful update, false if it failed to find 
+        /// the connection</returns>
+        internal void UpdateAuthToken(TokenRefreshedParams tokenRefreshedParams)
+        {
+            if (!this.TryFindConnection(tokenRefreshedParams.Uri, out ConnectionInfo connection))
+            {
+                Logger.Error($"Failed to find connection when updating refreshed token for URI {tokenRefreshedParams.Uri}");
```
Charles-Gagnon, line 309: return after the log message

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -556,6 +556,18 @@ public string AzureAccountToken
             }
         }
 
+        public int ExpiresOn
```
Charles-Gagnon, line 559: This should probably be a nullable int (`int?`) since it isn't going to exist in every request

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/SecurityTokenRequest.cs
```diff
@@ -41,6 +41,11 @@ class RequestSecurityTokenResponse
         /// Gets or sets the access token.
         /// </summary>
         public string Token { get; set; }
+
+        /// <summmary>
+        /// Gets or sets the token expiration
+        /// </summary>
+        public int ExpiresOn { get; set; }
```
Charles-Gagnon, line 48: Why is this being added anyways? You aren't changing anything with the security token request messages that I can see.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -452,11 +453,15 @@ internal async Task HandleSyntaxParseRequest(SyntaxParseParams param, RequestCon
                 }
                 else
                 {
-                    // get the current list of completion items and return to client
-                    ConnectionServiceInstance.TryFindConnection(
+                    ConnectionInfo connInfo = null;
+                    if (!await connectionService.TryRefreshAuthToken(scriptFile.ClientUri))
+                    {
+                        // if the auth token needs to be refreshed, we will disable intellisense until it has been refreshed.
```
Charles-Gagnon, line 459: The comment being in here is confusing - unless I'm misunderstanding this being inside the if block here means that we didn't need to refresh the auth token.

I'd instead suggest removing this and just put a comment above the if 

`Check if we need to refresh the auth token, and if we do then don't pass in the connection so that we only show the default options until the refreshed token is returned`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +234,84 @@ public ISqlConnectionFactory ConnectionFactory
...
+                        }
+                        if (String.IsNullOrEmpty(requestMessage.AccountId))
+                        {
+                            Logger.Error("No accountId in connection details when refreshing token for connection {ownerUri}");
+                            return false;
+                        }
+                        // Check if the token is updating already, in which case there is no need to request a new one
+                        if (!this.TokenUpdateUris.TryAdd(ownerUri, true))
+                        {
+                            return false;
```
Charles-Gagnon, line 281: Shouldn't this be true? Because we want this to return true if the token is invalid (regardless of whether we already requested a refresh or not)

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
cssuh: why do you way it won't exist in every request? It should be there

Charles-Gagnon: This is the connection details. Not all connections have an azure token.

cssuh: oh I see, okay that makes sense

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -556,6 +556,18 @@ public string AzureAccountToken
             }
         }
 
+        public int? ExpiresOn
+        {
+            get
+            {
+                return GetOptionValue<int>("expiresOn");
```
Charles-Gagnon, line 563: This should also be `int?`

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -33,6 +33,7 @@ public class ConnectionService
         public const string AdminConnectionPrefix = "ADMIN:";
         internal const string PasswordPlaceholder = "******";
         private const string SqlAzureEdition = "SQL Azure";
+        public const int maxTolerance = 2 * 60; // two minutes
```
alanrenmsft, line 36: 1. for const, you should follow Pascal casing just like other consts here. first letter to be upper case
2. is there a reason you choose 2 minutes? add it to the comment.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1211,7 +1224,7 @@ private bool RequiresReparse(ScriptParseInfo info, ScriptFile scriptFile)
             return prevSqlText.Length != currentSqlText.Length
                 || !string.Equals(prevSqlText, currentSqlText);
         }
-
+
```
alanrenmsft, line 1227: remove

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +234,83 @@ public ISqlConnectionFactory ConnectionFactory
...
+        /// <returns>true upon successful update, false if it failed to find 
+        /// the connection</returns>
+        internal void UpdateAuthToken(TokenRefreshedParams tokenRefreshedParams)
+        {
+            if (!this.TryFindConnection(tokenRefreshedParams.Uri, out ConnectionInfo connection))
+            {
+                Logger.Error($"Failed to find connection when updating refreshed token for URI {tokenRefreshedParams.Uri}");
+                return;
+            }
+            this.TokenUpdateUris.Remove(tokenRefreshedParams.Uri, out var result);
```
alanrenmsft, line 310: I just realized we should also delete the entry when the connection is disconnected.

cssuh: it only gets added once a token refresh is needed, so are you referring to the case where a refresh token is requested but before it gets returned the user closes the connection manually?

Charles-Gagnon: Right, or when the editor is closed. Otherwise a user could end up in a state where closing and re-opening an editor that's been disconnected will result in the new editor never getting connected autocompletion results since it's stuck waiting on the token refresh response. 

@cssuh

Charles-Gagnon: Yes, that's correct. Or when an error occurs as the other PR mentions (since that will result in the URI sitting in the cache forever currently)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -59,6 +60,11 @@ public class ConnectionService
         private readonly ConcurrentDictionary<CancelTokenKey, CancellationTokenSource> cancelTupleToCancellationTokenSourceMap =
                     new ConcurrentDictionary<CancelTokenKey, CancellationTokenSource>();
 
+        /// <summary>
+        /// A map containing the uris of editors whose azure tokens are being refreshed, these editors should have intellisense
+        /// disabled until the new refresh token is returned, upon which they will be removed from the map
+        /// </summary>
+        private readonly ConcurrentDictionary<string, Boolean> TokenUpdateUris = new ConcurrentDictionary<string, Boolean>();
```
Charles-Gagnon, line 67: `tokenUpdateUris`

Charles-Gagnon, line 67: `tokenUpdateUris`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -807,6 +885,11 @@ public virtual bool Disconnect(DisconnectParams disconnectParams)
                 return false;
             }
 
+            if (info.ConnectionDetails.AuthenticationType == "AzureMFA")
```
Charles-Gagnon, line 888: Is this also called when an editor is closed?

And please add a comment explaining why you're doing this

Charles-Gagnon, line 888: Is this also called when an editor is closed?

And please add a comment explaining why you're doing this

Charles-Gagnon (changes requested):
Waiting for confirmation that the URI is removed from the map when an editor is closed as well

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +234,83 @@ public ISqlConnectionFactory ConnectionFactory
...
+                        {
+                            Logger.Error("No tenant in connection details when refreshing token for connection {ownerUri}");
+                            return false;
+                        }
+                        if (String.IsNullOrEmpty(requestMessage.AccountId))
+                        {
+                            Logger.Error("No accountId in connection details when refreshing token for connection {ownerUri}");
+                            return false;
+                        }
+                        // Check if the token is updating already, in which case there is no need to request a new one
```
Charles-Gagnon, line 276: `no need to request a new one, but still return true so that autocompletion is disabled until the token is refreshed`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -59,6 +60,11 @@ public class ConnectionService
         private readonly ConcurrentDictionary<CancelTokenKey, CancellationTokenSource> cancelTupleToCancellationTokenSourceMap =
                     new ConcurrentDictionary<CancelTokenKey, CancellationTokenSource>();
 
+        /// <summary>
+        /// A map containing the uris of editors whose azure tokens are being refreshed, these editors should have intellisense
```
kburtram, line 64: probably the comment should say connections with expired tokens, since we don't know if they are actually being refreshed, or if the refreshed failed/aborted/etc.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -228,11 +234,83 @@ public ISqlConnectionFactory ConnectionFactory
...
+
+                        var requestMessage = new RefreshTokenParams
+                        {
+                            AccountId = connInfo.ConnectionDetails.GetOptionValue("azureAccount", string.Empty),
+                            TenantId = connInfo.ConnectionDetails.GetOptionValue("azureTenantId", string.Empty),
+                            Provider = "Azure",
+                            Resource = "SQL",
+                            Uri = ownerUri
+                        };
+                        if (String.IsNullOrEmpty(requestMessage.TenantId))
```
kburtram, line 266: nit: .Net style guides recommend to use primitive types when available.  e.g. `string.IsNullOrEmpty`.  Particularly since we use that convention above.  Not good to mix-and-match from a style perspective.

kburtram (approved):

kburtram:
@cssuh we need to get this fix available for Cx asap.  Could you please merge this and ADS PRs once you've completed whatever testing is remaining?  Thanks!

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -59,6 +60,11 @@ public class ConnectionService
         private readonly ConcurrentDictionary<CancelTokenKey, CancellationTokenSource> cancelTupleToCancellationTokenSourceMap =
                     new ConcurrentDictionary<CancelTokenKey, CancellationTokenSource>();
 
+        /// <summary>
+        /// A map containing the uris of connections with expired tokens, these editors should have intellisense
+        /// disabled until the new refresh token is returned, upon which they will be removed from the map
+        /// </summary>
+        public readonly ConcurrentDictionary<string, Boolean> tokenUpdateUris = new ConcurrentDictionary<string, Boolean>();
```
Charles-Gagnon, line 67: nit: now that it's public the convention should be `CapitalCased`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -807,6 +886,13 @@ public virtual bool Disconnect(DisconnectParams disconnectParams)
                 return false;
             }
 
+            // This clears the uri of the connection from the tokenUpdateUris map, which is used to track
+            // open editors that have requested a refreshed AAD token.
+            if (info.ConnectionDetails.AuthenticationType == "AzureMFA")
```
Charles-Gagnon, line 891: Probably don't need this check - if a connection is disconnected we can always just try to remove it from the map since it just won't do anything for non-AAD connections

Charles-Gagnon (approved):

PR #1479, opened 2022-04-26
Removes unnecessary using statement to fix build error

lewis-sanchez:
This PR resolves a build error that is caused by an unnecessary using directive.


Charles-Gagnon (approved):
Did you investigate why this isn't failing the product build pipeline? Please follow up with doing that so we can make sure this doesn't happen again.

chlafreniere (approved):

PR #1480, opened 2022-04-27
Add Tools Integration Test Project to STS Build Pipeline

lewis-sanchez:
This PR adds the integration environment config and test projects to the STS Build Pipeline,

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -43,11 +43,6 @@ steps:
     testRunTitle: Kusto.ServiceLayer.UnitTests
     arguments: '--configuration $(buildConfiguration) --collect "Code coverage"'
 
-- task: DotNetCoreCLI@2
-  displayName: Building Test Environment
-  inputs:
-    projects: '**/Microsoft.SqlTools.ServiceLayer.TestEnvConfig.csproj'
-
 - task: DownloadSecureFile@1
   displayName: Download Test Environment Configuration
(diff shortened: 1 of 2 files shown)


azure-pipelines/build.yml
```diff
@@ -43,6 +43,41 @@ steps:
...
+  inputs:
+    projects: '**/Microsoft.SqlTools.ServiceLayer.TestEnvConfig.csproj'
+
+- task: DownloadSecureFile@1
+  displayName: Download Test Environment Configuration
+  name: testEnvironmentPath
+  inputs:
+    secureFile: 'SQLConnectionInstances.xml'
+
+- task: DotNetCoreCLI@2
```
Charles-Gagnon, line 57: We shouldn't be adding this stuff here - if anything the integration tests should just be referenced in the build-and-release.yml as a separate stage there.

That being said it's still reasonable that build.cmd should be building everything. Did you confirm whether it is/isn't building the integration tests project? Because fixing that up would be the first step and we can follow up with adding integration tests running during the product build and PR pipelines

Charles-Gagnon (changes requested):

azure-pipelines/build.yml
Charles-Gagnon: Please remove the rest of this for now, we shouldn't need it for just fixing the build.

kburtram (approved):

PR #1481, opened 2022-05-01
Updating Kusto with Changes From #1434

benrr101:
**Description**: As far as I can tell, it looks like Kusto query execution is just a copy of the regular SQL Server query execution. As such, I'm thinking the changes I introduced for exporting headers (#1434 ) on otherwise empty export requests should be copied into Kusto codebase, as well. So, that's what this PR does. I basically took the patch for #1434, replaced the word "SqlTools" with "Kusto" and applied it to main.

**Testing**: Since there aren't any unit tests for query execution in Kusto, and I don't have any Kusto instances to test against (... and I don't really even know what Kusto is), these changes are pretty much untested aside from ensuring they build. I'd like to test it, but not sure how to do that yet, give me some suggestions.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/QueryExecution/DataStorage/IFileStreamFactory.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/QueryExecution/DataStorage/IFileStreamFactory.cs
@@ -3,6 +3,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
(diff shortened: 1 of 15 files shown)


Charles-Gagnon:
Thanks for following up with this @benrr101 

@JustinMDotNet Could you please find someone to review this change? Let us know if you have any questions about the context.

Charles-Gagnon:
Closing this for now until the Kusto team decides that this is something they want to take.

PR #1482, opened 2022-05-04
add a flag to make message processing parallel

alanrenmsft:
Currently by default, the messages are handled sequentially unless the service explicitly run the handler in a non-blocking way, this could be the root cause of a few GitHub issues, for example: intellisense not showing up.

I reviewed all the request handlers in STS and below is the summary table:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Service | Parallel handlers/Total handlers
-- | --
Admin service | 0/4
Agent | 0/36
Blob service | 0/1
Azure Functions   Service | 0/2
CMS | 6/6
Connection | 2/7
Dac | 9/9
DisasterRecovery | 2/6
EditData | 0/9
ExecutionPlan | 0/2
FileBrowser | 4/4
TSQLFormatter | 0/2
InsightsGenerator | 0/1
ExternalLanguage   (ML) | 0/4
LanguageService | 0/9
Metadata | 1/3
MigrationService | 0/5
ModelManagement   -ML | 0/7
Notebook Convert | 0/2
Object Explorer | 3/5
Profiler | 0/6
Query Execution | 8/14
Schema compare | 10/10
Scripting | 3/3
Security | 0/4
Server Config (ML) | 0/3
SQL Assessment | 3/3
Table Designer | 6/6
Task Service | 0/2
Workspace | 0/4
TOTAL | 47/179

</div>

</body>

</html>

To fix this we have to make the message handling non-blocking in the message dispatcher so that individual services don't need to worry about this. I chatted with @kburtram about this, we both think this is a good improvement but will be risky to enable it directly, so I am introducing a configuration option in ADS and pass it to STS when starting. draft PR for ADS (waiting for this PR to be merged and published): https://github.com/microsoft/azuredatastudio/pull/19279

I've done some basic testing and things are working fine in a more responsive way. 

with feature flag off (current behavior):
before-parallel-processing

with feature flag on:
parallel-processing

The plan going forward:

1. schedule a test pass for our team to test the product thoroughly with the feature flag turned on.
2. let the users who have experienced the effect of sequential processing. (by replying to the github issues)
3. enable it in insider
4. fix any issues related to this change
5. enable it by default in Aug release or later.

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
@@ -57,6 +57,12 @@ namespace Microsoft.SqlTools.Hosting.Protocol
         protected MessageWriter MessageWriter { get; private set; }
 
 
+        /// <summary>
+        /// Whether the message should be handled without blocking the main thread.
+        /// </summary>
+        public bool ParallelMessageProcessing { get; set; }
+
+
         #endregion
 
         #region Constructors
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -305,22 +311,37 @@ private async Task ListenForMessages(CancellationToken cancellationToken)
 
             if (handlerToAwait != null)
             {
-                try
+                if (this.ParallelMessageProcessing)
                 {
-                    await handlerToAwait;
+                    var task = Task.Run(async () =>
```
Charles-Gagnon, line 316: Just call RunTask without awaiting, no reason to wrap this if we're not doing anything with it.

alanrenmsft: there will be a build warning, to get rid of it, you need to either await it or assign it to a variable like the `var task = ` here.

image

Charles-Gagnon: You should also be able to add a suppression for it (VS is much better for this - VS Code doesn't have the auto-fix options available)

In the end it's just a warning though so as long as we make it clear that we intend now to await it I don't see a problem with that.

(you could also at the very least just set a var for the original task `var task = RunTask(...);`)

Charles-Gagnon: Oh yeah, that's a thing too 😁

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -305,22 +311,37 @@ private async Task ListenForMessages(CancellationToken cancellationToken)
 
             if (handlerToAwait != null)
             {
-                try
+                if (this.ParallelMessageProcessing)
                 {
-                    await handlerToAwait;
+                    _ = Task.Run(() =>
```
Charles-Gagnon, line 316: So are you still purposely using Task.Run to kick it off to a new thread? I suppose that makes sense, but please add a comment indicating that since right now it seems a bit odd to use Task.Run when you aren't awaiting the result.

alanrenmsft: added comment

PR #1489, opened 2022-05-10
Fix Agent actions for docker servers

Charles-Gagnon:
Was fixing the `TestAgentNotebookCreateHelper` test and found out that it was failing because of a bug - Docker servers may end up with a server name matching the container ID instead of the host name. In that case when we build these URNs we specify the connection name (localhost in my case) but then SFC throws because it expects the server name to match the "true" name (which is the `@@servername`).

The fix here is to remove the Name filter - SFC queries are always ran against a server in the first place so there's no reason to filter with the name.

For https://github.com/microsoft/azuredatastudio/issues/19323


aasimkhan30 (approved):
Thanks you for fixing these!

PR #1490, opened 2022-05-10
Fix Backup/Restore parameters defaulting to wrong DeviceType

Charles-Gagnon:
Fixing some broken tests which caught a regression - the DeviceType was previously hardcoded to be File but with the new changes it defaults to the default value for the enum (0 - which is LogicalDisk)

Introduced in https://github.com/microsoft/sqltoolsservice/pull/1428

Also formatted the test file and changed all the tests to use fluent assertions so we get detailed failure information instead of just messages like "Expected true but got false".


Charles-Gagnon:
@nemanja-milovancevic FYI, can you make sure that the tests you added are working as expected as well?

nemanja-milovancevic:
@Charles-Gagnon It looks like the tests are still working fine.

PR #1492, opened 2022-05-11
Revert SMO bump

kburtram:
Revert the SMO bump from this commit https://github.com/microsoft/sqltoolsservice/commit/6c97f41f7b61085091630aff297c219b5ccc3e37.  This SMO bump introduced this regression into ADS OE Tables expansion as discussed in https://github.com/microsoft/azuredatastudio/issues/19335.

@Charles-Gagnon please feel free to merge or close this PR based on how you'd like to address this.  I've confirmed this change resolves the ADS issue.


Charles-Gagnon:
Thanks, I'll kick off an STS build and get ADS updated so we aren't blocking people using insiders.

PR #1494, opened 2022-05-12
Add error link to period col error

abist:
Adds a link property to designer issues, and added a link for period column errors. 

ADS PR - https://github.com/microsoft/azuredatastudio/pull/19365

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
@@ -393,7 +393,7 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
                 errors.Add(new TableDesignerIssue()
                 {
                     Description = "System versioned table must have the period columns defined.",
-                    MoreInfoLink = "https://docs.microsoft.com/sql/relational-databases/tables/creating-a-system-versioned-temporal-table?view=sql-server-ver15"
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
```diff
@@ -392,7 +392,8 @@ public List<TableDesignerIssue> Run(Dac.TableDesigner designer)
             {
                 errors.Add(new TableDesignerIssue()
                 {
-                    Description = "System versioned table must have the period columns defined."
+                    Description = "System versioned table must have the period columns defined.",
+                    MoreInfoLink = "https://docs.microsoft.com/sql/relational-databases/tables/creating-a-system-versioned-temporal-table?view=sql-server-ver15"
```
kburtram, line 396: should this URL contain a parameter for a specific SQL Server version, such as `sql-server-ver15`?  If so, should it come from the version the user is connected to?  Perhaps if the parameter is omitted MSDN will auto-navigate to the latest version, which is probably fine.

kburtram (approved):
LGTM, but not sure the URL should have hard coded SQL Server version.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
Charles-Gagnon: +1 - remove that query parameter. Same as the LCIDs we should let the server handle redirecting to the appropriate final link for general links like this.

PR #1495, opened 2022-05-13
[Table Designer] Add apis to support computed column

caohai:
https://github.com/microsoft/azuredatastudio/issues/19117
This PR adds apis to support computed column.

image


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -816,6 +828,14 @@ private TableViewModel GetTableViewModel(TableInfo tableInfo)
                 columnViewModel.IdentityIncrement.Enabled = column.CanEditIdentityValues;
                 columnViewModel.IdentityIncrement.Value = column.IdentityIncrement?.ToString();
                 columnViewModel.CanBeDeleted = column.CanBeDeleted;
+                columnViewModel.IsComputed.Enabled = column.CanEditIsComputed;
+                columnViewModel.IsComputed.Checked = column.IsComputed;
+                columnViewModel.ComputedFormula.Enabled = column.CanEditComputedFormula;
+                columnViewModel.ComputedFormula.Value = column.ComputedFormula;
+                columnViewModel.IsComputedPersisted.Enabled = column.CanEditIsComputedPersisted;
+                columnViewModel.IsComputedPersisted.Checked = column.IsComputedPersisted == true;
```
alanrenmsft, line 836: you don't need the `==true` part.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -816,6 +828,14 @@ private TableViewModel GetTableViewModel(TableInfo tableInfo)
...
+                columnViewModel.IsComputedPersistedNullable.Enabled = column.CanEditIsComputedPersistedNullable;
+                columnViewModel.IsComputedPersistedNullable.Checked = column.IsComputedPersistedNullable == true;
```
alanrenmsft, line 838: same here.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
caohai: I defined `IsComputedPersisted ` in DacFx as `bool?` and it will be `null` if the column is not computed.

alanrenmsft (approved):

PR #1497, opened 2022-05-17
version bump

alanrenmsft:


Charles-Gagnon:
Are you going to fix it so that we aren't changing the revision # each time? I would have expected this first one to be `4.0.1-release.0`, the second one to be `4.0.2-release.0` etc. And then hotfixes would be `4.0.1-release.1` and so on. 

(although honestly I think we should consider removing the "release" and just go with `4.0.1.0` too...)

alanrenmsft:
I meant to create a draft pr while I test things out and I am going to email the team and collect feedback.

alanrenmsft:
as we chatted offline, things look good after the vbump.

Charles-Gagnon:
So the process is that we're going to manually update the version in this file to `4.1.0.0` after the next release? And then the version is always to be incrementing the revision version by 1 (so `4.0.0.0` -> `4.0.0.1` -> `4.0.0.2` and so on)?

Can we have this just be 3 numbers to match the actual semver spec? So `4.0.0`

alanrenmsft:
the third is patch number, we use it for hot fix, and the fourth number is build number, incremented automatically by the build script.

Charles-Gagnon:
Right, ok yeah that makes sense.

Charles-Gagnon (approved):

PR #1502, opened 2022-05-20
Add isNode and isEdge to properties list

abist:
Fixing April 2022 RCA bug to fetch isNode and isEdge properties by preloading them in the properties so fetching them doesn't require making extra metadata queries in the background.

TODO: Add Managed Instance flag for compatibility. (no draft PR's in this repo)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/TreeNodeDefinition.xml
```diff
@@ -83,6 +83,8 @@
       <Property Name="IsSystemVersioned" ValidFor="Sql2016|Sql2017|AzureV12"/>
 	    <Property Name="TemporalType" ValidFor="Sql2016|Sql2017|AzureV12"/>
       <Property Name="IsExternal" ValidFor="Sql2016|Sql2017|AzureV12|SqlOnDemand"/>
+      <Property Name="IsNode" ValidFor="Sql2017|AzureV12"/>
```
Charles-Gagnon, line 86: Did you look into how this is used? What happens for 2019/2022 servers? Are we correctly adding the property to the default init list for those?

alanrenmsft:
is this still needed? @abist

PR #1503, opened 2022-05-20
fix memory optimized option inconsistent state

caohai:
https://github.com/microsoft/azuredatastudio/issues/19425

The fix is to 
1) hide memory optimized options when it's not supported and not checked.
2) throw a validation error when `IsMemoryOptimized` is checked but we figure out it's actually not supported/enabled in current database (by running a query against the db). 

DacFx vbump to: https://www.nuget.org/packages/Microsoft.SqlServer.DacFx/160.6164.0-preview
disable_mem_check

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
@@ -567,7 +567,7 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
             {
                 errors.Add(new TableDesignerIssue()
                 {
-                    Description = string.Format("Memory-optimized is either not supported or not enabled in this database."),
+                    Description = string.Format("Memory-optimized table is not supported for this database."),
                     PropertyPath = new object[] { TablePropertyNames.IsMemoryOptimized }
                 });
             }
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
@@ -519,8 +519,9 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
             {
                 errors.Add(new TableDesignerIssue()
                 {
-                    Description = "The table has more than one edge constraint on it. This is only useful as a temporary state when modifying existing edge constraints, and should not be used in other cases. Please refer to https://docs.microsoft.com/sql/relational-databases/tables/graph-edge-constraints for more details.",
-                    Severity = Contracts.IssueSeverity.Warning
+                    Description = "The table has more than one edge constraint on it. This is only useful as a temporary state when modifying existing edge constraints, and should not be used in other cases.",
+                    Severity = Contracts.IssueSeverity.Warning,
+                    MoreInfoLink = "https://docs.microsoft.com/sql/relational-databases/tables/graph-edge-constraints"
                 });
             }
             return errors;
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
```diff
@@ -555,4 +556,22 @@ public List<TableDesignerIssue> Run(Dac.TableDesigner designer)
...
+    {
+        public List<TableDesignerIssue> Run(Dac.TableDesigner designer)
+        {
+            var table = designer.TableViewModel;
+            var errors = new List<TableDesignerIssue>();
+            if (!designer.IsMemoryOptimizedTableSupported && designer.TableViewModel.IsMemoryOptimized)
+            {
+                errors.Add(new TableDesignerIssue()
+                {
+                    Description = string.Format("Memory-optimized is either not supported or not enabled in this database."),
```
caohai, line 570: Not sure if I've asked this before, but do we need to use localized string for all these validation descriptions?

alanrenmsft: I was planning to move all the rules to dacfx, so I left these unlocalized. I'll create an issue in Aug milestone to track this task.

alanrenmsft, line 570: 1. change the description to: `Memory optimized table is not supported for this database`
2. use the new property that Aditya added to provide a more information link for this issue and get the link to a Microsoft docs website that covers this topic.

caohai: I'm thinking about adding a `IsAzure` property to `TableDesigner` and link to:
1. https://docs.microsoft.com/en-us/sql/relational-databases/in-memory-oltp/overview-and-usage-scenarios if the server is not Azure
2. https://docs.microsoft.com/en-us/azure/azure-sql/in-memory-oltp-overview if the server is Azure.

caohai: marking this as draft for now as `IsAzure` needs DacFx change.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
```diff
@@ -518,8 +519,9 @@ public List<TableDesignerIssue> Run(Dac.TableDesigner designer)
             {
                 errors.Add(new TableDesignerIssue()
                 {
-                    Description = "The table has more than one edge constraint on it. This is only useful as a temporary state when modifying existing edge constraints, and should not be used in other cases. Please refer to https://docs.microsoft.com/sql/relational-databases/tables/graph-edge-constraints for more details.",
```
caohai, line 521: I also updated this one using `MoreInfoLink` as I was making related chagnes. I've tested this one and the link is working correctly,

alanrenmsft:
the experience is definitely not good to show something first but then hide it later. I think a better experience is to always show it and just disable it if it is not supported. 
maybe remove line 1361 to 1375?
image

caohai:
> the experience is definitely not good to show something first but then hide it later. I think a better experience is to always show it and just disable it if it is not supported. maybe remove line 1361 to 1375? image

Are you saying removing from line 1361 to 1364? We need to keep the Memory Optimized checkbox, right?
Here is what it looks like after changing to disabling the checkbox.
disable_mem_check

alanrenmsft (approved):

caohai:
did sanity testing for dacpac, bacpac and schema compare features. things look good. cc @kisantia @zijchen

zijchen (approved):

PR #1507, opened 2022-05-23
Revert "Revert "Adding new Dac Deployment Options and getting descriptions fr…"

ssreerama:
Reverts microsoft/sqltoolsservice#1493

Reverting the revert PR
old PRs is here :
https://github.com/microsoft/sqltoolsservice/pull/1357 (merged)
https://github.com/microsoft/sqltoolsservice/pull/1493 (reverted the above)

This PR includes:

1. Newly added or missed Dac deployment options to the dacfx service, now dacpac/SC/SQL DB Proj extensions will have these options available.
2. Description for the options are now pulling from the DacFx API and added to the options model, now we don't need to duplicate the huge list of descriptions in all extensions.
3. Modified and fixed the tests cases accordingly.


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -2,192 +2,223 @@
...
+            this.Value = value;
+            this.Description = description;
+            this.DisplayName = displayName;
+        }
+
+        // Default and selected value of the deployment options
+        public T Value { get; set; }
+
+        // Description of the deployment options
+        public string Description { get; set; } = string.Empty;
```
kisantia, line 32: is setting this to string.Empty necessary? Doesn't description get set to "" by default in the constructor?

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Not required! its a duplicated one with no reason to have it, Thanks

PR #1509, opened 2022-05-25
ToolsService Integration tests - DacFx default options test validation fixes

ssreerama:
This PR fixes the default options actual and expected values test validations.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
@@ -859,10 +859,7 @@ Streaming query statement contains a reference to missing output stream 'Missing
                     Assert.True((defaultP == null && actualP == null) 
                      || (defaultP == null && String.IsNullOrEmpty(actualP as string))
                      || defaultP.Equals(actualP)
-                    , $"Actual Property from Service is not equal to default property for {v.Name}, Actual property: {actualP} and Default property: {defaultP}");
-
-                    // Verifying expected and actual deployment options property values are equal
-                    Assert.True((defaultPValue == null && actualPValue == null)
+                     || (defaultPValue == null && actualPValue == null)
                      || (defaultPValue == null && String.IsNullOrEmpty(actualPValue as string))
                      || (defaultPValue).Equals(actualPValue)
                     , $"Actual Property from Service is not equal to default property for {v.Name}, Actual value: {actualPValue} and Default value: {defaultPValue}");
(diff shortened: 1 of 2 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -859,10 +859,7 @@ private bool ValidateOptions(DeploymentOptions expected, DeploymentOptions actua
                     Assert.True((defaultP == null && actualP == null) 
                      || (defaultP == null && String.IsNullOrEmpty(actualP as string))
                      || defaultP.Equals(actualP)
-                    , $"Actual Property from Service is not equal to default property for {v.Name}, Actual property: {actualP} and Default property: {defaultP}");
-
-                    // Verifying expected and actual deployment options property values are equal
-                    Assert.True((defaultPValue == null && actualPValue == null)
+                     || (defaultPValue == null && actualPValue == null)
```
Benjin, line 862: Just making sure I'm reading this correctly: either the properties need to match, OR the property _values_ need to match for the test to pass?

What situation would the properties _not_ match, but the property values would still match?  With your changes, do we still care about the `P` now that we have `PValue`?

kisantia: this assert check is long and hard to read. Can it be simplified like Benjin mentioned?

ssreerama: Absolutely! there is no need of bunch of checks that required now as these tests are verifying the default values, I have updated and tested them and converted to simple assertions. Thanks :)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -731,7 +731,6 @@ public async Task GetOptionsFromProfile()
             expectedResults.BlockOnPossibleDataLoss = new DeploymentOptionProperty<bool>(true);
             expectedResults.AllowIncompatiblePlatform = new DeploymentOptionProperty<bool>(true);
             expectedResults.DisableIndexesForDataPhase = new DeploymentOptionProperty<bool>(false);
-            expectedResults.IgnoreTableOptions = new DeploymentOptionProperty<bool>(false);
```
ssreerama, line 734: Removing these as these are not required now :)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -859,15 +856,7 @@ private bool ValidateOptions(DeploymentOptions expected, DeploymentOptions actua
                 else
                 {
                     //Verifying expected and actual deployment options properties are equal
-                    Assert.True((defaultP == null && actualP == null)
```
Benjin, line 862: Removing defaultP/actualP and only checking defaultPValue/actualPValue was speculation/a question on my part; can you confirm that defaultP/actualP are no longer important to check?

ssreerama: We do not require now, as the defaultPValue get the value or null if defaulp (line 848), checking defaulPvalue is same here, and actuyalPValue is same it will be either null or "" depending on actualP which we do not needed in assertion. Thanks.

Benjin (approved):

PR #1510, opened 2022-05-25
[SQL Migration] Update assessment NuGet version + mark GAed VM SKUs as non-preview

raymondtruong:
In this PR, we:
- vbump Microsoft.SqlServer.Migration.Assessment from 1.0.20220211.24 to 1.0.20220527.33, including several changes/improvements to assessment and SKU recommendation
- update assessment/SKU recommendation metadata
- mark the Eb series SQL VM SKUs as GAed, no longer requiring the preview flag for them to be considered

Testing:
- verified latest assessment rules being consumed in ADS:
image
(with >100 databases, now shows the instance as an impacted object)
- verified SKU recommendation logic in ADS:
image
(no MI recommendations when >100 databases selected)


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -625,18 +625,16 @@ internal static List<AzureSqlSkuCategory> GetEligibleSkuCategories(string target
...
+                    if (File.Exists(computePreviewFilePath))
                     {
-                        // Eb series (in preview) capabilities stored separately 
-                        string computePreviewFilePath = Path.Combine(assemblyPath, RecommendationConstants.DataFolder, RecommendationConstants.SqlVmPreviewCapability);
-                        if (File.Exists(computePreviewFilePath))
-                        {
-                            jsonFile = File.ReadAllText(computePreviewFilePath);
-                            List<AzureSqlIaaSCapability> vmPreviewCapabilities = JsonConvert.DeserializeObject<List<AzureSqlIaaSCapability>>(jsonFile);
+                        jsonFile = File.ReadAllText(computePreviewFilePath);
+                        List<AzureSqlIaaSCapability> vmPreviewCapabilities = JsonConvert.DeserializeObject<List<AzureSqlIaaSCapability>>(jsonFile);
```
brian-harris, line 633: will these changes be backwards compatible?  or will the new nuget package be required?

raymondtruong: Previously, these specific VM SKUs (which are stored in a separate file) were only considered for recommendation if the preview flag was checked. Now that these SKUs are GAed, we consider them no matter what (hence removing the if statement). So for that change specifically, a new NuGet technically wasn't required since those SKUs already existed before this.

However, we do need a new NuGet because of changes made to the recommendation logic itself (which could how these SKUs even get recommended in the first place) as well as some other miscellaneous changes to SKU rec/assessment rules.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -625,18 +625,16 @@ internal static List<AzureSqlSkuCategory> GetEligibleSkuCategories(string target
                     string jsonFile = File.ReadAllText(Path.Combine(assemblyPath, RecommendationConstants.DataFolder, RecommendationConstants.SqlVmCapability));
                     List<AzureSqlIaaSCapability> vmCapabilities = JsonConvert.DeserializeObject<List<AzureSqlIaaSCapability>>(jsonFile);
 
-                    if (includePreviewSkus)
+                    // Eb series capabilities stored separately 
+                    string computePreviewFilePath = Path.Combine(assemblyPath, RecommendationConstants.DataFolder, RecommendationConstants.SqlVmPreviewCapability);
```
neetusingh-ms, line 629: nit: may be review 'preview' from variable name?

neetusingh-ms: just realized it's coming from a separate preview file :)... do we have plans to merge it. no harm in keeping as is. just curious.

neetusingh-ms (approved):

Packages.props
```diff
@@ -25,7 +25,7 @@
 		<PackageReference Update="Microsoft.Azure.Kusto.Data" Version="9.0.4" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Language" Version="9.0.4"/>
 		<PackageReference Update="Microsoft.SqlServer.Assessment"  Version="[1.0.305]" />
-		<PackageReference Update="Microsoft.SqlServer.Migration.Assessment"  Version="1.0.20220211.24" />
```
neetusingh-ms, line 28: Thank You! for saving my time :)

PR #1514, opened 2022-05-26
Fix integration test pipeline

Charles-Gagnon:
Fix indentation error in the DotNetCoreCLI task at the bottom (env was indented too much)

I also fixed up a number of other issues with invalid params or bad formatting, and updated the NuGetAuthenticate task since v0 was deprecated.

### DIFF
--- a/azure-pipelines/integrationtests.yml
+++ b/azure-pipelines/integrationtests.yml
@@ -1,6 +1,6 @@
 pool:
     name: ads-build-1es-hosted-pool
-    demands: 
+    demands:
     - ImageOverride -equals ADS-Linux_Image
 
 steps:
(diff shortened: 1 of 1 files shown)


azure-pipelines/integrationtests.yml
```diff
@@ -1,27 +1,23 @@
...
-    demands: 
+    demands:
     - ImageOverride -equals ADS-Linux_Image
 
 steps:
 
   - task: DockerInstaller@0
     displayName: Installing Docker
     inputs:
-      secureFile: 'testsettings.json'
```
Charles-Gagnon, line 11: This isn't a valid property for the task and doesn't seem to be necessary with how our tests are currently running

PR #1518, opened 2022-05-27
[Table Designer] Fix a few computed col issues

caohai:
https://github.com/microsoft/azuredatastudio/issues/19549
1. Add validation for table only containing computed columns
2. Fix type list for computed columns

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
@@ -472,11 +472,11 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
         {
             var table = designer.TableViewModel;
             var errors = new List<TableDesignerIssue>();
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
```diff
@@ -472,11 +472,11 @@ public List<TableDesignerIssue> Run(Dac.TableDesigner designer)
         {
             var table = designer.TableViewModel;
             var errors = new List<TableDesignerIssue>();
-            if (!table.IsEdge && table.Columns.Items.Count == 0)
+            if (!table.IsEdge && table.Columns.Items.Where(c => !c.IsComputed).Count() == 0)
```
caohai, line 475: Not sure if it's better to do the validation here or add a separate rule.

alanrenmsft: i think this is fine

alanrenmsft (approved):

PR #1521, opened 2022-05-27
Add route to GetAzureFunctions request + cleanup

Charles-Gagnon:
Need the route for the stuff I'm working on, so modifying the GetAzureFunctions request to add the route (if it exists)

Also did a number of cleanup things while I was here : 

- Moved helper functions into extension methods
- Fixed name query to handle interpolated strings
- Switched AF tests to use fluent assertions (much easier to read and gives better error messages)

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
@@ -60,11 +60,11 @@ namespace Microsoft.SqlTools.ServiceLayer.AzureFunctions
                 .FirstOrDefault()
                 ?.ChildNodes()
                 .OfType<ExpressionSyntax>() // Find the child identifier node with our value
-                .Where(le => le.Kind() != SyntaxKind.NullLiteralExpression)
+                .Where(le => le.Kind() != SyntaxKind.NullLiteralExpression) // Skip the null expressions so they aren't ToString()'d into "null"
                 .FirstOrDefault()
                 ?.ToString()
-                .TrimStart('$')
-                .Trim('\"'); // Trim off the quotes from the string value
+                .TrimStart('$') // Remove $ from interpolated string, since this will always be outside the quotes we can just trim
+                .Trim('\"'); // Trim off the quotes from the string value - additional quotes at the beginning and end aren't valid syntax so it's fine to trim
         }
 
         /// <summary>
@@ -87,8 +87,8 @@ namespace Microsoft.SqlTools.ServiceLayer.AzureFunctions
                 ?.Arguments
                 .FirstOrDefault() // The first argument is the function name
                 ?.ToString()
-                .TrimStart('$')
-                .Trim('\"') ?? ""; // Trim off the quotes from the string value
+                .TrimStart('$') // Remove $ from interpolated string, since this will always be outside the quotes we can just trim
+                .Trim('\"') ?? ""; // Trim off the quotes from the string value - additional quotes at the beginning and end aren't valid syntax so it's fine to trim
         }
 
         /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
@@ -93,7 +93,7 @@ namespace Microsoft.SqlTools.ServiceLayer.AzureFunctions
 
         /// <summary>
         /// Checks if any of the method declarations have .NET 5 style Azure Function attributes
-        /// .NET 5 AFs use the Function attribute, while .NET 3.1 AFs use FunctionName attritube
+        /// .NET 5 AFs use the Function attribute, while .NET 3.1 AFs use FunctionName attribute
         /// </summary>
         public static bool HasNet5StyleAzureFunctions(IEnumerable<MethodDeclarationSyntax> methodDeclarations)
         {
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
```diff
@@ -31,19 +34,71 @@ public static IEnumerable<MethodDeclarationSyntax> GetMethodsWithFunctionAttribu
...
+                ?.Arguments
+                .FirstOrDefault() // The first argument is the function name
+                ?.ToString()
+                .TrimStart('$')
+                .Trim('\"') ?? ""; // Trim off the quotes from the string value
+        }
+
         /// <summary>
         /// Checks if any of the method declarations have .NET 5 style Azure Function attributes
         /// .NET 5 AFs use the Function attribute, while .NET 3.1 AFs use FunctionName attritube
```
chlafreniere, line 96: can you fix this comment "attritube"

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
```diff
@@ -31,19 +34,71 @@ public static IEnumerable<MethodDeclarationSyntax> GetMethodsWithFunctionAttribu
...
+                    ).FirstOrDefault() // Get the first one available - there should only ever be 0 or 1
+                ).Where(a => // Filter out any that didn't have a FunctionName attribute
+                    a != null
+                ).FirstOrDefault() // Get the first one available - there should only ever be 0 or 1
+                ?.ArgumentList // Get all the arguments for the attribute
+                ?.Arguments
+                .FirstOrDefault() // The first argument is the function name
+                ?.ToString()
+                .TrimStart('$') // Remove $ from interpolated string, since this will always be outside the quotes we can just trim
+                .Trim('\"') ?? ""; // Trim off the quotes from the string value - additional quotes at the beginning and end aren't valid syntax so it's fine to trim
```
chlafreniere, line 91: Why is this line different than 67 above?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/AzureFunctions/AzureFunctionsServiceTests.cs
```diff
@@ -223,13 +218,39 @@ public void GetAzureFunctionsNet5()
...
+            GetAzureFunctionsResult result = operation.GetAzureFunctions();
+
+            Assert.That(result.AzureFunctions.Length, Is.EqualTo(7));
+            Assert.That(result.AzureFunctions[0].Route, Is.EqualTo("withRoute"));
+            Assert.That(result.AzureFunctions[1].Route, Is.EqualTo("{interpolated}String"));
+            Assert.That(result.AzureFunctions[2].Route, Is.EqualTo("$withDollarSigns$"));
+            Assert.That(result.AzureFunctions[3].Route, Is.EqualTo("withRouteNoSpaces"));
+            Assert.That(result.AzureFunctions[4].Route, Is.EqualTo("withRouteExtraSpaces"));
+            Assert.That(result.AzureFunctions[5].Route, Is.Null, "Route specified as null should be null");
+            Assert.That(result.AzureFunctions[6].Route, Is.Null, "No route specified should be null");
```
chlafreniere, line 253: how about another route with an empty string? is that valid?

Charles-Gagnon: What do you mean "another" route? Or you just mean add another one to test for `Route = ""`?

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
Charles-Gagnon: Because we expect routes to be null a lot of the time and that's valid, so we specifically return undefined in that case. But names should always have a value - even if that value is empty. Theoretically this should never happen, but the compiler doesn't know that so it complains that the value might be null at this point.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/AzureFunctions/AzureFunctionsServiceTests.cs
chlafreniere: Right, `Route = ""`

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
```diff
@@ -31,19 +34,74 @@ public static IEnumerable<MethodDeclarationSyntax> GetMethodsWithFunctionAttribu
...
 
             return methodsWithFunctionAttributes;
         }
 
+        /// <summary>
+        /// Gets the route from an HttpTrigger attribute if specified
+        /// </summary>
+        /// <param name="m">The method</param>
+        /// <returns>The name of the route, or null if no route is specified (or there isn't an HttpTrigger binding)</returns>
+        public static string? GetHttpRoute(this MethodDeclarationSyntax m)
```
kisantia, line 47: this is cool! I didn't know about extension methods before 😮

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
Charles-Gagnon: Extension methods are the best, I wish other languages had them.

PR #1522, opened 2022-05-27
Adding property type and rules to compare properties in Execution Plans 

aasimkhan30:
The rule denotes the better value when 2 similar properties are compared. 
There are 5 types of better values: 
       LowerNumber: Indicates a lower number is better than higher number
        HigherNumber: Indicates a higher number is better than lower number
        True: Indicates value true is better than false
        False: Indicates value false is better than true
        None: Default value. Indicates there is no meaningful way to compare the 2 values


src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -103,6 +104,14 @@ public class ExecutionPlanGraphPropertyBase
         /// </summary>
         /// <value></value>
         public string DisplayValue { get; set; }
+        /// <summary>
+        /// Indicates what kind of value is better amongst 2 values of the same property
+        /// </summary>
+        public BetterValue BetterValue { get; set; }
```
lewis-sanchez, line 110: What does "Better" mean in this context? Does it refer to something being greater/optimal, and is the value referring to the display value or a different value?

aasimkhan30: I think better means more optimal value in this case. The value refers to the value of the property. Display value is just the value with some additional formatting.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ShowPlan/Attributes.cs
```diff
@@ -283,5 +283,32 @@ public bool LongString
         private bool longString = false;
     }
 
+    public sealed class BetterValueAttribute: Attribute
```
lewis-sanchez, line 286: Don't usually comment on whitespace, but I think most would expect a space before the colon here.
`public sealed class BetterValueAttribute : Attribute`

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ShowPlan/Attributes.cs
```diff
@@ -283,5 +283,32 @@ public bool LongString
...
+        public BetterValueAttribute(BetterValue value){
+            this.value = value;
+        }
+
+        public BetterValue Value
+        {
+            get { return this.value; }
+        }
+
+        private BetterValue value;
```
lewis-sanchez, line 303: You can get rid of this field and use the `init` keyword to initialize `Value` in the constructor.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init

lewis-sanchez: The property could be rewritten like this: `public BetterValue Value { get; init; }`

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ShowPlan/Attributes.cs
```diff
@@ -283,5 +283,32 @@ public bool LongString
...
+    }
+
+    public enum BetterValue
+    {
+        LowerNumber = 0,
+        HigherNumber = 1,
+        True = 2,
+        False = 3,
```
lewis-sanchez, line 311: `BetterValue.False` isn't being used at the moment. What's the need for it?

aasimkhan30: If in case a false value is better than the true value. Even though it is not used here some other provider might have use for It in the future.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -189,4 +198,12 @@ public enum BadgeType
         CRITICALWARNING = 1,
         PARALLELISM = 2
     }
+
+    public enum PropertyValueDataType
+    {
+        NUMBER = 0,
```
kburtram, line 204: typically in C# naming convention enum values use PascalCasing (first char cap and subsequent work breaks are cap, rest lower-case), all caps is more C-style.

PR #1523, opened 2022-05-27
Add connection retry logic to integration tests

kburtram:
Some of the failures we're seeing in the lab are due to failed connection attempts.  It seems to be the test that goes down a particular connection path first (i.e. disabling tests causes the next test to fail).  Adding retry logic resolved these issues in ad-hoc lab runs.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
@@ -42,6 +42,7 @@ namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.Migration
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Migration/MigrationServiceTests.cs
```diff
@@ -42,6 +42,7 @@ public async Task TestHandleMigrationAssessmentRequest()
         }
 
         [Test]
+        [Ignore("Disable failing test")]
```
kburtram, line 45: this migration test isn't written correct, as the result return value isn't correctly being captures (and the service call is throwing an exception).  Disabling to let the owner team investigate.

aasimkhan30 (approved):

alanrenmsft (approved):

PR #1524, opened 2022-06-01
Save assessment/ SKU Recommendation reports

neetusingh-ms:
**Changes to assessment:**
      Adding support for auto saving assessment file every time assessment ran. To avoid overwriting issue we are appending timestamp at the end of file. Also, returning full path to saved file for displaying on UI. Tested using integration test. Didn't update assessment test (which requires an actual database name in the input) to avoid breaking integration tests. We need to understand STS build setup for integration tests (keeping it out of scope for this change).

**Changes to SKU Recommendation:**
    Adding a new call to save SKU Recommendation results. To be called using a 'Save Result' button on UX. This button will be enabled after the first 10 mins or if we are using already collected data for recommendation. Using timestamp to avoid overwriting. Returning the file names to be shown on UX (under notifications). 
Tested and updated integration tests. 

We could have worked on merging all target type recommendations into one single report. That would be work on template side. Our first goal is to bring the parity with DMA so keeping it as is for now.
image


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -435,15 +482,17 @@ internal async Task<MigrationAssessmentResult> GetAssessmentItems(string[] conne
             SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath = Path.GetDirectoryName(Logger.LogFileFullPath);
             DmaEngine engine = new DmaEngine(connectionStrings);
             ISqlMigrationAssessmentModel contextualizedAssessmentResult = await engine.GetTargetAssessmentResultsListWithCheck(System.Threading.CancellationToken.None);
-            engine.SaveAssessmentResultsToJson(contextualizedAssessmentResult, false);
+            var assessmentFileName = String.Format("SqlAssessmentReport-{0}.json", DateTime.UtcNow.ToString("yyyyMMddHHmmss"));
```
siyangMicrosoft, line 485: the file name should be provided by customer? if not, actually the NuGet will provide the default name like SqlAssessmentReport-{0}.json

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -402,6 +405,50 @@ public void InitializeService(ServiceHost serviceHost)
             }
         }
 
+        internal async Task HandleSaveSkuRecommendationsRequest(
+            SaveSkuRecommendationsParams parameters,
+            RequestContext<SaveSkuRecommendationsResult> requestContext)
+        {
+            const string SkuRecommendationJsonFileExt = ".json";
+            const string SkuRecommendationHtmlFileExt = ".html";
+            string timeStamp = DateTime.UtcNow.ToString("yyyyMMddHHmmss");
```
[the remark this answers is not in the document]
neetusingh-ms: we can skip ss but mm we need in case customer wants to save multiple sku recommendation files while data collection is still going on. And customer can do this if they want to see how (or if) recommendation varies with time.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
neetusingh-ms: yeah filename is very generic. ex - SqlAssessmentReport-20220602
should we add instance name etc? let's discuss offline

siyangMicrosoft (approved):
The PR looks good

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
raymondtruong, line 414: is `CultureInfo.InvariantCulture` needed here?

neetusingh-ms: always good to have.

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/GetSkuRecommendationsRequest.cs
```diff
@@ -96,4 +96,42 @@ public static readonly
...
+        /// </summary>
+        public SqlInstanceRequirements InstanceRequirements { get; set; }
+    }
+
+    public class SaveSkuRecommendationsResult
+    {
+        /// <summary>
+        /// Full file names where SKU Recommendation result is saved
+        /// </summary>
+        public string[] SkuRecommendationFiles { get; set; }
```
raymondtruong, line 127: maybe rename to `SkuRecommendationReportFileNames` -- also, not sure if there's any difference between using `string[]` and `List<string>`, but seems most of the other contracts use List so maybe use that to be consistent?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -402,6 +405,50 @@ public void InitializeService(ServiceHost serviceHost)
...
+                ExportRecommendationResultsAction.ExportRecommendationResults(reportMI, SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath, false, recommendationMIFileName);
+                skuRecommendationFileNames[0] = recommendationMIFileName + SkuRecommendationJsonFileExt;
+                skuRecommendationFileNames[1] = recommendationMIFileName + SkuRecommendationHtmlFileExt;
+
+                SkuRecommendationReport reportVM = new SkuRecommendationReport(
+                    new Dictionary<SqlInstanceRequirements, List<SkuRecommendationResult>> { [parameters.InstanceRequirements] = parameters.SqlVmRecommendationResults },
+                    AzureSqlTargetPlatform.AzureSqlManagedInstance.ToString());
+                string recommendationVMFileName = String.Format("SkuRecommendation{0}-{1}.json", AzureSqlTargetPlatform.AzureSqlVirtualMachine.ToString(), timeStamp);
+                ExportRecommendationResultsAction.ExportRecommendationResults(reportVM, SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath, false, recommendationVMFileName);
+                skuRecommendationFileNames[0] = recommendationVMFileName + SkuRecommendationJsonFileExt;
```
raymondtruong, line 434: `recommendationVMFileName` already is `SkuRecommendation{0}-{1}.json` -- wouldn't this be adding the file extension a second time?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -402,6 +405,50 @@ public void InitializeService(ServiceHost serviceHost)
...
 
+        internal async Task HandleSaveSkuRecommendationsRequest(
+            SaveSkuRecommendationsParams parameters,
+            RequestContext<SaveSkuRecommendationsResult> requestContext)
+        {
+            const string SkuRecommendationJsonFileExt = ".json";
+            const string SkuRecommendationHtmlFileExt = ".html";
+            string timeStamp = DateTime.UtcNow.ToString("yyyyMMddHHmmss");
+
+            string[] skuRecommendationFileNames = new string[6];
```
raymondtruong, line 416: I see what you're trying to do here but having it as an array of fixed size 6 can be a little confusing... particularly when it's unclear what 6 refers to, and especially further down here when you index into it directly... like `skuRecommendationFileNames[0]`, `skuRecommendationFileNames[1]`, etc... 

Would it be easier to have this as a List instead and just keep using `List.Add()`?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -402,6 +405,50 @@ public void InitializeService(ServiceHost serviceHost)
...
+                skuRecommendationFileNames[0] = recommendationMIFileName + SkuRecommendationJsonFileExt;
+                skuRecommendationFileNames[1] = recommendationMIFileName + SkuRecommendationHtmlFileExt;
+
+                SkuRecommendationReport reportVM = new SkuRecommendationReport(
+                    new Dictionary<SqlInstanceRequirements, List<SkuRecommendationResult>> { [parameters.InstanceRequirements] = parameters.SqlVmRecommendationResults },
+                    AzureSqlTargetPlatform.AzureSqlManagedInstance.ToString());
+                string recommendationVMFileName = String.Format("SkuRecommendation{0}-{1}.json", AzureSqlTargetPlatform.AzureSqlVirtualMachine.ToString(), timeStamp);
+                ExportRecommendationResultsAction.ExportRecommendationResults(reportVM, SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath, false, recommendationVMFileName);
+                skuRecommendationFileNames[0] = recommendationVMFileName + SkuRecommendationJsonFileExt;
+                skuRecommendationFileNames[1] = recommendationVMFileName + SkuRecommendationHtmlFileExt;
```
raymondtruong, line 435: I see you store the `skuRecommendationFileNames` in index 0 and 1 for the MI file names above, but then overwrite index 0 and 1 again for the VM file names here? Did you mean to use index 2 and 3 instead? Either way, see my other comment about using List instead of an array, maybe that would simplify things

raymondtruong (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -435,15 +482,17 @@ internal async Task<MigrationAssessmentResult> GetAssessmentItems(string[] conne
             SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath = Path.GetDirectoryName(Logger.LogFileFullPath);
             DmaEngine engine = new DmaEngine(connectionStrings);
             ISqlMigrationAssessmentModel contextualizedAssessmentResult = await engine.GetTargetAssessmentResultsListWithCheck(System.Threading.CancellationToken.None);
-            engine.SaveAssessmentResultsToJson(contextualizedAssessmentResult, false);
+            var assessmentFileName = String.Format("SqlAssessmentReport-{0}.json", DateTime.UtcNow.ToString("yyyyMMddHHmmss"));
+            engine.SaveAssessmentResultsToJson(contextualizedAssessmentResult, false, Path.Combine(SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath, assessmentFileName));
```
raymondtruong, line 486: so wait, we offer two separate functionalities for SKU rec - `GetSkuRecommendationsRequest` to generate the actual recommendation, and then `SaveSkuRecommendationsResultRequest` to save the report. Whereas for assessment we only offer one functionality, `MigrationAssessmentsRequest` which both generates the actual assessment and also saves it all in one. Why is SKU rec split into two calls but assessment is only one? Should it not be consistent for both?  i.e. we either offer a dedicated save report button for both SKU rec and assessment, or we build the save report functionality directly into the call that actually generates the results

raymondtruong (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
siyangMicrosoft: good catch. I think the previous behavior is automatically saving the assessment result in default folder. now we can remove this line.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -402,6 +406,75 @@ public void InitializeService(ServiceHost serviceHost)
...
 
+        internal async Task HandleSaveAssessmentRequest(
+            SaveAssessmentResultParams parameters,
+            RequestContext<SaveAssessmentResult> requestContext)
+        {
+            try
+            {
+                DmaEngine engine = new DmaEngine();
+                var assessmentReportFileName = String.Format("SqlAssessmentReport-{0}.json", DateTime.UtcNow.ToString("yyyyMMddHHmmss", CultureInfo.InvariantCulture));
+                engine.SaveAssessmentResultsToJson(parameters.AssessmentResult, false, Path.Combine(SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath, assessmentReportFileName));
```
siyangMicrosoft, line 417: the folder path should be passed by customer？ not from default folder path?

neetusingh-ms: for first iteration we are using default folder path. may be later we can give option to users to provide folder just for reports.

raymondtruong:
Closing this PR -- see this one instead https://github.com/microsoft/sqltoolsservice/pull/1613

PR #1527, opened 2022-06-02
Add Operations to GetAzureFunctions request

Charles-Gagnon:
Adding operations as well as routes.

Refactored the typing a bit so we can tell when a function has a binding (but not a route/operations specified directly) vs when it doesn't have a binding at all.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
@@ -5,6 +5,7 @@
 
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.SqlTools.ServiceLayer.AzureFunctions.Contracts;
 using System;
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/AzureFunctions/AzureFunctionsUtils.cs
```diff
@@ -39,12 +40,12 @@ public static IEnumerable<MethodDeclarationSyntax> GetMethodsWithFunctionAttribu
             return methodsWithFunctionAttributes;
         }
 
-        /// <summary>
-        /// Gets the route from an HttpTrigger attribute if specified
-        /// </summary>
-        /// <param name="m">The method</param>
-        /// <returns>The name of the route, or null if no route is specified (or there isn't an HttpTrigger binding)</returns>
-        public static string? GetHttpRoute(this MethodDeclarationSyntax m)
+        public static HttpTriggerBinding? GetHttpTriggerBinding(this MethodDeclarationSyntax m)
```
VasuBhog, line 43: Comments explaining this function?

VasuBhog (approved):

barbaravaldez (approved):

PR #1529, opened 2022-06-03
Adding more rules for execution plan comparison properties

aasimkhan30:


src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ShowPlan/PropertyFactory.cs
```diff
@@ -595,6 +595,48 @@ internal class PropertyFactory
...
+        [BetterValue(BetterValue.LowerNumber)]
+        public double WaitTimeMs { get { return 0; } }
+
+        [BetterValue(BetterValue.LowerNumber)]
+        public double WaitCount { get { return 0; } }
+
+        [BetterValue(BetterValue.LowerNumber)]
+        public double CpuTime { get { return 0; } }
+
+        [BetterValue(BetterValue.LowerNumber)]
```
lewis-sanchez, line 637: This is more out of curiosity, but what's the use case for the attributes?

lewis-sanchez (approved):

PR #1530, opened 2022-06-03
Fix backup closing parent connection

Charles-Gagnon:
Partner ran into an issue where they were attempting to backup multiple databases on a single server, but after the first one all the others would fail silently.

This is because currently the logic is disconnecting the connection from the URI of the original connection, NOT the connection that was opened to perform the backup (and in fact that connection was leaking, never properly being closed). It would also have other side effects, since the parent connection may have been from something else like a dashboard and so closing that would break any functionality there. 

Instead we should be closing the new connection opened instead.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
@@ -331,9 +331,9 @@ namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery
                 if (supported && connInfo != null)
                 {
                     DatabaseTaskHelper helper = AdminService.CreateDatabaseTaskHelper(connInfo, databaseExists: true);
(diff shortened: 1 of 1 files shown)


brian-harris (approved):

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -335,14 +334,30 @@ public void InitializeService(IProtocolEndpoint serviceHost)
                     SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo, "Backup");
```
alanrenmsft, line 334: what not wrap this with a using statement?

Charles-Gagnon: Because the usual case for it being closed is the task being completed - which can happen well after this request is handled. (the response is sent as soon as the task is kicked off, not when the task completes)

alanrenmsft (approved):

PR #1534, opened 2022-06-08
vbump STS

alanrenmsft:


Charles-Gagnon (approved):

erpett:
I was under the impression this was a *post* release step

alanrenmsft:
@erpett we noticed some issues with doing it post release, so I will need to work with you to correct it for both STS and ADS. I will reach out to you.

PR #1542, opened 2022-06-16
Sai/dac options map table

ssreerama:
This PR fixes 
1. Removed the approach of duplicating the same logic at two extensions for deployment options and Include objec types
2. Now these options are sending from STS as optionsMapTable-for deployment options, and includeObjectsTable-ObjectType Enum for include objects tab for both SC and SQL DB Project extension. 
3. All options names are coming from the DacFx API from the property_FriendlyName.
4. Tests are fixed accordingly, as optionsMapTable have the updated values and are now sending back to DacFx


ssreerama:
Tests are failing because if needs new DacFx nuget to get the names, but they all are passing on local(as i used the DacFx new dlls). DacFx PR is out and once its merged will updated the Nuget version here.

kisantia:
can you please split this into 2 PRs? 
- one PR for changing the option names to use property_FriendlyName
- one PR for the map changes

ssreerama:
since FriendlyName is being used in Map Table, cannot split it out of maptable, but may be I can try splitting include objects types related changes from map table, I want top use this PR as is for my development and will generate 2 new PRs. Thanks

PR #1550, opened 2022-06-21
OptionsMapTable preparing from DacFx description,value,displayname at…

ssreerama:
…tributes

This PR provides
1. OptionsMapTable {key:DisplayName, value: {value, description, propertyName}} to SQL DB Proj and SC
2. This change helps to remove the duplicate code form the extensions and utilize the direct useful information from the DacFx through the optionsMapTable.
3. Test fixes and added conditions for optionsMapTable values.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
@@ -27,17 +27,17 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         }
 
         /// <summary>
-        /// Default and selected value of the deployment options
+        /// Default and selected value of the deployment option
         /// </summary>
         public T Value { get; set; }
 
         /// <summary>
-        /// Description of the deployment options
+        /// Description of the deployment option
         /// </summary>
         public string Description { get; set; }
 
         /// <summary>
-        /// Display name of the property
+        /// Display name of the deployment option
         /// </summary>
         public string DisplayName { get; set; }
     }
@@ -45,7 +45,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     /// <summary>
     /// Class to define deployment options.
     /// The default property names here should also match the ADS UX
-    /// BooleanOptionsDict will automatically gets the newly added boolean properties from DacFx, All other types should be added here and ADS
+    /// BooleanOptionsDictionary will automatically gets the newly added boolean properties from DacFx, All other types should be added here and ADS
     /// </summary>
     public class DeploymentOptions
     {
@@ -97,16 +97,16 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         );
 
         /// <summary>
-        /// BooleanOptionsDict contains all boolean type deployment options
+        /// BooleanOptionsDictionary contains all boolean type deployment options
         /// </summary>
-        public Dictionary<string, DeploymentOptionProperty<bool>> BooleanOptionsDict { get; set; }
+        public Dictionary<string, DeploymentOptionProperty<bool>> BooleanOptionsDictionary { get; set; }
 
         #endregion
 
         public DeploymentOptions()
         {
             DacDeployOptions options = new DacDeployOptions();
-            BooleanOptionsDict = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
+            BooleanOptionsDictionary = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
 
             // Adding these defaults to ensure behavior similarity with other tools. Dacfx and SSMS import/export wizards use these defaults.
             // Tracking the full fix : https://github.com/microsoft/azuredatastudio/issues/5599
@@ -118,15 +118,15 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
             options.IgnoreKeywordCasing = false;
             options.IgnoreSemicolonBetweenStatements = false;
 
-            // Initializing the default boolean type options to the BooleanOptionsDict
+            // Initializing the default boolean type options to the BooleanOptionsDictionary
             InitializeBooleanTypeOptions(options);
 
-            // Excluding ExcludeObjectTypes to get the STS default values
             // Initialize all non boolean properties
             PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
             foreach (PropertyInfo deployOptionsProp in deploymentOptionsProperties)
             {
-                if (deployOptionsProp.Name != nameof(DeploymentOptions.ExcludeObjectTypes) && deployOptionsProp.Name != nameof(DeploymentOptions.BooleanOptionsDict))
+                // Except ExcludeObjectTypes, as it has some STS defaults which needs to be considered here
+                if (deployOptionsProp.Name != nameof(DeploymentOptions.ExcludeObjectTypes) && deployOptionsProp.Name != nameof(DeploymentOptions.BooleanOptionsDictionary))
                 { 
                     var prop = options.GetType().GetProperty(deployOptionsProp.Name);
                     object setProp = GetDeploymentOptionProp(prop, options);
@@ -137,7 +137,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 
         public DeploymentOptions(DacDeployOptions options)
         {
-            BooleanOptionsDict = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
+            BooleanOptionsDictionary = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
 
             SetOptions(options);
         }
@@ -184,7 +184,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         }
 
         /// <summary>
-        /// Populates BooleanOptionsDict with the boolean type properties
+        /// Populates BooleanOptionsDictionary with the boolean type properties
         /// </summary>
         /// <param name="options"></param>
         public void InitializeBooleanTypeOptions(DacDeployOptions options)
@@ -196,7 +196,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
                 if (prop.PropertyType == typeof(System.Boolean))
                 {
                     object setProp = GetDeploymentOptionProp(prop, options);
-                    this.BooleanOptionsDict[prop.Name] = (DeploymentOptionProperty<bool>)setProp;
+                    this.BooleanOptionsDictionary[prop.Name] = (DeploymentOptionProperty<bool>)setProp;
                 }
             }
         }
@@ -210,7 +210,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         }
 
         /// <summary>
-        /// Preparing all non boolean properties (except BooleanOptionsDict)
+        /// Preparing all non boolean properties (except BooleanOptionsDictionary)
         /// </summary>
         /// <param name="options"></param>
         public void InitializeNonBooleanTypeOptions(DacDeployOptions options)
@@ -219,7 +219,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
             PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
             foreach (PropertyInfo deployOptionsProp in deploymentOptionsProperties)
             {
-                if (deployOptionsProp.Name != nameof(DeploymentOptions.BooleanOptionsDict))
+                if (deployOptionsProp.Name != nameof(DeploymentOptions.BooleanOptionsDictionary))
                 {
                     var prop = options.GetType().GetProperty(deployOptionsProp.Name);
                     object setProp = GetDeploymentOptionProp(prop, options);
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
@@ -44,30 +44,12 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 
     /// <summary>
     /// Class to define deployment options.
-    /// The default property names here should also match the ADS UX
+    /// These property names will be given to the DeploymentOptions interface defined in ADS 'azuredatastudio\extensions\mssql\src\mssql.d.ts' and 'azuredatastudio\extensions\types\vscode-mssql.d.ts'
     /// BooleanOptionsDictionary will automatically gets the newly added boolean properties from DacFx, All other types should be added here and ADS
     /// </summary>
     public class DeploymentOptions
     {
         #region Properties
-
-        // Command timeout to 120 seconds when executing queries against SQL Server.
-        public DeploymentOptionProperty<int> CommandTimeout { get; set; } = new DeploymentOptionProperty<int>(120);
-
-        // LongRunningCommandTimeout 0 seconds to wait indefinitely.
-        public DeploymentOptionProperty<int> LongRunningCommandTimeout { get; set; } = new DeploymentOptionProperty<int>(0);
-
-        // Wait 60 seconds to lock database when executing queries against SQL Server.
-        public DeploymentOptionProperty<int> DatabaseLockTimeout { get; set; } = new DeploymentOptionProperty<int>(60);
-
-        public DeploymentOptionProperty<string> AdditionalDeploymentContributorArguments { get; set; }
-
-        public DeploymentOptionProperty<string> AdditionalDeploymentContributors { get; set; }
-
-        public DeploymentOptionProperty<string> AdditionalDeploymentContributorPaths { get; set; }
-
-        public DeploymentOptionProperty<ObjectType[]> DoNotDropObjectTypes { get; set; }
-
         public DeploymentOptionProperty<ObjectType[]> ExcludeObjectTypes { get; set; } = new DeploymentOptionProperty<ObjectType[]>
         (
             new ObjectType[] {
@@ -99,14 +81,13 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         /// <summary>
         /// BooleanOptionsDictionary contains all boolean type deployment options
         /// </summary>
-        public Dictionary<string, DeploymentOptionProperty<bool>> BooleanOptionsDictionary { get; set; }
+        public Dictionary<string, DeploymentOptionProperty<bool>> BooleanOptionsDictionary { get; set; } = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
 
         #endregion
 
         public DeploymentOptions()
         {
             DacDeployOptions options = new DacDeployOptions();
-            BooleanOptionsDictionary = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
 
             // Adding these defaults to ensure behavior similarity with other tools. Dacfx and SSMS import/export wizards use these defaults.
             // Tracking the full fix : https://github.com/microsoft/azuredatastudio/issues/5599
@@ -119,26 +100,12 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
             options.IgnoreSemicolonBetweenStatements = false;
 
             // Initializing the default boolean type options to the BooleanOptionsDictionary
+            // Not considering DacFx default ExcludeObjectTypes, as it has some STS defaults which needs to be considered here, DacFx defaults are only considered for InitializeFromProfile(), where options are loading from profile
             InitializeBooleanTypeOptions(options);
-
-            // Initialize all non boolean properties
-            PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
-            foreach (PropertyInfo deployOptionsProp in deploymentOptionsProperties)
-            {
-                // Except ExcludeObjectTypes, as it has some STS defaults which needs to be considered here
-                if (deployOptionsProp.Name != nameof(DeploymentOptions.ExcludeObjectTypes) && deployOptionsProp.Name != nameof(DeploymentOptions.BooleanOptionsDictionary))
-                { 
-                    var prop = options.GetType().GetProperty(deployOptionsProp.Name);
-                    object setProp = GetDeploymentOptionProp(prop, options);
-                    deployOptionsProp.SetValue(this, setProp);
-                }
-            }
         }
 
         public DeploymentOptions(DacDeployOptions options)
         {
-            BooleanOptionsDictionary = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
-
             SetOptions(options);
         }
 
@@ -189,11 +156,10 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         /// <param name="options"></param>
         public void InitializeBooleanTypeOptions(DacDeployOptions options)
         {
-            // To fill the options map table directly from the boolean type DacDeployoptions
             PropertyInfo[] dacDeploymentOptionsProperties = options.GetType().GetProperties();
             foreach (PropertyInfo  prop in dacDeploymentOptionsProperties)
             {
-                if (prop.PropertyType == typeof(System.Boolean))
+                if (prop != null && prop.PropertyType == typeof(System.Boolean))
                 {
                     object setProp = GetDeploymentOptionProp(prop, options);
                     this.BooleanOptionsDictionary[prop.Name] = (DeploymentOptionProperty<bool>)setProp;
@@ -203,7 +169,6 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 
         public void SetOptions(DacDeployOptions options)
         {
-            System.Reflection.PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
             // Set the default options properties
             InitializeBooleanTypeOptions(options);
             InitializeNonBooleanTypeOptions(options);
@@ -221,7 +186,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
             {
                 if (deployOptionsProp.Name != nameof(DeploymentOptions.BooleanOptionsDictionary))
                 {
-                    var prop = options.GetType().GetProperty(deployOptionsProp.Name);
+                    PropertyInfo prop = options.GetType().GetProperty(deployOptionsProp.Name);
                     object setProp = GetDeploymentOptionProp(prop, options);
                     deployOptionsProp.SetValue(this, setProp);
                 }
@@ -233,15 +198,17 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         /// </summary>
         /// <param name="prop"></param>
         /// <param name="options"></param>
-        public object GetDeploymentOptionProp(PropertyInfo? prop, DacDeployOptions options)
+        public object GetDeploymentOptionProp(PropertyInfo prop, DacDeployOptions options)
         {
             var val = prop.GetValue(options);
-            DescriptionAttribute descriptionAttribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault(); 
+            DescriptionAttribute descriptionAttribute = prop.GetCustomAttributes<DescriptionAttribute>().FirstOrDefault(); 
             DisplayNameAttribute displayNameAttribute = prop.GetCustomAttributes<DisplayNameAttribute>().FirstOrDefault();
             Type type = val != null ? typeof(DeploymentOptionProperty<>).MakeGenericType(val.GetType())
                 : typeof(DeploymentOptionProperty<>).MakeGenericType(prop.PropertyType);
 
-            object setProp = Activator.CreateInstance(type, val, descriptionAttribute.Description, displayNameAttribute.DisplayName);
+            object setProp = Activator.CreateInstance(type, val, 
+                (descriptionAttribute != null ? descriptionAttribute.Description : ""), 
+                (displayNameAttribute != null ? displayNameAttribute.DisplayName : ""));
             return setProp;
         }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
@@ -18,11 +18,11 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     /// </summary>
     public class DeploymentOptionProperty<T>
     {
-        public DeploymentOptionProperty(T value, string description = "", string propertyName = "")
+        public DeploymentOptionProperty(T value, string description = "", string displayName = "")
         {
             this.Value = value;
             this.Description = description;
-            this.PropertyName = propertyName;
+            this.DisplayName = displayName;
         }
 
         // Default and selected value of the deployment options
@@ -31,15 +31,14 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         // Description of the deployment options
         public string Description { get; set; }
 
-        // Property name which helps to get the selected values for preparing DacDeployOptions model for DacFx
-        public string PropertyName { get; set; }
+        // Display name of the deployment options
+        public string DisplayName { get; set; }
     }
 
     /// <summary>
     /// Class to define deployment options.
-    /// Keeping the order and defaults same as DacFx
-    /// The default values here should also match the default values in ADS UX
-    /// NOTE: When new deployment options are added in DacFx, they need to be added here too
+    /// The default property names here should also match the ADS UX
+    /// OptionsMapTable will automatically gets the newly added boolean properties from DacFx, All other types should be added here and ADS
     /// </summary>
     public class DeploymentOptions
     {
@@ -89,7 +88,10 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
                 ObjectType.AssemblyFiles
             }
         );
-
+        
+        /// <summary>
+        /// OptionsMapTable contains all boolean type deployment options
+        /// </summary>
         public Dictionary<string, DeploymentOptionProperty<bool>> OptionsMapTable { get; set; }
 
         #endregion
@@ -115,7 +117,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
             // Excluding ExcludeObjectTypes to get the STS default values
             // preparing all non boolean properties
             PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
-            foreach (var deployOptionsProp in deploymentOptionsProperties)
+            foreach (PropertyInfo deployOptionsProp in deploymentOptionsProperties)
             {
                 if (deployOptionsProp.Name != "ExcludeObjectTypes" && deployOptionsProp.Name != "OptionsMapTable")
                 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -504,11 +380,18 @@ public void SetOptions(DacDeployOptions options)
         public void SetGenericDeployOptionProps(PropertyInfo prop, DacDeployOptions options, PropertyInfo deployOptionsProp)
         {
             var val = prop.GetValue(options);
-            var attribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
+            var descriptionAttribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
+            var displayNameAttribute = prop.GetCustomAttributes<DisplayNameAttribute>(true).FirstOrDefault();
```
kisantia, line 384: what is the true getting passed in here? Can you add the parameter name?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -504,11 +380,18 @@ public void SetOptions(DacDeployOptions options)
         public void SetGenericDeployOptionProps(PropertyInfo prop, DacDeployOptions options, PropertyInfo deployOptionsProp)
         {
             var val = prop.GetValue(options);
-            var attribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
+            var descriptionAttribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
```
kisantia, line 383: will these description and display name attributes always be strings? If so, then it's clearer to have this be string rather than var

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -504,11 +380,18 @@ public void SetOptions(DacDeployOptions options)
...
+            var displayNameAttribute = prop.GetCustomAttributes<DisplayNameAttribute>(true).FirstOrDefault();
             Type type = val != null ? typeof(DeploymentOptionProperty<>).MakeGenericType(val.GetType()) 
                 : typeof(DeploymentOptionProperty<>).MakeGenericType(prop.PropertyType);
-            object setProp = Activator.CreateInstance(type, val, attribute.Description,_displayNameMapDict[deployOptionsProp.Name]);
+            object setProp = Activator.CreateInstance(type, val, descriptionAttribute.Description, deployOptionsProp.Name);
             deployOptionsProp.SetValue(this, setProp);
+
+            // All boolean options must go into optionsMapTable
```
kisantia, line 390: do non-boolean options not get included because only checkboxes are supported in the options dialog in ADS? Might be good to clarify this in the comment

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -28,6 +28,9 @@ internal static DacDeployOptions CreateSchemaCompareOptions(DeploymentOptions de
             PropertyInfo[] deploymentOptionsProperties = deploymentOptions.GetType().GetProperties();
```
kisantia, line 28: (I couldn't add this comment on an earlier line, but this is meant for line 26) Maybe we should consider renaming this function from `CreateSchemaCompareOptions()` to `CreateDacDeployOptions()` since this isn't used in only schema compare anymore. Doesn't have to be changed in this PR, but just wanted to bring it up since I noticed it :)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -842,24 +848,35 @@ private bool ValidateStreamingJobErrors(ValidateStreamingJobResult expected, Val
         private bool ValidateOptions(DeploymentOptions expected, DeploymentOptions actual)
         {
             System.Reflection.PropertyInfo[] deploymentOptionsProperties = expected.GetType().GetProperties();
+            var optionsMapTable = new Dictionary<string, DeploymentOptionProperty<bool>>();
```
kisantia, line 851: does this need to get initialized here if it gets set to something else on line 874?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1839,6 +1839,14 @@ private void CreateAndValidateScmpFile(SchemaCompareEndpointInfo sourceInfo, Sch
                 ExcludedTargetObjects = null,
             };
 
+            schemaCompareParams.DeploymentOptions.OptionsMapTable["Allow drop blocking assemblies"].Value = true;
+            schemaCompareParams.DeploymentOptions.OptionsMapTable["Drop constraints not in source"].Value = true;
+            schemaCompareParams.DeploymentOptions.OptionsMapTable["Ignore ANSI nulls"].Value = true;
+            schemaCompareParams.DeploymentOptions.OptionsMapTable["No alter statements to change Clr types"].Value = false;
+            schemaCompareParams.DeploymentOptions.OptionsMapTable["Populate files on File groups"].Value = false;
+            schemaCompareParams.DeploymentOptions.OptionsMapTable["Verify deployment"].Value = false;
+            schemaCompareParams.DeploymentOptions.OptionsMapTable["Disable indexes for data phase"].Value = false;
```
kisantia, line 1848: what is setting these options testing?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
```diff
@@ -158,24 +163,65 @@ internal static bool ValidateOptionsEqualsDefault(SchemaCompareOptionsResult opt
...
+                Assert.AreEqual(actualValue, expectedValue, $"Actual Property from Service is not equal to default property for {optionRow.Value.PropertyName}, Actual value: {actualValue} and Default value: {expectedValue}");
+            }
+        }
+
+        /// <summary>
+        /// Verify expected and actual DeploymentOptions OptionsMapTables values
+        /// </summary>
+        /// <param name="expectedOptionsMapTable"></param>
+        /// <param name="actualOptionsMapTable"></param>
+        private static void VerifyExpectedAndActualOptionsMapTables(Dictionary<string, DeploymentOptionProperty<bool>> expectedOptionsMapTable, Dictionary<string, DeploymentOptionProperty<bool>> actualOptionsMapTable)
```
kisantia, line 216: this function looks the same as the one in the DacServiceTests.cs. Can the tests use the same helper function so the code isn't copied in 2 places?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: It doesn't matter what we pass here true/false/nothing, but cannot pass a string type. Will make it empty.

prop.GetCustomAttributes<DescriptionAttribute>().FirstOrDefault()

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: These attributes are not a string types, and defined as Microsoft.SqlServer.Dac.OptionDescriptionAttribute and Microsoft.SqlServer.Dac.OptionDisplayNameAttribute. :)

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
ssreerama: I Agree with you! I'll create a task for this one and make sure to update all required calls changes to generic method names, Thanks :)

ssreerama: https://github.com/microsoft/sqltoolsservice/issues/1554 created on STS github issues board

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
ssreerama: yes, since the validation is taking care outside the for loop and optionsMapTable is collecting from the properties foreach loop, needed an initialization . so that I can do another iterate on optionsMapTable for the validation purpose. Thanks.

ssreerama: Noticed the validation is still inside, will move outside the loop for better complexity!

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
ssreerama: in line 1826, these values are changed to randomly and verifying buy comparing with the actual default options. to make them in consistent, I have added these values! Thanks

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -32,7 +32,7 @@ public DeploymentOptionProperty(T value, string description = "", string display
         public string Description { get; set; }
 
         // To display the options in ADS extensions UI in SchemaCompare/SQL-DB-Project/Dacpac extensions
-        public string DisplayName { get; set; }
+        public string PropertyName { get; set; }
```
kisantia, line 35: why is this getting changed from DisplayName to PropertyName?

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
kisantia: what's the difference between the display name and property name? Can you give an example?

ssreerama: DisplayName: "Restore sequence value" coming from the DacFx_Property_FriendlyName and
PropertName is nothing but a property "RestoreSequenceValue".

kisantia: ok and the PropertyName is what's used to convert from the map value back to the DacDeployOption?

ssreerama:
> Are the test failures because this PR needs a new DacFx preview nuget?

@kisantia , Yes, working on getting the Preview Nuget for the ADS, will updated it today once the Nuget is available. Thanks :)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -480,35 +174,67 @@ public async Task InitializeFromProfile(DacDeployOptions options, string profile
             SetOptions(options);
         }
 
+        /// <summary>
+        /// Populates OptionsMapTable with the boolean type properties
+        /// </summary>
+        /// <param name="options"></param>
+        public void PopulateOptionsMapTable(DacDeployOptions options)
+        {
+            // To fill the options map table directly fomr the boolean type DacDeployoptions
```
kisantia, line 183: typo - "fomr" should be "from"

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -480,35 +174,67 @@ public async Task InitializeFromProfile(DacDeployOptions options, string profile
...
             }
         }
 
         /// <summary>
-        /// Sets the Value and Description to all properties
+        /// Prepares and returns the Value and Description to all properties
         /// </summary>
         /// <param name="prop"></param>
         /// <param name="val"></param>
         /// <param name="deployOptionsProp"></param>
```
kisantia, line 228: these doc comments are out of date and don't match the actual parameters. Also, please add descriptions for these params

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -28,6 +28,9 @@ internal static DacDeployOptions CreateSchemaCompareOptions(DeploymentOptions de
...
 
             DacDeployOptions dacOptions = new DacDeployOptions();
+            var optionsMapTable = new Dictionary<string, DeploymentOptionProperty<bool>>();
+
+            // Get the optionsMapTable property which has the updated option values
             foreach (var deployOptionsProp in deploymentOptionsProperties)
```
kisantia, line 34: it looks like this look only does something if the deployOptionsProp is "ExcludeObjectTypes" or "OptionsMapTable". Does this need to loop through all the other options or can there just be logic for finding these two specific options and doing the actions needed for those two?

ssreerama: In this block, the first foreach loop sets the ExcludeObjectTypes values to the property and fetches the OptionsMapTable and iterate through the boolean type properties inside OptionsMapTable. Since all other properties are not using in the ADS and they can be takes as default values. Thanks

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -480,35 +174,66 @@ public async Task InitializeFromProfile(DacDeployOptions options, string profile
...
+        {
+            // To fill the options map table directly from the boolean type DacDeployoptions
+            PropertyInfo[] dacDeploymentOptionsProperties = options.GetType().GetProperties();
+            foreach (var prop in dacDeploymentOptionsProperties)
+            {
+                if (prop.PropertyType == typeof(System.Boolean))
+                {
+                    object setProp = GetDeploymentOptionProp(prop, options);
+                    var displayNameAttribute = prop.GetCustomAttributes<DisplayNameAttribute>().FirstOrDefault();
+                    this.OptionsMapTable[displayNameAttribute.DisplayName] = (DeploymentOptionProperty<bool>)setProp;
```
Benjin, line 191: Why key off of the display name (which gets localized) rather than the property name (which doesn't)?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
Benjin: Is there a reason DisplayName was selected for the key rather than PropertyName?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: So these displaNames as Keys() are using as labels to dsiplay on ADS options tab. If we set them as one of the properties in value, it would be costlier to get all the names to from a label array object. and on each checkbox onchange, we need to again loop all the values.displayName of optionsMapTable to fetch the right key to update the value.value. 

Using property as key wouldn't change the previous functionality of needing a map table again there for [key: "display name"; value:propertyname] to update the optionsMapTable.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -36,14 +39,27 @@ internal static DacDeployOptions CreateSchemaCompareOptions(DeploymentOptions de
...
 
+                // Exclude the direct properties values and considering the optionsMapTable values
+                if (deployOptionsProp.Name == "OptionsMapTable")
+                {
+                    optionsMapTable = deploymentOptions.OptionsMapTable as Dictionary<string, DeploymentOptionProperty<bool>>;
+                }
+            }
+
+            // Iterating through the updated boolean options coming from the optionsMapTable and assigning them to DacDeployOptions
+            foreach (var deployOptionsProp in optionsMapTable)
```
llali, line 57: please specify the type. try to avoid "var"

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
llali: how do you guarantee that none of the keys won't translate to same value in other languages causing duplicate key error? have you tested it for all languages? using localized value as hash key is risky

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -36,14 +39,27 @@ internal static DacDeployOptions CreateSchemaCompareOptions(DeploymentOptions de
                     var val = deployOptionsProp.GetValue(deploymentOptions);
                     var selectedVal = val.GetType().GetProperty("Value").GetValue(val);
 
-                    // JSON.NET by default reads Number type as Int64, deserializing an object type to dacOptions of Int32 type required to convert into Int32 from Int64.
-                    // If not converted setting value(Int64) to dacOption(Int32) will throw {"Object of type 'System.Int64' cannot be converted to type 'System.Int32'."}.
-                    // As these integer type options are non-editable and are not availbale in ADS to update, integer overflow exception will not be happening here.
-                    if (selectedVal != null && selectedVal.GetType() == typeof(System.Int64))
+                    // Set the excludeObjectTypes values to the DacDeployOptions
+                    if (selectedVal != null && deployOptionsProp.Name == "ExcludeObjectTypes")
```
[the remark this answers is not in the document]
ssreerama: ExcludeObjectTypes and optionsMapTable are two diff properties, when we loop the properties we need to get the excludeObjectTypes value and assign it to dacfx model, but for optionsMapTable, we need to loop it and assign properties inside optionsMapTable  to DacFx model.

So first iteration is excluding the optionsMapTable and collecting it to a local variable. and in second iteration processing the optionsMapTable.

Ill add the comments to the method. Thanks

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Converted to key:property and value has displayName, will test the funcitonality in Spanish machine that i got from Matteo to make sure the localization works as expected. Thanks

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -23,27 +23,49 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
...
+                {
+                    booleanOptionsDict = deploymentOptions.BooleanOptionsDict as Dictionary<string, DeploymentOptionProperty<bool>>;
+                }
+            }
+
+            // Iterating through the updated boolean options coming from the booleanOptionsDict and assigning them to DacDeployOptions
+            foreach (KeyValuePair<string, DeploymentOptionProperty<bool>> deployOptionsProp in booleanOptionsDict)
+            {
+                var prop = dacOptions.GetType().GetProperty(deployOptionsProp.Key);
+                if (prop != null)
```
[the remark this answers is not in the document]
ssreerama: Added a try catch and added exception message to the logger

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -25,122 +26,31 @@ public DeploymentOptionProperty(T value, string description = "", string display
             this.DisplayName = displayName;
         }
 
-        // Default and selected value of the deployment options
+        /// <summary>
+        /// Default and selected value of the deployment options
```
kisantia, line 30: nit: "option" singular since this class represents one deployment option

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -25,122 +26,31 @@ public DeploymentOptionProperty(T value, string description = "", string display
...
 
-        // Description of the deployment options
+        /// <summary>
+        /// Description of the deployment options
+        /// </summary>
         public string Description { get; set; }
 
-        // To display the options in ADS extensions UI in SchemaCompare/SQL-DB-Project/Dacpac extensions
+        /// <summary>
+        /// Display name of the property
```
kisantia, line 40: nit: also use "deployment option" instead of property for consistency

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -244,167 +96,17 @@ public class DeploymentOptions
...
-            _displayNameMapDict[nameof(d.ScriptNewConstraintValidation)] = "Script New Constraint Validation";
-            _displayNameMapDict[nameof(d.ScriptRefreshModule)] = "Script Refresh Module";
-            _displayNameMapDict[nameof(d.SqlCommandVariableValues)] = "Sql Command Variable Values";
-            _displayNameMapDict[nameof(d.TreatVerificationErrorsAsWarnings)] = "Treat Verification Errors As Warnings";
-            _displayNameMapDict[nameof(d.UnmodifiableObjectWarnings)] = "Unmodifiable Object Warnings";
-            _displayNameMapDict[nameof(d.VerifyCollationCompatibility)] = "Verify Collation Compatibility";
-            _displayNameMapDict[nameof(d.VerifyDeployment)] = "Verify Deployment";
-            #endregion
-        }
+        public Dictionary<string, DeploymentOptionProperty<bool>> BooleanOptionsDict { get; set; }
```
kisantia, line 102: nit: update the name to be "BooleanOptionsDictionary" so it's clear that it's a dictionary. No need to abbreviate it

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -416,25 +118,26 @@ public DeploymentOptions()
             options.IgnoreKeywordCasing = false;
             options.IgnoreSemicolonBetweenStatements = false;
 
-            PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
+            // Initializing the default boolean type options to the BooleanOptionsDict
+            InitializeBooleanTypeOptions(options);
 
-            foreach (var deployOptionsProp in deploymentOptionsProperties)
+            // Excluding ExcludeObjectTypes to get the STS default values
```
kisantia, line 124: I'm a little confused with this comment. Can you explain what "Excluding ExcludeObjectTypes to get the STS default values" means? Is it getting the default values for the deployment options that are not ExcludeObjectTypes because ExcludeObjectTypes has different defaults in STS that are not the same as the DacFx defaults?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -664,11 +667,13 @@ public async Task GenerateDeployScriptWithOptions()
                     DatabaseName = targetDb.DatabaseName,
                     DeploymentOptions = new DeploymentOptions()
                     {
-                        DropObjectsNotInSource = new DeploymentOptionProperty<bool>(false),
                         ExcludeObjectTypes = new DeploymentOptionProperty<ObjectType[]>(new[] { ObjectType.Views })
                     }
                 };
 
+                // Updating the BooleanOptionsDict since it has alla the boolean type properties
```
kisantia, line 674: nit: typo - "alla"

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Will update the comment clearly! 

Yes, STS has some default Excluding objects defined with the property declaration. We are considering them by explicitly not going to get the ExcludeObjectTypes from the DacFx API at this code location.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -416,25 +118,26 @@ public DeploymentOptions()
...
+            foreach (PropertyInfo deployOptionsProp in deploymentOptionsProperties)
             {
-                var prop = options.GetType().GetProperty(deployOptionsProp.Name);
-
-                // Note that we set excluded object types here since dacfx has this value as null;
-                if (prop != null && deployOptionsProp.Name != "ExcludeObjectTypes")
-                {
-                    // Setting DacFx default values to the generic deployment options properties.
-                    SetGenericDeployOptionProps(prop, options, deployOptionsProp);
+                // Except ExcludeObjectTypes, as it has some STS defaults which needs to be considered here
```
kisantia, line 128: it would be helpful to add the line where ExcludeObjectTypes gets set in STS to this comment

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -480,35 +183,66 @@ public async Task InitializeFromProfile(DacDeployOptions options, string profile
             SetOptions(options);
         }
 
+        /// <summary>
+        /// Populates BooleanOptionsDictionary with the boolean type properties
+        /// </summary>
+        /// <param name="options"></param>
+        public void InitializeBooleanTypeOptions(DacDeployOptions options)
+        {
+            // To fill the options map table directly from the boolean type DacDeployoptions
```
kisantia, line 192: nit: update "map table" with dictionary

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -480,35 +183,66 @@ public async Task InitializeFromProfile(DacDeployOptions options, string profile
...
         /// <summary>
-        /// Sets the Value and Description to all properties
+        /// Prepares and returns the value and description of a property
         /// </summary>
         /// <param name="prop"></param>
-        /// <param name="val"></param>
-        /// <param name="deployOptionsProp"></param>
-        public void SetGenericDeployOptionProps(PropertyInfo prop, DacDeployOptions options, PropertyInfo deployOptionsProp)
+        /// <param name="options"></param>
+        public object GetDeploymentOptionProp(PropertyInfo? prop, DacDeployOptions options)
```
kisantia, line 236: what's the `?` after `PropertyInfo?` for? Could `prop` be null?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -23,31 +25,62 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
     /// </summary>
     internal static class SchemaCompareUtils
     {
+        /// <summary>
+        /// BooleanOptionsDictionary has all the deployment options and is being processed separately in the second iteration by collecting it in first iteration.
```
kisantia, line 29: This comment is a little confusing. Specific implementation details like this can be in the body of the method itself, next to the related code. For the method summary comment, this can just say that this converts DeploymentOptions used in STS and ADS to DacDeployOptions which can be passed to the in the DacFx apis.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -25,122 +26,31 @@ public DeploymentOptionProperty(T value, string description = "", string display
...
-        public DeploymentOptionProperty<bool> DeployDatabaseInSingleUserMode { get; set; }
-
-        public DeploymentOptionProperty<bool> CreateNewDatabase { get; set; }
-
-        public DeploymentOptionProperty<bool> CompareUsingTargetCollation { get; set; }
-
-        public DeploymentOptionProperty<bool> CommentOutSetVarDeclarations { get; set; }
-
         // Command timeout to 120 seconds when executing queries against SQL Server.
         public DeploymentOptionProperty<int> CommandTimeout { get; set; } = new DeploymentOptionProperty<int>(120);
```
kisantia, line 55: these timeouts and additional deployment contributor options aren't in mssql.d.ts in ADS, probably since they aren't being used there. It would be helpful for there to be a comment in this file why they are here and not in ADS. Or if they aren't needed, remove them for now.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Not expecting null here! but i see intellisense green line warning "'prop' may be null here"! Should i consider them?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
ssreerama: I'll update it here and at below code, Benjin requested to add somewhere how we are handling these boolean and non-boolean options, so this is the place where we prepare the options to DacFx, and thought it would be nice to add here. Anyway the code it self has the comment explaining it.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -150,68 +60,10 @@ public class DeploymentOptions
         // Wait 60 seconds to lock database when executing queries against SQL Server.
         public DeploymentOptionProperty<int> DatabaseLockTimeout { get; set; } = new DeploymentOptionProperty<int>(60);
```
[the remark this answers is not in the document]
ssreerama: Or I would just delete them from here as these are not ever consider and if want to in future, anyway we need to set here or in ADS which needs the code change!

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
Charles-Gagnon: If intellisense is telling you that it can be null that means something upstream could potentially be returning null. 

In this case it's because GetProperty can return null if that property doesn't exist on the type - so you should make sure you're handling that appropriately.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -895,5 +910,21 @@ private void VerifyAndCleanup(string filePath)
...
+        /// <param name="expectedBooleanOptionsDictionary"></param>
+        /// <param name="actualBooleanOptionsDictionary"></param>
+        public void VerifyExpectedAndActualBooleanOptionsDictionary(Dictionary<string, DeploymentOptionProperty<bool>> expectedBooleanOptionsDictionary, Dictionary<string, DeploymentOptionProperty<bool>> actualBooleanOptionsDictionary)
+        {
+            foreach (KeyValuePair<string, DeploymentOptionProperty<bool>> optionRow in expectedBooleanOptionsDictionary)
+            {
+                var expectedValue = optionRow.Value.Value;
+                var actualValue = actualBooleanOptionsDictionary[optionRow.Key].Value;
+
+                Assert.That(actualValue, Is.EqualTo(expectedValue), $"Actual Property from Service is not equal to default property for {optionRow.Key}, Actual value: {actualValue} and Expected value: {expectedValue}");
```
Charles-Gagnon, line 926: The error message will already give the actual and expected value (that's part of the reason why using these assertions is better). So you can simplify this message to just the first part of the sentence.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -480,35 +183,66 @@ public async Task InitializeFromProfile(DacDeployOptions options, string profile
...
-        /// <param name="val"></param>
-        /// <param name="deployOptionsProp"></param>
-        public void SetGenericDeployOptionProps(PropertyInfo prop, DacDeployOptions options, PropertyInfo deployOptionsProp)
+        /// <param name="options"></param>
+        public object GetDeploymentOptionProp(PropertyInfo? prop, DacDeployOptions options)
         {
             var val = prop.GetValue(options);
-            var attribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
-            Type type = val != null ? typeof(DeploymentOptionProperty<>).MakeGenericType(val.GetType()) 
+            DescriptionAttribute descriptionAttribute = prop.GetCustomAttributes<DescriptionAttribute>(true).FirstOrDefault();
```
Charles-Gagnon, line 239: FirstOrDefault can return null in these cases - make sure you're handling that appropriately (even if that's just throwing an exception with more detailed information instead of generic NullReferenceException)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -480,35 +183,66 @@ public async Task InitializeFromProfile(DacDeployOptions options, string profile
...
+                {
+                    object setProp = GetDeploymentOptionProp(prop, options);
+                    this.BooleanOptionsDictionary[prop.Name] = (DeploymentOptionProperty<bool>)setProp;
+                }
+            }
+        }
+
         public void SetOptions(DacDeployOptions options)
         {
             System.Reflection.PropertyInfo[] deploymentOptionsProperties = this.GetType().GetProperties();
```
Charles-Gagnon, line 206: This line isn't being used anymore it seems.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
```diff
@@ -72,7 +72,7 @@ INSERT @returntable
         private DeploymentOptions GetIgnoreColumnOptions()
         {
             var options = new DeploymentOptions();
-            options.IgnoreColumnOrder = new DeploymentOptionProperty<bool>(true);
+            options.BooleanOptionsDictionary["IgnoreColumnOrder"].Value = true;
```
Charles-Gagnon, line 75: For all of these hardcoded strings it'd be preferable to use nameof when possible, for example this one should be able to be `nameof(DacDeployOptions.IgnoreColumnOrder)`

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -25,122 +26,31 @@ public DeploymentOptionProperty(T value, string description = "", string display
...
+        /// </summary>
         public string DisplayName { get; set; }
     }
 
     /// <summary>
     /// Class to define deployment options.
-    /// Keeping the order and defaults same as DacFx
-    /// The default values here should also match the default values in ADS UX
-    /// NOTE: When new deployment options are added in DacFx, they need to be added here too
+    /// The default property names here should also match the ADS UX
```
Charles-Gagnon, line 47: What do you mean "should match"? Is that defined somewhere else? A comment like this makes me think that there's something that needs to be kept in sync, if that's true then provide a link to that (and in the other one a link to here) otherwise reword this to make it clearer what you mean by this.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -416,25 +118,26 @@ public DeploymentOptions()
...
+                    deployOptionsProp.SetValue(this, setProp);
                 }
             }
         }
 
         public DeploymentOptions(DacDeployOptions options)
         {
-            // Setting Display names for all dacDeploy options
-            SetDisplayNameForOption();
+            BooleanOptionsDictionary = new Dictionary<string, DeploymentOptionProperty<bool>>(StringComparer.InvariantCultureIgnoreCase);
```
Charles-Gagnon, line 140: Just define this above with the property so you don't have to declare it in each constructor.

Charles-Gagnon (approved):
There's a lot of stuff here that I'm not the expert on with what properties exist and how they're passed around, but generally not seeing anything big.

I would very strongly suggest adding a README somewhere for the steps to take when new properties need to be added, covering both how the built-in boolean ones are handled and then how to add "custom" props.

kisantia (approved):

llali (approved):

PR #1553, opened 2022-06-21
Include object types getting from DacFx enum

ssreerama:
These changes will avoid the duplication of options on extensions by simply getting the name from the DacFx enum.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
@@ -274,6 +274,8 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 
         private Dictionary<string, string> _displayNameMapDict;
 
+        public Dictionary<string, int> IncludeObjectsTable;
+
         #endregion
 
         /// <summary>
@@ -405,6 +407,8 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 
             // Setting Display names for all dacDeploy options
             SetDisplayNameForOption();
+            // Preparing Include Object types options table
+            CreateIncludeObjectsTable();
 
             // Adding these defaults to ensure behavior similarity with other tools. Dacfx and SSMS import/export wizards use these defaults.
             // Tracking the full fix : https://github.com/microsoft/azuredatastudio/issues/5599
@@ -436,7 +440,20 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
             // Setting Display names for all dacDeploy options
             SetDisplayNameForOption();
 
+            // Preparing Include Object types options table
+            CreateIncludeObjectsTable();
+
             SetOptions(options);
+        }
+
+        /// <summary>
+        /// Sets include objects enum values and number in to the dictionary
+        /// </summary>
+        public void CreateIncludeObjectsTable()
+        {
+            // Set include objects table data
+            var objectTypeEnum = typeof(ObjectType);
+            IncludeObjectsTable = Enum.GetNames(objectTypeEnum).ToDictionary(t => t, t => (int)System.Enum.Parse(objectTypeEnum, t));
         }
 
         /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
@@ -274,7 +274,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 
         private Dictionary<string, string> _displayNameMapDict;
 
-        public Dictionary<string, int> IncludeObjectsTable;
+        public Dictionary<string, int> IncludeObjects;
 
         #endregion
 
@@ -408,7 +408,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
             // Setting Display names for all dacDeploy options
             SetDisplayNameForOption();
             // Preparing Include Object types options table
-            CreateIncludeObjectsTable();
+            GetIncludeObjectTypesDictionary();
 
             // Adding these defaults to ensure behavior similarity with other tools. Dacfx and SSMS import/export wizards use these defaults.
             // Tracking the full fix : https://github.com/microsoft/azuredatastudio/issues/5599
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -436,9 +440,22 @@ public DeploymentOptions(DacDeployOptions options)
...
         }
 
+        /// <summary>
+        /// Sets include objects enum values and number in to the dictionary
+        /// </summary>
+        public void CreateIncludeObjectsTable()
+        {
+            // Set include objects table data
+            var objectTypeEnum = typeof(ObjectType);
+            IncludeObjectsTable = Enum.GetNames(objectTypeEnum).ToDictionary(t => t, t => (int)System.Enum.Parse(objectTypeEnum, t));
```
kisantia, line 456: can you also please add a test that checks for a couple object types that are expected to be here?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -274,6 +274,8 @@ public class DeploymentOptions
 
         private Dictionary<string, string> _displayNameMapDict;
 
+        public Dictionary<string, int> IncludeObjectsTable;
```
kisantia, line 277: it's a little confusing that this has Table in the name but it's a Dictionary. Maybe just call it IncludeObjects?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -436,9 +440,22 @@ public DeploymentOptions(DacDeployOptions options)
...
             SetDisplayNameForOption();
 
+            // Preparing Include Object types options table
+            CreateIncludeObjectsTable();
+
             SetOptions(options);
         }
 
+        /// <summary>
+        /// Sets include objects enum values and number in to the dictionary
```
kisantia, line 450: what are the numbers of these include objects?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
chlafreniere, line 277: does this need to be public?

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -436,9 +440,22 @@ public DeploymentOptions(DacDeployOptions options)
             // Setting Display names for all dacDeploy options
             SetDisplayNameForOption();
```
chlafreniere, line 442: Instead of copying the same two calls in both constructors, having a helper method that gets called in both places will allow for logic to be reused.

Charles-Gagnon: Or even better, one should call the other (less specific calls more specific with appropriate defaults)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Actually its enum names and enum values, ill update the comment, Thanks

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Actually the first method is not more required and already removed in other main PR, this PR will merge later that PR, Thank you :)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Yes this Dictionary will be used in ADS SQLDB/SC extensions

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -436,9 +440,23 @@ public DeploymentOptions(DacDeployOptions options)
...
 
+            // Preparing Include Object types options table
+            GetIncludeObjectTypesDictionary();
+
             SetOptions(options);
         }
 
+        /// <summary>
+        /// Gets include objects enum names and values as a dictionary
+        /// Ex: {key: Aggregate, value: 0}
```
kisantia, line 451: like I mentioned in the other PR, does ADS need to know about the enum number of each object type? We should only need the name in ADS. If it's needed in in STS for converting before calling the DacServices apis, then a name to enum dictionary can be in STS.

kisantia: Now that the other options have to be converted anyway from the map to DacDeployOptions when coming back from ADS, it would make sense to only send the Schema Object names to ADS, then also do the conversion back to the enum values in STS

kisantia: what about the display names of these objects? Multiple word objects have a different display name from their enum name, like "StoredProcedure" has a space in the display name "Stored Procedure."

Right now, the display names are specified in the schema compare extension. Sql database projects will also need these, so it would be nice if the display names are also coming from STS so they aren't copied in 2 extensions.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -436,9 +440,23 @@ public DeploymentOptions(DacDeployOptions options)
...
+        /// </summary>
+        public void GetIncludeObjectTypesDictionary()
+        {
+            // Set include objects table data
+            var objectTypeEnum = typeof(ObjectType);
+            IncludeObjects = Enum.GetNames(objectTypeEnum).ToDictionary(t => t, t => (int)System.Enum.Parse(objectTypeEnum, t));
```
llali, line 457: using enum number is a bad idea. this means you cannot change the order of values in Enum in dacfx otherwise it can break ADs if the number is used in any extension. don't give any information you don't want any extension to use

ssreerama:
This PR was spliced into two PRs and one was merged into main
1. Boolean Option types - (Merged) https://github.com/microsoft/sqltoolsservice/pull/1550
2. Include Object Types - (Open) https://github.com/microsoft/sqltoolsservice/pull/1574

Issue #1558, opened 2022-06-24
Adding a row and reverting it does not decrement the rowID for EditData


smartguest:
This issue means that if a user wants to revert a pending row addition (which is done automatically by ADS), this means that the row index is permanently altered even if the user hasn't actually made an edit, and this makes tracking the rows very difficult to do once a user has made a mistake. More details can be found in the corresponding ADS issue:  https://github.com/microsoft/azuredatastudio/issues/19817

PR #1560, opened 2022-06-24
Fix for rowId not decrementing if reverting an create row

smartguest:
Fixes https://github.com/microsoft/sqltoolsservice/issues/1558 and it's associated issue https://github.com/microsoft/azuredatastudio/issues/19817

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
@@ -381,6 +381,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
                 throw new ArgumentOutOfRangeException(nameof(rowId), SR.EditDataUpdateNotPending);
             }
 
+            // Remove the increment of the NextRow ID if we're reverting a CreateRow, as we are not actually adding a row.
             if (typeof(RowCreate) == removedEdit.GetType())
             {
                 NextRowId--;
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
@@ -380,6 +380,11 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
             {
                 throw new ArgumentOutOfRangeException(nameof(rowId), SR.EditDataUpdateNotPending);
             }
+
+            if (typeof(RowCreate) == removedEdit.GetType())
+            {
+                NextRowId--;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -380,6 +380,11 @@ public void RevertRow(long rowId)
             {
                 throw new ArgumentOutOfRangeException(nameof(rowId), SR.EditDataUpdateNotPending);
             }
+
+            if (typeof(RowCreate) == removedEdit.GetType())
+            {
+                NextRowId--;
```
Charles-Gagnon, line 386: Add a comment explaining why you're doing this.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -380,6 +380,11 @@ public void RevertRow(long rowId)
             {
                 throw new ArgumentOutOfRangeException(nameof(rowId), SR.EditDataUpdateNotPending);
             }
+
+            if (typeof(RowCreate) == removedEdit.GetType())
```
Charles-Gagnon, line 384: This seems like a great opportunity to add some tests - you should be able to set up a unit test that creates a session and then verifies that reverting the row creation results in the expected state (number of rows, row ID, etc.)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/SessionTests.cs
```diff
@@ -606,6 +606,72 @@ public async Task RevertRowSuccess()
...
+
+            // Check that the nextRowId has a value that reflects the number of rows.
+            Assert.AreEqual(rows.Length, s.NextRowId);
+ 
+            // If: I revert the row that has a pending CreateRow (Such as when there's a row failure)
+            s.RevertRow(result1Id);
+
+            rows = await s.GetRows(0, 10);
+
+            // Check that the rowId has decremented after a revert.
```
Charles-Gagnon, line 643: All of these comments on the asserts should also be reflected in the message for the assert - right now if any of these fail it's going to be difficult to figure out exactly where it failed without debugging in it. 

So for example this one should be something like 

`Assert.That(s.NextRowId, Is.EqualTo(addRowNextRowId - 1), "NextRowId should not be incremented after reverting a CreateRow edit");`

See https://github.com/microsoft/azuredatastudio-docs/blob/main/Guides/Testing-Best-Practices.md for some general guidelines to follow (including using the fluent assertion language NUnit provides)

Charles-Gagnon (approved):

PR #1561, opened 2022-06-30
add parse t-sql script handler

alanrenmsft:
vbump DacFx and add parse T-SQL script handler.

I've tested dacfx, schema compare and project build features briefly.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -11,6 +11,7 @@ using Microsoft.SqlTools.ServiceLayer.Connection;
 using Microsoft.SqlTools.ServiceLayer.DacFx.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
 using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using DacTableDesigner = Microsoft.Data.Tools.Sql.DesignServices.TableDesigner.TableDesigner;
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -305,6 +307,14 @@ await requestContext.SendResult(new DacFxOptionsResult()
             }
         }
 
+        public async Task HandleParseTSqlScriptRequest(ParseTSqlScriptRequestParams requestParams, RequestContext<ParseTSqlScriptResult> requestContext)
+        {
+            await requestContext.SendResult(new ParseTSqlScriptResult()
```
kisantia, line 312: should this be in a try just in case `DacTableDesigner.ScriptContainsCreateTableStatements()` throws an error?

kisantia:
@ssreerama fyi Alan is bumping the DacFx version here already, so it won't be needed anymore in https://github.com/microsoft/sqltoolsservice/pull/1550 after this gets merged into main. The only DacFx change that affected options was changing a few options friendly names, so that shouldn't break anything, right?

ssreerama:
@kisantia , you are right! friendly name updates have no dependency here and it won't break or block anything! Thanks

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
alanrenmsft: done. good catch

caohai (approved):

PR #1562, opened 2022-06-30
use file path instead of content in the request

alanrenmsft:
use file path instead of script content to avoid sending large content over JSON RPC channel.


kisantia (approved):
Looks good as long as build error is fixed

PR #1564, opened 2022-07-01
BatchParser target netstandard2.0 and remove Hosting dependency

Charles-Gagnon:
@Matteo-T is looking to use this library but needs it to be targeting netstandard2.0. That by itself is simple enough, but currently there's a dependency on the Hosting library. This has a ton of stuff that ManagedBatchParser shouldn't need (and indeed isn't) using so I decided to remove that dependency.

There were two things being used from that library : the Validate helper methods and the Logging library. 

The Validate methods were easy enough to move over and are simple enough I don't see being worth making a separate library for or something. 

Logging was a bit more complicated - by using the Hosting library we got the whole TraceSource and listener set up for free - that means all of the service apps which used these then had a single source of truth for all logging. But for ManagedBatchParser with the goal being to make it a standalone thing that isn't possible. So at that point I had a couple of options : 

1. Remove the TraceSource and instead just use the raw Trace methods. This is by far the simplest but has the downside of not letting the thing using the library control the events or where they're output to easily. 
2. Make a separate TraceSource for the ManagedBatchParser library - then update the service apps (and whatever else is using the library) to make sure to hook into this TraceSource for whatever it wants. This is probably my "favorite" option but would require some refactoring of the Hosting logger to be able to hook into multiple TraceSources. Not a huge task but not something I wanted to take on right now
3. Make a separate "Microsoft.SqlTools.Logging" library that holds our common logging code to be shared among anything needing it. Not sure how useful this would be though - and if we went this route we should probably be spending the time first to make sure there aren't better libraries or patterns out there we could use (moving to EventSource being one of them)

So I went with 1 since it's the simplest and gets us up and running quickly, then we can work on improvements in this area down the line if we need. The main downside of this approach is the loss of fidelity of events since the Trace library only allows the 3 levels, but given how little this library is used currently that doesn't seem like a horrible tradeoff.


Matteo-T:
I'm ok with option 1.

Matteo-T (approved):
:shipit:

src/Microsoft.SqlTools.ManagedBatchParser/Utility/Validate.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Throws ArgumentException if the value is null or an empty string.
+        /// </summary>
+        /// <param name="parameterName">The name of the parameter being validated.</param>
+        /// <param name="valueToCheck">The value of the parameter being validated.</param>
+        public static void IsNotNullOrEmptyString(string parameterName, string valueToCheck)
+        {
+            if (string.IsNullOrEmpty(valueToCheck))
+            {
+                throw new ArgumentException(
+                    "Parameter contains a null, empty, or whitespace string.",
```
Matteo-T, line 39: >"Parameter contains a null, empty, or whitespace string.", [](http://example.com/codeflow?start=20&length=58)

Just out of curiosity: are there not localizable strings?

PR #1565, opened 2022-07-01
Clean up notebook export code

corivera:
Fixes https://github.com/microsoft/azuredatastudio/issues/14687

I originally was just going to fix that null reference issue above, but as I was testing the notebook export code I saw some opportunities for additional improvements. These mainly deal with the following export scenario.

Take this sample SQL comment:
```
/*
 * A
 * multiline
 * comment
 */
```

Currently, if you convert that SQL query into a notebook and then convert that back into a SQL query, you get the following result:
```
/*
* A

 * multiline

 * comment
*/
```
There are a bunch of unnecessary newlines, and the asterisks in the comments are misaligned. My changes remove those extra markdown newlines, as well as trimming each line of text to get the asterisks to line up on the first column.

Here's a rundown of the other changes:
1. Replaced newlines in the middle of text with Windows-style newlines on Windows, since trying test a sample piece of text would break on Windows because it had both Unix and Windows newlines in the output.
2. Add null argument handling to the notebookToSql and sqlToNotebook conversion methods.
3. Added various unit tests.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/NotebookConvertService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/NotebookConvertService.cs
@@ -201,7 +201,9 @@ namespace Microsoft.SqlTools.ServiceLayer.NotebookConvert
                         commentBlock = commentBlock.Remove(commentBlock.Length - 2);
                     }
 
-                    doc.Cells.Add(GenerateMarkdownCell(commentBlock.Trim()));
+                    // Trim off extra spaces for each line
+                    commentBlock = string.Join("\n", commentBlock.Trim().Split("\n").Select(comment => comment.Trim()));
+                    doc.Cells.Add(GenerateMarkdownCell(commentBlock));
                 }
                 else if (fragment.TokenType == NotebookTokenType.SinglelineComment)
                 {
--- a/src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/NotebookConvertService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/NotebookConvertService.cs
@@ -275,7 +275,7 @@ namespace Microsoft.SqlTools.ServiceLayer.NotebookConvert
                     {
                         // Markdown is text so wrapped in a comment block
                         "markdown" => $@"/*
-{string.Join(Environment.NewLine, cellSource)}
+{string.Join("", cellSource)}
 */",
                         // Everything else (just code blocks for now) is left as is
                         _ => string.Join("", cellSource),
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/NotebookConvertService.cs
```diff
@@ -200,8 +201,13 @@ internal static NotebookDocument ConvertSqlToNotebook(string sql)
                         commentBlock = commentBlock.Remove(commentBlock.Length - 2);
                     }
 
-                    doc.Cells.Add(GenerateMarkdownCell(commentBlock.Trim()));
-                } else if (fragment.TokenType == NotebookTokenType.SinglelineComment)
+                    // Trim off extra spaces for each line. This helps keep comment asterisks aligned on the
```
Charles-Gagnon, line 204: Do you know why the newlines are inserted?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/NotebookConvert/NotebookConvertServiceTests.cs
```diff
@@ -90,16 +87,82 @@ FROM sys.databases
...
+ending single line comment
+*/
+
+/*
+* Ending multiline  
+* comment
+*/";
+            var notebook = JsonConvert.DeserializeObject<NotebookDocument>(sampleNotebook);
+            var query = NotebookConvertService.ConvertNotebookDocToSql(notebook);
+            Assert.AreEqual(expectedSqlQuery, query);
```
Charles-Gagnon, line 158: We should be trying to use NUnit fluent assertions whenever possible `Assert.That(query, Is.EqualTo(expectedSqlQuery))`

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/NotebookConvert/NotebookConvertService.cs
corivera: For which part?

corivera: For which part?

Charles-Gagnon: The issue you called out - where newlines are added when you convert sql -> notebook -> back to sql

corivera: Not sure. It only affects multi-line markdown cells, but we already have newlines included for those in notebooks. We weren't inserting extra newlines for code cells. Seems like the sort of thing that was added as an afterthought. Usually you would need to insert your own newlines when combining an array of text.

corivera: Looks like you were the one who added the notebook conversion code a year ago.

Charles-Gagnon: Nope, this was all pretty quickly done hackathon-type stuff so I'm not surprised there's some issues like that. Not a huge deal, was just curious if this maybe pointed to a bigger issue.

PR #1571, opened 2022-07-11
Adding more top operations columns and fixing some column values

aasimkhan30:
In this PR, I am adding estimated rows, estimated row size and actual row size columns. I am also fixing the value for relative cost percentage column and fixing the name of subtree cost column.


src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
```diff
@@ -8565,11 +8565,11 @@ public static string EstimatedIO
             }
         }
 
-        public static string EstimatedAverageRowSize
+        public static string AverageRowSize
```
lewis-sanchez, line 8568: I think you might want to revert your changes to this file since the header comment states that we shouldn't make changes here:

https://github.com/microsoft/sqltoolsservice/blob/9ffbbe74d9c7238e222a010ec90df50a68a20d3b/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs#L4

aasimkhan30: These are the auto generated files that gets changed whenever we modify sr.strings file. Read the note in this section: https://github.com/microsoft/sqltoolsservice/wiki#building

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -4618,7 +4618,7 @@
  Parameters: 0 - cost (double), 1 - percentage (int) </comment>
   </data>
   <data name="ActualOfEstimated" xml:space="preserve">
-    <value>{0} of
+    <value>{0} of
```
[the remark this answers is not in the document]
aasimkhan30: Yeah, I have removed an extra white space. If you select the 2 lines you will see the difference.

image

lewis-sanchez (approved):

PR #1573, opened 2022-07-12
Add dotnet tool build steps for SQL Tools and Kusto services

corivera:
SQL Tools Service currently takes up almost 2 GB when used with .NET Interactive, since nuget pulls down all the different runtime versions. This change enables the SQL Tools and Kusto services to be built as dotnet tools (since we're guaranteed to have a dotnet runtime in .NET Interactive), which reduces the size of SQL Tools Service to less than 200 MB. I also had to exclude the migration service from the SQL Tools builds, since the associated Migration nuget package is incompatible with arm64 like on M1 Macs. Conveniently this also reduces the size of the final package by about a third, since the Migration service also includes several json files that are 40+ MB in size. Once the Migration team gets us an arm64-compatible build we can look into re-adding the service to this dotnet tool build.


Charles-Gagnon (approved):
How is this included in the product build - or are you going to make a separate pipeline just for the dotnet tool build?

corivera:
The build is controlled by a single environment variable (BUILD_DOTNET_TOOL), so I'll see if I can use the pre-existing pipeline.

corivera:
Waiting on some additional arm64 testing before merging

PR #1581, opened 2022-07-15
A few table designer fixes

caohai:
1. Fix `NoDuplicateConstraintNameRule` by comparing system name instead of display name to handle multiple unnamed constraints.
Before:
  image
After:
image

2. Re-enable index rename support. 
image
image


alanrenmsft:
assume you need a new dacfx version, right? what is systemname

caohai:
> assume you need a new dacfx version, right? what is systemname

Yes, a new DacFx version is needed.
I added `SystemName` to DacFx a while back, it gets underlying system name for unnamed element (something like `DF__mytable2__myvalu__6FB560CC `). For named element, it just gets the regular name like `Name`.

alanrenmsft:
sounds good, I've asked kim to create a new release of dacfx.

alanrenmsft (approved):

PR #1582, opened 2022-07-15
Add build task to create dotnet tools for SQL Tools and Kusto services.

corivera:
Dotnet tools are published on nuget, so we'll need to manually upload these once we add support for them in .NET Interactive. For now the build definition just uploads to our internal mssqltools nuget feed when RELEASE=true is set on the build. I also used the executable names for the package names for simplicity (MicrosoftSqlToolsServiceLayer, MicrosoftKustoServiceLayer), so let me know if I should change the naming,


azure-pipelines/build-and-publish-dotnet-tools.yml
```diff
@@ -0,0 +1,176 @@
+trigger: none
```
Charles-Gagnon, line 1: Is there any reason you have this as a separate pipeline instead of just always generating the tool version for each build? Seems simpler just to do that.

corivera: Currently we build the tool if the BUILD_DOTNET_TOOL environment variable is set, since we need different build properties for the tool vs the standalone application. The tool doesn't have a corresponding exe file in the resulting nupkg, for example, and we also need to exclude the Migration service for now since it's not supported on arm64 yet. I could try looking using build targets instead, or task-scoped variables.

corivera: Didn't know I could set an environment variable on a single build task, so that fixes the whole issue. I'll get rid of the new build yaml.

azure-pipelines/build.yml
```diff
@@ -110,6 +110,15 @@ steps:
     arguments: '-target=dotnetpackpublished -mono'
   continueOnError: true
 
+- task: BatchScript@1
+  displayName: "Package service tool projects"
```
Charles-Gagnon, line 114: `Build and Package service tools projects`

azure-pipelines/build.yml
```diff
@@ -110,6 +110,15 @@ steps:
...
+  env:
+    BUILD_DOTNET_TOOL: "true"
+  inputs:
+    filename: build.cmd
+    arguments: "-target=dotnetpackservicetools -mono"
+  continueOnError: true
```
Charles-Gagnon, line 120: I'd remove this - there shouldn't be any reason we'd expect it to fail unless I'm missing something

Charles-Gagnon (approved):

azure-pipelines/build.yml
corivera: We do the same thing with the existing publish task above this one.

Charles-Gagnon: And I'd suggest that we remove that too, except we'll get rid of that step as a follow up item anyways. We want this to be generated for every build so I'd rather the whole build fail if something breaks (which given that it's just building stuff locally seems...unlikely...to begin with)

PR #1583, opened 2022-07-15
Add 2019 and 2022 to SqlServerType enum

Charles-Gagnon:
At some point it'd be nice to have "minimum supported" versions like SMO does (or at least some additional flag enums such as `2015+`, `2016+`, etc) so we don't have to keep updating the tree node definition file for new versions, but for now just keeping this simple. Verified locally that things still seem to show up as expected. 

I also simplified the template generation logic - I had missed updating the helper method there but it isn't really necessary to have hardcoded checks (the build will fail if a value that doesn't exist is entered into the xml).


Charles-Gagnon:
@nofield FYI

alanrenmsft (approved):

Charles-Gagnon:
Always nice to see tests actually catch issues, turns out I had missed updating a helper function in the template so it didn't properly generate the property supported lookups for the node classes!

PR #1589, opened 2022-07-18
Add separate version numbers for dotnet tool projects.

corivera:
Looks like there's a common build props file in the src folder that was setting an old hard-coded version number for the dotnet tool builds: https://github.com/microsoft/sqltoolsservice/blob/main/src/Directory.Build.props#L8

This PR adds separate version numbers to the dotnet tool projects themselves to override that common version.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
+++ b/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
@@ -22,6 +22,7 @@
 	<Choose>
 		<When Condition="'$(BUILD_DOTNET_TOOL)' == 'true'">
 			<PropertyGroup>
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -22,6 +22,7 @@
 	<Choose>
 		<When Condition="'$(BUILD_DOTNET_TOOL)' == 'true'">
 			<PropertyGroup>
+				<PackageVersion>1.0.0</PackageVersion>
```
Charles-Gagnon, line 25: Sigh, reminded yet again that we still don't have the assembly versions match the zip version. This will do for now but we should really get that fixed at some point (and then use the same version here)

Charles-Gagnon (approved):

PR #1596, opened 2022-07-21
make the message processing async

alanrenmsft:
https://github.com/microsoft/azuredatastudio/issues/20130

there are two reasons that caused the hang described in the issue:
1. in this [PR](https://github.com/microsoft/sqltoolsservice/pull/1486), the wrapper task was removed for some request handlers, it is fine but for these 2 agent request handlers, calling task.Result will never return. changing it to use `await` solves the issue.
2. since we removed the wrapper task, up until the first `await` in the handlers, the code is executed as sync method on the main thread, and caused the message loop being blocked. the fix is to handle this in a centralized location to make sure the message processing is truely async.

also added logging for message processing to make it easier for us to debug the request performance issues.

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
@@ -124,25 +124,28 @@ namespace Microsoft.SqlTools.Hosting.Protocol
                 requestType.MethodName,
                 (requestMessage, messageWriter) =>
                 {
-                    var requestContext =
-                        new RequestContext<TResult>(
-                            requestMessage,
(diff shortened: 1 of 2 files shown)


Charles-Gagnon (approved):
Touching something as critical as the message dispatcher is risky, but right now we could easily have this issue happen elsewhere and not know if we just didn't happen to catch it in our testing. So this seems to be the most complete option. 

But maybe for this release we go with just fixing up the agent call to use await as a more targeted fix (and be keeping an eye out for other similar issues coming in for hotfix potential). It seems like a pretty big edge case right now.

alanrenmsft:
@Charles-Gagnon , yes, this is going to ADS main. for RC1, I am going to have a more targed fix.

PR #1598, opened 2022-07-22
Update DacFx version to get schema compare fix

kisantia:
Updates the DacFx version to [160.6231.0-preview](https://www.nuget.org/packages/Microsoft.SqlServer.DacFx/160.6231.0-preview#release-body-tab), which is a hotfix branch for 160.6208.0-preview with the fix for https://github.com/microsoft/azuredatastudio/issues/20143.

This is a commit only for the release branch since it had an older version of DacFx than main. I'll make a separate update for main with a DacFx preview nuget from main with the bug fix.


alanrenmsft (approved):

kisantia:
the one integration test failure is passing locally for me. It's failed a couple other times in the past week, so doesn't look related to this change

PR #1599, opened 2022-07-23
update DacFx version to get schema compare fix in main

kisantia:
This brings in the fix for #20143 from DacFx main branch. This DacFx perview nuget includes all the changes up to the previous DacFx version in  STS main, 160.6227.0-preview, plus some localization updates and the schema compare bug fix. 

The change made for the release branch was to use the hotfix off of the DacFx 160.6208.0-preview release in https://github.com/microsoft/sqltoolsservice/pull/1598 because it was using an older version of DacFx.


kisantia:
since I'll be out next week, @alanrenmsft or @Benjin can you finish this PR next week and get this updated in ADS?

alanrenmsft (approved):

PR #1601, opened 2022-07-25
Ledger Object Representation in Object Explorer

### LINKED ISSUE #1600: Ledger Object Representation in Object Explorer
Ledger objects should be represented in the Object Explorer by:
- a custom name identifier at the end
  - Tables: <table_name> (Updatable Ledger)
  - Tables: <table_name> (Append-Only Ledger)
  - History Tables: <table_name> (History)
  - Views: <view_name> (Updatable Ledger)
- custom ledger icons
- Nesting the history table as a sub-node of its respective updatable ledger table.

nofield:
This PR fixes #1600, adding support for Ledger objects in their representation in ADS. It includes:
- Adding ledger table, history table, view, and database icons to STS
- Updating ledger object names to indicate their status
- Nesting ledger history tables under their respective updatable ledger tables

For nesting the ledger history table under its updatable ledger table, support for an XML "Or" filter element was necessary, and is included in this PR as well.

Testing:
- Manual testing / visual verification in ADS
- Integration testing of the Object Explorer hierarchy objects.
  - in the Object Explorer tests I also included a code snippet to print the generated output to a bin folder, and display the file path in the test output, to help with debugging differences between the generated and expected output.

Visual:
ADS-OE-annotated


.gitattributes
```diff
@@ -2,4 +2,6 @@
 *.cs diff
 *.csproj diff
 *.sln diff
-*.resx diff
\ No newline at end of file
+*.resx diff
+
+* text=auto
```
Charles-Gagnon, line 7: Is this intended? Please open a separate PR for this if you wanted this changed

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/ChildFactory.cs
```diff
@@ -38,6 +38,11 @@ public abstract class ChildFactory
         /// </summary>
         public abstract IEnumerable<NodeFilter> Filters { get; }
 
+        /// <summary>
+        /// The list of filters that should be applied on the smo object list, concatenated by "or"
+        /// </summary>
+        public abstract IEnumerable<NodeFilter> OrFilters { get; }
```
Charles-Gagnon, line 44: Please open a separate PR for adding this OR support. It will be much simpler to review both PRs if they're separated out.

But also - an OR should just be another filter. It just has two parameters, predicate1 and predicate2 (each filters themselves). So we shouldn't need to be adding a separate list of or filters here since it'll just be added to the list of Filters.

Charles-Gagnon (commented):
It's odd that the AllSqlObjects.sql is being seen as "binary" content : 

image

Are you seeing this as well? What changes did you make to this file?

.gitattributes
nofield: It is- the line ending difference is what was causing me issues with the .resx files (all the strings related files, really). Every time they get regenerated, it's as CRLFs since I'm on Windows, and git was trying to push those as changes. text=auto tells git to check in files with the same line endings as they were checked out with.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/ChildFactory.cs
nofield: How would this look functionally, though? Filters is a list of  NodeFilters, with each their own property, values, type, etc.

I suppose I could add an Or property, which is itself a list of NodeFilters?

nofield:
> 



> It's odd that the AllSqlObjects.sql is being seen as "binary" content :
> 
> image
> 
> Are you seeing this as well? What changes did you make to this file?

I'm seeing that in the PR, yeah. I added two ledger table definitions, one Updatable and one Append-Only.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/ChildFactory.cs
Charles-Gagnon: Let's get this in a separate PR and we can discuss there - it might be that the current design makes what I'm thinking harder than is worth it. I'll take a look and see what I can come up with.

PR #1603, opened 2022-07-27
Add create model request

barbaravaldez:
Add create model request to generate a model from sql files. If there are no files provided, it will create an empty model instead.
Finally, we will store the project in memory under the project uri.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
@@ -1,0 +1,53 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlServer.Dac.Model;
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
+{
+    /// <summary>
+    /// Parameters to generate a SQL model
+    /// </summary>
+    public class GenerateTSqlModelParams
+    {
+        /// <summary>
+        /// Project uri
+        /// </summary>
+        public string ProjectUri { get; set; }
+
+        /// <summary>
+        /// The version of Sql Server to target
+        /// </summary>
+        public string ModelTargetVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Sql script file paths.
+        /// </summary>
+        public string[]? FilePaths { get; set; }
+    }
+
+    /// <summary>
+    /// Result for the GenerTSqlModel Request.
+    /// </summary>
+    public class GenerateTSqlModelResult : ResultStatus
+    {
+        public TSqlModel Model { get; set; }
+
+        public GenerateTSqlModelResult(TSqlModel model)
+        {
+            this.Model = model;
+        }
+    }
+
+    /// <summary>
+    /// Defines the generate sql model request
+    /// </summary>
+    class GenerateTSqlModelRequest
+    {
+        public static readonly RequestType<GenerateTSqlModelParams, GenerateTSqlModelResult> Type =
+            RequestType<GenerateTSqlModelParams, GenerateTSqlModelResult>.Create("dacFx/generateTSqlModel");
+    }
+}
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
@@ -30,7 +30,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     }
 
     /// <summary>
-    /// Result for the GenerTSqlModel Request.
+    /// Result for the GenerateTSqlModel Request.
     /// </summary>
     public class GenerateTSqlModelResult : ResultStatus
     {
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
@@ -4,7 +4,6 @@
 //
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Utility;
-using Microsoft.SqlServer.Dac.Model;
 
 namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
 {
@@ -14,7 +13,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     public class GenerateTSqlModelParams
     {
         /// <summary>
-        /// Project uri
+        /// URI of the project file this model is for
         /// </summary>
         public string ProjectUri { get; set; }
 
@@ -26,20 +25,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         /// <summary>
         /// Gets or sets the Sql script file paths.
         /// </summary>
-        public string[]? FilePaths { get; set; }
-    }
-
-    /// <summary>
-    /// Result for the GenerateTSqlModel Request.
-    /// </summary>
-    public class GenerateTSqlModelResult : ResultStatus
-    {
-        public TSqlModel Model { get; set; }
-
-        public GenerateTSqlModelResult(TSqlModel model)
-        {
-            this.Model = model;
-        }
+        public string[] FilePaths { get; set; }
     }
 
     /// <summary>
@@ -47,7 +33,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
     /// </summary>
     class GenerateTSqlModelRequest
     {
-        public static readonly RequestType<GenerateTSqlModelParams, GenerateTSqlModelResult> Type =
-            RequestType<GenerateTSqlModelParams, GenerateTSqlModelResult>.Create("dacFx/generateTSqlModel");
+        public static readonly RequestType<GenerateTSqlModelParams, ResultStatus> Type =
+            RequestType<GenerateTSqlModelParams, ResultStatus>.Create("dacFx/generateTSqlModel");
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -27,7 +27,9 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
         private static readonly Lazy<DacFxService> instance = new Lazy<DacFxService>(() => new DacFxService());
         private readonly Lazy<ConcurrentDictionary<string, DacFxOperation>> operations =
             new Lazy<ConcurrentDictionary<string, DacFxOperation>>(() => new ConcurrentDictionary<string, DacFxOperation>());
-        // Key for a projectModel is the project uri
+        /// <summary>
+        /// <see cref="Dictionary{String, TSqlModel}"/> that maps project uri to model
+        /// </summary>
         private Lazy<ConcurrentDictionary<string, TSqlModel>> projectModels =
             new Lazy<ConcurrentDictionary<string, TSqlModel>>(() => new ConcurrentDictionary<string, TSqlModel>());
 
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -11,6 +11,7 @@ using Microsoft.SqlTools.ServiceLayer.Connection;
 using Microsoft.SqlTools.ServiceLayer.DacFx.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
 using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlServer.Dac.Model;
 using DacTableDesigner = Microsoft.Data.Tools.Sql.DesignServices.TableDesigner.TableDesigner;
 
 namespace Microsoft.SqlTools.ServiceLayer.DacFx
@@ -25,6 +26,8 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
         private static readonly Lazy<DacFxService> instance = new Lazy<DacFxService>(() => new DacFxService());
         private readonly Lazy<ConcurrentDictionary<string, DacFxOperation>> operations =
             new Lazy<ConcurrentDictionary<string, DacFxOperation>>(() => new ConcurrentDictionary<string, DacFxOperation>());
+        private Lazy<ConcurrentDictionary<string, TSqlModel>> modelResults =
+            new Lazy<ConcurrentDictionary<string, TSqlModel>>(() => new ConcurrentDictionary<string, TSqlModel>());
 
         /// <summary>
         /// Gets the singleton instance object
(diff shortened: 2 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// The version of Sql Server to target
+        /// </summary>
+        public string ModelTargetVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Sql script file paths.
+        /// </summary>
+        public string[]? FilePaths { get; set; }
```
Charles-Gagnon, line 29: Should this be optional? We would expect an array of paths regardless, right? Even if it's just an empty one (although I'm not sure what that would accomplish, although I suppose starting with a "blank" model could be useful for something)

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
+{
+    /// <summary>
+    /// Parameters to generate a SQL model
+    /// </summary>
+    public class GenerateTSqlModelParams
+    {
+        /// <summary>
+        /// Project uri
```
Charles-Gagnon, line 17: Be more descriptive here, `URI of the project file this model is for`

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <summary>
+        /// Gets or sets the Sql script file paths.
+        /// </summary>
+        public string[]? FilePaths { get; set; }
+    }
+
+    /// <summary>
+    /// Result for the GenerateTSqlModel Request.
+    /// </summary>
+    public class GenerateTSqlModelResult : ResultStatus
```
Charles-Gagnon, line 35: Remove the ResultStatus - we don't need to use that.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        public TSqlModel Model { get; set; }
```
Charles-Gagnon, line 37: Can this actually be sent back over the JSON RPC channel? That generally is only able to handle simple POCO-type classes. 

To avoid potential issues we should have our own DTO (Data Transfer Object) defined that we create from the actual TSqlModel when sending it over JSON RPC

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -25,6 +26,8 @@ class DacFxService
         private static readonly Lazy<DacFxService> instance = new Lazy<DacFxService>(() => new DacFxService());
         private readonly Lazy<ConcurrentDictionary<string, DacFxOperation>> operations =
             new Lazy<ConcurrentDictionary<string, DacFxOperation>>(() => new ConcurrentDictionary<string, DacFxOperation>());
+        private Lazy<ConcurrentDictionary<string, TSqlModel>> modelResults =
```
Charles-Gagnon, line 29: `projectModels` would be a better name IMO

And please add a comment explaining what the key value is (the project URI in this case)

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                TSqlModelOptions options = new TSqlModelOptions();
+                SqlServerVersion version = (SqlServerVersion)Enum.Parse(typeof(SqlServerVersion), Parameters.ModelTargetVersion);
+
+                var model = new TSqlModel(version, options);
+                // read all sql files
+                if (Parameters.FilePaths?.Length > 0)
+                {
+                    string[] scripts = new string[Parameters.FilePaths.Length];
+                    for (var i = 0; i < Parameters.FilePaths.Length; i++)
```
Charles-Gagnon, line 43:
```
foreach (string filePath in Parameters.FilePaths)
{
   string fileContent = System.IO.File.ReadAllText(filePath);
   model.AddOrUpdateObjects(fileContent, filePath, null);
}
```

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GenerateTSqlModelRequest.cs
barbaravaldez: @Charles-Gagnon I've been thinking about this and I don't we need to return the model for the create call because we can access it through STS.

barbaravaldez: My plan was to add another request to getObjects from an existing model (returning them over the JSON RPC channel)

Charles-Gagnon: Yeah, I was trying to think if we wanted to do that. 

Pro:
Simplifies this and reduces chance of huge model causing perf issues

Con:
Just adds an extra call that needs to be done (and multiple calls if you need multiple types of objects)

What you suggested seems fine though, neither seem like the obvious choice and this avoids any potential problems with huge models.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
```diff
@@ -37,16 +36,15 @@ public GenerateTSqlModelResult GenerateTSqlModel()
 
                 var model = new TSqlModel(version, options);
                 // read all sql files
-                if (Parameters.FilePaths?.Length > 0)
+                if (Parameters.FilePaths.Length > 0)
```
Charles-Gagnon, line 39: You don't need this check, the foreach will just not do anything if there's no file paths

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -26,7 +27,8 @@ class DacFxService
         private static readonly Lazy<DacFxService> instance = new Lazy<DacFxService>(() => new DacFxService());
         private readonly Lazy<ConcurrentDictionary<string, DacFxOperation>> operations =
             new Lazy<ConcurrentDictionary<string, DacFxOperation>>(() => new ConcurrentDictionary<string, DacFxOperation>());
-        private Lazy<ConcurrentDictionary<string, TSqlModel>> modelResults =
+        // Key for a projectModel is the project uri
```
Charles-Gagnon, line 30: For stuff like this it's preferable to have it be an actual xml doc comment with a description of what it is so it shows up in hover text

Charles-Gagnon:
@llali @Benjin This look good to you?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Generate model from sql files, if no sql files are passed in then it generates an empty model.
+        /// </summary>
+        public TSqlModel GenerateTSqlModel()
+        {
+            try
+            {
+                TSqlModelOptions options = new TSqlModelOptions();
+                SqlServerVersion version = (SqlServerVersion)Enum.Parse(typeof(SqlServerVersion), Parameters.ModelTargetVersion);
```
llali, line 35: you should use TryParse instead and return error if fails

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
```diff
@@ -0,0 +1,53 @@
...
+
+                var model = new TSqlModel(version, options);
+                // read all sql files
+                foreach (string filePath in Parameters.FilePaths)
+                {
+                    string fileContent = System.IO.File.ReadAllText(filePath);
```
llali, line 41: maybe check if file exists before loading?

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
```diff
@@ -0,0 +1,53 @@
...
+            Validate.IsNotNull("parameters", parameters);
+            this.Parameters = parameters;
+        }
+
+        /// <summary>
+        /// Generate model from sql files, if no sql files are passed in then it generates an empty model.
+        /// </summary>
+        public TSqlModel GenerateTSqlModel()
+        {
+            try
```
llali, line 32: I don't see any tests. Are you planning to add any in this PR?

llali:
> @llali @Benjin This look good to you?

looks good but I was expecting to see some test to verify it

llali (changes requested):

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
Charles-Gagnon: Generally that's considered an anti-pattern since the file can be deleted between when you check and when you try to read regardless.

And I think it's reasonable to have this entire thing fail if one of the files doesn't exist. We don't want to just skip those because a file not existing implies there's some other issue and could result in an incorrect model.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
Charles-Gagnon: What are you suggesting here? Because it'll throw an exception regardless. I guess there's some slight benefit to throwing a custom exception here, but given that this is quite the edge case I'm not sure it's worth the extra logic.

llali: for extra logging. If ModelTargetVersion has invalid value

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
llali: I agree to fail all if one does not exist, I think we should check and add the log for debugging

llali: it's up  to you if you want to add extra logging really. it's useful for debugging later. I think this PR is just missing tests. I approve if adding tests later or in separate PR

Charles-Gagnon: I don't think we should get in the habit of adding extra overhead (checking for a file) just for logging. And the error from trying to read a non-existent file is clear enough as to what the issue is so it's not going to get us anything useful. 

If we're concerned about logging errors then we might as well just add logging to RequestContext.SendError so that we get it for all messages automatically.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
Charles-Gagnon: If the version doesn't exist it'll throw an exception with a message similar to this.

`[System.ArgumentException: Requested value 'Invalid' was not found.]`

I don't think we need to be more explicit, it'll be clear enough from the message what the issue was.

Charles-Gagnon: (if this was something that we expected to happen often then I'd definitely be in support of being more defensive here. But an invalid version being passed in should not be something we expect to happen normally so I'm not sure the extra logic here would be worth it)

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
Charles-Gagnon: I'll follow up with adding logging to SendError, that's a good idea and will help out everything in STS.

src/Microsoft.SqlTools.ServiceLayer/DacFx/GenerateTSqlModelOperation.cs
barbaravaldez: I'll add some tests

llali (approved):

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -926,5 +930,102 @@ public void VerifyExpectedAndActualBooleanOptionsDictionary(Dictionary<string, D
...
         }
+
+        /// <summary>
+        /// Verify the generate Tsql model operation
+        /// </summary>
+        [Test]
+        public void GenerateTSqlModelFromSqlFiles()
+        {
+            DacFxService service = new DacFxService();
+            Directory.CreateDirectory(TSqlModelTestFolder);
```
Charles-Gagnon, line 941: I guess this is ok since we're already doing it everywhere, but we should really be trying to use Mocks and dependency injection for stuff like this so we don't have to be writing and reading from disk (which may fail on its own and also ends up leaving a bunch of random stuff laying around on the users machine)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -926,5 +930,102 @@ public void VerifyExpectedAndActualBooleanOptionsDictionary(Dictionary<string, D
...
+                ProjectUri = Path.Join(TSqlModelTestFolder, "test.sqlproj"),
+                ModelTargetVersion = "Sql160",
+                FilePaths = new string[] { }
+            };
+
+            var requestContext = new Mock<RequestContext<ResultStatus>>();
+            requestContext.Setup((RequestContext<ResultStatus> x) => x.SendResult(It.Is<ResultStatus>((result) => result.Success == true))).Returns(Task.FromResult(new object()));
+            await service.HandleGenerateTSqlModelRequest(generateTSqlScriptParams, requestContext.Object);
+
+            Directory.Delete(TSqlModelTestFolder);
```
Charles-Gagnon, line 1026: Cleanup stuff should go in a finally so it happens even if the test fails

(or even consider using the TearDown attribute to create a helper that will clean everything up when the tests finish running)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -926,5 +930,102 @@ public void VerifyExpectedAndActualBooleanOptionsDictionary(Dictionary<string, D
...
+            GenerateTSqlModelOperation op = new GenerateTSqlModelOperation(generateTSqlScriptParams);
+            var model = op.GenerateTSqlModel();
+            var objects = model.GetObjects(DacQueryScopes.UserDefined, ModelSchema.Table).ToList();
+
+            VerifyAndCleanup(sqlTable1DefinitionPath);
+            VerifyAndCleanup(sqlTable2DefinitionPath);
+            Directory.Delete(TSqlModelTestFolder);
+
+            Assert.That(model.Version.ToString(), Is.EqualTo(generateTSqlScriptParams.ModelTargetVersion), $"Model version is not equal to {generateTSqlScriptParams.ModelTargetVersion}");
+            Assert.That(objects, Is.Not.Empty);
```
Charles-Gagnon, line 972: Message

Charles-Gagnon (approved):

barbaravaldez:
@Charles-Gagnon I didn't see your comments and merged the pr, I'll create another PR shortly addressing them

PR #1604, opened 2022-07-27
Use reserved namespace for dotnet tool projects.

corivera:
We have the Microsoft.SqlServer.* namespace already reserved on nuget, so I'm renaming these dotnet tool projects to use that for the new nuget packages.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
+++ b/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
@@ -22,7 +22,7 @@
 	<Choose>
 		<When Condition="'$(BUILD_DOTNET_TOOL)' == 'true'">
 			<PropertyGroup>
-				<PackageId>Microsoft.SqlTools.ServiceLayer.Tool</PackageId>
+				<PackageId>Microsoft.SqlServer.SqlToolsServiceLayer.Tool</PackageId>
 				<PackageVersion>1.0.0</PackageVersion>
 				<PackageDescription>.NET client SQL Tools Service application, usable as a dotnet tool. This package is intended to be used by internal applications only and should not be referenced directly.</PackageDescription>
 				<PackAsTool>true</PackAsTool>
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -22,7 +22,7 @@
 	<Choose>
 		<When Condition="'$(BUILD_DOTNET_TOOL)' == 'true'">
 			<PropertyGroup>
-				<PackageId>Microsoft.SqlTools.ServiceLayer.Tool</PackageId>
+				<PackageId>Microsoft.SqlServer.SqlToolsServiceLayer.Tool</PackageId>
 				<PackageVersion>1.0.0</PackageVersion>
```
shueybubbles, line 26: >>1.0.0</P [](http://example.com/codeflow?start=19&length=9)

can you not get the version from a build-time variable?

shueybubbles (approved):
:shipit:

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
Charles-Gagnon: We could...if it was set up to do that. Right now the version number is only generated during the release job, well after it's done building and packaging everything. 

It's a task we've had on our plate for a long time, and while it would be nice to actually fix it up (since right now our assemblies don't have the correct version either) we can just do this for now and then revisit it based on how often we think we'll need to be doing a new release of the tool (I don't expect very often though given the number of updates we've done over the past year to the current packages)

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
corivera: I think only our automated release pipeline for sql tools service does versioning. I can take a look once I push up the initial version.

Issue #1605, opened 2022-07-27
SQLCMD mode: Properly handle included files when parsing batches


alanrenmsft:
with the attached files, copy the create.sql and insert.sql to a folder and open exec.sql in ADS update the string `c:\test\` to the folder that the files were copied to. and execute the queries in exec.sql with SQLCMD enabled.

in the screenshots below, the line number information is not correct.

![image](https://user-images.githubusercontent.com/13777222/181385310-2628cd00-637b-4d2c-9278-7570777f5d7e.png)

[sqlcmd.zip](https://github.com/microsoft/sqltoolsservice/files/9204034/sqlcmd.zip)

PR #1606, opened 2022-07-27
handle null ref exceptions in managed parser

alanrenmsft:
This PR fixes: https://github.com/microsoft/azuredatastudio/issues/20119
The root cause of the issue is https://github.com/microsoft/sqltoolsservice/issues/1605, I am not familiar with this area to fix the root cause, so I opened an issue for it.

but meanwhile, we can unblock the user by properly handle the null reference exception.

### DIFF
--- a/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
@@ -229,6 +229,7 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
             }
 
             // get number of characters in the last line
+            // for some SQLCMD scenarios, the reader is already at the end of file and causing the prevLine to be null.
             int endColumn = prevLine == null ? 0 : prevLine.ToCharArray().Length;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
```diff
@@ -229,7 +229,7 @@ private static List<int> GetOffsets(string content, IList<BatchInfo> batchInfos)
             }
 
             // get number of characters in the last line
-            int endColumn = prevLine.ToCharArray().Length;
+            int endColumn = prevLine == null ? 0 : prevLine.ToCharArray().Length;
```
Charles-Gagnon, line 232: Update comment explaining what the situation would be where this would be null

Charles-Gagnon (approved):
Would be good to add tests for this while you're at it.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/BatchParserWrapper.cs
alanrenmsft: updated comment

PR #1608, opened 2022-07-28
Or Filtering on Object Explorer Nodes

nofield:
Introduces support for a more robust filtering system in the Object Explorer xml, allowing for "Or"ing different filter properties together for use in URN querying for objects, e.g.:

```
<Filters>
  <Or>
    <Filter A>
    <Filter B>
  </Or>
  <Filter C>
</Filters>
```

Which will produce a URN query like so:

[(@ A or @ B) and (@ C)]

"Or" xml nodes can contain an arbitrary number of Filter nodes, all of which will be concatenated like above.
An INodeFilter interface is introduced, which is inherited by both NodePropertyFilter (regular filters), and NodeOrFilter, allowing arbitrary concatenation of 'or'd and 'and'ed properties.


.gitattributes
```diff
@@ -2,4 +2,6 @@
 *.cs diff
 *.csproj diff
 *.sln diff
-*.resx diff
\ No newline at end of file
+*.resx diff
+
+* text=auto
```
Charles-Gagnon, line 7: Merge in latest from main so this doesn't show up as changed (looks like there's some other stuff showing as changed as well, if you do a merge from main it should fix all those)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            var orNode = new NodeOrFilter {
+                FilterList = new List<NodePropertyFilter> {
+                    TemporalFilter,
+                    LedgerHistoryFilter
+                }
+            };
+
+            string validForBoth = orNode.ToPropertyFilterString(typeof(SqlHistoryTableQuerier), ValidForFlag.Sql2022);
+            string expectedBothFilters = "((@TemporalType = 1) or (@LedgerType = 1))";
+            Assert.AreEqual(expectedBothFilters, validForBoth);
```
Charles-Gagnon, line 69: Asserts should use NUnit fluent assertions (`Assert.That(validForBoth, Is.EqualTo(expectedBothFilters))`) and then also generally include a message indicating what exactly failed (especially when there's multiple asserts in a test).

.gitattributes
Charles-Gagnon: (fixed it for you)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
nofield: Updated accordingly. Previously I was just following coding standards for the rest of the ObjectExplorer unit tests.

Charles-Gagnon: Yeah, unfortunately most of the original tests were written with the base assert methods. The fluent assertions are a lot easier to read and provide better failure messages so we've moved to using those for newer tests but haven't had time to go back and update all the older tests yet.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
```diff
@@ -116,20 +116,96 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
-					WriteLine("        public override IEnumerable<NodeFilter> Filters");
+					WriteLine("        public override IEnumerable<INodeFilter> Filters");
 					WriteLine("        {");
-					WriteLine("           get");
-					WriteLine("           {");
+					WriteLine("            get");
+					WriteLine("            {");
+					WriteLine("                var filters = new List<INodeFilter>();");
+
+					foreach (var orNode in filtersNode.SelectNodes("Or"))
```
Charles-Gagnon, line 136: Please add comments here for this logic explaining what each step is trying to do

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
```diff
@@ -116,20 +116,96 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+						if (or == null)
+						{
+							continue;
+						}
+
+						WriteLine("                filters.Add(new NodeOrFilter");
+						WriteLine("                {");
+						WriteLine("                    FilterList = new List<NodePropertyFilter> {");
+
+						var orFilterNodes = or.GetElementsByTagName("Filter");
```
Charles-Gagnon, line 148: So it looks like this won't be able to handle nested OR nodes? That's fine for now, but if so please make sure to document that in the comments so we know what the expected behavior is.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
```diff
@@ -116,20 +116,96 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+						foreach(var orFilterNode in or.GetElementsByTagName("Filter"))
```
Charles-Gagnon, line 149: This should be `var orFilterNode in orFilterNodes`

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
```diff
@@ -116,20 +116,96 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+						{
+							XmlElement orFilter = orFilterNode as XmlElement;
+							if (orFilter == null)
+							{
+								continue;
+							}
+
+							var propertyName = orFilter.GetAttribute("Property");
```
Charles-Gagnon, line 157: This looks like duplicated code from the below filter logic - would be good to add a helper method that is used for parsing each type of node and call those as appropriate.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -0,0 +1,196 @@
...
+                orNode2
+            };
+
+            string allFiltersValid = INodeFilter.ConcatProperties(nodeList, typeof(SqlHistoryTableQuerier), ValidForFlag.Sql2022);
+            string expectedAllFilters = "[((@TemporalType = 1) or (@LedgerType = 1)) and (@IsSystemObject = 1) and ((@IsSystemObject = 1) or (@LedgerType = 1) or (@TemporalType = 1))]";
+            Assert.That(allFiltersValid, Is.EqualTo(expectedAllFilters), "ConcatProperties did not construct the URN filter string as expected");
+
+            string sql2016ServerVersion = INodeFilter.ConcatProperties(nodeList, typeof(SqlHistoryTableQuerier), ValidForFlag.Sql2016);
+            string expectedSql2016Filters = "[((@TemporalType = 1)) and (@IsSystemObject = 1) and ((@IsSystemObject = 1) or (@TemporalType = 1))]";
+            Assert.AreEqual(expectedSql2016Filters, sql2016ServerVersion);
```
Charles-Gagnon, line 188: `Assert.That(...)`

nofield: Yes, that was below. Just missed deleting the AreEqual.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
nofield: Done

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
nofield: orFilterNodes wasn't really necessary. I've parsed this logic into its own helper function to write the node definition.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
nofield: Done!

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
nofield: I've left a detailed explanation of how the <Filter> node is defined / used in the comments, let me know if there's anything else I should add to that.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodeOrFilter.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+            return canApplyFilter;
+        }
+
+        /// <summary>
+        /// Creates a string representation of a node "or" filter, which is combined in the INodeFilter interface to construct the filter used in the URN to query the SQL objects.
+        /// Example of the output: ((@TableTemporalType = 1) or (@LedgerTableType = 1))
+        /// </summary>
+        /// <returns></returns>
+        public string ToPropertyFilterString(Type type, ValidForFlag validForFlag)
```
Charles-Gagnon, line 41: Can you just use INodeFilter.GetPropertyFilter and have that take in the and/or value to use between filter values? The logic seems to be the same. 

(looks like you'd have to have a value for the types of brackets to use too)

I would suggest probably just making a static helper class at that point to keep all that generic logic in one place.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
```diff
@@ -116,96 +116,110 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 			WriteLine(string.Format("        public override IEnumerable<string> ApplicableParents() {{ return new[] {{ \"{0}\" }}; }}", type));
 
 			List<XmlElement> children = GetChildren(xmlFile, type);
-			List<XmlElement> filters = GetNodeFilters(xmlFile, type);
 			List<XmlElement> smoProperties = GetNodeSmoProperties(xmlFile, type);
 
-			if (filters.Count > 0)
+			// Load and parse Filters node
+			// A <Filters> node is comprised of <Or> and <Filter> nodes
+			//   - A <Filter> node defines the properties to construct a NodePropertyFilter object
```
Charles-Gagnon, line 123: It seems like we should probably rename `Filter` to `PropertyFilter` to make it clear what type of filter it is since we have multiple types right now.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
```diff
@@ -116,96 +116,110 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+						{
+							continue;
+						}
 
+						// Start declaration of Filter node
+						WriteLine("                filters.Add(new NodePropertyFilter");
 
+						// Parse the elements in the <Filter> node into a string, and write
+						string filterNodeString = ParseFilterNode(xmlFile, type, filter);
+						Write(filterNodeString);
```
Charles-Gagnon, line 196: Given that you always write the filter nodes as soon as you parse them I'd say you can just save the extra steps and have the parse method also write it (but probably update to `ParseAndWritePropertyFilterNode`)

Charles-Gagnon (approved):
Don't see any big issues, just a number of cleanup items to take a look at.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/INodeFilter.cs
```diff
@@ -0,0 +1,72 @@
...
+            var list = filters.ToList();
+            for (int i = 0; i < list.Count; i++)
+            {
+                var value = list[i];
+
+                var propertyFilterString = value.ToPropertyFilterString(type, validForFlag);
+                if (propertyFilterString != string.Empty)
+                {
+                    string andPrefix = filter == string.Empty ? string.Empty : "and";
+                    filter = $"{filter} {andPrefix} {propertyFilterString}";
```
alanrenmsft: If I read this correctly, with this method, if we have 4 filter strings, the output will be something like this:
`[{{{{filter1} and {filter2}} and {filter3}} and {filter4}}]`

will `[{filter1} and {filter2} and {filter3} and {filter4}]` work this same way? if so, i'd recommend to use the latter one. easier to read and easier to use StringBuilder.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/INodeFilter.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            if (filter != string.Empty)
+            {
+                filter = $"[{filter}]";
+            }
+            return filter;
+        }
+
+        public static string ConcatProperties(IEnumerable<INodeFilter> filters, Type type, ValidForFlag validForFlag)
```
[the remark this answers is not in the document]
nofield: Yep! Accidentally added this when I was deciding to remove the GetPropertyFilter function from the SmoChildFactoryBase.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/INodeFilter.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// Creates a string from the filter property and values to be used in the Urn to query the SQL objects
+        /// Example of the output:[@ IsSystemObject = 0]
+        /// </summary>
+        /// <returns></returns>
+        string ToPropertyFilterString(Type type, ValidForFlag validForFlag);
+
+        public static string GetPropertyFilter(IEnumerable<INodeFilter> filters, Type type, ValidForFlag validForFlag)
+        {
+            string filter = "";
+            var list = filters.ToList();
```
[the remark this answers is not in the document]
nofield: Oof, yeah. all these string generator functions were really messy. I've gone and cleaned them all up.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/INodeFilter.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <summary>
+        /// Creates a string from the filter property and values to be used in the Urn to query the SQL objects
+        /// Example of the output:[@ IsSystemObject = 0]
+        /// </summary>
+        /// <returns></returns>
+        string ToPropertyFilterString(Type type, ValidForFlag validForFlag);
+
+        public static string GetPropertyFilter(IEnumerable<INodeFilter> filters, Type type, ValidForFlag validForFlag)
+        {
+            string filter = "";
```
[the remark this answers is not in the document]
nofield: Done!

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs
```diff
@@ -55,52 +55,40 @@ public bool CanApplyFilter(Type type, ValidForFlag validForFlag)
...
 
         /// <summary>
-        /// Creates a string from the filter property and values to be used in the Urn to query the SQL objects
-        /// Example of the output:[@ IsSystemObject = 0]
+        /// Creates a string representation of a given node filter, which is combined in the INodeFilter interface to construct the filter used in the URN to query the SQL objects.
+        /// Example of the output: (@ IsSystemObject = 0)
         /// </summary>
         /// <returns></returns>
-        public string ToPropertyFilterString()
+        public string ToPropertyFilterString(Type type, ValidForFlag validForFlag)
```
alanrenmsft, line 62: we already have similar method above, right?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodeOrFilter.cs
```diff
@@ -0,0 +1,63 @@
...
+    public class NodeOrFilter : INodeFilter
+    {
+        /// <summary>
+        /// Filter values
+        /// </summary>
+        public List<NodePropertyFilter> FilterList { get; set; }
+
+        /// <summary>
+        /// Returns true if any of the filters within the FilterList apply to the type and server type.
+        /// Note: Not currently used, because the validity of the filter is essentially only possible while
```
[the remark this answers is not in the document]
nofield: So, how the CanApplyFilter functions are used has changed a bit. Previously, the list of filters would be filtered down based on CanApplyFilter. This approach doesn't work for NodeOrFilter: if some but not all of the filters apply, we still want to generate the logic for the ones that do apply, and just ignore the others.
Now CanApplyFilter is called within NodePropertyFilter's ToPropertyFilterString, before each filter is processed and added to the all-up filter.

That's what I was trying to get at by saying it's "Not currently used".

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
nofield: Done

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs
nofield: Where?

alanrenmsft: https://github.com/microsoft/sqltoolsservice/pull/1608/files#diff-3411bb2896907eccf0500c3b3c71db52d3fb27d69497c5cceb8cb249b9a88bc2R36

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
Charles-Gagnon: On second thought, that can be a follow-up item. Will help avoid making this PR even more messy than it already is.

Charles-Gagnon (commented):
Please also update https://github.com/microsoft/sqltoolsservice/blob/main/docs/guide/object_explorer.md#smotreenodesdefinitionxml with details about the new types of filters supported.

PR #1609, opened 2022-07-28
Add functionality to shutdown Kusto process when parent process exits

corivera:
I noticed when starting the Kusto service from .NET Interactive that the Kusto process was staying open after closing ADS. To fix this I copied over the parent process polling code from the SQL Tools service that lets the service exit when the parent process exits. We already do a similar kind of process polling in ADS in the vscode language client layer, but since Interactive doesn't have a middleman layer like that I added the polling directly to the SQL/Kusto services.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/Program.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/Program.cs
@@ -8,7 +8,6 @@ using Microsoft.SqlTools.ServiceLayer.SqlContext;
 using Microsoft.Kusto.ServiceLayer.Utility;
 using Microsoft.SqlTools.Utility;
 using System.Diagnostics;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Kusto.ServiceLayer/Program.cs
```diff
@@ -60,5 +69,20 @@ internal static void Main(string[] args)
                 Logger.Close();
             }
         }
+
+        private static void CheckParentStatusLoop(Process parent)
```
alanrenmsft, line 73: I'd recommend building the logic directly in the Hosting dll to avoid duplicate code.

alanrenmsft (approved):

Charles-Gagnon (approved):

PR #1610, opened 2022-07-29
Catch Request/Event handler errors at dispatcher level

Charles-Gagnon:
Adding logic to catch any uncaught request/event handler exceptions in the dispatcher. This has a couple of benefits : 

1. Simplifies handler code since we don't have to have a try/catch around every handler now just to send the exception
2. Guarantees that a request doesn't get "stuck" waiting if an exception occurs and the handler wasn't written correctly

There's a lot of changes but the majority of them are just deleting the try/catches in code like this : 

```csharp
try
{
   // do handler logic
}
catch (Exception ex)
{ 
   await request.SendError(ex.ToString());
}
```

### DIFF
--- a/src/Microsoft.SqlTools.Credentials/Credentials/CredentialService.cs
+++ b/src/Microsoft.SqlTools.Credentials/Credentials/CredentialService.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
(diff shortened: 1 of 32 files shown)


alanrenmsft (approved):

Charles-Gagnon:
@kburtram I disabled one of the agent tests because it was failing https://github.com/microsoft/sqltoolsservice/blob/457b6056eb1691f6444a45f3ea3e2866d08fa4c8/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Agent/AgentServiceTests.cs#L47

This is because of a number of issues with the test itself : 

1. The handler is looking at JobName but that's never passed in the params itself (just the ID)
2. Nothing is ever creating the jobs so it's likely there won't even be one on the server unless something else already happened to create it

This was never failing before because the catch was catching the exception being thrown and just sending the error - but given that this test is supposed to actually be verifying the functionality here these changes are just exposing the problems the test had since the beginning that we just never noticed. 

I can file an issue for you if you want to follow up with fixing these to properly test the feature.

PR #1611, opened 2022-07-29
Improvements to GenerateTSqlModel request and tests

barbaravaldez:
Address missed comments from previous PR https://github.com/microsoft/sqltoolsservice/pull/1603

- [X] Change request return type to bool
- [X] Add teardown attribute to tests

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
@@ -17,7 +17,6 @@ using Microsoft.SqlTools.ServiceLayer.DacFx.Contracts;
 using Microsoft.SqlTools.ServiceLayer.IntegrationTests.Utility;
 using Microsoft.SqlTools.ServiceLayer.TaskServices;
 using Microsoft.SqlTools.ServiceLayer.Test.Common;
-using Microsoft.SqlTools.ServiceLayer.Utility;
 using Microsoft.SqlServer.Dac.Model;
 using NUnit.Framework;
 using Moq;
@@ -28,7 +27,6 @@ namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.DacFx
     public class DacFxServiceTests
     {
         private string publishProfileFolder = Path.Combine("..", "..", "..", "DacFx", "PublishProfiles");
-        private string TSqlModelTestFolder = Path.Combine("..", "..", "..", "DacFx", "TSqlModels");
         private const string SourceScript = @"CREATE TABLE [dbo].[table1]
 (
     [ID] INT NOT NULL PRIMARY KEY,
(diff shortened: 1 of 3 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -930,102 +928,109 @@ public void VerifyExpectedAndActualBooleanOptionsDictionary(Dictionary<string, D
...
+        Directory.Delete(TSqlModelTestFolder);
+    }
+
+    /// <summary>
+    /// Verify the generate Tsql model operation
+    /// </summary>
+    [Test]
+    public void GenerateTSqlModelFromSqlFiles()
+    {
+        string sqlTable1DefinitionPath = Path.Join(TSqlModelTestFolder, "table1.sql");
```
Charles-Gagnon, line 959: We should be careful when making statically-named files in a single location. I'd suggest having each test run generate a random folder name (usually using a timestamp is useful for this) instead of always using TSqlModels

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -930,102 +928,109 @@ public void VerifyExpectedAndActualBooleanOptionsDictionary(Dictionary<string, D
...
-            var model = op.GenerateTSqlModel();
-            var objects = model.GetObjects(DacQueryScopes.UserDefined, ModelSchema.Table).ToList();
+        GenerateTSqlModelOperation op = new GenerateTSqlModelOperation(generateTSqlScriptParams);
+        var model = op.GenerateTSqlModel();
+        var objects = model.GetObjects(DacQueryScopes.UserDefined, ModelSchema.Table).ToList();
 
-            VerifyAndCleanup(sqlTable1DefinitionPath);
-            VerifyAndCleanup(sqlTable2DefinitionPath);
-            Directory.Delete(TSqlModelTestFolder);
+        File.Delete(sqlTable1DefinitionPath);
```
Charles-Gagnon, line 984: And instead of having each test have to cleanup (which again can fail and cause false alarms) I'd just give the test files timestamp names and then they'll all be deleted at the end of the test run anyways.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -930,102 +928,109 @@ public void VerifyExpectedAndActualBooleanOptionsDictionary(Dictionary<string, D
...
-            var requestContext = new Mock<RequestContext<ResultStatus>>();
-            requestContext.Setup((RequestContext<ResultStatus> x) => x.SendResult(It.Is<ResultStatus>((result) => result.Success == true))).Returns(Task.FromResult(new object()));
-            await service.HandleGenerateTSqlModelRequest(generateTSqlScriptParams, requestContext.Object);
+        await service.HandleGenerateTSqlModelRequest(generateTSqlScriptParams, requestContext.Object);
 
-            Directory.Delete(TSqlModelTestFolder);
-
-            Assert.True(service.projectModels.Value.Keys.Contains(generateTSqlScriptParams.ProjectUri), "Model was not stored under project uri");
-        }
+        Assert.True(service.projectModels.Value.Keys.Contains(generateTSqlScriptParams.ProjectUri), "Model was not stored under project uri");
```
Charles-Gagnon, line 1034: `Assert.That(service.projectModels.Value.Keys, Contains.Value(generateTSqlScriptParams.ProjectUri), "....")`

Charles-Gagnon (approved):
You should be able to just get rid of the File.Delete calls now too since you're already deleting the entire test folder at the end

PR #1613, opened 2022-07-30
[SQL Migration] Enable saving assessment/recommendation reports

raymondtruong:
In this PR, we:
- modify the `migration/getassessments` request handler to return the file path of the generated assessment report (which defaults to the log folder), so that ADS can expose a button which allows the user to save it to a folder of their choosing 
  - this also fixes an issue where previously, running multiple assessments in a short time (e.g. by going backwards, changing the database list, and assessing again) would fail
- modify the `migration/getskurecommendations` request handler to create recommendation reports, one for each target platform, and also to return the file paths of these reports (which also default to the log folder), so that ADS can expose a button which allows the user to save them to a folder of their choosing

----------

See corresponding PR in azuredatastudio: https://github.com/microsoft/azuredatastudio/pull/20212


alanrenmsft:
I don't think STS is the right place for these html related stuff, is it possible to add these in the extension directly? 


@aasimkhan30  FYI.

raymondtruong:
> @alanrenmsft 
> I don't think STS is the right place for these html related stuff, is it possible to add these in the extension directly?
> 
> @aasimkhan30 FYI.

Yep, currently working on changing the way our NuGet is packed in order to remove the need to have this (or really, any of the metadata, like those 40+ MB json files) checked into sqltoolsservice at all. Still working on it, but see this PR for reference:  https://github.com/microsoft/sqltoolsservice/pull/1617 -- will update this one once that's done.

brian-harris (approved):

PR #1616, opened 2022-08-03
Open Query Plan from XML

lewis-sanchez:
This PR sends information back to Azure Data Studio, so that it can determine if the XML data present in a table cell should be opened using the execution plan viewer.

The added functionality to the SqlToolsService in this PR enables the following in Azure Data Studio: https://github.com/microsoft/azuredatastudio/pull/20243

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/DbCellValue.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/DbCellValue.cs
@@ -23,6 +23,11 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
         public bool IsNull { get; set; }
 
         /// <summary>
+        /// The file extension to be used for any XML cell values representing an execution plan.
+        /// </summary>
+        public string ExecutionPlanFileExtension { get; set; }
+
+        /// <summary>
         /// Culture invariant display value for the cell, this value can later be used by the client to convert back to the original value.
         /// </summary>
         public string InvariantCultureDisplayValue { get; set; }
@@ -51,6 +56,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
             other.IsNull = IsNull;
             other.RawObject = RawObject;
             other.RowId = RowId;
+            other.ExecutionPlanFileExtension = ExecutionPlanFileExtension;
         }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
@@ -212,7 +212,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
         /// <summary>
         /// Returns the file extension for XMl values that represent an execution plan.
         /// </summary>
-        /// <param name="cellValue"></param>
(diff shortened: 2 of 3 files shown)


alanrenmsft:
what is this PR for? which issue?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/ServiceBufferFileStreamReader.cs
```diff
@@ -202,12 +202,36 @@ public IList<DbCellValue> ReadRow(long fileOffset, long rowId, IEnumerable<DbCol
...
+        private string GetDbCellValueExecutionPlanFileExtension(DbCellValue cellValue, DbColumnWrapper column)
+        {
+            if (!column.IsXml && !column.IsJson)
+            {
+                return string.Empty;
+            }
+
+            var planFileExtension = string.Empty;
+
+            if (cellValue.DisplayValue.Contains("ShowPlanXML")) // Should xmlns value be used instead? "http://schemas.microsoft.com/sqlserver/2004/07/showplan"
```
lewis-sanchez, line 227: Question: SQL execution plan XML files contain a root XML node named "ShowPlanXML". Is it okay to use this value or should the XML namespace value be used instead to determine the file extension the Plan Viewer in ADS should use to view the graphical representation of the data?

lewis-sanchez: @alanrenmsft, this is an enhancement we wanted to add to query execution plans in ADS. There's no issue associated with this PR. I can make one if needed.

lewis-sanchez:
@alanrenmsft, this is an enhancement we wanted to add to query execution plans in ADS. There's no issue associated with this PR. I can make one if needed.

alanrenmsft:
we already have similar code in ADS: https://github.com/microsoft/azuredatastudio/blob/67ecd6d3d9b280b90965e600d5fc199b80cfbb4f/src/sql/workbench/contrib/query/browser/gridPanel.ts#L406

lewis-sanchez:
> we already have similar code in ADS: https://github.com/microsoft/azuredatastudio/blob/67ecd6d3d9b280b90965e600d5fc199b80cfbb4f/src/sql/workbench/contrib/query/browser/gridPanel.ts#L406

The existing code only renames the column header from "Microsoft SQL Server 2005 XML Showplan" to a localized version of "XML Showplan".

The added code to the SQL Tools Service will help to provide an file extension value so that ADS is able to open a graphical view of an XML showplan instead of plain XML for a plan: 
image

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/DbCellValue.cs
```diff
@@ -22,6 +22,11 @@ public class DbCellValue
         /// </summary>
         public bool IsNull { get; set; }
 
+        /// <summary>
+        /// The file extension to be used for any XML cell values representing an execution plan.
+        /// </summary>
+        public string ExecutionPlanFileExtension { get; set; }
```
kburtram, line 28: I think this what we're trying to avoid with some of the comments on the ADS PR, of special casing execution plan too much into the contract.  Having a execution plan file extension attached to every cell of data in a resultset doesn't seem like a flexible, generic enough design.  I wonder if there is a way to have something like "result set auxiliary data" that says something like "this column contains a value that should be displayed with this render" and the auxiliary data is only returned in the case some subset of the data needs special handling in the UI.

kburtram (changes requested):
I think we need to think through the DMP contract changes here to not special case execution plan so particularly to each cell value.  Let's sync up to chat about alternatives that send this data in a more extensible way.

PR #1617, opened 2022-08-03
[SQL Migration] Update assessment NuGet version, removing the need to check in content files into STS

raymondtruong:
In this PR, we:
- update the version of the `Microsoft.SqlServer.Migration.Assessment` NuGet package from `1.0.20220527.33` to `1.0.20220804.38`
  - feature-wise, this contains the following changes (among others):
    - add retry mechanism for transient errors during assessment
    - enable xevent assessment
    - consider log storage requirement for MI SKU recommendation  
    - fix DBNull casting which could cause UnhandledSqlException
- remove all the NuGet content files/metadata that used to be checked into `src/Microsoft.SqlTools.ServiceLayer/Migration/Metadata`, including the 40+ MB json files 
  - this is because the latest version of the NuGet utilizes `contentFiles` in the nuspec to define how these files should unpack themselves, removing the need to check them into this repo directly (note the files themselves still live in the nupkg, but heavily compressed -- the whole package is still under 4 MB)

Tested locally and verified with `dotnet publish` that the NuGet is being consumed as intended and that all assessment/SKU recommendation functionality still works properly in ADS.


raymondtruong:
+@Charles-Gagnon, @alanrenmsft, @corivera,

There have been multiple discussions/comments on some of our PRs over the last little while, mainly concerning 1) the unnecessary content files that we have to check in along with our NuGet (like [here](https://github.com/microsoft/sqltoolsservice/pull/1613)) and 2) the size of these files (particularly the 40+ mb json files) jacking up the size of sqltoolsservice. This PR should address all of that without any major fundamental changes. I was able to successfully test the changes locally, but would you guys mind taking a look in case I may have missed something obvious or if I'm doing something wrong here? Thanks!

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -39,10 +39,6 @@
 			</PropertyGroup>
 			<ItemGroup>
 				<PackageReference Include="Microsoft.SqlServer.Migration.Assessment" />
-				<Content Include="./Migration/Metadata/**">
-					<Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
-					<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-				</Content>
 				<Compile Include="**/*.cs" Exclude="**/obj/**/*.cs" />
```
raymondtruong, line 42: +@corivera,

especially concerning this csproj -- all I removed was the Content Include tag, but not sure if anything else was needed? 

(Once we get the NuGet to be arm64 compatible maybe we can revisit the workaround you added here -- but for now just curious about the current way the package is being consumed)

corivera: There's another reference to the migration metadata folder in the integration tests csproj that could be removed: https://github.com/microsoft/sqltoolsservice/blob/main/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Microsoft.SqlTools.ServiceLayer.IntegrationTests.csproj#L27 

You can probably remove that whole Otherwise block, since it only deals with copying the metadata files.

corivera (approved):

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
raymondtruong: Got it, thanks! I'll have a look at that one too. When you say "remove the whole Otherwise block", do you mean the one in the IntegrationTests csproj specifically? Or can I do it in this one too? I notice there's the `<Compile Include="**/*.cs" Exclude="**/obj/**/*.cs" />` line right below my changes, which I didn't want to touch

corivera: Only the IntegrationTests one. The compile elements exclude the migration source files in the dotnet tool build, since it's not supported on arm64 yet.

raymondtruong: @corivera Cool, thanks. Pushed a new iteration to remove the reference from IntegrationTests

corivera (approved):

alanrenmsft (approved):

brian-harris (approved):

PR #1625, opened 2022-08-08
Dropped Ledger Tables Folder

### LINKED ISSUE #1624: Add Dropped Ledger Tables sub-folder to Tables in Object Explorer
[Task 1852883](https://msdata.visualstudio.com/Database%20Systems/_workitems/edit/1852883): Subfolders for Dropped Ledger Tables and Dropped Ledger Views

By design, Ledger tables that are dropped are not removed from the database. Instead, they are renamed, marked as "is_dropped_ledger_table", and marked as read-only. This is a constraint by the hashing functionality of the ledger blockchain which does not allow true "deletion", and also as posterity for any ledger objects inside a specific database. As such, dropped tables need to be presented to users, but not most of the time, so they are filed under a separate folder specifically for dropped tables.

nofield:
This PR resolves #1624 by adding a sub-folder under "Tables" titled "Dropped Ledger Tables" which contains ledger tables that have been dropped.

Testing:
- Manual testing / visual verification in ADS
- Integration testing of the Object Explorer hierarchy objects.

Visual:
image


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodeTypes.cs
```diff
@@ -133,6 +133,7 @@ public enum NodeTypes
         SecurityPolicies,
         DatabaseScopedCredentials,
         ExternalTables,
+        DroppedLedgerTables,
```
alanrenmsft, line 136: why adding here? it should be either alphabetically or add it to the end.

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodeTypes.cs
nofield: I couldn't figure out the ordering from the given file, so I added it next to a similar item. Is there a preference between alphabetical or appending?

alanrenmsft: it seems the values here are not sorted alphabetically, so in general, adding it to the end of the list is safer in case the int value is used somewhere.

PR #1627, opened 2022-08-10
Add modifiable sort priority to nodes

nofield:
Adding support for defining different sort priorities for nodes.

The current relevant scenario is for Dropped Ledger Tables/Views folders, which are expected to be organized at the bottom of their respective Tables and Views folders.
This PR does a few things:
- modifies the SmoTreeNode generator files to parse a SortPriority attribute
- modifies the CompareTo logic to handle placing objects without defined priorities between lowest-priority and highest-priority objects.
- requires sorting of node lists if the list contains either a DroppedLedgerViews or DroppedLedgerTables node.

It then implements these changes for the two custom priority defined nodes.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
@@ -413,23 +413,16 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.Nodes
                 return CompareSamePriorities(this, other);
             }
 
-            if (this.SortPriority.HasValue &&
-                !other.SortPriority.HasValue)
-            {
-                return -1; // this is above other
-            }
-            if (!this.SortPriority.HasValue)
-            {
-                return 1; // this is below other
-            }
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
```diff
@@ -413,23 +413,16 @@ public int CompareTo(TreeNode other)
...
-            if (priDiff < 0)
-                return -1; // this is below other
-            if (priDiff == 0)
-                return 0;
-            return 1;
+            // Higher SortPriority = lower in the list. A couple nodes are defined with SortPriority of Int16.MaxValue
+            // so they're placed at the bottom of the node list (Dropped Ledger Tables and Dropped Ledger Views folders)
+            // Individual objects, like tables and views, don't have a SortPriority defined, so their values need
+            // to be resolved. If a node doesn't have a SortPriority, set it to the second-highest value.
+            int thisPriority = this.SortPriority ?? Int16.MaxValue - 1;
```
alanrenmsft, line 420: why Int16? it should be Int32, right?

alanrenmsft (approved):

PR #1629, opened 2022-08-11
Add a delay to avoid race condition for sql project generate script to new Azure db

kisantia:
This fixes https://github.com/microsoft/azuredatastudio/issues/20133. When generating a script for deploying a dacpac, there can be 2 scripts generated when the target is Azure - one for updating the database and one for operations that need to be done against master. 

As reported in the issue, sometimes only the master script to create the new Azure database and other times one blank editor was opened. When I attached a debugger, both scripts were getting opened in ADS, so it looks like it was a race condition causing both scripts to not get opened. To avoid this, I added a 500ms delay between the two `SqlTask.AddScript()` calls, which is what tells ADS to open the provided scripts.


alanrenmsft:
let's also add logging in ADS as well.

alanrenmsft (approved):

PR #1641, opened 2022-08-16
Defines new endpoint to Assist with Opening Execution Plan XML from DMV in Graphical View.

lewis-sanchez:
This PR adds a new endpoint that Azure Data Studio can use to determine if an XML column value represents an execution plan and returns the appropriate file extension, so that it can render on the front end client correctly.

The Azure Data Studio PR that uses this new endpoint: https://github.com/microsoft/azuredatastudio/pull/20377


kburtram (changes requested):
Let's follow-up on this one.  I think we should either have a generic extensibility API (e.g. file opener handler) or make this an internal implementation detail in the front-end.

PR #1653, opened 2022-08-25
[Table Designer] Add validators for computed col in PK and FK

caohai:
- Add validators for computed col in PK and FK
- PK: pk_col
- FK: fk_col


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
```diff
@@ -664,6 +666,57 @@ public List<TableDesignerIssue> Run(Dac.TableDesigner designer)
...
+            {
+                for (int j = 0; j < table.ForeignKeys.Items[i].Columns.Count; j++)
+                {
+                    var col = table.Columns.Items.First(item => item.Name == table.ForeignKeys.Items[i].Columns[j]);
+                    if (col != null && col.IsComputed && col.IsComputedPersisted != true)
+                    {
+                        errors.Add(new TableDesignerIssue()
+                        {
+                            Description = SR.ComputedColumnNeedToBePersistedInForeignKeyRuleDescription(col.Name, table.ForeignKeys.Items[i].SystemName),
+                            PropertyPath = new object[] { TablePropertyNames.ForeignKeys, i, ForeignKeyPropertyNames.ColumnMapping, j, ForeignKeyColumnMappingPropertyNames.Column }
```
caohai, line 711: this PropertyPath is set to the path of column specification in the FK, I'm not sure if this is better or should I point the user to the `IsPersisited` setting of the column directly.

src/Microsoft.Kusto.ServiceLayer/Localization/sr.cs
```diff
@@ -1,5 +1,5 @@
 // WARNING:
-// This file was generated by the Microsoft DataWarehouse String Resource Tool 5.0.0.0
+// This file was generated by the Microsoft DataWarehouse String Resource Tool 4.0.0.0
```
caohai, line 2: I tried to build with latest `main` and still got this change.

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerValidator.cs
alanrenmsft: either one is fine, I don't have a preference.

src/Microsoft.Kusto.ServiceLayer/Localization/sr.cs
alanrenmsft: manually revert it for now.

alanrenmsft (approved):

smartguest:
@alanrenmsft The reason for the 5.0.0.0 version of the string resource tool is that you still have net 3.1 installed and it is using that version to generate the strings. If you uninstall net 3.1 (which isn't needed anymore), it should work as expected.

PR #1660, opened 2022-08-29
Added error message to validate connect params.

smartguest:
It turns out the error message isn't added to the error message field, which is what ADS uses to get the error message, since this is technically an error message, it should be added to the error message field. 

This fixes: https://github.com/microsoft/azuredatastudio/issues/20111

### DIFF
--- a/external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
+++ b/external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
@@ -220,17 +220,11 @@ namespace Microsoft.SqlTools.CoreServices.Connection
         public ConnectionCompleteParams ValidateConnectParams(ConnectParams connectionParams)
         {
             string paramValidationErrorMessage;
-            if (connectionParams == null)
-            {
-                return new ConnectionCompleteParams
-                {
-                    Messages = SR.ConnectionServiceConnectErrorNullParams
-                };
-            }
             if (!connectionParams.IsValid(out paramValidationErrorMessage))
             {
                 return new ConnectionCompleteParams
                 {
+                    ErrorMessage = paramValidationErrorMessage,
                     OwnerUri = connectionParams.OwnerUri,
                     Messages = paramValidationErrorMessage
                 };
--- a/external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
+++ b/external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
@@ -225,8 +225,7 @@ namespace Microsoft.SqlTools.CoreServices.Connection
                 return new ConnectionCompleteParams
                 {
                     ErrorMessage = paramValidationErrorMessage,
-                    OwnerUri = connectionParams.OwnerUri,
-                    Messages = paramValidationErrorMessage
+                    OwnerUri = connectionParams.OwnerUri
                 };
             }
 
--- a/external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
+++ b/external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
@@ -220,12 +220,19 @@ namespace Microsoft.SqlTools.CoreServices.Connection
         public ConnectionCompleteParams ValidateConnectParams(ConnectParams connectionParams)
         {
             string paramValidationErrorMessage;
+            if (connectionParams == null)
+            {
+                return new ConnectionCompleteParams
+                {
+                    Messages = SR.ConnectionServiceConnectErrorNullParams
+                };
+            }
             if (!connectionParams.IsValid(out paramValidationErrorMessage))
             {
                 return new ConnectionCompleteParams
                 {
-                    ErrorMessage = paramValidationErrorMessage,
-                    OwnerUri = connectionParams.OwnerUri
+                    OwnerUri = connectionParams.OwnerUri,
+                    Messages = paramValidationErrorMessage
                 };
             }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -325,8 +325,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                 return new ConnectionCompleteParams
                 {
                     ErrorMessage = paramValidationErrorMessage,
-                    OwnerUri = connectionParams.OwnerUri,
-                    Messages = paramValidationErrorMessage
+                    OwnerUri = connectionParams.OwnerUri
                 };
             }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -319,7 +319,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         /// null upon validation success</returns>
         public ConnectionCompleteParams ValidateConnectParams(ConnectParams connectionParams)
         {
-            string paramValidationErrorMessage;
             if (connectionParams == null)
             {
                 return new ConnectionCompleteParams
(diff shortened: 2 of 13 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -324,13 +324,15 @@ public ConnectionCompleteParams ValidateConnectParams(ConnectParams connectionPa
             {
                 return new ConnectionCompleteParams
                 {
+                    ErrorMessage = SR.ConnectionServiceConnectErrorNullParams,
```
Charles-Gagnon, line 327: Well this is kind of a mess...

So according to https://github.com/Microsoft/sqlops-dataprotocolclient/blob/c03fa98d19be3c1fba195f861c842f309056b1c0/src/types.ts#L161 (and other matching types like https://github.com/Microsoft/azuredatastudio/blob/466b80fb21351c10632b6e8f5c121cfbe46e1289/src/sql/azdata.d.ts#L415) this is the definition for these two : 

`messages` : `any diagnostic messages return from the service host`
`errorMessage` : `Error message returned from the engine, if any`

Which isn't exactly clear about how these should each be used. Should non-engine errors only go in messages? What are "diagnostic" messages?

Looking in STS it seems like generally ErrorMessage is used to contain just the Message portion of an exception, while ErrorMessage usually contains the full stack trace of the exception. This isn't consistent though...

So as far as this fix goes - what you have is technically correct but we should take the time to figure out exactly what is the correct way to handle these fields and ensure that all our code is using that correctly. And there's two things that'd be good to look into while we're fixing this : 

1. Updating the documentation for the sqlopsclient and related types - they don't seem correct based on how I see the fields from the messages being used
2. Remove the messages field from this error - that shouldn't be necessary to duplicate (and seems like a code smell to me). If it is let's figure out why and whether there's any way we can remove that dependency without breaking anything

Charles-Gagnon: @alanrenmsft If you could take a quick look over the code and see if I'm missing anything too that'd be great

alanrenmsft: I agree about the clean up comments overall.

@smartguest  @cheenamalhotra the string for this key is `Connection parameters cannot be null`, it doesn't tell the user which field is missing, are we ok with it?

cheenamalhotra: That is certainly not sufficient.
We need the username message, which would come from here I think: 
https://github.com/microsoft/sqltoolsservice/blob/ccf95aed77898c0b0dbc804c29a486ce092cd41e/external/Microsoft.SqlTools.CoreServices/Connection/ConnectParamsExtensions.cs#L38-L45

cheenamalhotra: About the `ErrorMessage` field, I think the only other place it's assigned to is here: 

https://github.com/microsoft/sqltoolsservice/blob/ccf95aed77898c0b0dbc804c29a486ce092cd41e/external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs#L533-L551

Which makes this change a reasonable fix.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -320,19 +320,12 @@ internal void UpdateAuthToken(TokenRefreshedParams tokenRefreshedParams)
         public ConnectionCompleteParams ValidateConnectParams(ConnectParams connectionParams)
         {
             string paramValidationErrorMessage;
-            if (connectionParams == null)
```
alanrenmsft, line 323: why removing the null check here? is it never going to be null?

external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
```diff
@@ -220,19 +220,12 @@ public virtual bool TryFindConnection(string ownerUri, out ConnectionInfo connec
         public ConnectionCompleteParams ValidateConnectParams(ConnectParams connectionParams)
         {
             string paramValidationErrorMessage;
-            if (connectionParams == null)
```
alanrenmsft, line 223: is this code being used? I've never touched this library before.

alanrenmsft:
@smartguest assuming you don't have to make changes in the connection management service in ADS core, right?

alanrenmsft (approved):

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -247,6 +247,27 @@ public void CanCancelConnectRequestByDisconnecting()
             Assert.False(disconnectResult);
         }
 
+        /// <summary>
+        /// Verify that we cannot connect to the default database when no username is
+        /// provided as a parameter.
+        /// </summary>
+        [Test]
+        public async Task CantConnectWithoutUsername([Values(null, "")]string databaseName)
```
Charles-Gagnon, line 255: You don't appear to actually be using the databaseName parameter. Copy/paste error?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
smartguest: That is correct, according to the test code, it appears to be a non-nullable type when I attempted to add a null test.

external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
smartguest: Not sure if it's being used, but it appears to be identical to ServiceLayer's function, if those changes need to be reverted, I'll be happy to do so.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: If you're talking about the green squigglies (and compiler warnings) you should be aware that those are purely from a null-checking perspective. I.e. the fact that it's declared as a non-nullable parameter means that the compiler assumes that it's not being passed a null value. 

BUT this is not (yet) guaranteed to actually be the case. This is only enabled as a warning for now because the code base is so massive that trying to fix all these would be a huge task. The warnings are there mostly as a guide for when you're writing new code you use the nullable types correctly (like in typescript with strict mode on). 

In this specific case though I think it's probably fine. I would dearly hope that we never get passed null parameters (and if we did it should be an error farther up the chain) so removing the check here is fine by me.

external/Microsoft.SqlTools.CoreServices/Connection/ConnectionServiceCore.cs
smartguest: Reverted it as it appears to be external and the string localization tool does not recognize it currently.

cheenamalhotra (approved):

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -247,6 +247,27 @@ public void CanCancelConnectRequestByDisconnecting()
             Assert.False(disconnectResult);
         }
 
+        /// <summary>
+        /// Verify that we cannot connect to the default database when no username is
```
Charles-Gagnon, line 251: `when no username is provided and the authentication type is basic auth`

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -247,6 +247,27 @@ public void CanCancelConnectRequestByDisconnecting()
...
+            connectionDetails.UserName = "";
+            var connectionResult = await
+                TestObjects.GetTestConnectionService()
+                .Connect(new ConnectParams()
+                {
+                    OwnerUri = "file:///my/test/file.sql",
+                    Connection = connectionDetails
+                });
+
+            Assert.AreEqual(SR.ConnectionParamsValidateNullSqlAuth("UserName"), connectionResult.ErrorMessage);
```
Charles-Gagnon, line 268: `Assert.That(connectionResult.ErrorMessage, Is.EqualTo(SR.Connection...));`

Using fluent style assertions is easier to read and generally provides better error messages so it's preferred to use them when possible.

Charles-Gagnon (approved):

Charles-Gagnon (changes requested):
Actually looks like a bunch of tests are failing now. Please fix those

Charles-Gagnon (approved):

PR #1664, opened 2022-08-30
Add external projects to srgen

smartguest:
In case any changes are made to the external strings, they will be localized as well now.


build.cake
```diff
@@ -669,7 +674,11 @@ Task("SRGen")
                 var xlfFile = xlfDoc.Files.Single();
 
                 // load a language template
-                var templateFileLocation = System.IO.Path.Combine(locTemplateDir, System.IO.Path.GetFileName(docName) + ".template");
+                var templateDirToUse = locTemplateDir;
+                if(parentDir.Equals("external")){
```
Charles-Gagnon, line 678: Do we actually need this check? The loc folder is the same root folder for both it seems

smartguest: Yeah you're right about that, it's completely unnecessary

Charles-Gagnon (approved):

smartguest:
Update from Karl, localization to the external folder is not necessary to do so due to the files being archived, this PR will be closed.

Charles-Gagnon:
@kburtram Any issues with just deleting that stuff then at this point? It's been long enough that I don't really see huge value in keeping this around since we don't have any intention in bringing it back (or if we do it'll probably be quite different)

PR #1669, opened 2022-09-02
Alex/fixforretrysleeping

smartguest:
Adds retry logic to the TryOpenConnection function call to account for sleeping serverless databases, currently the connection info fed to the connection function does not determine whether a connection is serverless or not so only the error number is available to determine whether to retry or not. The loop will keep retrying until the counter is up or if we get a response other than the sleeping error.

Information on the serverless sleep error comes from here: https://docs.microsoft.com/en-us/azure/azure-sql/database/serverless-tier-overview?view=azuresql#connectivity

Fixes: https://github.com/microsoft/azuredatastudio/issues/18993


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,28 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
             return completeParams;
         }
 
+        private async Task<ConnectionCompleteParams> getResponseConnect(ConnectionInfo connectionInfo, ConnectParams connectionParams)
```
cheenamalhotra, line 405: I'd recommend renaming method 'getResponseConnect' to "TryOpenConnectionWithRetry" and 'firstRun" to "isRetriable".

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,28 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
...
         }
 
+        private async Task<ConnectionCompleteParams> getResponseConnect(ConnectionInfo connectionInfo, ConnectParams connectionParams)
+        {
+            Boolean firstRun = true;
+            int counter = 0;
+            ConnectionCompleteParams response = null;
+            while (firstRun && counter <= MaxTries)
+            {
+                response = await TryOpenConnection(connectionInfo, connectionParams);
```
alanrenmsft, line 412: So TryOpenConnection will never throw, right? if not you have to wrap this with try catch block.

smartguest: It already catches for errors in the areas that it is expected in, so adding another Try Catch block is redundant.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,28 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
...
+                // If a serverless database is sleeping, it will return this error number and will need to be retried.
+                if (response?.ErrorNumber == 40613)
+                {
+                    counter++;
+                }
+                else
+                {
+                    // Every other response, we can stop.
+                    isRetriable = false;
```
Charles-Gagnon, line 421: Just break out (or even directly return the response here) in this case, no need to have a separate var just for this

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -35,6 +35,8 @@ public class ConnectionService
         private const string SqlAzureEdition = "SQL Azure";
         public const int MaxTolerance = 2 * 60; // two minutes - standard tolerance across ADS for AAD tokens
 
+        public const int MaxTries = 10; // Max number of tries to wait for a serverless database to start up when its paused before giving up.
```
Charles-Gagnon, line 38: Let's make this more clear - `MaxServerlessReconnectTries`

Charles-Gagnon, line 38: Where is 10 from? That seems like a large number - typically retries are only 3 - 5 times. Especially if there's a timeout delay for each attempt.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,28 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
             return completeParams;
         }
 
+        private async Task<ConnectionCompleteParams> TryOpenConnectionWithRetry(ConnectionInfo connectionInfo, ConnectParams connectionParams)
+        {
+            Boolean isRetriable = true;
+            int counter = 0;
+            ConnectionCompleteParams response = null;
+            while (isRetriable && counter <= MaxTries)
```
Charles-Gagnon, line 410: Make sure there's appropriate logging for each retry. Include the number retry it's on, and have a final message if it gets to the max retries and still isn't able to connect.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,28 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
...
+        private async Task<ConnectionCompleteParams> TryOpenConnectionWithRetry(ConnectionInfo connectionInfo, ConnectParams connectionParams)
+        {
+            Boolean isRetriable = true;
+            int counter = 0;
+            ConnectionCompleteParams response = null;
+            while (isRetriable && counter <= MaxTries)
+            {
+                response = await TryOpenConnection(connectionInfo, connectionParams);
+                // If a serverless database is sleeping, it will return this error number and will need to be retried.
+                if (response?.ErrorNumber == 40613)
```
Charles-Gagnon, line 414: @shueybubbles @Matteo-T Is this what we do in SSMS as well? Or is there some built in functionality in SqlClient we can use for this?

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
smartguest: Randomly picked, just to ensure things work, I'm in the process of trying to figure out what's the minimum amount of tries required, I'll lower it to 5 for now.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: I searched around in both SSMS and SMO but didn't find any reference to 40613, serverless or sleeping...

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: Serverless DBs generally resume in ~ 45 seconds.. so 2 retries should be enough, but we can also ceil at 3 or 5 as we want to choose..

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,28 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
...
 
+        private async Task<ConnectionCompleteParams> TryOpenConnectionWithRetry(ConnectionInfo connectionInfo, ConnectParams connectionParams)
+        {
+            Boolean isRetriable = true;
+            int counter = 0;
+            ConnectionCompleteParams response = null;
+            while (isRetriable && counter <= MaxTries)
+            {
+                response = await TryOpenConnection(connectionInfo, connectionParams);
+                // If a serverless database is sleeping, it will return this error number and will need to be retried.
```
Charles-Gagnon, line 413: Add link to the docs you provided in the description here

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon, line 410: @cheenamalhotra Is there a reason you know of why we can't/shouldn't use the built-in retry logic of SqlClient? @smartguest was this something you investigated before going with a custom retry solution?

https://docs.microsoft.com/en-us/sql/connect/ado-net/configurable-retry-logic?view=sql-server-ver16

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: It should be done by SSMS as well.. since serverless DBs are new, it's always better to let them resume and retry silently..!

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: We need to evaluate all aspects of enabling it before we enable it, as the doc says, there are some limitations to it specially with transaction management, and it may cause disruptions rather than being useful. Retry logic is mainly for enterprise apps, where queries are in control.

smartguest: I did not investigate using SqlClient yet, but the OpenAsync DbConnection function that TryOpenConnection uses doesn't retry for it currently, so with working with the existing code before implementing SqlClient, this is the option I have taken.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: Also to add, the overall connection time may exceed timeout limit of 30 seconds, and if we want to stay within timeout limit, this change should be avoided. But if we can prompt user that server is resuming timely, and continue with connection, that should be acceptable I think.

So my ask would be is there a way to provide user feedback to end user that "The server is resuming, please wait while we continue to connect .. "?

smartguest: I would need to design a way to relay information from STS back to ADS, which will require some work to do.

smartguest:
When I tried the latest version of ADS using regular STS, it was able to connect immediately to the sleeping database (that was just paused) if the connection was already saved. I'm not sure why it works, maybe it takes time for the connection to really be paused.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: Alright, seems reasonable to just do this given how straightforward it is to implement the retry ourselves. I'd suggest adding a comment explaining this though for future context.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,33 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
...
+                }
+                else
+                {
+                    // Every other response, we can stop.
+                    break;
+                }
+            }
+            if (response != null && counter == MaxServerlessReconnectTries)
+            {
+                response.ErrorMessage = SR.ConnectionServiceServerlessRetryFailed(MaxServerlessReconnectTries, response.ErrorMessage);
```
Charles-Gagnon, line 427: What's the error message from the sqlclient error? I think that'll probably be good enough (not sure we need to tell the user that we've hit max retries, that isn't really relevant to them)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,33 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
...
+            int counter = 0;
+            ConnectionCompleteParams response = null;
+            while (counter <= MaxServerlessReconnectTries)
+            {
+                response = await TryOpenConnection(connectionInfo, connectionParams);
+                // If a serverless database is sleeping, it will return this error number and will need to be retried.
+                // See here for details: https://docs.microsoft.com/en-us/azure/azure-sql/database/serverless-tier-overview?view=azuresql#connectivity
+                if (response?.ErrorNumber == 40613)
+                {
+                    Logger.Write(TraceEventType.Information, "Attempt to retry serverless DB connection no. " + (counter + 1) + " for " + connectionInfo.OwnerUri);
```
Charles-Gagnon, line 416: FYI - I added shortcut extension methods for all the logger stuff so you can just do `Logger.Information(...)`

And I'd suggest this for the message

`$"Database for connection {connectionInfo.OwnerUri} is paused, retrying connection. Attempt #{counter}"` (and move it below the counter++ (you'll also want to add an if check so this isn't printed out the last time when counter == MaxServerlessReconnectTries)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,33 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
...
+                    counter++;
+                }
+                else
+                {
+                    // Every other response, we can stop.
+                    break;
+                }
+            }
+            if (response != null && counter == MaxServerlessReconnectTries)
```
Charles-Gagnon, line 425: If you do keep this - the check is wrong. Counter will be MaxServerlessReconnectTries + 1 when it exits the loop if it hit the max retries.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionServerlessRetryNotification.cs
```diff
@@ -0,0 +1,19 @@
...
+//
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
+{
+    /// <summary>
+    /// ConnectionComplete notification mapping entry 
+    /// </summary>
+    public class ConnectionServerlessRetryNotification
```
Charles-Gagnon, line 13: Please revert this for now - this kind of thing doesn't need to be specific to serverless connection retries so if we're going to do it (which I'm not sure we need to) then we should have a separate discussion of how to do it in a more generic manner. 

For the purposes of this PR just fixing the issue by adding the retry is enough - even if that means it just sits there spinning for a bit longer.

smartguest: I was following what Cheena suggested, but I agree it's not necessary.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -400,6 +402,33 @@ public virtual async Task<ConnectionCompleteParams> Connect(ConnectParams connec
             return completeParams;
         }
 
+        private async Task<ConnectionCompleteParams> TryOpenConnectionWithRetry(ConnectionInfo connectionInfo, ConnectParams connectionParams)
+        {
+            int counter = 0;
+            ConnectionCompleteParams response = null;
```
Charles-Gagnon, line 408: You should be able to make this nullable to be stricter about whether values can be null - `ConnectionCompleteParams? response = null`. And then update the return param too. 

(if you're using VS you should see these show up as yellow squigglies)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: Chatted with @shueybubbles offline, it looks like SMO/SSMS isn't doing any specific retry logic for Azure DB serverless (there's one for synapse though)

@smartguest Could you try connecting to a sleeping serverless instance with SSMS to verify what the behavior is there? It'll be good to verify what the behavior is across our tools for comparison.

Charles-Gagnon (approved):
The retry logic looks fine now, although there's a few smaller follow up items for you to look at.

PR #1672, opened 2022-09-06
[Table Designer] support hash index and column store index

caohai:
This PR adds hash index and column store index support to table designer.

image
image


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -1283,6 +1411,109 @@ private void SetIndexesViewInfo(TableDesignerView view)
             view.IndexColumnSpecificationTableOptions.CanInsertRows = false;
         }
 
+        private void SetColumnStoreIndexesViewInfo(TableDesignerView view)
```
alanrenmsft, line 1414: no sql server version requirement for this feature?

caohai: All supported version should work, do we need to handle unsupported versions as well?
image
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-columnstore-index-transact-sql?view=sql-server-ver16

alanrenmsft: sounds good, just making sure.

alanrenmsft (approved):

PR #1674, opened 2022-09-07
remove copying Microsoft.Data.Tools.Schema.SqlTasks.targets

kisantia:
Removing this since this targets file is no longer needed for building sql projects after the change in https://github.com/microsoft/azuredatastudio/pull/20447 to get all the files needed for building sql projects from Microsoft.Build.Sql instead of copying them from STS.


zijchen (approved):

alanrenmsft (approved):

Charles-Gagnon (approved):
🎉

PR #1681, opened 2022-09-09
[SKU Recommendation] Remove the creation of malformed SqlDB recommendations

junierch:
Remove the creation of malformed SqlDB recommendations when creating a SKU recommendation report. The report logic for SqlDB is already adding empty entries for databases that didn't get a recommendation. Will create a task to improve the documentation on the SKU `SkuRecommendationReport` class.

image


brian-harris (approved):

raymondtruong (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -344,27 +344,7 @@ public void InitializeService(ServiceHost serviceHost)
...
-                                TargetSku = null,
-                                MonthlyCost = null,
-                                Ranking = -1,
-                                PositiveJustifications = null,
-                                NegativeJustifications = null,
-                            });
-                        }
-                    }
-
-                    sqlDbStopwatch.Stop();
```
raymondtruong, line 367: looks like you removed the stopwatch `Stop()` call here by accident, we'll need that back

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -344,27 +344,7 @@ public void InitializeService(ServiceHost serviceHost)
                     prefs.EligibleSkuCategories = GetEligibleSkuCategories("AzureSqlDatabase", parameters.IncludePreviewSkus);
                     resultSet.sqlDbResults = provider.GetSkuRecommendation(prefs, req);
 
-                    if (resultSet.sqlDbResults.Count < parameters.DatabaseAllowList.Count)
```
raymondtruong, line 347: this change makes sense and works well as discussed offline - but do we need to make the same change for MI/VM as well? seems we have the same logic there

junierch: We need to find the right data to trigger those scenarios, but they will probably fail. We may need to create an item to test and address this.

PR #1683, opened 2022-09-09
Additional Execution Plan Node Properties for Highlighting Expensive Operators

lewis-sanchez:
This PR adds additional properties to Query Execution Plan Nodes for the purposes of highlighting expensive operators based on the following additional properties:

* Elapsed CPU Time in ms
* Row Metrics:
    - Number of Rows Read
    - Estimated Number of Rows to be Read
    - Actual Number of Rows for All Executions
    - Estimated Number of Rows for All Executions

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
@@ -63,9 +63,13 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         /// </summary>
         public double RelativeCost { get; set; }
         /// <summary>
-        /// Time take by the node operation in milliseconds
+        /// Time taken by the node operation in milliseconds
         /// </summary>
         public long? ElapsedTimeInMs { get; set; }
+        /// <summary>
+        /// CPU Time taken by the node operation in milliseconds
+        /// </summary>
+        public long? ElapsedCpuTimeInMs { get; set; }
         /// <summary>
         /// Node properties to be shown in the tooltip
         /// </summary>
@@ -95,7 +99,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         /// <summary>
         /// Cost metrics for the node
         /// </summary>
-        public List<ExecutionPlanCostMetrics> CostMetrics { get; set; }
+        public List<ExecutionPlanCostMetrics> RowReadMetrics { get; set; }
     }
 
     public class ExecutionPlanGraphPropertyBase
--- a/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
@@ -97,9 +97,9 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         /// </summary>
         public List<TopOperationsDataItem> TopOperationsData { get; set; }
         /// <summary>
-        /// Cost metrics for the node
+        /// Rows read metrics for the node
         /// </summary>
-        public List<ExecutionPlanCostMetrics> RowReadMetrics { get; set; }
+        public Dictionary<string, string?> RowMetrics { get; set; }
     }
 
     public class ExecutionPlanGraphPropertyBase
--- a/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
@@ -92,6 +92,10 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         /// Top operations table data for the node
         /// </summary>
         public List<TopOperationsDataItem> TopOperationsData { get; set; }
+        /// <summary>
+        /// Cost metrics for the node
+        /// </summary>
+        public List<ExecutionPlanCostMetrics> CostMetrics { get; set; }
     }
 
     public class ExecutionPlanGraphPropertyBase
@@ -125,6 +129,18 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan.Contracts
         /// Indicates the data type of the property
         /// </summary>
         public PropertyValueDataType DataType { get; set; }
+    }
+
+    public class ExecutionPlanCostMetrics
+    {
+        /// <summary>
+        /// Name of the cost metric
+        /// </summary>
+        public string? Name { get; set; }
+        /// <summary>
+        /// Value held by the cost metric
+        /// </summary>
+        public string? Value { get; set; }
     }
 
     public class NestedExecutionPlanGraphProperty : ExecutionPlanGraphPropertyBase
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -63,10 +63,14 @@ public class ExecutionPlanNode
...
         public double RelativeCost { get; set; }
         /// <summary>
-        /// Time take by the node operation in milliseconds
+        /// Time taken by the node operation in milliseconds
         /// </summary>
         public long? ElapsedTimeInMs { get; set; }
         /// <summary>
+        /// CPU Time taken by the node operation in milliseconds
+        /// </summary>
+        public long? ElapsedCpuTimeInMs { get; set; }
```
aasimkhan30, line 72: Why do we still need this, if are passing the Rowmetrics array

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -92,6 +96,10 @@ public class ExecutionPlanNode
         /// Top operations table data for the node
         /// </summary>
         public List<TopOperationsDataItem> TopOperationsData { get; set; }
+        /// <summary>
+        /// Rows read metrics for the node
+        /// </summary>
+        public Dictionary<string, string?> RowMetrics { get; set; }
```
aasimkhan30, line 102: Is it really rowMetrics or costMetrics?

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
```diff
@@ -127,6 +135,18 @@ public class ExecutionPlanGraphPropertyBase
         public PropertyValueDataType DataType { get; set; }
     }
 
+    public class ExecutionPlanCostMetrics
```
aasimkhan30, line 138: Why is this object not passed?

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
```diff
@@ -100,6 +102,61 @@ public static ExecutionPlanEdges ConvertShowPlanEdgeToExecutionPlanEdge(Edge edg
             };
         }
 
+        public static Dictionary<string, string?> GetRowMetrics(PropertyDescriptorCollection properties)
+        {
```
aasimkhan30, line 106: Have you gotten these properties reviewed from @erinstellato-ms? 
Please switch it to cost metrics object here.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
lewis-sanchez: I did it for consistency since elapsed time in milliseconds is being passed at the same level. I can definitely move it into the dictionary, if that's preferable.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
lewis-sanchez: I renamed it to rowMetrics since this dictionary only contains values that pertain to rows for things like:
* Number of Rows Read
* Estimated Number of Rows to be Read
* Actual Number of Rows for All Executions
* Estimated Number of Rows for All Executions

If metrics that didn't just pertain to rows was also included than I would've named it costMetrics in that case.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
lewis-sanchez: I meant to get rid of this. Originally I was using an array, but switched over to using a dictionary and no longer needed this. The dictionary is simpler to look things up in than having to sift through elements of an array to find the correct name/value pair.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
erinstellato-ms: Are these the ONLY metrics we are looking at to determine what is most expensive?  If so, we do want cost also.

lewis-sanchez: Properties like cost and subtree cost were already present. The properties added here are ones that were missing.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
```diff
@@ -100,6 +102,61 @@ public static ExecutionPlanEdges ConvertShowPlanEdgeToExecutionPlanEdge(Edge edg
...
+            var metrics = new Dictionary<string, string?>();
+
+            if (properties["EstimateRowsAllExecs"] != null)
```
Charles-Gagnon, line 109: You can simplify these checks by just doing 

```csharp
var estimatedRowsAllExecsProperty = properties["EstimateRowsAllExecs"] as PropertyValue;
if (estimatedRowsAllExecsProperty != null) {
...
}
```

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanGraph.cs
Charles-Gagnon: Is there a reason you're passing a dictionary of the row metrics instead of just adding them as additional properties like we have for all the other values? Generally being explicit like that is going to be easier to work with on the TS side

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
```diff
@@ -100,6 +109,54 @@ public static ExecutionPlanEdges ConvertShowPlanEdgeToExecutionPlanEdge(Edge edg
...
+
+            if (property != null)
+            {
+                actualRows = GetPropertyDisplayValue(property);
+            }
+
+            return actualRows;
+        }
+
+        private static string GetActualRowsRead(PropertyValue? property)
```
Charles-Gagnon, line 148: These all appear to be doing exactly the same thing - I'd just rename this as `GetPropertyDisplayValue` and only have a single one.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
```diff
@@ -412,8 +422,13 @@ private static string ParseMissingIndexQueryText(string fileName, string impact,
 ";
         }
 
-        private static string GetPropertyDisplayValue(PropertyValue property)
+        private static string GetPropertyDisplayValue(PropertyValue? property)
         {
+            if (property == null)
+            {
+                return String.Empty;
```
kburtram, line 429: nit: Framework coding guidelines recommends using language primitive types when possible, e.g., `string.Empty`.

kburtram (approved):

PR #1687, opened 2022-09-12
Enable table rename and schema change for offline table designer

kisantia:
Fixes https://github.com/microsoft/azuredatastudio/issues/19557. Rename and schema change were previously disabled for offline table designer. Since refactoring is not currently supported in sql projects in ADS (https://github.com/microsoft/azuredatastudio/issues/13920), DacFx treats a table rename or schema change as a drop table and create new table. This is not ideal since the new table definition will be created in a new file, rather than updating the table in its current file.

To support table rename and schema change in the meantime, this changeset makes it so that table renames and schema changes for offline table designer are kept track of in STS and those changes are made after DacFx commits any other table changes. This will update the table and/or schema name in the "CREATE TABLE" statement only in its current file, but will not update not any references to the table.

offlineRename

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
@@ -5,12 +5,14 @@
 
 using System;
 using System.Collections.Generic;
-using System.Threading.Tasks;
+using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using Microsoft.Data.SqlClient;
+using Microsoft.Data.Tools.Sql.DesignServices;
 using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
-using Microsoft.Data.Tools.Sql.DesignServices;
 using Microsoft.SqlTools.ServiceLayer.TableDesigner.Contracts;
 using Dac = Microsoft.Data.Tools.Sql.DesignServices.TableDesigner;
 using STSHost = Microsoft.SqlTools.ServiceLayer.Hosting.ServiceHost;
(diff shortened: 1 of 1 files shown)


alanrenmsft:
how about the file that got added, that needs to be removed, right?

kisantia:
> how about the file that got added, that needs to be removed, right?

No new file gets added - the existing "CREATE TABLE" statement gets updated with the new table name and/or schema.

alanrenmsft:
@kisantia IIRC, a new file will be created with the new table name, you have to refresh the the project for it to show up, we also need to remove it.

kisantia:
> @kisantia IIRC, a new file will be created with the new table name, you have to refresh the the project for it to show up, we also need to remove it. see: [microsoft/azuredatastudio#19557 (comment)](https://github.com/microsoft/azuredatastudio/issues/19557#issuecomment-1139933194)

Updating here after our call: the new file no longer gets created with this approach because the table/schema name gets changed back to its previous name right before DacFx commits the changes, then STS updates the name in the "CREATE TABLE" statement afterwards.

kisantia:
Currently changing the table name and schema is disabled for offline sql projects table designer:
image

As mentioned in https://github.com/microsoft/azuredatastudio/issues/19557#issuecomment-1227582655, until refactoring is supported, we won't be able to have complete support for table renaming. The approach in this PR will enable renaming and schema change for sql projects, but it has limitations:
- only the table and/or schema name get updated in the "CREATE TABLE" statement
- references, like indexes, views, etc, don't get updated with the new table/schema name

Since it's a sql project, they'll have to build the project before publishing and see the errors they need to fix about the old table name not being found anymore. The main benefit I see of enabling the schema to be changed in the designer is that the schema dropdown shows all the available schemas. For changing the table name, I don't see a big benefit since it isn't hard to just edit the .sql file in the project directly.

@dzsquared @erinstellato-ms do you think we should go ahead with this approach or should we keep table rename and schema change disabled for sql projects table designer?

alanrenmsft (approved):

erinstellato-ms:
@kisantia Due to the limitations that exist without refactoring, I vote that we keep the table rename and schema change disabled for the sql projects table designer.  I don't love that the process would be build project, see errors, then fix.  In general, I feel we should avoid having errors as part of the workflow, so I would rather we document how to make this change in the script. // @dzsquared

alanrenmsft:
If we are going with what Kim has here, I'd like to add a warning in the issues tab when name or schema is edited for sql db project scenario. 

image

kisantia:
Sounds good I'll go ahead and close this PR then and update https://github.com/microsoft/azuredatastudio/issues/19557 with the discussion here. Thanks @erinstellato-ms and @dzsquared!

PR #1688, opened 2022-09-13
Dropped Ledger Columns folders for ledger primary tables

nofield:
This PR introduces the Dropped Ledger Columns folder for ledger tables. Once a column on a ledger table is dropped, it appears under the Dropped Ledger Columns folder and is removed from the regular 'Columns' folder list.

Like the other 'Dropped Ledger' folders, this one is organized underneath its superset folder type (Columns > Dropped Ledger Columns), and at the bottom of it.

*Note: the sub-folder only pertains to the primary ledger table itself, because the 'IsDroppedLedgerColumn' property is only set on the primary table's column, not the history table or view*

## Testing

Manual verification

Updatable Ledger table:
image

Append-Only Ledger table:
image

Non-ledger table does not include the dropped columns folder:
image


alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -1644,6 +1649,35 @@ internal partial class ColumnsChildFactory : SmoChildFactoryBase
...
+                    ValidFor = ValidForFlag.Sql2022|ValidForFlag.AzureV12,
+                    Values = new List<object> { 0 },
+                });
+                return filters;
+            }
+        }
+
+        protected override void OnExpandPopulateFolders(IList<TreeNode> currentChildren, TreeNode parent)
+        {
+            currentChildren.Add(new FolderNode {
```
Charles-Gagnon, line 1670: Couldn't you just check the parent here and only add the folder if the parent is a ledger table?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
```diff
@@ -98,6 +98,11 @@ public string NodePathName
         /// </summary>
         public bool IsSystemObject { get; set; }
 
+        /// <summary>
+        // True if the node includes ledger objects
+        /// </summary>
+        public bool IsLedger { get; set; } = false;
```
Charles-Gagnon, line 104: I'd rather we try to avoid having feature-specific flagsin a general class like TreeNode if we can. Most TreeNodes have nothing to do with Ledger stuff.

Charles-Gagnon (changes requested):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
nofield: Maybe, is there precedent for something like that? This is the auto-generated code from the SmoTreeNodesDefinition.

Charles-Gagnon: Not that I'm aware, this is pretty unprecedented territory for ADS unfortunately =)

This being here is problematic though, what I'd say then is see if there's somewhere else that you can do the check instead that still keeps the same result (maybe on where this is called?)

nofield: That's where I made the change, in OnExpandPopulateFoldersAndFilter :)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
nofield: I just pushed an update that removes the need for the IsLedger property by just checking for the node type id of DroppedLedgerColumns, is that a better solution?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
Charles-Gagnon: Much better, thank you 😄

PR #1686, opened 2022-09-10
Feature rename sql objects

M-Patrone:
I’m trying to implement [this feature request](https://github.com/microsoft/azuredatastudio/issues/8511) from the Azure Data  Studio. So, I wanted to add the renaming function to the sqltoolsservice project. To possible enable that feature, I added a service class, which executes the renaming operation. The class “RenameUtils” is my helper class, which validates the parameters. In addition, it creates the SQL-command for renaming objects. In the feature request it is only mentioned to enable renaming tables and columns. If this PR is accepted and the first draft of the feature for the Azure Data Studio is implemented, there is the possibility to extend to database and indexes.  As I am a newbie on this project, I would love to get some feedback. 😊  

Also have a look if I declared the custom error message on the right file. Because I am not 100% sure If I get how the translation works. 

Furthermore, I would love to add my docker file to the wiki, so it would be easier for newbies to execute the integration tests on a SQL server. Is there a specific reason why this would not be wanted? 

Checklist Contribution Guidelines
- [x] Unit- and integrationtests are added and run successfuly
- [x] The 'Release' build and unit tests pass locally 
- [ ] The are two integration tests which didn't run successfully

1. BackupRestoreUrlTests.BackupDatabaseToUrlAndRestoreFromUrlTest() --> I don't have any Azure environment to execute this test
2. ObjectExplorerServiceTests.VerifyAllSqlObjects() -->I didn't check why this isn't working. This test doesn’t work on the main branch either. For that matter, the other test also. Therefore, I didn’t invest time in that. 

Finally, sorry for my bad english 😉

PR #1689, opened 2022-09-14
add docker file for testing and development

M-Patrone:
It is a docker file to set up a quick SQL Server. With this container you can execute the integrationtests and it is also useful for development. 

It was discussed on this [PR](https://github.com/microsoft/sqltoolsservice/pull/1686#discussion_r970134362).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+#This is a docker file to generate a SQL server. It helps you to execute the integration tests and for development
+#For more information about the possibility for SQL server on docker, check this site out: https://hub.docker.com/_/microsoft-mssql-server 
+
+FROM mcr.microsoft.com/mssql/server:2019-latest 
+
+ENV ACCEPT_EULA=Y 
+ENV SA_PASSWORD=abcDEF123#
+ENV MSSQL_PID=Developer
+ENV MSSQL_TCP_PORT=1433 
+
+
+WORKDIR /src 
+RUN (/opt/mssql/bin/sqlservr --accept-eula & ) | grep -q "Service Broker manager has started" 
+
+#To create the image, execute the following command in the directory of the docker file 
+#docker build . -t sqlserver_sqltoolsservice --no-cache 
+
+#To execute the docker container, use the following command 
+#docker run -d -p 1433:1433 --name sqlserver sqlserver_sqltoolsservice 
+
+#Now you have a working SQL server for development and testing purposes 


Dockerfile
```diff
@@ -0,0 +1,21 @@
+#This is a docker file to generate a SQL server. It helps you to execute the integration tests and for development
+#For more information about the possibility for SQL server on docker, check this site out: https://hub.docker.com/_/microsoft-mssql-server 
+
+FROM mcr.microsoft.com/mssql/server:2019-latest 
+
+ENV ACCEPT_EULA=Y 
+ENV SA_PASSWORD=abcDEF123#
```
Charles-Gagnon, line 7: Let's comment this line out and add a comment saying that they either need to uncomment & fill it in here or set the environment variable themselves. We shouldn't be checking in passwords (even temp ones) into the repo.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ENV MSSQL_PID=Developer
+ENV MSSQL_TCP_PORT=1433 
+
+
+WORKDIR /src 
+RUN (/opt/mssql/bin/sqlservr --accept-eula & ) | grep -q "Service Broker manager has started"
```
Charles-Gagnon, line 13: You can just set the ACCEPT_EULA env var too

Charles-Gagnon (commented):
Thinking about it - do we really need a Dockerfile? Just running this command by itself should set up and run the instance without needing the Dockerfile itself : 

`docker run -e ACCEPT_EULA=Y -e "SA_PASSWORD=******" -p 1433:1433 --name sqlserver -d mcr.microsoft.com/mssql/server:2019-latest`

It might be interesting to set up a full dev container for building and running the tests - but for just getting a test SQL instance stood up we probably don't actually need this. I can just update the docs to provide the command above.

Dockerfile
M-Patrone: Yeah you are absolutely right. What do you think of that?
```Dockerfile
#ENV SA_PASSWORD=ADD_YOUR_SA_PASSWORD
```
So it would be clear to change here the password for the sa user.

M-Patrone:
Of course we could add this in the wiki. I was also not sure, if it makes sense to add it in the repo. I thought it would be a file like this "test\CreateConnectionSettings.cmd". An "executable" file with some explanations.

Charles-Gagnon:
I added instructions for using docker [here](https://github.com/microsoft/sqltoolsservice/wiki#one-time-setup-to-run-integration-performance-and-testdriver-tests) so I'm going to close this for now since that's a simpler way to get a container instance running.

PR #1691, opened 2022-09-15
Remove dropped ledger columns from Edit Data columns

nofield:
Dropped ledger columns should not be exposed to the Edit Data view, because the columns are non-editable and soft-deleted (marked as dropped and hidden, and removed from all views unless the user explicitly requests them). This PR filters dropped columns out from the list returned by GetObjectMetadata so it's never exposed in Edit Data.

## Testing

Manual verification. I attempted to create a unit test for this specific scenario, but it got tricky with IsDroppedLedgerColumn not being exposed in DbColumn (only SMO Column), and trying to set specific properties in a column passed to GetObjectMetadata. If you have suggestions for how to work around these issues I'd be happy to write an automated test.

Table:
image

Edit Data does not expose the dropped column:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
@@ -86,16 +86,16 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
                     throw new ArgumentOutOfRangeException(nameof(objectType), SR.EditDataUnsupportedObjectType(objectType));
             }
 
-            // Filter out dropped ledger columns from the list of columns to be returned
-            //
-            smoResult.Columns.ClearAndInitialize("[(@IsDroppedLedgerColumn=0)]", new [] {nameof(Column.IsDroppedLedgerColumn), nameof(Column.DataType)});
-
             // A bug in SMO makes it necessary to call refresh to attain certain properties (such as IsMemoryOptimized)
             smoResult.Refresh();
             if (smoResult.State != SqlSmoState.Existing)
             {
                 throw new ArgumentOutOfRangeException(nameof(objectNamedParts), SR.EditDataObjectNotFound);
             }
+
+            // Filter out dropped ledger columns from the list of columns to be returned
+            // and grab the specific column properties checked below
+            smoResult.Columns.ClearAndInitialize("[(@IsDroppedLedgerColumn=0)]", new [] { nameof(Column.DataType), nameof(Column.DefaultConstraintName) });
 
             // Generate the edit column metadata
             List<EditColumnMetadata> editColumns = new List<EditColumnMetadata>();
@@ -107,7 +107,7 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
                 try
                 {
                     // The default value may be escaped
-                    defaultValue = smoColumn.DefaultConstraint == null
+                    defaultValue = string.IsNullOrEmpty(smoColumn.DefaultConstraintName)
                         ? null
                         : FromSqlScript.UnwrapLiteral(smoColumn.DefaultConstraint.Text);
                 }
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
@@ -99,6 +99,12 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
             {
                 Column smoColumn = smoResult.Columns[i];
 
+                // Don't return columns that are dropped
+                if (smoColumn.IsDroppedLedgerColumn)
+                {
+                    continue;
+                }
+
                 string defaultValue = null;
                 try
                 {
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
@@ -86,6 +86,10 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
                     throw new ArgumentOutOfRangeException(nameof(objectType), SR.EditDataUnsupportedObjectType(objectType));
             }
 
+            // Filter out dropped ledger columns from the list of columns to be returned
+            //
+            smoResult.Columns.ClearAndInitialize("[(@IsDroppedLedgerColumn=0)]", new [] {nameof(Column.IsDroppedLedgerColumn), nameof(Column.DataType)});
+
             // A bug in SMO makes it necessary to call refresh to attain certain properties (such as IsMemoryOptimized)
             smoResult.Refresh();
             if (smoResult.State != SqlSmoState.Existing)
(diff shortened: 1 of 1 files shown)


alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -99,6 +99,12 @@ public EditTableMetadata GetObjectMetadata(DbConnection connection, string[] obj
             {
                 Column smoColumn = smoResult.Columns[i];
 
+                // Don't return columns that are dropped
+                if (smoColumn.IsDroppedLedgerColumn)
```
Charles-Gagnon, line 103: @shueybubbles - This shouldn't have any perf implications just querying the value on the column, right? (especially since we already refresh it above)

@nofield Did you test this with a older sql version that doesn't have ledger stuff? I thought the default for unsupported properties was to throw but I may be misremembering that.

smartguest (approved):

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
shueybubbles: `IsDroppedLedgerColumn` is hardcoded to false on old versions
See https://github.com/microsoft/sqlmanagementobjects/blob/73cd2cb929409cb9ddc2c7b416c5d7c6ef621d5f/src/Microsoft/SqlServer/Management/SqlEnum/xml/column.xml#L247

shueybubbles:
>             for (int i = 0; i < smoResult.Columns.Count; i++)

I highly recommend using smoREsult.Columns.ClearAndInitialize, passing a filter to exclude the undesirable rows on the server side and also passing the names of the properties you reference.
Otherwise perf is going to be really bad for objects with many columns.

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs:98 in 21470b0. [](commit_id = 21470b0929c9d8c31fd4022637e9f57cc9da0d60, deletion_comment = False)

shueybubbles:
>             for (int i = 0; i < smoResult.Columns.Count; i++)

[https://github.com/microsoft/sqlmanagementobjects/wiki/Refresh-vs-ClearAndInitialize](https://github.com/microsoft/sqlmanagementobjects/wiki/Refresh-vs-ClearAndInitialize)

---
In reply to: [1249850114](https://github.com/microsoft/sqltoolsservice/pull/1691#issuecomment-1249850114) [](http://example.com/codeflow?ancestors=1249850114)

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs:98 in 21470b0. [](commit_id = 21470b0929c9d8c31fd4022637e9f57cc9da0d60, deletion_comment = False)

shueybubbles (changes requested):
:clock1:

Charles-Gagnon:
Thanks, I created a separate issue to track the perf improvement since this PR isn't changing anything there.

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
nofield: In SMO the property is defined for all versions of SQL, just defaulted to 0 if it's an unsupported sql version. https://msdata.visualstudio.com/SQLToolsAndLibraries/_git/SqlManagementObjects?path=/src/Microsoft/SqlServer/Management/SqlEnum/xml/column.xml&version=GBmaster&line=248&lineEnd=249&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

nofield:
> > ```
> >         for (int i = 0; i < smoResult.Columns.Count; i++)
> > ```
> 
> I highly recommend using smoREsult.Columns.ClearAndInitialize, passing a filter to exclude the undesirable rows on the server side and also passing the names of the properties you reference. Otherwise perf is going to be really bad for objects with many columns.
> 
> Refers to: src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs:98 in [21470b0](https://github.com/microsoft/sqltoolsservice/commit/21470b0929c9d8c31fd4022637e9f57cc9da0d60). [](commit_id = [21470b0](https://github.com/microsoft/sqltoolsservice/commit/21470b0929c9d8c31fd4022637e9f57cc9da0d60), deletion_comment = False)

I moved the dropped ledger column filtering into ClearAndInitialize as recommended, thanks @shueybubbles. @Charles-Gagnon, David was saying that the dropped ledger column filtering as it was would be a perf impact, better to handle the filtering in SMO itself.

I'm not sure if I'm now duplicating refresh calls on the columns, though, since there's both a call to ClearAndInitialize on the Columns and the Refresh() on the entire smoResult object. I looked into if the call to Refresh() could be replaced, but the 'State' property that's being checked against on line 95 is being set during that call.

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -86,6 +86,10 @@ public EditTableMetadata GetObjectMetadata(DbConnection connection, string[] obj
                     throw new ArgumentOutOfRangeException(nameof(objectType), SR.EditDataUnsupportedObjectType(objectType));
             }
 
+            // Filter out dropped ledger columns from the list of columns to be returned
+            //
+            smoResult.Columns.ClearAndInitialize("[(@IsDroppedLedgerColumn=0)]", new [] {nameof(Column.IsDroppedLedgerColumn), nameof(Column.DataType)});
+
             // A bug in SMO makes it necessary to call refresh to attain certain properties (such as IsMemoryOptimized)
```
shueybubbles, line 93: >  // A bug in SMO [](http://example.com/codeflow?start=10&length=17)

what bug is that?

nofield: Yeah, I'm not sure what that's referring to.

src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs
```diff
@@ -86,6 +86,10 @@ public EditTableMetadata GetObjectMetadata(DbConnection connection, string[] obj
                     throw new ArgumentOutOfRangeException(nameof(objectType), SR.EditDataUnsupportedObjectType(objectType));
             }
 
+            // Filter out dropped ledger columns from the list of columns to be returned
+            //
+            smoResult.Columns.ClearAndInitialize("[(@IsDroppedLedgerColumn=0)]", new [] {nameof(Column.IsDroppedLedgerColumn), nameof(Column.DataType)});
```
shueybubbles, line 91: >.ClearAndInitialize [](http://example.com/codeflow?start=29&length=19)

Move this after the check for Existing State.
No need to include IsDroppedLedgerColumn in the property list parameter either, just in the filter

shueybubbles:
>                         : FromSqlScript.UnwrapLiteral(smoColumn.DefaultConstraint.Text);

Add nameof(Column.DefaultConstraintName) to the property list in ClearAndInitialize, and don't try to fetch DefaultConstraint unless smoColumn.DefaultConstraintName is non-empty.

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs:112 in b6e5707. [](commit_id = b6e57076dd00ccec9a8ec3d03a5d82942f237bf5, deletion_comment = False)

shueybubbles (approved):
:shipit:

Charles-Gagnon (approved):

PR #1705, opened 2022-09-21
Export to Markdown Support

benrr101:
**Description**: This PR adds support in the tools service for exporting query results and serialization requests to Markdown tables. This is a feature I'm working on as a hackathon project, born out of a discussion with a fellow engineer about the tediousness of writing Markdown tables.

The functionality is implemented the same as export to CSV, Excel, JSON, and XML functionality, by adding a new `IFileStreamWriter` and associated `IFileStreamFactory` in the `Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage` namespace`. A new class for the request parameters has been added and it has been wired up in both the QueryExecutionService (as `query/saveMarkdown`) and the SerializationService.

Encoding is a little bit different than existing export formats. Since the intended result of parsing Markdown is to get HTML, cell values are sent through the `HttpUtility.HtmlEncode` method to convert special characters to HTML entities. Markdown tables do not support multiple lines per cell, so if a cell contains a newline, it is converted to a `<br />` tag that will insert a new line after parsing. Finally, I've left a note about how multiple consecutive spaces and tabs should be converted to `&nbsp;`, since on conversion to html, any excess whitespace will be treated as a single space. However, since this is subject to user preference (tabs vs spaces, 2 vs 4 vs 8, `nbsp` vs `ensp` vs `emsp`), I'll sidestep the holy wars and leave until someone requests that functionality.

One small note, headers are made optional in the request params. Depending on the Markdown flavor and the renderer, the headers may or may not be required for rendering of the table (eg, Github flavored Markdown requires headers). As such, although this parameter is optional, I'll expose the configuration in ADS as defaulting to `true`.

_No refactoring was applied :)_

**Testing**: Unit tests were added for the query execution service, the writer, the factory, and the serialization service. (a tiny bit of refactoring was done in the serialization service tests to remove some redundancy in CSV tests).

**Example Output**
Generated:
```
|column0|column1|column2|column3|
|---|---|---|---|
|item0|item1|item2|item3|
|item0|item1|item2|item3|
|item0|item1|item2|item3|
```

Rendered:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsMarkdownFileStreamFactory.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsMarkdownFileStreamFactory.cs
@@ -1,0 +1,70 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts;
+using Microsoft.SqlTools.ServiceLayer.SqlContext;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
+{
+    public class SaveAsMarkdownFileStreamFactory : IFileStreamFactory
+    {
+        private readonly SaveResultsAsMarkdownRequestParams saveRequestParams;
+
+        /// <summary>
+        /// Constructs and initializes a new instance of <see cref="SaveAsMarkdownFileStreamFactory"/>.
+        /// </summary>
+        /// <param name="settings"></param>
+        /// <param name="requestParams"></param>
+        public SaveAsMarkdownFileStreamFactory(QueryExecutionSettings settings, SaveResultsAsMarkdownRequestParams requestParams)
+        {
+            this.QueryExecutionSettings = settings;
+            this.saveRequestParams = requestParams;
+        }
+
+        /// <inheritdoc />
+        public QueryExecutionSettings QueryExecutionSettings { get; set; }
+
+        /// <inheritdoc />
+        /// <exception cref="InvalidOperationException">Throw at all times.</exception>
+        [Obsolete("Not implemented for export factories.")]
+        public string CreateFile()
+        {
+            throw new InvalidOperationException();
+        }
+
+        /// <inheritdoc />
+        /// <remarks>
+        /// Returns an instance of the <see cref="ServiceBufferFileStreamreader"/>.
+        /// </remarks>
+        public IFileStreamReader GetReader(string fileName)
+        {
+            return new ServiceBufferFileStreamReader(
+                new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite),
+                this.QueryExecutionSettings);
+        }
+
+        /// <inheritdoc />
+        /// <remarks>
+        /// Returns an instance of the <see cref="SaveAsMarkdownFileStreamWriter"/>.
+        /// </remarks>
+        public IFileStreamWriter GetWriter(string fileName, IReadOnlyList<DbColumnWrapper> columns)
+        {
+            return new SaveAsMarkdownFileStreamWriter(
+                new FileStream(fileName, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite),
+                this.saveRequestParams,
+                columns);
+        }
+
+        /// <inheritdoc />
+        public void DisposeFile(string fileName)
+        {
+            FileUtilities.SafeFileDelete(fileName);
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsMarkdownFileStreamFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            this._saveRequestParams = requestParams;
+        }
+
+        /// <inheritdoc />
+        public QueryExecutionSettings QueryExecutionSettings { get; set; }
+
+        /// <inheritdoc />
+        /// <exception cref="InvalidOperationException">Throw at all times.</exception>
+        [Obsolete("Not implemented for export factories.")]
```
Charles-Gagnon, line 33: Let's put this in the exception message instead 

`InvalidOperationException("CreateFile not implemented for export factories");`

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsMarkdownFileStreamWriterTests.cs
```diff
@@ -0,0 +1,383 @@
...
+            using var outputStream = new MemoryStream(output);
+
+            // If: I construct a Markdown file writer
+            var writer = new SaveAsMarkdownFileStreamWriter(outputStream, requestParams, columns);
+            writer.Dispose();
+
+            // Then:
+            // ... It should have written a line
+            string[] lines = ParseWriterOutput(output, Environment.NewLine);
+            Assert.AreEqual(2, lines.Length);
```
Charles-Gagnon, line 71: Tests should use [constraint style](https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertion-models/constraint.html) (fluent) assertions. These are : 

- Easier to read
- Have better overloads (allow custom messages, necessary when there's multiple assertions in a single test to be able to tell what part exactly failed)
- Are harder to mess up the parameter order from (expected vs actual)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsMarkdownFileStreamWriterTests.cs
```diff
@@ -0,0 +1,383 @@
...
+            // ... Create a set of columns
+            // --- Create a memory location to store the output
+            var requestParams = new SaveResultsAsMarkdownRequestParams { IncludeHeaders = true };
+            var (columns, _) = GetTestValues(2);
+            byte[] output = new byte[8192];
+            using var outputStream = new MemoryStream(output);
+
+            // If: I construct a Markdown file writer
+            var writer = new SaveAsMarkdownFileStreamWriter(outputStream, requestParams, columns);
+            writer.Dispose();
```
Charles-Gagnon, line 66: `using (var writer...)`

Here and elsewhere

Charles-Gagnon (commented):
Some cleanup in the tests, but otherwise looks fine

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsMarkdownFileStreamWriterTests.cs
benrr101: Originally I wasn't doing that because I wanted to make sure disposal didn't do anything weird. Since `using var xyz;` (I think) disposes on method exit and `using (var writer) { }` would have this pointless empty block. But, I guess since there isn't anything fancy written in `Dispose`, I guess we can just use the `using var xyz;`.

benrr101:
@Charles-Gagnon I think I got your changes addressed. Let me know what else you have in mind.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/SerializationService.cs
```diff
@@ -256,7 +259,7 @@ private void EnsureWriterCreated()
                         };
                         break;
                     default:
-                        throw new Exception(SR.SerializationServiceUnsupportedFormat(this.requestParams.SaveFormat));
+                        throw new Exception($"FUCK: '{this.requestParams.SaveFormat}'");
```
Charles-Gagnon, line 262: Uhh... 😮

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/SaveResults/SerializationServiceTests.cs
```diff
@@ -267,6 +302,39 @@ private static void AssertLineEquals(string line, string[] expected)
             }
         }
 
+        public static void VerifyMarkdownMatchesData(
+            DbCellValue[][] data,
+            ColumnInfo[] columns,
+            bool includeHeaders,
+            string filePath)
+        {
+            Assert.True(File.Exists(filePath));
```
Charles-Gagnon, line 311: You should be able to use `Assert.That(filePath, Does.Exist)` here

test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/SaveResults/SerializationServiceTests.cs
```diff
@@ -267,6 +302,39 @@ private static void AssertLineEquals(string line, string[] expected)
...
+            string[] lines = File.ReadAllLines(filePath);
+
+            int expectedLength = includeHeaders ? data.Length + 2 : data.Length;
+            Assert.AreEqual(expectedLength, lines.Length);
```
Charles-Gagnon, line 315: `Assert.That(...)` (with message) - same for other checks in here

Charles-Gagnon (approved):
I'll merge it in once the checks pass!

PR #1707, opened 2022-09-23
[Table Designer] Fix columnstore wording

caohai:
https://github.com/microsoft/azuredatastudio/issues/20659
- Fix columnstore wording
- Fix aria label for columnstore indexes table
image


alanrenmsft (approved):

smartguest:
@caohai please run build.cmd so that the generated sr.cs file is included as well

caohai:
> @caohai I will need to add the generated sr.cs file to this, it is needed for proper localization

@smartguest I did run `build.cmd` for this and I don't think sr.cs need any update as the string keys are not touched, only the contents are updated.

smartguest:
@caohai Ah I see, yeah you're right, if the string is updated, but not the name of the string itself, it is fine.

PR #1708, opened 2022-09-23
Update Microsoft.Data.SqlClient to v5.0.1

cheenamalhotra:
Brings in all necessary changes to update Microsoft.Data.SqlClient to v5.0.1.

:exclamation:**Breaking changes**:exclamation:
- `Microsoft.SqlTools.ManagedBatchParser` now targets .NET 6 and .Net Framework 4.7.2 (_earlier .NET Standard 2.0_). Since DacFx also targets .NET Standard 2.1, STS can no longer build for .NET Standard 2.0. MSBuild was picking .NET Framework DLLs from DacFx for .NET Standard 2.0 build.
- `ConnectionDetails.Encrypt` is now a `string` (previously bool) - backwards compatible with ADS Connection Options
  - Accepted values: `'Optional', 'Mandatory', 'Strict', 'True', 'False', 'Yes', 'No'`

**New:**
- Adds new `ConnectionDetails` property: `HostNameInCertificate: string` to align with `SqlConnection` properties.

Todo (will do in another PR):
- Update DacFx NuGet to contain HostNameInCertificate support [Waiting]
  - _Failures captured in DacFx Service tests when HostNameInCertificate is required to be provided._


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1299,14 +1299,18 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
...
+            if (connectionDetails.Encrypt != null)
             {
-                connectionBuilder.Encrypt = connectionDetails.Encrypt.Value;
+                connectionBuilder.Encrypt = connectionDetails.Encrypt;
             }
             if (connectionDetails.TrustServerCertificate.HasValue)
             {
                 connectionBuilder.TrustServerCertificate = connectionDetails.TrustServerCertificate.Value;
             }
+            if (connectionDetails.HostNameInCertificate != null)
```
Charles-Gagnon, line 1310: Is this something we'll then have to bring over to any other service (Import/Data virt) that creates its own connection as well?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1299,14 +1299,18 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
 
                 connectionBuilder.EnclaveAttestationUrl = connectionDetails.EnclaveAttestationUrl;
             }
-            if (connectionDetails.Encrypt.HasValue)
+            if (connectionDetails.Encrypt != null)
```
Charles-Gagnon, line 1302: So this is going to require updating the code in the other services too, right? (any anything else that creates its own connections)

cheenamalhotra: All consuming libraries would need to update the same in order to work with TDS 8 using MDS 5.0. 
If you continue to set these to true/false, it won't break apps, but TDS 8 support would be missing.

Directory.Build.props
```diff
@@ -3,6 +3,7 @@
     <HighEntropyVA>true</HighEntropyVA>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
     <TargetFramework>net6.0</TargetFramework>
+    <NoWarn>NU1605</NoWarn>
```
Charles-Gagnon, line 6: This doesn't seem like something we should be ignoring. Can we not fix whatever the downgrade being flagged is? Or if there's a reason why this is OK then at the least add a comment explaining why this needs to be done.

src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj
```diff
@@ -33,6 +33,7 @@
     <PackageReference Include="Microsoft.Azure.OperationalInsights" />
     <PackageReference Include="Microsoft.SqlServer.DACFx" />
     <PackageReference Include="Microsoft.SqlServer.Management.SmoMetadataProvider" />
+	<PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" />
```
Charles-Gagnon, line 36: fix indentation

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<!-- Targeting netstandard2.0 so that other things such as PS cmdlets can use this which need to support a wider range of machines -->
-		<TargetFramework>netstandard2.0</TargetFramework>
+		<!-- Targeting netstandard2.1 so that other things such as PS cmdlets can use this which need to support a wider range of machines -->
+		<TargetFramework>netstandard2.1</TargetFramework>
```
Charles-Gagnon, line 4: @Matteo-T FYI

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
```diff
@@ -155,7 +147,16 @@ internal static ConnectionProviderOptions BuildConnectionProviderOptions()
...
                         GroupName = "Security",
-                        ValueType = ConnectionOption.ValueTypeBoolean
+                        ValueType = ConnectionOption.ValueTypeBoolean,
+                    },
+                    new ConnectionOption
+                    {
+                        Name = "strictEncryption",
+                        DisplayName = "Strict Encryption",
+                        Description =
+                        "When true (and encrypt=true), SQL Server uses strict SSL encryption for all data sent between the client and server if the servers has a certificate installed",
```
Charles-Gagnon, line 157: `if the server has a certificate installed`

test/Microsoft.SqlTools.ServiceLayer.Test.Common/ConnectionSetting.cs
```diff
@@ -58,6 +58,12 @@ public InstanceInfo(string versionKey)
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public string ProfileName { get; set; }
 
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public string Encrypt { get; set; }
```
Charles-Gagnon, line 62: Why not use the boolean encrypt and strictEncryption options like you added above? Not a huge deal so if you'd prefer to keep this closer to the SqlClient implementation then that's fine - I just usually find it easier to keep things as consistent as possible within a code base.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionDetailsTests.cs
```diff
@@ -226,27 +235,25 @@ public void ConnectTimeoutShouldReturnNullIfSetToNull()
...
         {
             ConnectionDetails details = new ConnectionDetails();
-
-            string encrypt = "True";
-            bool? expectedValue = true;
-            details.Options["encrypt"] = encrypt;
-
-            Assert.AreEqual(details.Encrypt, expectedValue);
+            details.Options["Encrypt"] = true.ToString();
+            Assert.AreEqual(details.Encrypt, true);
```
Charles-Gagnon, line 242: `Assert.That(details.Encrypt, Is.True, "Encrypt should be true);`

(and same for below check). 

1. You should always add messages to failures when there's more than one assert so that it's easy to tell where it actually failed
2. You should try to use Constraint (fluent) assertions whenever possible. They're a lot easier to read, have better error messages and provide much more robust ways of checking results.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -647,6 +648,31 @@ public void BuildConnectionStringWithInvalidOptions(string propertyName, object
...
+            new object[] {true, true, SqlConnectionEncryptOption.Strict },
+            new object[] {true, false, SqlConnectionEncryptOption.Mandatory },
+            new object[] {false, true, SqlConnectionEncryptOption.Optional },
+            new object[] {false, false, SqlConnectionEncryptOption.Optional },
+        };
+        
+        /// <summary>
+        /// Verify that Strict Encryption parameters can be built into a connection string for connecting.
+        /// </summary>
+        [Test, TestCaseSource(nameof(EncryptionCombinations))]
```
Charles-Gagnon, line 662: You can just use TestCase here to inline these - shouldn't need to use a TestCaseSource

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Connection/ConnectionServiceTests.cs
```diff
@@ -1643,18 +1669,44 @@ public void ParseConnectionStringTest()
             // If we make a connection to a live database
             ConnectionService service = ConnectionService.Instance;
 
-            var connectionString = "Server=tcp:{servername},1433;Initial Catalog={databasename};Persist Security Info=False;User ID={your_username};Password={your_password};MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;";
+            var connectionString = "Server=tcp:{servername},1433;Initial Catalog={databasename};Persist Security Info=False;User ID={your_username};Password={your_password};MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;HostNameInCertificate={servername}";
 
             var details = service.ParseConnectionString(connectionString);
+            Assert.AreEqual("tcp:{servername},1433", details.ServerName);
```
Charles-Gagnon, line 1675: `Assert.That(details.ServerName, Is.EqualTo("..."), "Unexpected ServerName value")`

(same for below)

Charles-Gagnon (approved):
Changes look fine - just need verification that all connection types across all platforms have been tested before merging in.

test/Microsoft.SqlTools.ServiceLayer.Test.Common/ConnectionSetting.cs
cheenamalhotra: I wanted to leave the surface API as close to how it will be captured by connection strings by ADS end users and MDS as well. The bool props are mainly intended to keep the internal models backwards compatible without surfacing MDS new encryption object.

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
Matteo-T: Any chance this one project can be made available for, say, net472 as well as netstandard2.0?

What is that it is making it so netstandard2.1-specific, which would essentially make it unable to interop (without taking risks) with NetFx (full framework)?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
```diff
@@ -614,10 +614,7 @@ internal void RaiseError(ErrorCode errorCode, string message = null)
 
         internal static void RaiseError(ErrorCode errorCode, Token token, string message = null)
         {
-            if (message == null)
-            {
-                message = string.Format(CultureInfo.CurrentCulture, SR.BatchParser_IncorrectSyntax, token.Text);
-            }
+            message ??= string.Format(CultureInfo.CurrentCulture, SR.BatchParser_IncorrectSyntax, token.Text);
```
Matteo-T, line 617: >??= [](http://example.com/codeflow?start=20&length=3)

Glad to learn about this new operator, but... isn't this code just a 1-liner:

```csharp
throw new BatchParserException(errorCode, token, message ?? string.Format(CultureInfo.CurrentCulture, SR.BatchParser_IncorrectSyntax, token.Text));
```

src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj
```diff
@@ -33,6 +33,7 @@
     <PackageReference Include="Microsoft.Azure.OperationalInsights" />
     <PackageReference Include="Microsoft.SqlServer.DACFx" />
```
Matteo-T, line 34: >DACFx [](http://example.com/codeflow?start=51&length=5)

It wont matter much, but just use the same casing as the official name of the package: Microsoft.SqlServer.DacFx

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/BatchParserSqlCmd.cs
```diff
@@ -155,11 +155,7 @@ private string GetFilePath(string fileName)
                     }
 
                 }
-                if (fullFileName == null)
-                {
-                    fullFileName = Path.GetFullPath(fileName);
-                }
-
+                fullFileName ??= Path.GetFullPath(fileName);
                 return fullFileName;
```
Matteo-T, line 159: Similar to the other one:

  return fullFileName ?? Path.GetFullPath(fileName);

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ScriptExecutionArgs.cs
```diff
@@ -130,11 +130,7 @@ internal int StartingLine
         {
             get
             {
-                if (cmdVariables == null)
-                {
-                    cmdVariables = new Dictionary<string, string>(StringComparer.CurrentCultureIgnoreCase);
-                }
-
+                cmdVariables ??= new Dictionary<string, string>(StringComparer.CurrentCultureIgnoreCase);
                 return cmdVariables;
```
Matteo-T, line 134: Maybe I'm missing something, but... do we really think we save much with "a lazy init" of the backing field?

I would just:
1) Drop the backing field at line 18
2) Declare the internal property like this:

```csharp
   internal Dictionary<string, string> Variables { get;  } = new Dictionary<string, string>(StringComparer.CurrentCultureIgnoreCase);
```

(I would have make it return an IDictionary, but then again... we love "internal" accessors in our codebases, so we could break somebody...)

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -50,10 +50,11 @@
 		<PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
 		<PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" />
 		<PackageReference Include="Microsoft.SqlServer.DACFx" />
```
Matteo-T, line 52: >DACFx [](http://example.com/codeflow?start=49&length=5)

DacFx

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1299,14 +1300,26 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
 
                 connectionBuilder.EnclaveAttestationUrl = connectionDetails.EnclaveAttestationUrl;
             }
-            if (connectionDetails.Encrypt.HasValue)
+
+            if (!string.IsNullOrEmpty(connectionDetails.Encrypt))
             {
-                connectionBuilder.Encrypt = connectionDetails.Encrypt.Value;
+                connectionBuilder.Encrypt = connectionDetails.Encrypt.ToLower() switch
```
Matteo-T, line 1306: >ToLower [](http://example.com/codeflow?start=70&length=7)

I would probably go with .ToLowerInvariant()

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Connection/ReliableConnectionTests.cs
```diff
@@ -242,8 +242,19 @@ private SqlConnectionStringBuilder CreateTestConnectionStringBuilder()
                 csb.Password = connectParams.Connection.Password;
             }
             csb.ConnectTimeout = connectParams.Connection.ConnectTimeout.HasValue ? connectParams.Connection.ConnectTimeout.Value: 30;
-            csb.Encrypt = connectParams.Connection.Encrypt.HasValue ? connectParams.Connection.Encrypt.Value : false;
+            
+            csb.Encrypt = connectParams.Connection.Encrypt?.ToLower() switch {
```
Matteo-T, line 246: >ToLower [](http://example.com/codeflow?start=60&length=7)

here too

test/Microsoft.SqlTools.ServiceLayer.Test.Common/TestConnectionProfileService.cs
```diff
@@ -106,7 +106,7 @@
                         }
                     }
                 }
-                if (settings != null)
+                if (settings != null && settings.Connections != null)
```
Matteo-T, line 109: >if (settings != null && settings.Connections != null) [](http://example.com/codeflow?start=16&length=53)

is this not
   settings?.Connections != null
?

Matteo-T (changes requested):
:clock1:

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
cheenamalhotra: > What is that it is making it so netstandard2.1-specific, which would essentially make it unable to interop (without taking risks) with NetFx (full framework)?

Recently DacFx modified their support for .NET Standard to 2.1, that drove this change.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
cheenamalhotra: Now it is :)

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ScriptExecutionArgs.cs
cheenamalhotra: We have a lot of opportunities here, I agree. I'll consider that for next time :)

Matteo-T (approved):
:shipit:

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/BatchParserSqlCmd.cs
```diff
@@ -155,12 +155,7 @@ private string GetFilePath(string fileName)
                     }
 
                 }
-                if (fullFileName == null)
-                {
-                    fullFileName = Path.GetFullPath(fileName);
-                }
-
-                return fullFileName;
+                return fullFileName ??= Path.GetFullPath(fileName);
```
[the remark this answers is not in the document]
cheenamalhotra:
```suggestion
                return fullFileName ?? Path.GetFullPath(fileName);
```

alanrenmsft (approved):

PR #1717, opened 2022-10-05
Add more logging to GetSignatureHelp

Charles-Gagnon:
More logging to try and figure out why the test is still failing.

Along the way fixed up some null handling.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
@@ -87,10 +87,10 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         /// <summary>
         /// Queue a binding request item
         /// </summary>
-        public virtual QueueItem QueueBindingOperation(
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -89,18 +89,12 @@ public bool IsCancelRequested
...
-            Func<IBindingContext, CancellationToken, object> bindOperation,
-            Func<IBindingContext, object> timeoutOperation = null,
-            Func<Exception, object> errorHandler = null,
+            Func<IBindingContext, CancellationToken, object?> bindOperation,
+            Func<IBindingContext, object>? timeoutOperation = null,
+            Func<Exception, object>? errorHandler = null,
             int? bindingTimeout = null,
             int? waitForLockTimeout = null)
         {
-            // don't add null operations to the binding queue
```
Charles-Gagnon, line 98: Verified that we're never calling this anywhere that doesn't pass in a bindOperation. As long as no one does something weird and specifically pass in null here we don't need this.

cheenamalhotra (approved):

PR #1723, opened 2022-10-10
Add option to import permissions for dacfx extract

kisantia:
This adds the option to include permissions in an extract, needed for fixing https://github.com/microsoft/azuredatastudio/issues/15071. By default, permissions are ignored in DacFx extract. After this is checked in, I'll make a PR in ADS for adding this option to the create project from db dialog, which is an option that SSDT has in its create project from db dialog.

SSDT's import permissions option:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/ExtractRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/ExtractRequest.cs
@@ -30,7 +30,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         /// <summary>
         /// Gets or sets whether permissions should be included in the extract
         /// </summary>
-        public bool? ImportPermissions { get; set; }
+        public bool? IncludePermissions { get; set; }
     }
 
     /// <summary>
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/ExtractRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/ExtractRequest.cs
@@ -26,6 +26,11 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx.Contracts
         /// Gets or sets the target for extraction
         /// </summary>
         public DacExtractTarget ExtractTarget { get; set; }
+
+        /// <summary>
+        /// Gets or sets whether permissions should be included in the extract
+        /// </summary>
+        public bool? ImportPermissions { get; set; }
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/ExtractRequest.cs
```diff
@@ -26,6 +26,11 @@ public class ExtractParams : DacFxParams
         /// Gets or sets the target for extraction
         /// </summary>
         public DacExtractTarget ExtractTarget { get; set; }
+
+        /// <summary>
+        /// Gets or sets whether permissions should be included in the extract
+        /// </summary>
+        public bool? ImportPermissions { get; set; }
```
zijchen, line 33: nit: Since Import has its own meaning in DacFx and this is used with Extract, `IncludePermissions` may be a better name?

zijchen (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/ExtractRequest.cs
kisantia: makes sense - updated!

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
```diff
@@ -26,7 +26,7 @@ public ExtractOperation(ExtractParams parameters, ConnectionInfo connInfo) : bas
         public override void Execute()
         {
             Version version = ParseVersion(this.Parameters.ApplicationVersion);
-            DacExtractOptions extractOptions = GetExtractOptions(this.Parameters.ExtractTarget);
+            DacExtractOptions extractOptions = GetExtractOptions(this.Parameters.ExtractTarget, this.Parameters.IncludePermissions);
```
SakshiS-harma, line 29: Are there any tests for this where we should update it?

SakshiS-harma (approved):
Are we following the default that is been used in SSDT? Until now, it isn't the case, how about after this/ADS PR?

src/Microsoft.SqlTools.ServiceLayer/DacFx/ExtractOperation.cs
kisantia: since it's an optional parameter, the existing tests aren't affected and they're passing in the Integration tests PR check :)

kisantia:
> Are we following the default that is been used in SSDT? Until now, it isn't the case, how about after this/ADS PR?

Yes the SSDT default is not including permissions, which is the same here and will also be the same in the next ADS PR to add it the option in the create project from db dialog.

PR #1725, opened 2022-10-18
Fixing names of execution plan operator icons

aasimkhan30:
This is part of fixing the issue: https://github.com/microsoft/azuredatastudio/issues/20747

I am just renaming the icon strings to match the pattern we are following for other operators. This does not actually fixes the issue but it is just a small refactor / cleanup.


src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ShowPlan/OperationTable.cs
```diff
@@ -259,7 +259,7 @@ static OperationTable()
                 new Operation("DeletedScan",            SR.Keys.LogicalOpDeletedScan),
                 new Operation("DistinctSort",           SR.Keys.LogicalOpDistinctSort),
                 new Operation("Distinct",               SR.Keys.LogicalOpDistinct),
-                new Operation("DistributeStreams",      SR.Keys.LogicalOpDistributeStreams, SR.Keys.DistributeStreamsDescription, "Parallelism_distribute.ico"),
+                new Operation("DistributeStreams",      SR.Keys.LogicalOpDistributeStreams, SR.Keys.DistributeStreamsDescription, "parallelismDistribute"),
```
Charles-Gagnon, line 262: Why are there only a couple of these with these additional parameters? Especially since the physical operations seem to all have the 4 parameters.

Charles-Gagnon (commented):
Please make sure to add context to the descriptions of PRs e.g. a linked issue and/or a description of the problem and how this is solving it.

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ShowPlan/OperationTable.cs
aasimkhan30: I am guessing there isn't an icon for other operations. This code is ported from SSMS so I do not have the original  context about it.

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ShowPlan/OperationTable.cs
Charles-Gagnon: I would highly encourage finding that information out and documenting it. Even if you don't know the original "reason" for why you should still find out what the result is and try to reason out why it's done that way (for example the main follow up question I would have is why isn't there an icon for these other operations?)

PR #1726, opened 2022-10-18
Make SerializeDataStartRequestParams properties public so they can be set properly when deserializing RPC request.

corivera:
This is a service facing change needed to fix the following issue: https://github.com/microsoft/azuredatastudio/pull/20874

We had several properties in this request object that were internal, which was keeping things like Delimiter from being set properly when deserializing the RPC request from ADS.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/Serialization.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/Serialization.cs
@@ -114,43 +114,43 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
             this.IsLastBatch = isLast;
         }
 
-        internal bool IncludeHeaders
+        public bool IncludeHeaders
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/Serialization.cs
```diff
@@ -114,43 +114,43 @@ public SerializeDataStartRequestParams()
             this.IsLastBatch = isLast;
         }
 
-        internal bool IncludeHeaders
+        public bool IncludeHeaders
```
Charles-Gagnon, line 117: Huh, so these being internal meant that they weren't being assigned values from the request when being deserialized?

corivera: Yeah, the request object gets created in Microsoft.SqlTools.Hosting, but the Params object is in ServiceLayer. All the internal fields always showed up as null, even though the corresponding properties had real values in the RPC message.

Charles-Gagnon (approved):

PR #1727, opened 2022-10-20
fix schema comparing empty project showing more differences than expected

kisantia:
Fixes https://github.com/microsoft/azuredatastudio/issues/19528. When comparing an empty project to a database or another project that isn't empty, there were unexpected differences showing up. These differences aren't excludable and also aren't "included" in the schema compare, so it doesn't make sense to show them.

I still need to dig more in DacFx to figure out why these even show up as differences that aren't excludable and included, but this is the fix. I was wondering why I wasn't able to repro this in the DacFx test I wrote, and then figured out it's because in the DacFx schema compare tests we filter out differences that aren't included by doing `(this.ComparisonResult.Differences as List<SchemaDifference>).RemoveAll(d => !d.Included)` before checking that the difference count is as expected, so someone must have also run into this before when writing those tests in DacFx.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
@@ -113,6 +113,9 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                 this.Differences = new List<DiffEntry>();
                 if (this.ComparisonResult.Differences != null)
                 {
+                    // filter out not included and not excludeable differences
+                    (this.ComparisonResult.Differences as List<SchemaDifference>).RemoveAll(d => !d.Included && !d.IsExcludable);
+
                     foreach (SchemaDifference difference in this.ComparisonResult.Differences)
                     {
                         DiffEntry diffEntry = SchemaCompareUtils.CreateDiffEntry(difference, null);
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
@@ -22,6 +22,8 @@ namespace Microsoft.SqlTools.ServiceLayer.IntegrationTests.SchemaCompare
 {
     internal static class SchemaCompareTestUtils
     {
+        private static string sqlProjectsFolder = Path.Combine("..", "..", "..", "SchemaCompare", "SqlProjects");
+
         internal static void VerifyAndCleanup(string path)
         {
             // verify it was created...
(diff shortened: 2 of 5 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
```diff
@@ -61,12 +63,17 @@ internal static string CreateDacpac(SqlTestDb testdb)
...
+        /// </summary>
+        /// <param name="testdb">Database to create the sql project from</param>
+        /// <param name="projectName">Name of the project</param>
+        /// <returns>Full path to the project folder</returns>
         internal static string CreateProject(SqlTestDb testdb, string projectName)
         {
             var result = GetLiveAutoCompleteTestObjects();
-            string folderPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "SchemaCompareTest", projectName);
-            Directory.CreateDirectory(folderPath);
-            File.Create(Path.Combine(folderPath, projectName + ".sqlproj")).Close();
```
kisantia, line 69: this was previously creating an empty file with no contents that had the file extension `.sqlproj`. The actual .sqlproj isn't used in the schema compare, but I updated this to actually create a .sqlproj with the expected xml so it isn't confusing in the future.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -113,6 +113,9 @@ public void Execute(TaskExecutionMode mode)
                 this.Differences = new List<DiffEntry>();
                 if (this.ComparisonResult.Differences != null)
                 {
+                    // filter out not included and not excludeable differences
+                    (this.ComparisonResult.Differences as List<SchemaDifference>).RemoveAll(d => !d.Included && !d.IsExcludable);
+
```
SakshiS-harma, line 118: Would this also fix the issues where the script generation has the excludable stuff in?

kisantia: I'm not sure. It might if those actually aren't excludable. We can ask the customer to try in insiders after this is checked in or ask them for a repro so we can test it out.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareTestUtils.cs
Benjin: Constructing these sorts of tests should be much easier with the XML stuff I'm working on!

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareOperation.cs
SakshiS-harma: Sounds good.

SakshiS-harma (approved):

PR #1728, opened 2022-10-24
Akma/login migrations

AkshayMata:
In this PR, we make the appropriate backend service changes in order to enable the login migrations feature in the SQL migration extension.

Changes include:
- updating the Microsoft.SqlServer.Migration.Login NuGet version to the latest version
- adding a new request handler for StartLoginMigrations calls, which makes the appropriate calls to the login NuGet
- adding ExtensionMethod helper to properly combine exception maps login migration nuget calls


src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/StartLoginMigration.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Connection string to connect to target
+        /// </summary>
+        public string TargetConnectionString { get; set; }
+
+        /// <summary>
+        /// List of logins to migrate
+        /// </summary>
+        public List<string> LoginList { get; set; }
```
kavitham10, line 27: Would you also need something like CSVFilePath of logins in case user wants to send a large batch of logins for migrations? That is what the NuGet uses for console app/PowerShell.

src/Microsoft.SqlTools.ServiceLayer/Migration/Helper/ExtensionMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+        public static void AddExceptions(this IDictionary<string, IEnumerable<ReportableException>> exceptionMap1, IDictionary<string, IEnumerable<ReportableException>> exceptionMap2)
+        {
+            foreach (var keyValuePair2 in exceptionMap2)
+            {
+                // If the dictionary already contains the key then merge them
+                if (exceptionMap1.ContainsKey(keyValuePair2.Key))
+                {
+                    foreach (var value in keyValuePair2.Value)
+                    {
+                        exceptionMap1[keyValuePair2.Key].Append(value);
```
kavitham10, line 23: Does this not give an error? I tried this and it said IEnumerable doesn't have a definition of Append.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -305,6 +311,40 @@ public void InitializeService(ServiceHost serviceHost)
             }
         }
 
+        internal async Task HandleStartLoginMigration(
+            StartLoginMigrationParams parameters,   
+            RequestContext<StartLoginMigrationResults> requestContext)
+        {
+            try
+            {
+                // TODO AKMA: Do we need aad domain name and sql cred? If so, i'll need to set up login options as a param
```
kavitham10, line 320: Regarding the comment, yes we do need AAD domain name because that is what is used for Windows logins migrations. We probably don't need Sql creds.

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/StartLoginMigration.cs
AkshayMata: So the UI doesn't give CX the option of passing in a CSVFilePath. Right now the workflow lists all the logins and the CX needs to select which logins to migrate. Do we want to add this functionality as well?

kavitham10: Oh you're right sorry, forgot that the UI will have table list of logins that are selectable. Please ignore comment!

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
kavitham10: So just to elaborate a bit more, if the table list of logins on the UI has Windows logins and user wants to migrate those Windows logins, then we need to have a provision to ask user for AAD domain name so that we can do the Windows user to AAD user conversion in the backend. I am not sure how in terms of UI this will translate but just wanted to let you know the use case for AAd domain name.

src/Microsoft.SqlTools.ServiceLayer/Migration/Helper/ExtensionMethods.cs
AkshayMata: i'm not seeing any build issues. Did you have build errors or run time errors?

kavitham10: For a similar change of dictionary check in the NuGet side, I was just trying this change out in a separate console app and this was what I got, hence I used Concat:

image

If it is working on your end then that's good, maybe I am missing something.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
AkshayMata: Added the AAD domain name as a param and I pass it into the LoginMigrations construcor now. The UI will control whether the param is "" or has a value (for windows auth case)

src/Microsoft.SqlTools.ServiceLayer/Migration/Helper/ExtensionMethods.cs
AkshayMata: hmm i see. not seeing any errors on my end: 
image

not sure where the discrepancy comes from. i'll try running an end to end test to make sure it's working.

Packages.props
```diff
@@ -27,8 +27,9 @@
 		<PackageReference Update="Microsoft.Azure.Kusto.Data" Version="9.0.4" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Language" Version="9.0.4" />
 		<PackageReference Update="Microsoft.SqlServer.Assessment" Version="[1.1.9]" />
-		<PackageReference Update="Microsoft.SqlServer.Migration.Assessment" Version="1.0.20221014.16" />
-		<PackageReference Update="Microsoft.SqlServer.Management.SqlParser" Version="160.22519.0" />
+		<PackageReference Update="Microsoft.SqlServer.Migration.Assessment" Version="1.0.20220816.40" />
```
raymondtruong, line 30: maybe pull from latest main, `Microsoft.SqlServer.Migration.Assessment` should already be updated in my last PR: https://github.com/microsoft/sqltoolsservice/pull/1734

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/StartLoginMigration.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// List of logins to migrate
```
raymondtruong, line 30: nit: update comment

Packages.props
raymondtruong: (I see you already did that yesterday, I just merged mine this morning)

src/Microsoft.SqlTools.ServiceLayer/Migration/Helper/ExtensionMethods.cs
AkshayMata: tested this and it is working

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/StartLoginMigration.cs
AkshayMata: done

Packages.props
AkshayMata: done!

brian-harris (approved):

PR #1731, opened 2022-10-25
Remove testing cancel request from VerifySchemaCompareServiceCalls

kisantia:
I wasn't able to repro the test failure locally, but the VerifySchemaCompareServiceCalls test has been flaky in the integration test pipeline. From the stack trace in the test pipeline failure, this looks like it's failing due to a race condition where the cancellation request hasn't been handled before the validation happens.

Removing this part of the test for now to see if it helps with stabilizing the test.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
@@ -1217,25 +1217,6 @@ WITH VALUES
                 };
 
                 await SchemaCompareService.Instance.HandleSchemaCompareRequest(schemaCompareParams, schemaCompareRequestContext.Object);
-
-                // Schema compare Cancel call
-                var schemaCompareCancelRequestContext = new Mock<RequestContext<ResultStatus>>();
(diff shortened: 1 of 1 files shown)


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1216,25 +1216,6 @@ public async Task VerifySchemaCompareServiceCalls()
                     DeploymentOptions = new DeploymentOptions()
                 };
 
-                await SchemaCompareService.Instance.HandleSchemaCompareRequest(schemaCompareParams, schemaCompareRequestContext.Object);
```
Charles-Gagnon, line 1219: Instead of just deleting (presumably) useful test code - could you instead add more logging so that if it does fail you can better identify where the failure is occurring?

Or is this not testing anything useful right now?

kisantia: There's already another test that tests schema compare cancellation that's been reliable, so testing it here doesn't add much value

SakshiS-harma (approved):

alanrenmsft (approved):

Charles-Gagnon (approved):

PR #1733, opened 2022-10-26
Update dependencies

cheenamalhotra:


caohai (approved):

aasimkhan30 (approved):

src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="Microsoft.SqlServer.DacFx" />
     <PackageReference Include="Microsoft.SqlServer.Management.SmoMetadataProvider" />
     <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" />
+	<PackageReference Include="System.Configuration.ConfigurationManager" />
```
Charles-Gagnon, line 37: Fix spacing

Charles-Gagnon (approved):
It should be mentioned that DacFX is currently bringing in 6.0.0 of System.Security.Permissions which is what is being used at runtime so this should be safe - it's more about being explicit with our dependencies.

Is that the same reason you updated System.Configuration.ConfigurationManager?

src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj
cheenamalhotra:
```suggestion
    <PackageReference Include="System.Configuration.ConfigurationManager" />
```

PR #1736, opened 2022-10-27
tde migration integration

junierch:
This PR adds the backend functionality to perform certificate migration to managed instances through the SQL migration extension.

Changes include:
* Add the  _Microsoft.SqlServer.Migration.Tde_ NuGet package.
* Add a new request _migration/tdemigration_ to handle the migration of multiple database certificates during one request.


Packages.props
```diff
@@ -44,5 +44,10 @@
 		<PackageReference Update="Azure.Storage.Blobs" Version="12.10.0" />
 		<PackageReference Update="coverlet.collector" Version="3.1.2" />
 		<PackageReference Update="coverlet.msbuild" Version="3.1.2" />
+
+		<PackageReference Update="Azure.ResourceManager" Version="1.3.1" />
```
Charles-Gagnon, line 48: Why are you adding these other references? They aren't being added in any of the projects so the additions here aren't going to do anything.

Charles-Gagnon:
We really should avoid checking nuget packages into the source - is there a reason you're not putting these up on nuget?

Charles-Gagnon:
This is a public repo so we can't use internal feeds. 

It would be highly preferable to put these packages both in nuget. Is there a reason you're unable to do that?

raymondtruong:
@Charles-Gagnon,

Our thinking is that these nupkgs that we are checking into the source (already for assessments/SKU recommendations, now for TDE migration, and soon for login migration) were not intended to be published publicly as official Microsoft packages, but rather as internal packages only meant for use with our own tools and services, hence our existing internal NuGet feed in ADO. For example, every `Microsoft.SqlServer.*` package published on the NuGet.org feed has hundreds of thousands of downloads, whereas our assessments/recommendations functionality were specifically designed and built for consumption in ADS and Azure Migrate, so I don't know if it's appropriate to publish them to the world as official Microsoft packages. 

Of course checking the nupkg into the source doesn't make it any less visible to the world, the point is just that they weren't really developed with public consumption in mind. Is there another option to publish to some other public feed? What would you suggest in this situation?

Charles-Gagnon:
We publish "internal" stuff on nuget all the time - e.g. https://www.nuget.org/packages/Microsoft.Data.SqlClient.SNI.runtime/5.1.0-preview1.22278.1

Just because it's on nuget doesn't mean it's something we have to officially support external customers using. And then you get the benefits of at least making it possible for external users to get the package and use it themselves if they want (which as you already mentioned they can get the package through this repo anyways so it's not like it's actually hidden) which is generally something customers appreciate and we should strive to be open to. 

So my recommendation is to add it to nuget.

raymondtruong:
@Charles-Gagnon

I see, thanks for calling that out. I also see that the package you linked to also mentions this in their readme: 
image

We can take a look into publishing to NuGet, but the internal documentation mentions that this process can take a while (making sure the package is compliant, then onboarding the team, etc). At least in this iteration, we want to have these packages available in the next release of ADS and as we understand the deadline for sqltoolsservice changes is the end of this month. Can we proceed with manually checking in the nupkg for now while we look into publishing?

Charles-Gagnon:
Yeah it's fine to go forward with this now, but I definitely would like to see this process started now. We can chat offline about it - I can help with getting stuff set up if needed too.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -305,140 +308,143 @@ public void InitializeService(ServiceHost serviceHost)
...
-                // generate SQL DB recommendations, if applicable
-                if (parameters.TargetPlatforms.Contains("AzureSqlDatabase"))
-                {
-                    Stopwatch sqlDbStopwatch = new Stopwatch();
-                    sqlDbStopwatch.Start();
-
-                    prefs.EligibleSkuCategories = GetEligibleSkuCategories("AzureSqlDatabase", parameters.IncludePreviewSkus);
-                    resultSet.sqlDbResults = provider.GetSkuRecommendation(prefs, req);
+        internal RecommendationResultSet GenerateBaselineRecommendations(SqlInstanceRequirements req, GetSkuRecommendationsParams parameters)
+        {
```
raymondtruong, line 312: I guess git messed up the diff -- presumably there's no actual change to any of this SKU recommendation code besides just reformatting the brackets here?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -827,5 +833,42 @@ public void Dispose()
                 disposed = true;
             }
         }
+
+        internal async Task HandleCertificateMigrationRequest(
```
raymondtruong, line 837: I would suggest wrapping the whole request handler in a try/catch then use `requestContext.SendError()` to send back any exceptions, sort of like this:

https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs#L302

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/CertificateMigrationRequest.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.Migration.Contracts
+{
+    public class CertificateMigrationParams
+    {
+        /// <summary>
+        /// List of databses to migrate the certificates
+        /// </summary>
+        public List<string> EncryptedDatabases { get; set; } = new List<string>();
+
```
raymondtruong, line 17: nit: blank line

src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/CertificateMigrationRequest.cs
```diff
@@ -0,0 +1,68 @@
...
+    public class CertificateMigrationResult
+    {
+        public List<CertificateMigrationEntryResult> MigrationStatuses { get; set; } = new List<CertificateMigrationEntryResult>();
+    }
+
+    public class CertificateMigrationEntryResult
+    {
+        public string DbName { get; set; }
+        public bool Success { get; set; }
+
```
raymondtruong, line 55: nit: blank line, also can you add docs for all the other public fields

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -305,140 +308,143 @@ public void InitializeService(ServiceHost serviceHost)
...
-                    // if no result was generated, create a result with a null SKU
-                    if (!resultSet.sqlMiResults.Any())
+                    // if there are fewer recommendations than expected, find which databases didn't have a result generated and create a result with a null SKU
+                    List<string> databasesWithRecommendation = resultSet.sqlDbResults.Select(db => db.DatabaseName).ToList();
+                    foreach (var databaseWithoutRecommendation in parameters.DatabaseAllowList.Where(db => !databasesWithRecommendation.Contains(db)))
                     {
-                        resultSet.sqlMiResults.Add(new SkuRecommendationResult()
+                        resultSet.sqlDbResults.Add(new SkuRecommendationResult()
                         {
+                            //SqlInstanceName = sqlDbResults.FirstOrDefault().SqlInstanceName,
```
brian-harris, line 340: is this still needed?

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
junierch: Yep, git is showing too much. The only change was the opening curly brace moved to the next line.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -305,140 +308,142 @@ public void InitializeService(ServiceHost serviceHost)
...
+            {
+                EligibleSkuCategories = null,       // eligible SKU list will be adjusted with each recommendation type
+                ScalingFactor = parameters.ScalingFactor / 100.0,
+                TargetEnvironment = TargetEnvironmentType.Production
+            };
 
-                    sqlDbStopwatch.Stop();
-                    resultSet.sqlDbDurationInMs = sqlDbStopwatch.ElapsedMilliseconds;
+            // generate SQL DB recommendations, if applicable
+            if (parameters.TargetPlatforms.Contains("AzureSqlDatabase"))
```
brian-harris, line 324: is there an enum / constant for "AzureSqlDatabase" to use instead.

brian-harris (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
junierch: Agree, I didn't modify this code, only changed the line for the opening `{` while applying format to the document. We may need to create an item to remove all hardcoded strings in this file.

Packages.props
```diff
@@ -44,5 +44,7 @@
 		<PackageReference Update="Azure.Storage.Blobs" Version="12.10.0" />
 		<PackageReference Update="coverlet.collector" Version="3.1.2" />
 		<PackageReference Update="coverlet.msbuild" Version="3.1.2" />
+
```
raymondtruong, line 47: nit: blank line

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -563,117 +565,119 @@ public void InitializeService(ServiceHost serviceHost)
...
+                resultSet.sqlDbReportPath = sqlDbRecommendationReportFullPath + ".html";
+            }
 
-                // generate SQL MI recommendations, if applicable
-                if (parameters.TargetPlatforms.Contains("AzureSqlManagedInstance"))
-                {
-                    Stopwatch sqlMiStopwatch = new Stopwatch();
-                    sqlMiStopwatch.Start();
+            // generate SQL MI recommendations, if applicable
+            if (parameters.TargetPlatforms.Contains("AzureSqlManagedInstance"))
```
AbrantesMar, line 598: this string repeats itself, wouldn't it be better to put it in a constant

"AzureSqlManagedInstance"

Charles-Gagnon:
Bulk closing stale PRs, please update and reopen if you plan on continuing this work!

PR #1737, opened 2022-10-27
Support Boolean type for Encrypt option value

cheenamalhotra:
As a feedback from https://github.com/microsoft/azuredatastudio/pull/21010, adding bool support to encrypt property for backward compatibility

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
@@ -147,7 +147,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


alanrenmsft:
what are the expected values?

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -153,7 +153,13 @@ public string Encrypt
         {
             get
             {
-                return GetOptionValue<string>("encrypt");
+                string value = GetOptionValue<string>("encrypt");
```
Charles-Gagnon, line 156: Update doc comment

cheenamalhotra:
> what are the expected values?

For Encrypt, all these continue to be valid:
"optional, "mandatory", "strict", "yes", "no", "true" and "false"

kisantia (approved):

PR #1739, opened 2022-10-29
Fix Comparable implementation for ConnectionDetails

cheenamalhotra:
Fixes edit connection behavior in Azure Data Studio reported in https://github.com/microsoft/azuredatastudio/issues/20985

The design of comparable was wrong as it didn't consider all connection properties for identifying change in connection settings.
This led to same old connection settings (cached ConnectionInfo) being fetched from OwnerToConnectionMap that maps ownerUri with connectionInfo in Connection Service class.

We always need to compare all connection settings so we can update stored connection Info to latest settings configured by user.

Expected behavior with fix:

https://user-images.githubusercontent.com/13396919/198814651-75d1f15f-753c-4495-afd2-17021d0466be.mp4

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
@@ -593,40 +593,40 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection.Contracts
 
         public bool IsComparableTo(ConnectionDetails other)
             => other != null
-            && ApplicationIntent == other.ApplicationIntent
-            && ApplicationName == other.ApplicationName
-            && AttachDbFilename == other.AttachDbFilename
-            && AuthenticationType == other.AuthenticationType
-            && AzureAccountToken == other.AzureAccountToken
-            && ColumnEncryptionSetting == other.ColumnEncryptionSetting
-            && ConnectionString == other.ConnectionString
+            && string.Equals(ApplicationIntent, other.ApplicationIntent, System.StringComparison.InvariantCultureIgnoreCase)
+            && string.Equals(ApplicationName, other.ApplicationName, System.StringComparison.InvariantCultureIgnoreCase)
+            && string.Equals(AttachDbFilename, other.AttachDbFilename, System.StringComparison.InvariantCultureIgnoreCase)
+            && string.Equals(AuthenticationType, other.AuthenticationType, System.StringComparison.InvariantCultureIgnoreCase)
+            && string.Equals(AzureAccountToken, other.AzureAccountToken, System.StringComparison.InvariantCultureIgnoreCase)
+            && string.Equals(ColumnEncryptionSetting, other.ColumnEncryptionSetting, System.StringComparison.InvariantCultureIgnoreCase)
+            && string.Equals(ConnectionString, other.ConnectionString, System.StringComparison.InvariantCultureIgnoreCase)
             && ConnectRetryCount == other.ConnectRetryCount
             && ConnectRetryInterval == other.ConnectRetryInterval
             && ConnectTimeout == other.ConnectTimeout
-            && CurrentLanguage == other.CurrentLanguage
-            && DatabaseDisplayName == other.DatabaseDisplayName
-            && DatabaseName == other.DatabaseName
-            && EnclaveAttestationProtocol == other.EnclaveAttestationProtocol
-            && EnclaveAttestationUrl == other.EnclaveAttestationUrl
-            && Encrypt == other.Encrypt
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -592,31 +592,41 @@ public string AzureAccountToken
...
-            if (!string.IsNullOrEmpty(DatabaseName)
-                && !string.IsNullOrEmpty(other.DatabaseName)
-                && DatabaseName != other.DatabaseName)
-            {
-                return false;
-            }
-
-            return true;
-        }
+            => other != null
```
kburtram, line 595: I'm a little concerned if we have a different "connection equality" check in STS than in ADS that we can get weird side-effects in various scenarios when editing the connections in ADS.  If you've done sufficient testing across these kinds of end-to-end scenarios then this is fine.

kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
```diff
@@ -592,31 +592,41 @@ public string AzureAccountToken
         }
 
         public bool IsComparableTo(ConnectionDetails other)
```
Charles-Gagnon, line 594: Since it's now comparing all properties we should probably just make this `public override bool Equals(object? obj)`

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
Charles-Gagnon: My main concern here is that if STS is making changes to the connection details itself that could result in unexpected disconnects/reconnects since it would have different properties than what ADS has which could break customers if they're using temp objects or other transient functionality. 

It would be great to have a better sense of how/when changes to the connection details could happen (both in ADS and STS) to make sure we aren't going to be regressing any normal behavior here.

Charles-Gagnon (approved):
This seems pretty risky to do since it'll be hard to know for sure that there aren't any scenarios that will be affected by this but give how it currently works with editing connections I think it's reasonable to do and just make sure we're taking extra care to look over places where connections are made/modified (such as the database dropdown in the query editor) to make sure things don't expect the current behavior to stay. 

One thing I think we need to do is start writing down the specific goals and behavior we expect STS/ADS to follow with regards to connections - and then making tests to make sure those contracts are held. That will help make decisions like this easier and less risky in the future - as well as make sure that we don't have regressions in such a critical path.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
cheenamalhotra: > I'm a little concerned if we have a different "connection equality" check in STS than in ADS that we can get weird side-effects in various scenarios when editing the connections in ADS.

That shouldn't be of concern, as ADS seems to delegate all settings comparisons to STS layer and only works with Connection Profile (that is identity) with whatever settings it's provided with. STS is responsible to cache identity and settings combo in `OwnerToConnectionMap` and that's the heart of all saved settings that apply to connection profiles. ADS would always show new properties on modifying connection settings even if STS did not update the settings in the map so that was a big design problem and affected reliability of ADS connections. ADS is always doing the right thing.
 
> My main concern here is that if STS is making changes to the connection details itself that could result in unexpected disconnects/reconnects since it would have different properties than what ADS has which could break customers if they're using temp objects or other transient functionality.

There won't be any unexpected disconnects, as the connections will only be updated with new settings on a successful save and will only affect new connections created from that profile. And that's exactly what is needed to be able to edit a connection. We should be forcing a reconnect with new advanced options before saving it, and on saving, new properties must apply on everything new that starts with this edited connection.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionDetails.cs
cheenamalhotra: I thought about that, but we are still not covering 'GroupId' in this comparison, as it's managed by ADS layer and didn't seem to make a difference when editing a connection from STS layer.

Charles-Gagnon: Gotcha, makes sense

Charles-Gagnon, line 594: I'd suggest adding a doc comment to this explaining what properties it is/isn't comparing and what the general purpose of this is so others know what to expect.

PR #1655, opened 2022-08-27
Revert  async dispatcher change 08c74be

kburtram:
This reverts the async dispatcher change in https://github.com/microsoft/sqltoolsservice/commit/08c74be51feb0ae1f04652a13fe3dcebe9cf70c6.  Unfortunately there were some merge conflicts so not a clean `git revert`.  This change appears to be the underlying cause of https://github.com/microsoft/azuredatastudio/issues/20472, which appears to be related to OE/Connection events getting processed out-of-order/dropped on slower SQL Connections.

PR #1741, opened 2022-11-01
Re-enable parallel message processing

caohai:
Follow-up PR for https://github.com/microsoft/sqltoolsservice/pull/1596 and https://github.com/microsoft/sqltoolsservice/pull/1655

This PR adds back the parallel message processing and adds an additional flag at handler level to control if this handler supports running in parallel.

Some handlers are already using `Task.Run()` so they are good without this change.

If the user doesn't choose to enable the `ParallelMessageProcessing` flag, the behavior should be exactly the same as the behavior without this change.

| Service Name      | Marked As Supporting Parallel Processing | Already using Task.Run() |Comment|
| ----------- | ----------- |----------- |-|
|AdminService.cs |YES|YES|Enabling(already async)|
|AgentService.cs |YES|NO|Enabling (no concern for running in parallel)|
|BlobService.cs |YES|NO|Enabling (no concern for running in parallel)|
|AzureFunctionsService.cs |YES|NO|Enabling (no concern for running in parallel)|
|CmsService.cs |YES |YES|Enabling(already async)|
|ConnectionService.cs |YES|YES|Enabling(already async)|
|DacFxService.cs |YES|YES|Enabling(already async)|
|DisasterRecoveryService.cs |YES|YES|Enabling(already async)|
|EditDataService.cs |YES|NO|Enabling (no concern for running in parallel)|
|ExecutionPlanService.cs |YES|NO|Enabling (no concern for running in parallel)|
|FileBrowserService.cs |YES|YES|Enabling(already async)|
|TSqlFormatterService.cs |YES|YES|Enabling(already async)|
|ExternalLanguageService.cs |YES|NO|Enabling (no concern for running in parallel)|
|LanguageService.cs |NO||order of msgs can matter, example: HandleTokenRefreshedNotification|
|MetadataService.cs |YES|NO|Enabling (no concern for running in parallel)||
|MigrationService.cs |YES|NO|Enabling (no concern for running in parallel)||
|ModelManagementService.cs |YES|NO|Enabling (no concern for running in parallel)||
|NotebookConvertService.cs |YES|NO|Enabling (no concern for running in parallel)||
|ObjectExplorerService.cs |NO|YES|Not enabling as OE could hang||
|ObjectManagementService.cs |YES|NO|Enabling (no concern for running in parallel)|
|ProfilerService.cs |YES|NO|Enabling (no concern for running in parallel)|
|QueryExecutionService.cs |YES|NO| Enabling(already async)|
|SerializationService.cs |YES|YES|Enabling(already async)|
|SchemaCompareService.cs |YES|YES|Enabling(already async)|
|ScriptingService.cs |YES|YES|Enabling(already async)|
|SecurityService.cs |YES|YES|Enabling(already async)|
|ServerConfigService.cs |YES|NO|Enabling (no concern for running in parallel)|
|SqlAssessmentService.cs |YES|YES|Enabling(already async)|
|TableDesignerService.cs |YES|YES|Enabling(already async)|
|TaskService.cs |YES|YES|Enabling(already async)|
|WorkspaceService.cs |NO|   |Not enabling as there could be doc out of sync error |


alanrenmsft:
@caohai  Thanks for following up on this. Could you please add another column Reason explaining why it is not enabled/enabled? also, for the services/handlers do not support parallel processing, should also add comment to the code.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/IProtocolEndpoint.cs
```diff
@@ -18,15 +18,18 @@ public interface IProtocolEndpoint : IEventSender, IRequestSender
     {
         void SetRequestHandler<TParams, TResult>(
             RequestType<TParams, TResult> requestType,
-            Func<TParams, RequestContext<TResult>, Task> requestHandler);
+            Func<TParams, RequestContext<TResult>, Task> requestHandler,
+            bool isParallelProcessingSupported = false);
```
Charles-Gagnon, line 22: Please add doc comments for these methods and at the very least add a param comment for the new param being added (but if you could do the others while you're here that'd be great)

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -27,14 +27,22 @@ public class MessageDispatcher
...
 
+        internal Dictionary<string, bool> eventHandlerParallelismMap =
+            new Dictionary<string, bool>();
+
         private Action<Message> responseHandler;
 
         private CancellationTokenSource messageLoopCancellationToken =
             new CancellationTokenSource();
 
+        private SemaphoreSlim semaphore = new SemaphoreSlim(10); // Limit to 10 threads
```
Charles-Gagnon, line 44: Please add the reason why 10 was chosen to the comment

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -27,14 +27,22 @@ public class MessageDispatcher
         internal Dictionary<string, Func<Message, MessageWriter, Task>> requestHandlers =
             new Dictionary<string, Func<Message, MessageWriter, Task>>();
 
+        internal Dictionary<string, bool> requestHandlerParallelismMap =
```
Charles-Gagnon, line 30: Instead of having a separate dictionary just to keep track of bools I would suggest using a tuple type in the requestHandlers dictionary

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples

`internal Dictionary<string, (Func<Message, MessageWriter, Task>, bool)> requestHandlers =
            new Dictionary<string, (Func<Message, MessageWriter, Task>, bool)>();`

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -324,13 +354,15 @@ private async Task ListenForMessages(CancellationToken cancellationToken)
 
             if (handlerToAwait != null)
             {
-                if (this.ParallelMessageProcessing)
+                if (this.ParallelMessageProcessing && isParallelProcessingSupported)
```
Charles-Gagnon, line 357: This currently means that any non-parallel enabled messages will block all other messages from being processed until it's completed. But that means we're only getting some of the benefit of going parallel in the first place - ideally we should be having all parallel messages kicked off as soon as they're read in while any non-parallel ones will be handled sequentially.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -324,13 +354,15 @@ private async Task ListenForMessages(CancellationToken cancellationToken)
...
                 {
                     // Run the task in a separate thread so that the main
                     // thread is not blocked.
+                    await semaphore.WaitAsync();
```
Charles-Gagnon, line 361: I'd suggest adding a comment explaining why you're doing this.

Charles-Gagnon (commented):
I'm quite wary of bringing this back without knowing what the problem was in the first place. Especially with so many other huge changes going in this month already.

1. Are you planning on having this in the November release?
2. Have you done any more investigation into what the original problem was?
3. What's the plan for if we start seeing more issues?

Given that we've lived with the current functionality this entire time I don't see this being something we need to rush back into until we have more confidence there aren't regressions, so my suggestion is we work some more on building up that confidence before moving ahead with this.

alanrenmsft:
@Charles-Gagnon makes sense, @caohai let's do it after the Nov release so that we could ask the user to try the insider build and see whether they could reproduce the issue with Insider builds.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -112,43 +120,55 @@ public void Stop()
...
+                        catch (Exception ex)
+                        {
+                            Logger.Error($"{requestType.MethodName} : {ex}");
+                            await requestContext.SendError(ex.Message);
                         }
+                    };
 
-                        await requestHandler(typedParams, requestContext);
+                    if (!this.ParallelMessageProcessing || !isParallelProcessingSupported) {
+                        await func();
```
caohai, line 167: This part is to make sure if a user does not explicitly turn on the `ParallelMessageProcessing` flag, the message handling behavior will remain unchanged.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
caohai: I gave this a try but personally the readability feels worse to me (even with the use of field name properties), I might stick with the additional dict here.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -166,39 +186,52 @@ public void Stop()
...
-                    try
-                    {                
-                        if (eventMessage.Contents != null)
-                        {
-                            try
-                            {
-                                typedParams = eventMessage.Contents.ToObject<TParams>();
-                            }
-                            catch (Exception ex)
+                    Func<Task> func = async () =>
```
cheenamalhotra, line 203: What you're defining here with `async (eventMessage, messageWriter) => { }` is already a function, why are you creating another function `Func<Task>` inside and awaiting on it?

Same comment for above method.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -324,28 +360,31 @@ private async Task ListenForMessages(CancellationToken cancellationToken)
...
+                if (this.ParallelMessageProcessing && isParallelProcessingSupported)
                 {
                     // Run the task in a separate thread so that the main
-                    // thread is not blocked.
+                    // thread is not blocked. Use semaphore to limit the degree of parallelism.
+                    await semaphore.WaitAsync();
                     _ = Task.Run(() =>
                     {
-                        _ = RunTask(handlerToAwait);
+                        _ = RunTask(handlerToAwait, messageToDispatch);
```
cheenamalhotra, line 370: Shouldn't this be awaited before releasing semaphore?

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
caohai: Yes we are only getting part of the benefit here. I attempted to get msg loop on its own thread but that will be a big change so we are going this route for now.

alanrenmsft:
@caohai please work with @cheenamalhotra and make sure the logs can capture the query hang issue she is facing.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
cheenamalhotra: Pasting a simple example for reference, no code would be executed until you await on the function, as this is an async task object you're passing around.

```
Func<int, Task> func = async(i) =>
{
    Console.WriteLine("2. Non awaited result");
    await Console.Out.WriteLineAsync("3. Awaited result");
};
var function = func;
await Task.Delay(100);
await Console.Out.WriteLineAsync("1. Before await");
await function(1);
```

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Messaging/MessageDispatcherTests.cs
```diff
@@ -84,5 +86,67 @@ public void OnListenTaskCompletedFaultedTaskTest()
...
+            bool okayToEnd = false;
+            mockChannel.Setup(c => c.MessageReader.ReadMessage())
+                .Returns(Task.FromResult(Message.Request("1", "test/requestType", null)));
+            var dispatcher = new MessageDispatcher(mockChannel.Object);
+            dispatcher.ParallelMessageProcessing = parallelMessageProcessing;
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            var handler = async (int _, RequestContext<int> _) =>
+            {
+                Thread.Sleep(msForEachRequest);
+                unfinishedRequestCount.Wait();
```
cheenamalhotra, line 123: What's the purpose of waiting sync here?
You can await with `WaitAsync()` on it.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
cheenamalhotra: Also tested the original code from @alanrenmsft , it doesn't have negative consequences from execution perspective. 
It works normally as expected:

```
Func<int, Task> func = (i) => Task.Run(async () => 
{
    Console.WriteLine("2. Non awaited result");
    await Console.Out.WriteLineAsync("3. Awaited result");
});
var function = func;
await Task.Delay(100);
await Console.Out.WriteLineAsync("1. Before await");
await function(1);
```

caohai: The difference here is that the function actually gets invoked here with `handlerToAwait = requestHandler(messageToDispatch, messageWriter);`, I'll move this call to where await happens.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Messaging/MessageDispatcherTests.cs
```diff
@@ -84,5 +86,67 @@ public void OnListenTaskCompletedFaultedTaskTest()
...
+                    stopwatch.Stop();
+                    okayToEnd = true;
+                }
+                await Task.CompletedTask;
+            };
+
+            dispatcher.SetRequestHandler(requestType, handler, false, true);
+            dispatcher.Start();
+
+            while (true)
```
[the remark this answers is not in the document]
cheenamalhotra: This can be ignored, thanks for meeting offline and explaining the requirement!

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -284,19 +300,13 @@ private async Task ListenForMessages(CancellationToken cancellationToken)
             Message messageToDispatch,
             MessageWriter messageWriter)
         {
-            Task handlerToAwait = null;
+            Func<Message, MessageWriter, Task> handlerToAwait = null;
```
caohai, line 303: changed hanlderToAwait to a Func instead of a Task and removed `handlerToAwait = requestHandler(messageToDispatch, messageWriter);` so we don't have to wait for the potential slow sync code in the handler. The invocation of hanlderToAwait happens later when we awaits.

caohai:
@alanrenmsft This PR should be ready for another review now.

alanrenmsft (approved):

alanrenmsft:
there are still test failures.

caohai:
> nfinishedRequestCount.Wait();

The test failed was the one I added, the assertion is a bit strict and can become flaky if the build machine has small amount of logical cores. I'll take a look to loose the assertion a bit.

PR #1747, opened 2022-11-04
Build Managed Batch Parser NuGet package

cheenamalhotra:
This PR enables building `Microsoft.SqlTools.ManagedBatchParser` NuGet package with below changes:
- Fixes versioning scheme to not use hardcoded version string.
  - New variables supported in pipelines:
    - **Major**: specifies Major version
    - **Minor**: specifies Minor version
    - **StableRelease**: true/false: specifies if Stable build must be done instead.
- Introduces Nuspec based packaging + symbols indexing
- Strong name key signing for .NET Framework 4.7.2 DLL.
  - We also should ideally sign the .NET 6.0 DLL but I had issues due to the `InternalsVisibleTo` attribute and its train of dependencies. Since its not critical, we can do it another time.


Directory.Build.props
```diff
@@ -3,8 +3,15 @@
...
     <TargetFramework>net6.0</TargetFramework>
-  </PropertyGroup>
-  <PropertyGroup>
+    
+    <Build_BuildNumber Condition="'$(Build_BuildNumber)' == ''">1.dev</Build_BuildNumber>
+    <AssemblyVersion>3.0.0.0</AssemblyVersion>
+    <VersionPrefix>3.0.0</VersionPrefix>
+    <VersionSuffix>preview$(Build_BuildNumber)</VersionSuffix>
+    <PackageVersion>$(VersionPrefix)-$(VersionSuffix)</PackageVersion>
+    <AssemblyFileVersion>$(Version)</AssemblyFileVersion>
```
Matteo-T, line 12: Do you have the commitID available somewhere in your pipeline? If you do,  you could add 

```xml
<InformationalVersion>$(AssemblyFileVersion)$(CommitIdForCurrentBuild)</InformationalVersion><InformationalVersion>$(AssemblyFileVersion)$(CommitIdForCurrentBuild)</InformationalVersion>
```

that would put the commidID right into the binary

azure-pipelines/build.yml
```diff
@@ -49,6 +56,22 @@ steps:
...
     summaryFileLocation: '$(Agent.TempDirectory)/coverlet/reports/Cobertura.xml'
 
+- task: PowerShell@2
+  displayName: 'Generate Nuspec Version'
+  inputs:
+    targetType: 'inline'
+    script: |
+      $propsFile = '$(Build.SourcesDirectory)\Directory.Build.Props'
+      $props = New-Object XML
+      $props.Load($propsFile)
```
Matteo-T, line 66: Is this just a complicated way to do:

$props = [xml](Get-Content -Raw $propsFile)

?

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
```diff
@@ -0,0 +1,60 @@
...
+            </group>
+            <group targetFramework="net6.0">
+                <dependency id="Microsoft.Data.SqlClient" version="5.0.1" />
+                <dependency id="Microsoft.SqlServer.SqlManagementObjects" version="170.8.0" />
+                <dependency id="Newtonsoft.Json" version="13.0.1" />
+                <dependency id="System.Configuration.ConfigurationManager" version="6.0.0" />
+            </group>
+        </dependencies>
+    </metadata>
+    <files>
```
Matteo-T, line 29: > <files> [](http://example.com/codeflow?start=3&length=8)

Is a simple
  <IsPackable>True</IsPackable>
in the .csproj file + a handful properties for the nuget pack target not good enough for you?

This seems like a straight-off-the-book care where you want to delegate all the package management to msbuild and not be in the bisiness of hand-crafting nuspec files

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
```diff
@@ -0,0 +1,60 @@
...
+        <file src="..\..\artifacts\publish\Microsoft.SqlTools.ServiceLayer\default\net472\de\Microsoft.SqlTools.ManagedBatchParser.resources.dll" target="lib\net472\de" />
```
Matteo-T, line 30: >\net472\de\ [](http://example.com/codeflow?start=82&length=11)

FWIW, you should be able to use
   ...\net472\**\*.dll
and the like

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
```diff
@@ -34,11 +33,13 @@
...
\ No newline at end of file
+[assembly: AssemblyVersion("3.0.0.0")]
+[assembly: AssemblyFileVersion("3.0.0.0")]
+#if NET6_0_OR_GREATER
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Microsoft.SqlTools.ServiceLayer.UnitTests")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Microsoft.SqlTools.ServiceLayer.IntegrationTests")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Microsoft.SqlTools.ServiceLayer.Test.Common")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("MicrosoftSqlToolsServiceLayer")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("MicrosoftKustoServiceLayer")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Microsoft.SqlTools.ManagedBatchParser.IntegrationTests")]
```
Matteo-T, line 44: These kind of things should really be in the .csproj file as
  <InternalsVisibleTo>
so they'd be right there with your build logic (you can make it conditional so that the key may or may not be specified based on the target framework...

FWIW, I tried to do it in this project but it did not work for me. Normal .csproj files (outside of the STS solution) work just fine... so I don't really know what custom logic you guys have in your .targets/.props that is overriding the behavior...

Matteo-T:
Most of my comments are just "possible improvements" and should not be considered blockers to the PR, though I strongly suggest getting out of the business of handcrafting your nuspec file, particularly for a such mainline scenario. :)

Matteo-T (approved):
:shipit:

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
shueybubbles, line 44: please just make everything public and get rid of these attributes.
We want to consume this batchparser in other apps and we shouldn't have to guess which "internal" components those other apps might want to use.

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
```diff
@@ -15,7 +14,7 @@
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("")]
 [assembly: AssemblyProduct("Microsoft.SqlTools.ManagedBatchParser")]
-[assembly: AssemblyCopyright("Copyright ©  2019")]
+[assembly: AssemblyCopyright("Copyright © 2022")]
```
shueybubbles, line 17: all these attributes should be msbuild properties instead

Directory.Build.props
```diff
@@ -3,8 +3,15 @@
     <HighEntropyVA>true</HighEntropyVA>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
     <TargetFramework>net6.0</TargetFramework>
-  </PropertyGroup>
-  <PropertyGroup>
+    
+    <Build_BuildNumber Condition="'$(Build_BuildNumber)' == ''">1.dev</Build_BuildNumber>
+    <AssemblyVersion>3.0.0.0</AssemblyVersion>
```
shueybubbles, line 8: > <AssemblyVers [](http://example.com/codeflow?start=3&length=14)

Shouldn't these version numbers be overridable by msbuild /p or by environment variables?

<MajorVersion Condition='$(MajorVersion)'==''>3</MajorVersion>
<AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>

shueybubbles:
>     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>

move this to the directory.build.props in the root so every project has the same copyright and no need for separate attribute in the assemblyinfo.cs file

---
Refers to: packages/Directory.Build.props:17 in cbcba85. [](commit_id = cbcba85e1bfc423d8877e05276b1b9d5b119e85f, deletion_comment = False)

azure-pipelines/build.yml
```diff
@@ -49,6 +56,22 @@ steps:
     codeCoverageTool: 'Cobertura'
     summaryFileLocation: '$(Agent.TempDirectory)/coverlet/reports/Cobertura.xml'
 
+- task: PowerShell@2
```
shueybubbles, line 59: >task: PowerShell@2 [](http://example.com/codeflow?start=2&length=18)

if you built your nupkg from a csproj you wouldn't need this task at all

Matteo-T:
>     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>

You speak like a veteran who's been thru dozens of migrations and "how to get the version right in my repos"... I can see the scars on your face... :-)

---
In reply to: [1303564129](https://github.com/microsoft/sqltoolsservice/pull/1747#issuecomment-1303564129) [](http://example.com/codeflow?ancestors=1303564129,1303564129)

---
Refers to: packages/Directory.Build.props:17 in cbcba85. [](commit_id = cbcba85e1bfc423d8877e05276b1b9d5b119e85f, deletion_comment = False)

Directory.Build.props
Matteo-T: Ideally coming from the pipeline... yes

shueybubbles (approved):
:shipit:

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
Matteo-T: Glad to know I'm not the only one who thinks that InternalsVisibleTo are just evil...

Matteo-T: Of course, being public makes one be more diligent at implementing things... in that you need to honor a contract. But, then again, you want to honor the contract even for your "friends" that you grant access to your internals...

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
Matteo-T: AssemblyInfo.cs file should probably not even exist

shueybubbles:
>     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>

I'm Rutger Hauer in the rain in Blade Runner :smile: 

---
In reply to: [1303568257](https://github.com/microsoft/sqltoolsservice/pull/1747#issuecomment-1303568257) [](http://example.com/codeflow?ancestors=1303568257,1303564129)

---
Refers to: packages/Directory.Build.props:17 in cbcba85. [](commit_id = cbcba85e1bfc423d8877e05276b1b9d5b119e85f, deletion_comment = False)

Directory.Build.props
```diff
@@ -3,8 +3,15 @@
...
+    <VersionPrefix>3.0.0</VersionPrefix>
+    <VersionSuffix>preview$(Build_BuildNumber)</VersionSuffix>
```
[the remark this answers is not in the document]
shueybubbles: Another option to consider is building the package twice, once with a preview suffix and once without in the same build.

Then you can decide at release time which version to push to the feed. 
We build both preview and non-preview packages for SMO with each official build.

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
Charles-Gagnon: +1 to putting everything in csproj, it's much easier to use and modify at build time and lets you share common properties so much easier.

azure-pipelines/build.yml
```diff
@@ -4,12 +4,19 @@ steps:
...
+    secureFile: sqlToolsKeyPair.snk
+    retryCount: 5
+
 - task: BatchScript@1
   displayName: 'Run script build.cmd'
   inputs:
     filename: build.cmd
-    arguments: '-target=all -mono'
-
+    arguments: '-target=all -mono -signAssembly=true -signAssemblyKeyFile="$(sqltoolssigningkey.secureFilePath)"'
```
Charles-Gagnon, line 18: In general I don't think we should be signing the assemblies ourselves and instead using ESRP for that - see https://github.com/Azure/azure-functions-sql-extension/pull/301 for an example

azure-pipelines/build.yml
cheenamalhotra: I had some trouble accessing properties with Get-Content, but maybe casting to XML would have helped. 
I think it's fine to use this unless there's a risk with xml.load

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
cheenamalhotra: Packing is not done on csproj since we need to sign the built binaries first, that's why we need an extra step. 
We would need either another csproj, like there are in packages folder or nuspec. Nuspec is the best way to go for IMO.

Charles-Gagnon: Here's ours for sql binding - easy to pull these properties out into a common props file for this repo

https://github.com/Azure/azure-functions-sql-extension/blob/main/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
cheenamalhotra: But we don't want all the net472 dlls

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
Charles-Gagnon: We should remove the copyright year too, we never use it anywhere else. `Copyright (c) Microsoft Corporation`

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
cheenamalhotra: I hear you. It's a web of internal dependencies as of now, in every single project. 
It needs to be done separately.

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
Charles-Gagnon: Build first, then sign, then package. 

Example : 

[Build](https://github.com/Azure/azure-functions-sql-extension/blob/main/builds/azure-pipelines/template-steps-build-test.yml#L67) (GeneratePackageOnBuild=false)

[Package](https://github.com/Azure/azure-functions-sql-extension/blob/main/builds/azure-pipelines/template-steps-build-test.yml#L236) (--no-build)

src/Microsoft.SqlTools.ManagedBatchParser/Properties/AssemblyInfo.cs
cheenamalhotra: Can't agree more! https://github.com/microsoft/sqltoolsservice/pull/1747#discussion_r1014346676

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
shueybubbles: signing of DLLs should happen in one task of the pipeline, then you run
```
msbuild /p:nobuild=1 /t:pack 
or
dotnet pack -nobuild (not sure of the exact syntax/params)
```
so it creates the nuget package with the signed DLLs

then you have to sign the nupkg using esrp so it can go to nuget.org

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
Charles-Gagnon: If you used csproj you wouldn't have to do any of this, it'll pick up everything it needs by itself 😄

Directory.Build.props
cheenamalhotra: They can be overridden by pipeline variables. They're not as of now. I'm yet to discuss the vision of how we would like to maintain versioning scheme for longer run so we can design it that way, privately.

azure-pipelines/build.yml
cheenamalhotra: Commented: https://github.com/microsoft/sqltoolsservice/pull/1747#discussion_r1014343548
Given that, you must be signing your binaries from output folder and not a separate publish directory. In this case its separate, so signed DLLs are not consumed by csproj when packing.

azure-pipelines/build.yml
cheenamalhotra: This is a requirement for strong name key (SNK) signing that applies to .NET Framework DLLs. 
It's not actually 'signing' the DLL but creating a unique identity. 

The DLLs are ofcourse signed with ESRP Codesigning later.

Charles-Gagnon: ESRP does strong name signing. Look at the line the link brings you to.

shueybubbles: i think Cheena means this step is delay signing the DLL so ESRP can apply the full signing later.

Charles-Gagnon: In general all SQL assemblies are signed with the same key and I don't see any reason we should change that.

shueybubbles: though afaik you can just add these to the props to do it
```
      <SignAssembly>true</SignAssembly>  
      <DelaySign>true</DelaySign>
<AssemblyOriginatorKeyFile>mykey.snk</AssemblyOriginatorKeyFile>
```

cheenamalhotra: I can enable delay sign and include the ESRP signing operation, that makes sense! 
I can address that with key alongside too.

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
cheenamalhotra: > [Package](https://github.com/Azure/azure-functions-sql-extension/blob/main/builds/azure-pipelines/template-steps-build-test.yml#L236) (--no-build)

We'd still need "signed" binaries in the same location to provide 'no-build'. DLLs are published to artifacts/publish directory, where they are signed and packaged from. 

I don't disagree, but with current setup I don't want to disturb lot of things. As of now separate csproj files in [packages](https://github.com/microsoft/sqltoolsservice/tree/main/packages) folder are used for packaging (view readme), and the only thing I did differently for this NuGet was to use nuspec instead.

Directory.Build.props
cheenamalhotra: image

Got that working, I figured `GenerateAssemblyFile` was turned off in this project, which was not generating this before.

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
```diff
@@ -4,21 +4,36 @@
 		<TargetFrameworks>net6.0;net472</TargetFrameworks>
 		<LangVersion>9.0</LangVersion>
 		<Nullable>disable</Nullable>
-		<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+		<AssemblyName>Microsoft.SqlTools.ManagedBatchParser</AssemblyName>
+		<Guid>82dd9738-2ad3-4eb3-9f80-18b594e03621</Guid>
+		<DelaySign>True</DelaySign>
+		<!-- Explicitly disable since it leads to compilation errors. The .NET 6.0 target is used in tests with internalsVisibleTo attribute.-->
+		<SignAssembly Condition="$(TargetFramework) == 'net472'">True</SignAssembly>
+		<AssemblyOriginatorKeyFile>$(RootDir)\SqlToolsKeyPair.snk</AssemblyOriginatorKeyFile>
```
Charles-Gagnon, line 12: Where did you get this from? Make sure you're using the SQL 2003 one, that's what the ESRP service will expect. 

I would also check in the key file - otherwise won't this fail for people locally since the key file doesn't exist?

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
cheenamalhotra: As discussed in [comment](https://github.com/microsoft/sqltoolsservice/pull/1747#discussion_r1014412695) signed DLLs are not available in same output folder for packing with `no-build` using the same csproj file. This is old school design using build.cake targets, but shouldn't be a blocker for this PR.

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
cheenamalhotra: I've added the file too in root folder, its generated using sn.exe. 
I'm not sure if it needs to be a particular snk, since I've not used the ESRP delay sign before, but I'll let build fail and will get the other one if we need that one.

shueybubbles: I'm fairly certain there's a small number of valid public key tokens ESRP will sign.

azure-pipelines/build.yml
```diff
@@ -49,6 +49,22 @@ steps:
...
+- task: PowerShell@2
+  displayName: 'Generate Nuspec Version'
+  inputs:
+    targetType: 'inline'
+    script: |
+      $propsFile = '$(Build.SourcesDirectory)\Directory.Build.Props'
+      $props = New-Object XML
+      $props.Load($propsFile)
+      $propGroup = $props.Project.PropertyGroup;
+      $version = $propGroup.VersionPrefix + '-' + 'preview' + $(Build.BuildNumber)
```
Charles-Gagnon, line 61: As David noted above, the preview should generally be the last part of the version string. So `3.0.0.1234-preview` instead of `3.0.0-preview1234`

packages/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.nuspec
```diff
@@ -0,0 +1,60 @@
...
+        <projectUrl>https://github.com/Microsoft/sqltoolsservice/</projectUrl>
+        <license type="expression">MIT</license>
+        <iconUrl>http://s.gravatar.com/avatar/6f39d8562df0a3509a8240fb281bc5bd?s=80</iconUrl>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
+        <description>Managed Batch parser for Sql Tools Services.</description>
+        <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+        <tags>SQL Tools XPLAT Managed Batch Parser</tags>
+        <dependencies>
+            <group targetFramework="net472">
+                <dependency id="Microsoft.Data.SqlClient" version="5.0.1" />
```
Charles-Gagnon, line 16: Another reason to use csproj, not having to remember to update this if these dependencies ever change...

If you're going to keep this for now though please put these packages in a separate section in the Packages.props and add a comment telling people to update this nuspec as well if those ever change

src/Directory.Build.props
```diff
@@ -2,17 +2,15 @@
...
 	</PropertyGroup>
 	<PropertyGroup>
 		<Authors>Microsoft</Authors>
+		<EmbedUntrackedSources>true</EmbedUntrackedSources>
+		<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
 		<PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
-		<Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-		<PackageLicenseUrl>https://github.com/Microsoft/sqltoolsservice/blob/master/license.txt</PackageLicenseUrl>
+		<Copyright>Copyright © Microsoft Corporation. All rights reserved.</Copyright>
+		<PackageLicense>https://github.com/Microsoft/sqltoolsservice/blob/master/license.txt</PackageLicense>
```
Charles-Gagnon, line 13: `master` -> `main`

Charles-Gagnon (approved):
Lots of follow up items to consider - most notably the nuspec -> csproj. Please make issues to track those follow up tasks so we don't forget about them and have all the context in one place.

Also make sure you check out the last few comments I left - some stuff I think we should be fixing before merging this in.

PR #1755, opened 2022-11-09
Update .NET SDK to 6.0.403

cheenamalhotra:


Charles-Gagnon (approved):
Is there a specific reason you're doing this? Context in PRs is helpful - even if it's just updating to get the latest version of something.

cheenamalhotra:
> Is there a specific reason you're doing this? Context in PRs is helpful - even if it's just updating to get the latest version of something.

Yes, with Windows updates, we get .NET SDK updated from time to time, and suddenly builds stop working locally. Keeping this updated helps building as-is, and I think its generally a good idea to bring in latest security updates from .NET in our pipelines. Since we're forked for 4.3, it seemed a right opportunity to do so!

PR #1760, opened 2022-11-15
Update to latest .Net SDK 7.0

kburtram:
Update to latest stable SDK for general product maintenance and health.


src/Microsoft.InsightsGenerator/DataTransformation.cs
```diff
@@ -34,6 +34,7 @@ public DataArray Transform(DataArray array)
 
         private string[] GetColumnLabels(DataArray array, out DataArray.DataType[] columnDataType)
         {
+            #pragma warning disable CA1854
```
alanrenmsft, line 37: comments explaining why?

alanrenmsft (approved):

src/Microsoft.InsightsGenerator/DataTransformation.cs
kburtram: sure, I can a comment.  The SDK update is bringing in a bunch of additional static analysis, that I'd rather fix incrementally than as part of the bump.  I'll also create an issue to track following up on reenabling the disabled rules.

cheenamalhotra:
.NET 7 is not an LTS release.. do we not need an LTS version?
https://dotnet.microsoft.com/en-us/platform/support/policy

caohai (approved):

kburtram:
> .NET 7 is not an LTS release.. do we not need an LTS version? https://dotnet.microsoft.com/en-us/platform/support/policy

@cheenamalhotra we generally update to each new major .Net SDK unless there is some reason not to.  Given this we don't worry about whether the version is LTS.

cheenamalhotra (approved):
LGTM, just eyeing into the test failures, cannot find an obvious reason... can we try rerunning?

PR #1766, opened 2022-11-22
Bump SMO to 170.9.0

Charles-Gagnon:
For https://github.com/microsoft/azuredatastudio/issues/21245

This is reverting the change in SMO that's breaking the table expansion, see the issue for more details

I also added a logging statement to make explicitly clear when we're falling back to master

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -28,6 +28,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             // We detect that here and fall back to master
             if (db.State == SqlSmoState.Creating && !IsDWGen3(db))
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -28,14 +28,15 @@ public DatabaseTreeNode(ServerNode serverNode, string databaseName): this()
             // We detect that here and fall back to master
             if (db.State == SqlSmoState.Creating && !IsDWGen3(db))
             {
```
shueybubbles, line 30: log a bug to only do this fallback for non-Azure connections

shueybubbles (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
Charles-Gagnon: https://github.com/microsoft/sqltoolsservice/issues/1767

PR #1773, opened 2022-11-26
Send Error Code with session failed notification

cheenamalhotra:
The PR updates session failed notification to contain 'ErrorNumber' as received from SqlClient.
Companion PR for new property: https://github.com/microsoft/sqlops-dataprotocolclient/pull/72

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
@@ -12,6 +12,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
 using Microsoft.Data.Tools.Sql.DesignServices.TableDesigner;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlTools.Extensibility;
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -822,6 +830,11 @@ public static ObjectExplorerSession CreateSession(ConnectionCompleteParams respo
                     session.Root = databaseNode;
                 }
 
+                if(response?.ErrorMessage != null)
```
Charles-Gagnon, line 833: Format these files

Charles-Gagnon (approved):
Make sure you're updating [sqlops-dataprotocolclient](https://github.com/microsoft/sqlops-dataprotocolclient/blob/main/src/types.ts#L7) too

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -543,25 +547,27 @@ private async Task<ConnectionCompleteParams> Connect(ConnectParams connectParams
...
             {
-                await SendSessionFailedNotification(uri, ex.ToString());
+                int? errorNum = ex is SqlException sqlEx ? sqlEx.Number : null;
+                await SendSessionFailedNotification(uri, ex.ToString(), errorNum);
                 return null;
             }
         }
 
-        private async Task SendSessionFailedNotification(string uri, string errorMessage)
+        private async Task SendSessionFailedNotification(string uri, string errorMessage, int? errorCode = null)
```
Charles-Gagnon, line 563: I'd suggest removing the default here so there's less chance of someone forgetting to add that parameter when making changes in the future.

alanrenmsft (approved):

PR #1775, opened 2022-12-06
Remove big data cluster references from connections.

corivera:
Part of the general BDC cleanup in ADS.


Charles-Gagnon (approved):
Looks good, but wait to merge this in until after the ADS side of things is removed so we aren't breaking anything on insiders

PR #1782, opened 2022-12-13
Use error code checking and simplified messages

### LINKED ISSUE #1777: Change password error handling shouldn't use localized error strings
https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs#L1169

We're checking the error messages using localized strings here - but there's no guarantee the user language is going to match what the server sends. 

We should probably be checking the error number instead.

smartguest:
Now using SQL error code checking and also removed "password cannot be changed" message from returned SqlClient error:

Empty Password error is in STS itself (and does not have an error code of its own as it's a generic exception I created), so it will still work regardless of language.

Fixes #1777

Password Change Dialog Empty Password
Password Change Dialog password cannot be changed
Password Change Dialog Same Password

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -22,7 +22,6 @@ using Microsoft.SqlTools.ServiceLayer.LanguageServices.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Utility;
 using Microsoft.SqlTools.Utility;
 using System.Diagnostics;
-using System.Text.RegularExpressions;
 
 namespace Microsoft.SqlTools.ServiceLayer.Connection
 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -38,6 +37,11 @@ public class ConnectionService
 
         public const int MaxServerlessReconnectTries = 5; // Max number of tries to wait for a serverless database to start up when its paused before giving up.
 
+        // SQL Error Code Constants
+        private const int DoesNotMeetPWReqs = 18466;
```
cheenamalhotra, line 41: Can you please add comment here with error message or link to error info so it's easy to refer from.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1159,18 +1163,24 @@ private void RunConnectRequestHandlerTask(ConnectParams connectParams)
             catch (Exception ex)
             {
                 newResponse.Result = false;
-                newResponse.ErrorMessage = ex.InnerException != null ? (ex.Message + Environment.NewLine + Environment.NewLine + ex.InnerException.Message) : ex.Message;
-                newResponse.ErrorMessage = Regex.Replace(newResponse.ErrorMessage, @"\r?\nChanged database context to '\w+'\.", "");
-                newResponse.ErrorMessage = Regex.Replace(newResponse.ErrorMessage, @"\r?\nChanged language setting to \w+\.", "");
-                if (newResponse.ErrorMessage.Equals(SR.PasswordChangeEmptyPassword))
+                newResponse.ErrorMessage = ex.Message;
+                int errorCode = 0;
+                SqlError endError = (ex.InnerException as SqlException)?.Errors[0];
```
cheenamalhotra, line 1168: Would recommend checking if 'Errors' is not empty too, just to avoid any possible NREs

PR #1789, opened 2022-12-16
Update login migrations nuget with perf improvements

AkshayMata:
This PR updates the login migration nuget version, which has perf improvements. Additionally, this PR fixes a null exception bug with the exception map.


raymondtruong:
looks like you removed the old nupkg but didn't check in the new one

raymondtruong (approved):

brian-harris (approved):

PR #1791, opened 2022-12-16
Add support for firewall rule name in request

cheenamalhotra:
New API introduced:
```
public class CreateFirewallRuleParams {
    public string FirewallRuleName { get; set; }
}
```


alanrenmsft (approved):

src/Microsoft.SqlTools.ResourceProvider.Core/Firewall/FirewallRuleService.cs
```diff
@@ -55,31 +56,40 @@ public class FirewallRuleService : IFirewallRuleService
...
-            IPAddress startIpAddress = ConvertToIpAddress(startIpAddressValue);
-            IPAddress endIpAddress = ConvertToIpAddress(endIpAddressValue);
-            return await CreateFirewallRuleAsync(serverName, startIpAddress, endIpAddress);
+            IPAddress startIpAddress = ConvertToIpAddress(firewallRuleParams.StartIpAddress);
+            IPAddress endIpAddress = ConvertToIpAddress(firewallRuleParams.EndIpAddress);
+            FirewallRuleRequest firewallRuleRequest = new FirewallRuleRequest()
+            {
+                StartIpAddress = ConvertToIpAddress(firewallRuleParams.StartIpAddress),
+                EndIpAddress = ConvertToIpAddress(firewallRuleParams.EndIpAddress),
+                FirewallRuleName = string.Format(CultureInfo.InvariantCulture, firewallRuleParams.FirewallRuleName ?? "ClientIPAddress_{0}",
```
Charles-Gagnon, line 67: You shouldn't be formatting the passed in FirewallRuleName, right? Or if it's intentional that you'll be inserting the date you should document that in the API.

PR #1798, opened 2023-01-04
Exclude isLedger property check for Azure DBs for improving OE nodes expansion perf

cheenamalhotra:
Additional changes to fix issue https://github.com/microsoft/azuredatastudio/issues/21475

Checking 'IsLedger' leads to making new connections on every node in database list.
We're disabling it for Azure databases to avoid making node-level connections.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
@@ -4,8 +4,11 @@
 //
 
 using System.Collections.Generic;
+using Microsoft.SqlTools.Utility;
 using Microsoft.SqlServer.Management.Smo;
 using Microsoft.SqlTools.ServiceLayer.ObjectExplorer.Nodes;
+using System.Diagnostics;
+using System;
 
 namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 {
@@ -35,14 +38,15 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             try
             {
                 Database? db = smoObject as Database;
-                if (db != null && IsPropertySupported("IsLedger", smoContext, db, CachedSmoProperties) && db.IsLedger)
+                if (db != null && IsPropertySupported(nameof(db.IsLedger), smoContext, db, CachedSmoProperties) && db.IsLedger)
                 {
                     return "Ledger";
                 }
             }
-            catch
+            catch (Exception e)
             {
                 //Ignore the exception and just not change create custom name
+                Logger.Write(TraceEventType.Verbose, $"Error ignored when reading node subtype: {e.Message}");
             }
 
             return string.Empty;
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
@@ -32,9 +32,6 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 
         public override string GetNodeSubType(object smoObject, SmoQueryContext smoContext)
         {
-            // IsLedger property check is disabled on Azure databases to avoid making connections to databases 
-            // individually when listing databases in Object Explorer since it's value is not available from 'master' DB for all databases..
-
             try
             {
                 Database? db = smoObject as Database;
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
@@ -32,18 +32,21 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 
         public override string GetNodeSubType(object smoObject, SmoQueryContext smoContext)
         {
-            try
-            {
-                Database? db = smoObject as Database;
-                if (db != null && IsPropertySupported("IsLedger", smoContext, db, CachedSmoProperties) && db.IsLedger)
-                {
-                    return "Ledger";
-                }
-            }
-            catch
-            {
-                //Ignore the exception and just not change create custom name
-            }
+            // IsLedger property check is disabled to avoid making connection to databases individually when listing databases in OE
+            // since it's value is not available from 'master' DB for all databases > v12.
+
+            // try
+            // {
+            //     Database? db = smoObject as Database;
+            //     if (db != null && IsPropertySupported("IsLedger", smoContext, db, CachedSmoProperties) && db.IsLedger)
+            //     {
+            //         return "Ledger";
+            //     }
+            // }
+            // catch
+            // {
+            //     //Ignore the exception and just not change create custom name
+            // }
 
             return string.Empty;
         }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
@@ -32,21 +32,21 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
 
         public override string GetNodeSubType(object smoObject, SmoQueryContext smoContext)
         {
-            // IsLedger property check is disabled to avoid making connection to databases individually when listing databases in OE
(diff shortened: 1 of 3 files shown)


aasimkhan30 (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
```diff
@@ -32,18 +32,21 @@ protected override void InitializeChild(TreeNode parent, TreeNode child, object
 
         public override string GetNodeSubType(object smoObject, SmoQueryContext smoContext)
```
shueybubbles, line 33: > public ove [](http://example.com/codeflow?start=7&length=11)

you could still do this for onprem

shueybubbles (approved):
:shipit:

Charles-Gagnon (changes requested):
Are you planning on finding a better fix for this and this is just a temporary workaround? And have you worked with @nofield on what we can do here to both keep the Ledger functionality and not cause perf regressions?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
Charles-Gagnon: Ledger is available on Azure too so I don't think that's going to be a good long term solution here unfortunately.

shueybubbles:
I'm not sure what a "better fix" will look like for Azure. You can't find out if a DB is Ledger in Azure without logging in to it.
So any menu items or whatever that are based on IsLedger==true will have to defer that calculation until the user logs in to the database.

---
In reply to: [1236583480](https://github.com/microsoft/sqltoolsservice/pull/1798#issuecomment-1236583480) [](http://example.com/codeflow?ancestors=1236583480,1236583480)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
nofield: It would at least allow the functionality to continue as intended for on-prem. The primary impact is on the Ledger database icon. As another data point, in SSMS we don't have the icon enabled for Azure, but we do for on-prem.

Charles-Gagnon:
Yes, the better fix will unfortunately probably involve a decent chunk of work to refactor how we populate these properties (only populate some on node expansion for example). But this is work that we'll almost certainly want to do at some point - there will be other times that we want to fetch DB-specific properties only after a node is expanded and so having the ability to defer that lookup is going to be something we'll want to see about doing.

Do you recall how SSMS does it? Does it just not look up these properties until the node is expanded and then dynamically update menu items/icons?

It may be that we do something like this for now to address the linked issue in the short term, but let's at least have a discussion with the feature owner before ripping this stuff out.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
Charles-Gagnon: That's good to know - then going that approach for now here seems reasonable as well to at least keep parity between the two.

cheenamalhotra: I can enable it, but I think it still is a performance issue.

Customers who have 200s of databases (from feedback), often see performance hit due to OE connecting to every single database just for the icon. Time it takes is relative to where the database is hosted, and if network is latent, we're going to continue to see timeouts in OE. So is icon worth over OE timing out?

Charles-Gagnon: SMO doesn't connect to every DB for on-prem servers - it can get that at the server level for all DBs through the sys.databases DMV.

cheenamalhotra: Didn't catch that, thanks :)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
```diff
@@ -32,6 +32,9 @@ protected override void InitializeChild(TreeNode parent, TreeNode child, object
...
         {
+            // IsLedger property check is disabled on Azure databases to avoid making connections to databases
```
Charles-Gagnon, line 35: You're still calling db.IsLedger below - isn't that still resulting in querying SMO for the property and causing it to run the DB query?

cheenamalhotra: No, because it didn't exist in the [SmoTreeNodesDefinition.xml](https://github.com/microsoft/sqltoolsservice/pull/1798/files#diff-90e1717d1de73f5827918374b648cfb7f8374fa071d7f6b993303cfab4c68157) file before, IsPropertySupported always returned true with a warning.. we have had that warning in our logs.

Enabling for specific databases, returns true only for those.

Charles-Gagnon: Whoops, thought this was SqlSmoObject.IsSupportedProperty!

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs
```diff
@@ -32,6 +32,9 @@ protected override void InitializeChild(TreeNode parent, TreeNode child, object
 
         public override string GetNodeSubType(object smoObject, SmoQueryContext smoContext)
         {
+            // IsLedger property check is disabled on Azure databases to avoid making connections to databases 
+            // individually when listing databases in Object Explorer since it's value is not available from 'master' DB for all databases..
```
Charles-Gagnon, line 36: I would suggest adding this as a comment in the XML instead - that's where this is actually being done and someone could easily later on go change that back to All and not realize what they're doing.

Charles-Gagnon (approved):
One suggestion I think you should do, but otherwise looks good!

shueybubbles:
SSMS OE uses the fewest number of properties it can in the initial expansion, that don't require PostProcess. Then individual menu handlers may implement their own property fetch, which ends up causing delays opening the context menu while it tries to connect. 
I think it would be a reasonable model for both apps to require context menu handlers NOT make their own connections, and instead connect to the database on node expansion and populate any missing properties at that time. Then the user would see more enabled menu items than they would before the node has expanded.


---
In reply to: [1371529614](https://github.com/microsoft/sqltoolsservice/pull/1798#issuecomment-1371529614) [](http://example.com/codeflow?ancestors=1371529614,1371529614)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -213,6 +213,11 @@ public override IEnumerable<NodeSmoProperty> SmoProperties
                    Name = "Status",
                    ValidFor = ValidForFlag.All
                 });
+                properties.Add(new NodeSmoProperty
+                {
+                   Name = "IsLedger",
```
shueybubbles, line 218: >IsLedger", [](http://example.com/codeflow?start=27&length=10)

nameof(Database.IsLedger) and nameof(Database.Status) above

shueybubbles:
>                 if (db != null && IsPropertySupported("IsLedger", smoContext, db, CachedSmoProperties) && db.IsLedger)

nameof(db.IsLedger)

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs:41 in 3ef8afc. [](commit_id = 3ef8afc72abb6186a5aa5dc39f8e9f51fa5b21f0, deletion_comment = False)

shueybubbles:
>             }

when would this ever throw?

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoDatabaseCustomNode.cs:49 in 3ef8afc. [](commit_id = 3ef8afc72abb6186a5aa5dc39f8e9f51fa5b21f0, deletion_comment = False)

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
nofield: This file is auto-generated. You'd need to update the code gen to make this change.

nofield (approved):

Issue #1802, opened 2023-01-09
Object Explorer design changes for property support


cheenamalhotra:
Feedback from https://github.com/microsoft/sqltoolsservice/pull/1801#issuecomment-1374521242, below optimization should be done to keep up with SMO's latest APIs and design:

- Try to remove the redundancy of the XML and C# code. Couldn't the node builder pick up the property list from the deserialized XML objects instead of requiring a duplicate code object that returns the same property list?

- Try to get rid of the "ValidFor" stuff and delay the check until issuing the SMO query. The Server object in SMO now has a version of IsSupportedProperty for this. The node just needs a way to determine which properties were actually set during initialization so it doesn't try to use the unset ones. I suggest this because Azure gets new stuff pretty quickly, and properties that were only valid on sql box for a while become valid on cloud, and ideally we could just pick up a new SMO drop for that and have it "just work" with no STS code changes.

```
 /// <summary>
 /// Returns true if the given property name is valid for the object of Type type for the Server's version and edition.
 /// </summary>
 /// <param name="type">A type that is assignable to SqlSmoObject</param>
 /// <param name="propertyName"></param>
 /// <param name="databaseEngineEdition">Specific database edition to check for property supported. If Unknown or not specified, the check will use the edition of the master database.</param>
 /// <returns>true if the property can be requested, false otherwise</returns>
 /// <exception cref="ArgumentException">When type is not a SqlSmoObject or if the SqlSmoObject does not have property visibility based on the Server version</exception>
 public bool IsSupportedProperty(Type type, string propertyName, DatabaseEngineEdition databaseEngineEdition = DatabaseEngineEdition.Unknown)
```
and a generic version

```
public bool IsSupportedProperty<T>(string propertyName, DatabaseEngineEdition databaseEngineEdition = DatabaseEngineEdition.Unknown) where T : SqlSmoObject
```

Alternatively, `SetDefaultInitFields/ClearAndInitialize` should ignore unsupported properties, so can be tried.

Benjin:
Test link to AB#3197041

Benjin:
Another test link to AB#3197034

PR #1804, opened 2023-01-09
Enable IsLedger only for SQL 2022

cheenamalhotra:
It turns out SMO queries fail for non-supported properties that do not apply for SQL Server versions we are connecting to, as in case of `IsLedger` for Database list, we are getting below error for SQL Server 19 and below:
```
sqltools Error: 0 : Failed getting smo objects. parent:<server>/Databases querier: Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel.SqlDatabaseQuerier error:unknown property IsLedger
 inner: stacktrace:   at Microsoft.SqlServer.Management.Sdk.Sfc.EnumObject.GetProperty(String name, ObjectPropertyUsages usage)
   at Microsoft.SqlServer.Management.Smo.SqlObjectBase.AddRequestProperties()
   at Microsoft.SqlServer.Management.Smo.SqlObjectBase.RetrieveParentRequestLinks(SqlRequest sr)
   at Microsoft.SqlServer.Management.Smo.SqlObjectBase.RetrieveParentRequest()
   at Microsoft.SqlServer.Management.Smo.DatabaseLevel.RetrieveParentRequest()
   at Microsoft.SqlServer.Management.Sdk.Sfc.Environment.InitObjects(Request req)
   at Microsoft.SqlServer.Management.Sdk.Sfc.Environment.GetData(Request req, Object ci)
   at Microsoft.SqlServer.Management.Sdk.Sfc.Enumerator.GetData(Object connectionInfo, Request request)
   at Microsoft.SqlServer.Management.Smo.ExecutionManager.GetEnumeratorDataReader(Request req)
   at Microsoft.SqlServer.Management.Smo.SqlSmoObject.DoDatabaseSpecialCase(Request levelQuery, Urn levelFilter, Boolean forScripting, List`1 urnList, Int32 startLeafIdx, IList`1 defaultFields)
   at Microsoft.SqlServer.Management.Smo.SqlSmoObject.InitChildLevel(Urn levelFilter, ScriptingPreferences sp, Boolean forScripting, IEnumerable`1 extraFields)
   at Microsoft.SqlServer.Management.Smo.SmoCollectionBase.InitializeChildCollection(Boolean refresh, ScriptingPreferences sp, String filterQuery, IEnumerable`1 extraFields)
   at Microsoft.SqlServer.Management.Smo.SmoCollectionBase.ClearAndInitialize(String filterQuery, IEnumerable`1 extraFields)
   at Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel.SqlDatabaseQuerier.Query(SmoQueryContext context, String filter, Boolean refresh, IEnumerable`1 extraProperties) in E:\sts\src\Microsoft.SqlTools.ServiceLayer\ObjectExplorer\SmoModel\SmoQueryModel.cs:line 38
   at Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel.SmoChildFactoryBase.OnExpandPopulateNonFolders(IList`1 allChildren, TreeNode parent, Boolean refresh, String name, CancellationToken cancellationToken) in E:\sts\src\Microsoft.SqlTools.ServiceLayer\ObjectExplorer\SmoModel\SmoChildFactoryBase.cs:line 141
sqltools Error: 0 : Failed expanding oe children. parent:<server>/Databases error:unknown property IsLedger
...
sqltools Error: 0 : Failed populating oe children. error:unknown property IsLedger
...
```

This PR fixes that by enabling 'IsLedger' only for SQL 2022 instead of 'AllOnPrem' - I probably didn't realize it applies to SQL 2022 only.. (ref: https://github.com/microsoft/sqltoolsservice/pull/1798#discussion_r1061946247)
cc @shueybubbles 

Thanks @caohai for capturing this!


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -216,7 +216,7 @@ public override IEnumerable<NodeSmoProperty> SmoProperties
                 properties.Add(new NodeSmoProperty
                 {
                    Name = "IsLedger",
-                   ValidFor = ValidForFlag.AllOnPrem
+                   ValidFor = ValidForFlag.Sql2022
```
Charles-Gagnon, line 219: This means we're going to have to update this every time a new version comes out which is going to be a pain to maintain. I would suggest adding a "Sql2022OrHigher" value to the enum so that we only have to update it in one place when a new version is added. 

(another option would be to do like SMO does and add a "MinValidForFlag" that takes in a single value and returns true if the server version is that or higher. This has the advantage of not having to remember to update the "OrHigher" values when a new version is added. But the first option seems more straightforward and that extra bit of overhead shouldn't be too bad.

cheenamalhotra: I think with #1802 we should be able to address that.

Charles-Gagnon: Probably, but I wouldn't suggest relying on that being done anytime soon so better to add in the enum value now (it's only a single line added and then update value in xml) so that we don't forget and end up paying for it down the road when the next version is released.

nofield: Are there examples of properties that are only valid for a certain subset of SQL versions? Like valid in a previous version but not a later version? I took a quick look through the xml and didn't see any like that. 

This issue with the ValidFor enums came up when I was making changes for Ledger; it was a pain adding "Sql2022" everywhere. I think ideally there'd be a "Sql20xxOrHigher" for each SQL version, so the only place that would need to be updated for future editions would be the ValidFor enum definition. #1802 is the cleanest long-term solution though, I agree.

nofield (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
cheenamalhotra: I just added more flags to streamline the pattern.. please check again @nofield :)

Charles-Gagnon: In the future we may need to support features that are only valid for a certain range, but currently that's not something we support so I think it's fine to just go with the minimum required here and then #1802 can cover special scenarios like that in a much easier way.

Charles-Gagnon (approved):
Thanks for cleaning up the other property filters as well! 😄

PR #1806, opened 2023-01-11
Tde migration integration

junierch:
Adding certificate migration for encrypted databases.

Uses notifications to send events back to the client every time each database certificate is migrated.

image


src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -131,6 +134,7 @@ public void InitializeService(ServiceHost serviceHost)
             this.ServiceHost.SetRequestHandler(MigrateLoginsRequest.Type, HandleMigrateLogins, true);
             this.ServiceHost.SetRequestHandler(EstablishUserMappingRequest.Type, HandleEstablishUserMapping, true);
             this.ServiceHost.SetRequestHandler(MigrateServerRolesAndSetPermissionsRequest.Type, HandleMigrateServerRolesAndSetPermissions, true);
+            this.ServiceHost.SetRequestHandler(CertificateMigrationRequest.Type, HandleCertificateMigrationRequest);
```
AkshayMata, line 137:
```suggestion
            this.ServiceHost.SetRequestHandler(CertificateMigrationRequest.Type, HandleTdeCertificateMigrationRequest);
```

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -1043,6 +1050,77 @@ internal static List<AzureSqlSkuCategory> GetEligibleSkuCategories(string target
...
+            await requestContext.SendResult(result);
+        }
+
+        /// <summary>
+        /// Individual certificate migration operation
+        /// </summary>
+        /// <param name="tdeMigrationClient">Instance of the migration client</param>
+        /// <param name="dbName">Name of the database to migrate</param>
+        /// <returns></returns>
+        private async Task<CertificateMigrationEntryResult> MigrateCertificate(TdeMigration tdeMigrationClient,string dbName)
```
AkshayMata, line 1104:
```suggestion
        private async Task<CertificateMigrationEntryResult> MigrateCertificate(TdeMigration tdeMigrationClient, string dbName)
```

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -1043,6 +1050,77 @@ internal static List<AzureSqlSkuCategory> GetEligibleSkuCategories(string target
...
+        {
+            try
+            {
+                var result = await tdeMigrationClient.MigrateTdeCertificate(dbName, CancellationToken.None);
+
+                if (result is TdeExceptionResult tdeExceptionResult)
+                {
+                    return new CertificateMigrationEntryResult { DbName = dbName, Success = result.IsSuccess, ErrorMessage = tdeExceptionResult.Exception.Message };
+                }
+                else {
```
AkshayMata, line 1114:
```suggestion
                else
                {
```

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -1043,6 +1050,77 @@ internal static List<AzureSqlSkuCategory> GetEligibleSkuCategories(string target
...
+                    return new CertificateMigrationEntryResult { DbName = dbName, Success = result.IsSuccess, ErrorMessage = result.UserFriendlyMessage };
```
AkshayMata, line 1115: param name ErrorMessage doesn't make sense for this case

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -1043,6 +1050,77 @@ internal static List<AzureSqlSkuCategory> GetEligibleSkuCategories(string target
...
+                   parameters.TargetResourceGroupName,
+                   parameters.TargetManagedInstanceName,
+                   parameters.NetworkSharePath,
+                   parameters.NetworkShareDomain,
+                   parameters.NetworkShareUserName,
+                   parameters.NetworkSharePassword,
+                   credentials
+                   );
+
+            foreach (var dbName in parameters.EncryptedDatabases)
```
AkshayMata, line 1080: I remember one of Steven's demo talked about parallelizing migrating for dbs. just curious what the outcome of that was?

AkshayMata (changes requested):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
junierch: Renamed.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
junierch: Fixed

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
junierch: Fixed

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
junierch: Renamed all to Message

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
junierch: The operation that consumes most of the time, is getting an MI instance from the ARM client. We are getting that one once and reusing it later. 
Parallelizing would be for the next phase when we will move these steps inside the nuget package and try to review how retries are handled by the ARM client or whether they can be configured, so we don't get 429 from azure.

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
```diff
@@ -560,120 +564,123 @@ public void InitializeService(ServiceHost serviceHost)
             }
         }
 
-        internal RecommendationResultSet GenerateBaselineRecommendations(SqlInstanceRequirements req, GetSkuRecommendationsParams parameters) {
-                RecommendationResultSet resultSet = new RecommendationResultSet();
+        internal RecommendationResultSet GenerateBaselineRecommendations(SqlInstanceRequirements req, GetSkuRecommendationsParams parameters)
```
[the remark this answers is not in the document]
junierch: Not related. Probably got modified when the documented got autoformatted and moved the curly brackets down. Not sure why it didn't add the space on the other _MigrateCertificate_ method parameters lol.

src/Microsoft.SqlTools.ServiceLayer/Migration/Models/StaticTokenCredential.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        private readonly AccessToken _token;
+
+        /// <summary>
+        /// Build credentials using a token that will not change.
+        /// </summary>
+        /// <param name="accessToken">Newly created token that should last for the duration of the whole operation.</param>
+        public StaticTokenCredential(string accessToken) => _token = new AccessToken(
+                   accessToken: accessToken,
+                   expiresOn: DateTimeOffset.Now.AddHours(1));  //Default to an hour, the current duration of a newly create token.
```
raymondtruong, line 27: curious, what are the implications of this token expiry? Does this cap the duration of our migrations to 1 hour at a time?

src/Microsoft.SqlTools.ServiceLayer/Migration/Models/StaticTokenCredential.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core;
+
+namespace Microsoft.SqlTools.ServiceLayer.Migration.Models
+{
+    /// <summary>
+    /// Temp token creadentials to interact with ArmClient class. 
+    /// The token passed to this class should be a newly request token, because this class doesn't renew the token.
+    /// Once MSAL is rolled out on ADS, we will implement a way to use the same ADS token cache configured by ADS.
```
raymondtruong, line 15: the Azure Core extension apparently recently migrated to using MSAL instead of ADAL - not sure if this opens the door for us to start using MSAL as well

src/Microsoft.SqlTools.ServiceLayer/Migration/Models/StaticTokenCredential.cs
junierch: That is the expiration time token are getting on token ADS is requesting. A new token is requested on ADS before making this call which should end in less than 5 minutes for around 100 dbs.

src/Microsoft.SqlTools.ServiceLayer/Migration/Models/StaticTokenCredential.cs
junierch: Yes, we'll use the same configuration to use information stored in the cache and replace the token parameter for the tenant.

src/Microsoft.SqlTools.ServiceLayer/Migration/Models/StaticTokenCredential.cs
raymondtruong: got it, if those are the types of numbers we're seeing (around 5 mins max) then this shouldn't be an issue. If hitting the 1 hour limit was a possibility I would suggest increasing it but doesn't look like we'll need to

raymondtruong (approved):

src/Microsoft.SqlTools.ServiceLayer/Migration/MigrationService.cs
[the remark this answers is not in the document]
AkshayMata: gotcha, that's what i suspected, but it was hard for me to tell from the github diff. thanks for clarifying!

AkshayMata (approved):

PR #1808, opened 2023-01-14
Adding extension service host for extension services 

aasimkhan30:
This extension host will allow extension services to create a JSON RPC server and add services by MEF or registering individual services. 
This feature is added as a replacement to the old Hosting.V2 framework that we had in STS. Once merged, I will replace the Hosting.V2 framework in FlatFileImport and ScaleOutData with this.

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceHost.cs
+++ b/src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceHost.cs
@@ -31,9 +31,6 @@ namespace Microsoft.SqlTools.Extensibility
         ExtensibleServiceHostOptions<T> options
         ) : base(new StdioServerChannel())
         {
-            // Initialize the shutdown activities
-            shutdownCallbacks = new List<ShutdownCallback>();
-            initializeCallbacks = new List<InitializeCallback>();
             this.options = options;
 
             this.Initialize();
@@ -100,14 +97,7 @@ namespace Microsoft.SqlTools.Extensibility
 
         private bool IsServiceInitialized(T service)
         {
-            foreach(T s in this.initializedServices)
-            {
-                if(s.GetType() == service.GetType())
-                {
-                    return true;
-                }
-            }
-            return false;
+            return this.initializedServices.Any(s => s.GetType() == service.GetType());
         }
 
         /// <summary>
@@ -120,9 +110,9 @@ namespace Microsoft.SqlTools.Extensibility
         /// </summary>
         public delegate Task InitializeCallback(InitializeRequest startupParams, RequestContext<InitializeResult> requestContext);
 
-        private readonly List<ShutdownCallback> shutdownCallbacks;
+        private readonly List<ShutdownCallback> shutdownCallbacks = new List<ShutdownCallback>();
 
-        private readonly List<InitializeCallback> initializeCallbacks;
+        private readonly List<InitializeCallback> initializeCallbacks = new List<InitializeCallback>();
 
         private readonly Version serviceVersion = Assembly.GetEntryAssembly().GetName().Version;
 
--- a/src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceHost.cs
+++ b/src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceHost.cs
@@ -1,0 +1,226 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+using Microsoft.SqlTools.Hosting;
+using Microsoft.SqlTools.Hosting.Contracts;
+using Microsoft.SqlTools.Hosting.Protocol;
+using Microsoft.SqlTools.Hosting.Protocol.Channel;
+using Microsoft.SqlTools.ServiceLayer.SqlContext;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.Extensibility
+{
+    public class ExtensionServiceHost : ServiceHostBase
+    {
+
+        private ExtensibleServiceHostOptions options;
+        private static bool isLoaded;
+        public ExtensionServiceProvider serviceProvider;
+
+        public ExtensionServiceHost(
+        ExtensibleServiceHostOptions options
+        ) : base(new StdioServerChannel())
+        {
+            // Initialize the shutdown activities
+            shutdownCallbacks = new List<ShutdownCallback>();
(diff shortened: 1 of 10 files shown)


alanrenmsft:
more context for this PR?

aasimkhan30:
> more context for this PR?

@alanrenmsft, added more context.

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
```diff
@@ -6,7 +6,7 @@
 using System;
 using System.Globalization;
 
-namespace Microsoft.SqlTools.Hosting.Utility
+namespace Microsoft.SqlTools.Utility
```
Charles-Gagnon, line 9: Any specific reason you changed this namespace?

src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -119,16 +119,107 @@ private void EnsureExtensionStoreRegistered<T>()
         /// <summary>
         /// Merges in new assemblies to the existing container configuration.
         /// </summary>
-        public void AddAssembliesToConfiguration(IEnumerable<Assembly> assemblies)
+        public void AddAssembliesToConfiguration<T>(IEnumerable<Assembly> assemblies)
```
Charles-Gagnon, line 122: Update the doc comment with the generic param and explain what it's used for

Charles-Gagnon (commented):
I'm still not sure what all the changes in this PR are doing. What was missing/incorrect before that you're changing here? And how are consumers going to be using this?

It would be generally helpful for us to have docs on this stuff if we're expecting other people to use this stuff, as a start I would suggest adding a README to the project folder that gives an overview of what the library is for and some examples of how someone might use it.

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
aasimkhan30: Only this class was present in the namespace. I changed it so it can be consistent with other utility classes.

Charles-Gagnon: Hm, I think it'd make more sense to move the others to Microsoft.SqlTools.Hosting. By convention the namespace should usually match the library name as well - and we don't want to conflict with other general "Microsoft.SqlTools" namespaces.

aasimkhan30: That will create too many file changes for this PR. I will create a follow up PR for that.

Charles-Gagnon: Sounds good, thanks. You can leave this change in here for now then so at least things are consistent.

src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceHost.cs
```diff
@@ -0,0 +1,306 @@
...
+                }
+                Logger.Verbose("Initializing service: " + service.GetType());
+                // Initialize all hosted services, and register them in the service provider for their requested
+                // service type. This ensures that when searching for the ConnectionService you can get it without
+                // searching for an IHostedService of type ConnectionService
+                this.InitializeService(service);
+            }
+        }
+
+        private bool isServiceInitiazlied(T service)
```
Charles-Gagnon, line 101: `isServiceInitialized`

src/Microsoft.SqlTools.Hosting/README.md
```diff
@@ -0,0 +1,67 @@
+Hosting project contains the necessary classes for implementing a JSON - RPC server. It also contains the Utility classes for Logging, Parsing command line arguments from the RPC client.
```
Charles-Gagnon, line 1: It's important to call out that this isn't just a generic JSON-RPC server, but rather a specialized one (language server that communicates with the language client). Please make sure the docs all refer to this correctly.

And I haven't gone through all the docs yet, but if there's anything here specialized even further to the ADS data protocol then make sure to provide links to that as well. 

```suggestion
This library contains the necessary classes for implementing an Azure Data Studio [language server](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide). It also contains the Utility classes for Logging, Parsing command line arguments from the language client. 
```

src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -119,16 +119,108 @@ private void EnsureExtensionStoreRegistered<T>()
...
+                    continue;
+                }
+
+                try
+                {
+                    assemblies.Add(AssemblyLoadContext.Default.LoadFromAssemblyPath(path));
+                }
+                catch (Exception)
+                {
+                    // we expect exceptions trying to scan all DLLs since directory contains native libraries
```
Charles-Gagnon, line 183: We should still log this since with the inclusion list we expect that only assemblies that are able to be loaded will make it this far.

src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -119,16 +119,108 @@ private void EnsureExtensionStoreRegistered<T>()
...
+        /// <summary>
+        /// Creates a service provider by loading a set of named assemblies, expected to be <paramref name="directory"/>
+        /// </summary>
+        /// <param name="directory">Directory to search for included assemblies</param>
+        /// <param name="inclusionList">full DLL names, case insensitive, of assemblies to include</param>
+        /// <returns><see cref="ExtensionServiceProvider"/> instance</returns>
+        public static ExtensionServiceProvider CreateFromAssembliesInDirectory(string directory, IList<string> inclusionList)
+        {
+            //AssemblyLoadContext context = new AssemblyLoader(directory);
+            var assemblyPaths = Directory.GetFiles(directory, "*.dll", SearchOption.TopDirectoryOnly);
```
Charles-Gagnon, line 156: A good follow up task would be to add more debug logging to this startup stuff (for example "Loading service assemblies from <directory>...") to help track down issues later on.

Charles-Gagnon (approved):
I didn't get to spend too much time going over the code itself - it seems like most of this is ripped from other places in STS. Is that true? But I didn't see anything too majorly concerning there - just lots of cleanup that would be nice to do while you're working on this.

I would like to see the docs expanded on though - right now there's an inherent assumption that the person reading this understands ADS/VS Code and how the language server/client works and why they need to be used. So a couple things I think would help avoid future questions : 

1. Brief description of why this is needed and/or how it fits into the ADS extension story (doesn't have to be super complete, just a sentence or two with some links to other docs)
2. Links to the data protocol library with messages it deals with (make sure to call out that's only if adding a new provider)
3. General overview of how to add a custom message (on both extension & server side)

alanrenmsft (approved):

kburtram (approved):

PR #1813, opened 2023-01-18
Added Error Handling for STS

smartguest:
Currently only supports MSSQL Password change, but more errors can be added.

Companion PR for: https://github.com/microsoft/azuredatastudio/pull/21627


cheenamalhotra:
It seems this RPC call to STS is avoidable as decision based on error code can be made by MSSQL extension itself. Can we skip changes to STS since we don't have a use-case where we need to communicate with server during error handling?

Any thoughts @alanrenmsft ?

alanrenmsft:
> It seems this RPC call to STS is avoidable as decision based on error code can be made by MSSQL extension itself. Can we skip changes to STS since we don't have a use-case where we need to communicate with server during error handling?
> 
> Any thoughts @alanrenmsft ?

@cheenamalhotra agreed, this can be in the mssql extension.

cheenamalhotra:
Closing as PR not needed anymore

PR #1823, opened 2023-01-25
Fix datetimeoffset formatting

caohai:
https://github.com/microsoft/azuredatastudio/issues/21011
With current implementation. `datetimeoffset` type will always be formatted with 7 digits regardless of the scale. This PR changes the format string to honor the scale of the column.

Unit tests and integration tests have been added.

Before:
image

After:
image

SQLCMD:
image


alanrenmsft (approved):

alanrenmsft:
@caohai please remember to add size label to the issue.

PR #1827, opened 2023-01-26
Adding missing nuget references for SqlProjects functionality

Benjin:
Also adds better error messaging in the unit test failures

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.Test.Common/RequestContextMocking/RequestContextMocks.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.Test.Common/RequestContextMocking/RequestContextMocks.cs
@@ -7,7 +7,9 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
 using Moq;
+using NUnit.Framework;
 
 namespace Microsoft.SqlTools.ServiceLayer.Test.Common.RequestContextMocking
 {
@@ -61,7 +63,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Test.Common.RequestContextMocking
         }
     }
 
-    public class MockRequest<T>
+    public class MockRequest<T> where T : ResultStatus
     {
         private T? result;
         public T Result => result ?? throw new InvalidOperationException("No result has been sent for the request");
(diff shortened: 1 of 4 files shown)


alanrenmsft:
so you are trying to add more logging to troubleshoot the test failures, right? @Benjin

alanrenmsft (approved):

test/Microsoft.SqlTools.ServiceLayer.Test.Common/RequestContextMocking/RequestContextMocks.cs
```diff
@@ -73,5 +75,15 @@ public MockRequest()
...
         }
+
+        /// <summary>
+        /// Asserts that this request was successful
+        /// </summary>
+        /// <param name="handlerName">Name of the handler, recommended to use nameof(service.MyHandler), used in the failure message</param>
+        /// <param name="descriptor">Optional extra descriptor, parenthesized in the failure message</param>
+        public void AssertSuccess(string handlerName, string? descriptor = null)
+        {
+            Assert.IsTrue(this.Result.Success, $"{handlerName} {(descriptor != null ? $"(descriptor) " : String.Empty)}expected to succeed, but failed with error: '{this.Result.ErrorMessage}'");
```
kisantia, line 86: nit: space before `expected`

kisantia (approved):

test/Microsoft.SqlTools.ServiceLayer.Test.Common/RequestContextMocking/RequestContextMocks.cs
Benjin: Oops, good catch.  Stuck the optional space in the wrong spot.

Benjin:
@alanrenmsft 
> so you are trying to add more logging to troubleshoot the test failures, right? @Benjin

Yup, not just these failures, but also any in the future.  "List of Scripts was expected to contain <path>" is so much more helpful of a failure message than "got 'false', but expected 'true'"

test/Microsoft.SqlTools.ServiceLayer.Test.Common/RequestContextMocking/RequestContextMocks.cs
Benjin: Wait, dangit.  This was right the first time: the spaces were after `{handlerName}` and (if descriptor was set) after `(descriptor)`.

But I did find that I did `(descriptor)` instead of `({descriptor})`, so not a total loss, lol.

PR #1828, opened 2023-01-27
Separating Migration into its own project 

aasimkhan30:
This PR creates a new project for Migration services. It includes the following changes.
1. Moving migration service into its own project with its own JSON - RPC server.
2. Removing dependency of Migration service on connection service. Now MDS is used to manipulate connection strings in migration service. 
3. I have modified the build script to create archives for migration service. Now, these archives will be published with STS archives during a release for the migration extension to download. 
4. Created a separate Integration Test project for Migration service. 

ADO Build: 
https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=187809&view=results


alanrenmsft:
@aasimkhan30 please queue a new STS build so that we can check the output.

src/Microsoft.SqlTools.Migration/Contracts/ValidateNetworkFileShare.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.Migration.Contracts
+{
+    public class ValidateNetworkFileShareRequestParams
+    {
+        public string Path { get; set; }
+        public string Username { get; set; }
+        public string Password { get; set; }
+    }
```
brian-harris, line 15: Should passwords/credentials be SecureString?

src/Microsoft.SqlTools.Migration/Contracts/ValidateWindowsAccount.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.Migration.Contracts
+{
+    public class ValidateWindowsAccountRequestParams
+    {
+        public string Username { get; set; }
+        public string Password { get; set; }
+    }
```
brian-harris, line 14: same here, should Password or other credentials be SecureString

src/Microsoft.SqlTools.Migration/Contracts/ValidateNetworkFileShare.cs
Charles-Gagnon: Given that this is already doing that in the existing code changing it now seems out of scope of this PR

https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Migration/Contracts/ValidateNetworkFileShare.cs#L14

So I'd suggest filing an issue for your team to follow up with after if you think it's a concern - although my understanding is that given that accessing process memory is going to require elevated privileges anyways (at which point you're already compromised) that it isn't a security concern to have it in plaintext in places like this. Especially since this isn't being stored long term anywhere (I hope) - but rather just a transient thing that's only around for the lifetime of the request. 

@shueybubbles You also mentioned something about SecureString being not so great anyways, do you happen to have a link to where you saw that from?

shueybubbles: in the horse's mouth!

https://learn.microsoft.com/en-us/dotnet/api/system.security.securestring?view=net-7.0#remarks

aasimkhan30:
@alanrenmsft and @Charles-Gagnon https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=187809&view=results

alanrenmsft (approved):

PR #1832, opened 2023-01-30
Support for database references in SqlProjects service

Benjin:


src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/DatabaseReferences/AddSqlProjectReferenceParams.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// SQLCMD variable name for specifying the other server this reference is to, if different from that of the current project.
+        /// If this is set, DatabaseVariable must also be set.
+        /// </summary>
+        public string? ServerVariable { get; set; }
+    }
+
+
+    public class AddSqlProjectReferenceRequest
+    {
+        public static readonly RequestType<AddSqlProjectReferenceParams, ResultStatus> Type = RequestType<AddSqlProjectReferenceParams, ResultStatus>.Create("sqlprojects/addSqlProjectReference");
```
alanrenmsft, line 36: As far as I am aware, we are only using lower case in the request name in all other places, you might want to be consistent with the current pattern.

lewis-sanchez: I was once told they're easier to read if you camel case them. I think @Charles-Gagnon told me that, so I ended up doing it for query execution plan comparison request endpoints..
https://github.com/microsoft/sqltoolsservice/blob/e2962390bc6b521ee33a410611d08384f89cba43/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/Contracts/ExecutionPlanComparisonRequest.cs#L46

Charles-Gagnon: It's definitely easier to read IMO, but it doesn't really matter (JSON RPC message names are case insensitive) so whatever is your preference. But readability helps avoid silly spelling mistakes and the like too so there's something to be said for that...

Benjin: @alanrenmsft when choosing from multiple preexisting patterns, I support the use of the one with better readability. :)

alanrenmsft: :) fine by me.

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/DatabaseReferences/AddDacpacReferenceParams.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// SQLCMD variable name for specifying the other server this reference is to, if different from that of the current project.
+        /// If this is set, DatabaseVariable must also be set.
+        /// </summary>
+        public string? ServerVariable { get; set; }
+    }
+
+    public class AddDacpacReferenceRequest
+    {
+        public static readonly RequestType<AddDacpacReferenceParams, ResultStatus> Type = RequestType<AddDacpacReferenceParams, ResultStatus>.Create("sqlprojects/addDacpacReference");
```
lewis-sanchez, line 30: Feel free to capitalize the P in "sqlprojects".

lewis-sanchez (approved):

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/DatabaseReferences/AddSystemDatabaseReference.cs
```diff
@@ -0,0 +1,27 @@
...
+//
+
+using Microsoft.SqlServer.Dac.Projects;
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.SqlProjects.Contracts
+{
+    /// <summary>
+    /// Parameters for adding a System Database
```
kisantia, line 13: `...adding a reference to a System Database`

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
```diff
@@ -157,6 +159,95 @@ await service.HandleDeleteSqlObjectScriptRequest(new SqlProjectScriptParams()
...
+        {
+            // Setup
+            SqlProjectsService service = new();
+            string projectUri = await service.CreateSqlProject();
+
+            SqlCmdVariable databaseVar = new SqlCmdVariable("$(OtherDb)", "OtherDbDefaultValue", "OtherDbValue");
+            SqlCmdVariable serverVar = new SqlCmdVariable("$(OtherServer)", "OtherServerDefaultValue", "OtherServerValue");
+
+            service.Projects[projectUri].SqlCmdVariables.Add(databaseVar);
+            service.Projects[projectUri].SqlCmdVariables.Add(serverVar);
```
kisantia, line 173: just to make sure I understand how this will work - ADS will be the one responsible for making the call to STS for adding any sqlcmd variables needed for a database reference? Will the database reference not get added if there's an error trying to add a sqlcmd var (like if the sqlcmd variable already exists) or will it still get added? Or do we need to add validation in ADS for that?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
```diff
@@ -157,6 +159,95 @@ await service.HandleDeleteSqlObjectScriptRequest(new SqlProjectScriptParams()
...
+
+            Assert.AreEqual(0, service.Projects[projectUri].DatabaseReferences.Count, "Baseline number of database references");
+
+            // Validate adding a system database reference
+            MockRequest<ResultStatus> requestMock = new();
+            await service.HandleAddSystemDatabaseReferenceRequest(new AddSystemDatabaseReferenceParams()
+            {
+                ProjectUri = projectUri,
+                SystemDatabase = SystemDatabase.MSDB,
+                DatabaseVariable = "$(EmEssDeeBee)",
```
kisantia, line 183: I was so confused when I first read this haha 😆

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
```diff
@@ -157,6 +159,95 @@ await service.HandleDeleteSqlObjectScriptRequest(new SqlProjectScriptParams()
...
+
+            requestMock = new();
+            await service.HandleAddSqlProjectReferenceRequest(new AddSqlProjectReferenceParams()
+            {
+                ProjectUri = projectUri,
+                ProjectPath = mockReferencePath,
+                ProjectGuid = TEST_GUID,
+                SuppressMissingDependencies = false,
+                DatabaseVariable = databaseVar.Name,
+                ServerVariable = serverVar.Name
```
kisantia, line 226: can one of the references in this test not have DatabaseVariable and ServerVariable to cover the "Same database" reference case?

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/DatabaseReferences/AddDacpacReferenceParams.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Parameters for adding a Dacpac reference to a SQL project
+    /// </summary>
+    public class AddDacpacReferenceParams : AddDatabaseReferenceParams
+    {
+        /// <summary>
+        /// Path to the .dacpac file
+        /// </summary>
+        public string DacpacPath { get; set; }
```
kisantia, line 19: do we enforce this being a relative path to the .sqlproj? That's what we did when adding support for dacpac references in ADS to avoid having full paths to dacpacs in .sqlprojs get checked into source control

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
Benjin: > ADS will be the one responsible for making the call to STS for adding any sqlcmd variables needed for a database reference?

Correct.

> Will the database reference not get added if there's an error trying to add a sqlcmdvar?  Or do we need validation in ADS for that?

We'd need validation in ADS; the database reference will throw an `InvalidOperationException` when being added to a project if that project doesn't already contain a variable by that name.

---

Since the XML for database references only contain the name of the variable(s), but not the value or default value, there isn't enough information in _just_ the database reference to automatically add those variables, right?  Unless we wanted to do some sorta smart default sorta thing, but idk _what_ a smart default for `$(OtherServer)` and `$(OtherDatabase)` would even be...

Another possibility is that we could change the method signature to accept `SqlCmdVariable` objects for the database and server variables, and use those to add new vars if they aren't there.  That'd probably require us to add some new `Get()` functions to the collections (easy enough), but also to define the behavior if the passed variable matches on name but not value/default value (do we update? do we fail?).

Thoughts?

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/DatabaseReferences/AddDacpacReferenceParams.cs
Benjin: Not in the DacFx library 'cause absolute paths are technically still valid, and that's meant to be general-purpose.  Think we should convert to relative paths at either the STS or ADS level?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
kisantia: The SSDT and ADS "Add database reference" dialogs get the sqlcmd variable name and value so that there's the information needed for adding a new sqlcmd variable. 

I also just noticed that the database variable isn't required, so if only database name is given, then `DatabaseVariableLiteralValue` is used instead of `DatabaseSqlCmdVariable`. Is this case also handled?
image

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/DatabaseReferences/AddDacpacReferenceParams.cs
kisantia: ADS makes sense since the paths are coming from there anyway, just have to remember to do that later :)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
Benjin: I was thinking about these tests as validating the unique parts of the Tools Service layer, so they'd need to make sure that all parameters are copied from the `Add____ReferenceParams` object to the constructor for the `____Reference` objects in their respective handlers.

Same database reference cases are handled and tested down in DacFx here, so I don't _think_ there's anything new to test up here.  Maybe I missed something, though?

https://msdata.visualstudio.com/SQLToolsAndLibraries/_git/DacFx?path=/Test/UnitTests/UTProjects/DatabaseReferenceTests.cs&version=GBmain&line=28&lineEnd=30&lineStartColumn=13&lineEndColumn=1&lineStyle=plain&_a=contents

Ofc, if you think it'd be better to cover those cases here as well, just let me know and I can add them!

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
Benjin: Err, right.  All of these cases should be covered, as long as the caller (whether that's STS directly or ADS via STS) creates the variables first.  I thought it made more sense to have ADS do two simple calls (one for variables, one for the reference) than one more complicated one that covers all the cases of both aspects.

Oh, shoot - I don't think it does.  I'll message you via teams to make sure I understand how that works.

kisantia (approved):

PR #1837, opened 2023-01-31
Fix broken STS language service tests

Charles-Gagnon:
RebuildIntellisenseCacheClearsScriptParseInfoCorrectly started failing since https://github.com/microsoft/sqltoolsservice/pull/1825. Investigating this lead me to find out that it was an issue with stuff from previous tests not being cleaned up correctly - in this case I was able to repro it consistently if I ran one of the ExpandSqlStart tests before this one.

Because all the tests in this file share a common "starter" file (sqltest.sql) what would happen is one test would open the file and do some stuff, which then caused the service to start parsing that file in the background for intellisense. Since we weren't cleaning up the service, when another test came along and started running it would be using the same service instance that still had tasks running in the background to add and update ScriptParseInfos - so when the new test did anything it would often get the incorrect ScriptParseInfo from the previous test (because it shared the same file name).

Best solution here is to make each test self-contained, i.e. create a new language service instance just for that test. That way we're guaranteed that nothing between the tests will affect each other (at least as long as they aren't using global objects directly themselves. For example I had to also set up a new workspace service instance because otherwise it would default to the global workspace instance and run into the same problem). 

This was actually already being done in the HandleRequestToChangeToSqlcmdFile test - although unfortunately that one is still broken to some other issue (null ref error) so I'm leaving it as disabled.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -181,7 +181,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             }
         }
 
-        private CancellationTokenSource existingRequestCancellation;
+        private CancellationTokenSource? existingRequestCancellation;
 
         /// <summary>
         /// Gets or sets the current workspace service instance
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -181,7 +181,7 @@ internal ConnectionService ConnectionServiceInstance
             }
         }
 
-        private CancellationTokenSource existingRequestCancellation;
+        private CancellationTokenSource? existingRequestCancellation;
```
Charles-Gagnon, line 184: Did a bit of other minor cleanup while I was looking through these files too.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -218,7 +218,6 @@ private string CopyFileWithNewModifiedTime(string filePath)
         /// provide signature help.
         /// </summary>
         [Test]
-        [Ignore("this test is not stable")]
```
Charles-Gagnon, line 221: Not 100% sure that this will fix this test, but worth a shot.

alanrenmsft (approved):

PR #1841, opened 2023-02-01
Normalize string to fix test case on Unix

Benjin:
Also lowers a rule for handling all cases for a switch to a suggestion, as it ignores the presence of a discard branch that intentionally handles the remaining cases.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.Test.Common/FileUtils.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.Test.Common/FileUtils.cs
@@ -76,5 +76,25 @@ namespace Microsoft.SqlTools.ServiceLayer.Test.Common
                 }
             }
         }
+
+        /// <summary>
+        /// Normalizes Windows, Unix, and mixed paths to the same slash direction, specified by <paramref name="separatorType"/>
+        /// </summary>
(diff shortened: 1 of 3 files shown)


kisantia (approved):

test/Microsoft.SqlTools.ServiceLayer.Test.Common/FileUtils.cs
```diff
@@ -76,5 +76,25 @@ public static string TestServerNamesDefaultDirectory
...
+        /// <param name="separatorType">Win32NT for \, Unix for /</param>
+        /// <returns></returns>
+        public static string NormalizePath(string path, PlatformID separatorType)
+        {
+            return separatorType switch
+            {
+                PlatformID.Win32NT => path.Contains('/')
+                                ? String.Join('\\', path.Split('/', StringSplitOptions.RemoveEmptyEntries))
+                                : path,
+                PlatformID.Unix => path.Contains('\\')
```
Charles-Gagnon, line 93: What about Mac? Or does that get set as Unix?

Benjin: Mac and Linux are both Unix-based OSes, so that covers both.

PR #1846, opened 2023-02-07
Enable local logging for login migrations

AkshayMata:
This change allows the login migration nuget to log changes locally:
image


junierch (approved):

raymondtruong:
should this be updated in the old code (pre- migration service separation) as well? I guess it doesn't matter much for back compat since login migration was only just released recently?

raymondtruong (approved):

AkshayMata:
@raymondtruong that's a great question. Agreed for login migrations it is not as relevant since it is a new feature. 

Let's discuss offline. There will be some edge cases for backwards compat, but I think by and large it should flow through smoothly. We should probs figure out a plan on if we want to support both code paths and for how long.

PR #1852, opened 2023-02-08
Support None Attestation Protocol

cheenamalhotra:
Small change to support 'None' attestation protocol with MDS 5+


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1364,6 +1364,9 @@ public static SqlConnectionStringBuilder CreateConnectionStringBuilder(Connectio
                     case "HGS":
                         connectionBuilder.AttestationProtocol = SqlConnectionAttestationProtocol.HGS;
                         break;
+                    case "None":
```
Charles-Gagnon, line 1367: What about using [Enum.TryParse](https://learn.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-7.0) instead?

cheenamalhotra: I'll get it done in another PR (wip)

PR #1855, opened 2023-02-09
fix db list issue for contained users

alanrenmsft:
for issue: https://github.com/microsoft/azuredatastudio/issues/21884
we attempted to fix the issue with this PR: https://github.com/microsoft/sqltoolsservice/pull/957, but introduced the issue: https://github.com/microsoft/azuredatastudio/issues/10816, so it was reverted.

I am fixing the issue by first try the master if fails then fall back to the original database to cover both scenarios.

scenario 1: have access to other databases while connected to a particular database directly, this works like before.
image

scenario 2: user doesn't have access to master.

before:
image

after:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
@@ -8,11 +8,13 @@ using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
+using System.Diagnostics;
 using System.Linq;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlTools.ServiceLayer.Admin.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.Connection
 {
@@ -90,6 +92,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                     // https://learn.microsoft.com/sql/relational-databases/errors-events/mssqlserver-18456-database-engine-error
                     if (i != databasesToTry.Count - 1 && ex.Number == 18456)
                     {
+                        Logger.Write(TraceEventType.Information, string.Format("Failed to get database list from database '{0}', will fallback to original database.", databasesToTry[i]));
                         continue;
                     }
                     else
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
@@ -12,6 +12,7 @@ using System.Linq;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlTools.ServiceLayer.Admin.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.Connection
 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -64,39 +65,74 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
...
+                    connectionDetails.DatabaseName = databasesToTry[i];
+                    var results = this.GetResults(connectionFactory, connectionDetails);
+                    SetResponse(response, results);
+                    break;
+                }
+                catch (Microsoft.Data.SqlClient.SqlException ex)
+                {
+                    // Retry when login attempt failed.
+                    // https://learn.microsoft.com/sql/relational-databases/errors-events/mssqlserver-18456-database-engine-error
+                    if (i != databasesToTry.Count - 1 && ex.Number == 18456)
```
kburtram, line 91: it may be worth logging a debug message here in case this results in any unexpected behavior.

kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
alanrenmsft: added logging.

Charles-Gagnon:
@shueybubbles Have you looked at contained DB issues in SSMS at all? Alan mentioned that there's an issue with them in SSMS as well where the dropdown will show master but selecting it freezes SSMS.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
```diff
@@ -64,39 +67,75 @@ abstract class ListDatabaseRequestHandler<T> : IListDatabaseRequestHandler
 
         public ListDatabasesResponse HandleRequest(ISqlConnectionFactory connectionFactory, ConnectionInfo connectionInfo)
         {
+            ListDatabasesResponse response = new ListDatabasesResponse();
             ConnectionDetails connectionDetails = connectionInfo.ConnectionDetails.Clone();
+            // Running query against sys.databases view will only return a subset of databases the current login/user might have access to, we need to
+            // query the master database to get the full database list, but for users without master db access, we have to query the
+            // original database as a fallback.
+            var databasesToTry = new List<string>() { CommonConstants.MasterDatabaseName };
```
shueybubbles, line 75: For Azure SQL DB it's way more common for users not to have any access to databases other than the specific one they connected to. Querying `master` is likely to be a waste of time. Even if they do have limited access to master, a successful query will likely lead to the list being populated with potentially a huge number of databases, many of which the user can't access. Can't we just fail gracefully if the attempt to access master fails?

The other option is if the connection string specifies a non-master database and the database is Azure DB, we just remove master from the list for switching. I consider it somewhat unfortunate that `sys.databases` in Azure DB includes `master` when querying a user database.

shueybubbles:
> @shueybubbles Have you looked at contained DB issues in SSMS at all? Alan mentioned that there's an issue with them in SSMS as well where the dropdown will show master but selecting it freezes SSMS.

We fixed all the issues with Object Explorer trying to query `master` but I don't recall anything specific about the query editor.

src/Microsoft.SqlTools.ServiceLayer/Connection/ListDatabaseRequestHandler.cs
alanrenmsft: Agreed, it is all because of SQL server includes master and tempdb (on prem only) when querying a userdb with contained user. I was also thinking about only return the current database name for this scenario.

But we can't simply remove master from the list when connecting to a user database directly, see the issue I mentioned in the PR description: https://github.com/microsoft/azuredatastudio/issues/10816. we have to query the master first and then fallback.

alanrenmsft:
this is how SSMS behaves:
connect with a contained user, you can see master in the db dropdown.
image

select master, the following 2 errors will pop up and then the connection is closed.
image
image

PR #1857, opened 2023-02-09
Minor follow-ups for user management support classes

kburtram:
A couple minor issues missed when merging https://github.com/microsoft/sqltoolsservice/pull/1856/files.  Only the ResourceManager change is fixing a bug.  Other is deleting some dead-code.


kburtram:
Trivial follow-up change

PR #1858, opened 2023-02-09
Adding support for database literal references for Dacpac and SqlProj references

Benjin:
Also removes some `nullable` suppressions that aren't necessary.


src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/DatabaseReferences/AddDatabaseReferenceParams.cs
```diff
@@ -18,8 +16,8 @@ public abstract class AddDatabaseReferenceParams : SqlProjectParams
         public bool SuppressMissingDependencies { get; set; }
 
         /// <summary>
-        /// SQLCMD variable name for specifying the other database this reference is to, if different from that of the current project
+        /// Database
```
kisantia, line 19: Match the comment in DacFx for more detail on when this is used vs DatabaseVariable? `Literal name used to reference another database in the same server, if not using SQLCMD variables`

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
```diff
@@ -304,7 +304,7 @@ await service.HandleAddSystemDatabaseReferenceRequest(new AddSystemDatabaseRefer
             {
                 ProjectUri = projectUri,
                 SystemDatabase = SystemDatabase.MSDB,
-                DatabaseVariable = "$(EmEssDeeBee)",
+                DatabaseLiteral = "$(EmEssDeeBee)",
```
kisantia, line 307: I thought DatabaseLiterals don't need the `$()`

kisantia (approved):
There's an integration test failing that needs to be fixed before checking in

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
Benjin: Oh yeah, good catch.

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/Folders/AddFolder.cs
```diff
@@ -3,6 +3,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+#nullable disable
```
SakshiS-harma, line 6: Added by mistake? Or is it necessary?

SakshiS-harma (approved):

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/Folders/AddFolder.cs
Benjin: Intentional.  It's a bit of a funky balance: we want to be able to turn on full nullable checks, but because the JRPC deserializer uses a blank constructor followed by property sets, we can't enforce that "required" properties (like `Path` below) are actually set.  But it's also not meant to be nullable, so I don't want to mark it as `string?` to imply that it's okay if it's null.

The choice then becomes either setting Path to be `string?` and force everything that uses it to handle null cases, or add the `#nullable disable` hint back.

... Or we could probably do something like having a secondary field with proper nullability (or lack thereof) that actually gets used in Tools Service:

```
public string? Path { get; set; } // set by the JRPC deserializer, never called except below
public string GetPath => Path ?? throw new ArgumentNullException(); // used in code, handles the null check for everything that uses it
```

... Or maybe if the JRPC deserializer uses init syntax to set these, they we can use new accessor scopes:

```
public string Path {get; init set; }
```

But that's something I'll have to play with.

Benjin:
Integration test is failing due to a cross-platform bug in the Projects library.  Fix is in PR: https://msdata.visualstudio.com/SQLToolsAndLibraries/_git/DacFx/pullrequest/986511

PR #1860, opened 2023-02-10
Introduce AAD interactive auth mode

cheenamalhotra:
Introduces "ActiveDirectoryInteractive" authentication support with `SqlAuthenticationProvider` integration.

### Azure Data Studio

A new setting is introduced in ADS with https://github.com/microsoft/azuredatastudio/pull/21903 that enables this auth mode when user selects "AzureMFA" authentication mode.

image

The setting is disabled by default, and can be enabled by users for initial release. This should provide a safe fallback in case there're extensions that get impacted and feature can be primarily tested by users most interested in this support. This setting  applies to MSAL authentication only.

### VS Code MSSQL Extension

A new setting is introduced in VSCode MSSQL Extension with https://github.com/microsoft/vscode-mssql/pull/17562 that enables this auth mode when user selects "AzureMFA" authentication mode. This setting  applies to MSAL authentication only.

image

The setting is **enabled** by default.

### Microsoft.SqlTools.Authentication
This project is kept isolated with Authentication related types, so it can be also leveraged in SqlMigration and Kusto projects in future, and also can be extended to be used by AzureCore extension to transfer authentication calls to MSAL.NET.

#### Cache Persistence:
Same cache location is used as with MSAL.js - in order to share accounts authenticated with MSAL.js with MSAL.NET without the need to reauthenticate.

Pending Tasks:
- [ ] Verify Cache path is correct and cache is shared in Linux (@cheenamalhotra)
- [ ] Verify Cache path is correct and cache is shared in macOS (@cssuh)


src/Microsoft.SqlTools.Authentication/Authenticator.cs
```diff
@@ -0,0 +1,242 @@
...
+
+namespace Microsoft.SqlTools.Authentication
+{
+    /// <summary>
+    /// Provides APIs to acquire access token using MSAL.NET v4 with provided <see cref="AuthenticationParams"/>.
+    /// </summary>
+    public class Authenticator
+    {
+        private const string ApplicationClientId = "a69788c6-1d43-44ed-9ca3-b83e194da255";
+        private const string ApplicationName = "azuredatastudio";
```
kburtram, line 19: this library will be used in other applications, such as vscode or potentially other CLIs (mssql-scripter, etc., though this is a less important use-case).  Should this be configurable?

src/Microsoft.SqlTools.Authentication/Authenticator.cs
```diff
@@ -0,0 +1,242 @@
...
+            {
+                var appData = Environment.GetEnvironmentVariable("APPDATA");
+                var userProfile = Environment.GetEnvironmentVariable("USERPROFILE");
+                if (appData != null)
+                {
+                    return appData;
+                }
+                else if (userProfile != null)
+                {
+                    return String.Join(Environment.GetEnvironmentVariable("USERPROFILE"), "AppData", "Roaming");
```
kburtram, line 208: nit: .Net Framework coding guidelines suggests using primitive types when available, e.g., `string.Join`, etc.

src/Microsoft.SqlTools.Authentication/Authenticator.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        public string BuildDirectoryPath()
+        {
+            var homedir = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+            // Windows
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                var appData = Environment.GetEnvironmentVariable("APPDATA");
+                var userProfile = Environment.GetEnvironmentVariable("USERPROFILE");
+                if (appData != null)
```
kburtram, line 202: I guess this logic matches what's in MSAL.JS to share the cache?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -154,6 +155,12 @@ public static async Task<string> AzureActiveDirectoryAuthenticationCallback(stri
             return response.Token;
         }
 
+        /// <summary>
+        /// Enables configured 'Sql Authentication Provider' for 'Active Directory Interactive' authentication mode to be used 
+        /// when user chooses 'Azure MFA'.
+        /// </summary>
+        public static bool EnableSqlAuthenticationProvider { get; set; }
```
kburtram, line 162: the services are singleton so we generally don't statics for state

src/Microsoft.SqlTools.ServiceLayer/Program.cs
```diff
@@ -49,13 +52,21 @@ internal static async Task Main(string[] args)
...
 
+
                 // set up the host details and profile paths 
                 var hostDetails = new HostDetails(version: new Version(1, 0));
 
                 SqlToolsContext sqlToolsContext = new SqlToolsContext(hostDetails);
                 ServiceHost serviceHost = HostLoader.CreateAndStartServiceHost(sqlToolsContext);
                 serviceHost.MessageDispatcher.ParallelMessageProcessing = commandOptions.ParallelMessageProcessing;
 
+                if (commandOptions.EnableSqlAuthenticationProvider)
```
kburtram, line 63: this doesn't seem to be the typical pattern we're using to get state into a service.  Maybe passing into the service constructor is more scalable then special-casing a particular option in the main method?

kburtram (commented):
Looks good overall.  Given the amount of non-trivial changes I'll take a second pass through the code later today when I have more time.

Packages.props
```diff
@@ -15,6 +15,8 @@
 		<PackageReference Update="Microsoft.Extensions.DependencyModel" Version="3.1.4" />
 		<PackageReference Update="Microsoft.Extensions.FileSystemGlobbing" Version="7.0.0" />
 
+		<PackageReference Update="Microsoft.Identity.Client" Version="4.49.1" />
```
kburtram, line 18: can you confirm these dependencies are compatible with macos arm64?  I assume they likely are but these has been a problem with other dependencies we're trying to fix.

src/Microsoft.SqlTools.Authentication/AuthenticationParams.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <summary>
+        /// <see cref="Guid"/> Connection Id, that will be passed to Azure AD when requesting access token.
+        /// It can be used for tracking AAD request status if needed.
+        /// </summary>
+        public Guid ConnectionId { get; set; }
+
+        /// <summary>
+        /// User name to be provided as userhint when acquiring access token.
+        /// </summary>
+        public string Username { get; set; }
```
kburtram, line 43: I think most other places in this project we can this `UserName`.  Maybe better to be consistent if there isn't a good reason not to be?

src/Microsoft.SqlTools.Authentication/Authenticator.cs
```diff
@@ -0,0 +1,242 @@
...
+    /// </summary>
+    public class Authenticator
+    {
+        private const string ApplicationClientId = "a69788c6-1d43-44ed-9ca3-b83e194da255";
+        private const string ApplicationName = "azuredatastudio";
+        private const string AzureTokenFolder = "Azure Accounts";
+        private const string MSAL_CacheName = "azureTokenCacheMsal-azure_publicCloud";
+        private const string RedirectUri = "http://localhost";
+
+        private static ConcurrentDictionary<string, IPublicClientApplication> s_pcaMap
```
kburtram, line 24: we aren't generally using type/scoping variable name prefixes, such as `s_`, in this codebase.  Generally better to use similar style throughout the code if possible, IMO.

src/Microsoft.SqlTools.Authentication/Authenticator.cs
```diff
@@ -0,0 +1,242 @@
...
+            return clientApplicationInstance;
+        }
+
+        private IPublicClientApplication CreatePublicClientAppInstance(string authority, string audience) =>
+            PublicClientApplicationBuilder.Create(ApplicationClientId)
+                .WithAuthority(authority, audience)
+                .WithClientName(ApplicationName)
+                .WithRedirectUri(RedirectUri)
+                .WithClientId(ApplicationClientId)
+                .WithLogging((logLevel, message, pii) =>
```
kburtram, line 143: I thought we had a setting to turn on PII logging, IIRC.  Do we no longer want to support that or am I misremembering what we currently support?

src/Microsoft.SqlTools.Authentication/Authenticator.cs
```diff
@@ -0,0 +1,242 @@
...
+                 * but once the user finishes authentication, the web page is redirected in such a way that MSAL can intercept the Uri.
+                 * MSAL cannot detect if the user navigates away or simply closes the browser. Apps using this technique are encouraged
+                 * to define a timeout (via CancellationToken). We recommend a timeout of at least a few minutes, to take into account
+                 * cases where the user is prompted to change password or perform 2FA.
+                 *
+                 * https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/System-Browser-on-.Net-Core#system-browser-experience
+                 */
+                ctsInteractive.CancelAfter(180000);
+
+                return await publicClientApplication.AcquireTokenInteractive(scopes)
```
kburtram, line 182: IIUC, this will cause a browser to be launched from STS?  Currently STS is designed to explicitly be a non-UI process, to support scenarios such as remote extension host or webmode.  We haven't made a decision that we won't support those scenarios in the future and having STS directly interact UI would not work in those scenarios.

kburtram: FWIW, thought what we were planning to support was ADS gettting token through the browser and then providing that token to SqlClient.  Not having STS show the browser itself.

kburtram (commented):
If I understand correctly this PR is adding direct UI interaction to STS, which isn't something that this process was intended to support and is a significant design change.  Please let me know if I'm misunderstanding what this PR is doing.

src/Microsoft.SqlTools.Authentication/AuthenticationParams.cs
```diff
@@ -0,0 +1,64 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.Authentication
+{
+    /// <summary>
+    /// Parameters to be passed to <see cref="Authenticator"/> to request an access token
+    /// </summary>
+    public class AuthenticationParams
```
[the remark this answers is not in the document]
cheenamalhotra: There are no new warnings in new projects.

src/Microsoft.SqlTools.Authentication/Authenticator.cs
cheenamalhotra: Agreed, I can accept this in constructor, I was also thinking of extracting full cache location out of authenticator if so it can be controlled by caller, and as we're looking at other tools, I think it makes most sense to do so..

src/Microsoft.SqlTools.Authentication/Authenticator.cs
cheenamalhotra: I couldn't find PII logging available in STS, we have the setting in Azure Core extension, I also thought it's generally not enabled when requesting logs from end users, so I disabled it.

We'll have to transfer setting here in order to enable it. Let me know if want me to implement it.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: I needed to use this in static API, that's why made it static.. will attach it to instance instead.

src/Microsoft.SqlTools.Authentication/Authenticator.cs
```diff
@@ -0,0 +1,242 @@
...
+                            {
+                                account = currentVal;
+                                break;
+                            }
+                        }
+                        while (accounts.MoveNext());
+                    }
+                    else
+                    {
+                        account = accounts.Current;
```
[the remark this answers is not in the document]
cheenamalhotra: The authorized accounts in the list are applicable to a particular tenant and resource combination (as it's unique for a public client app), but in our case we can have multiple accounts here, I'll remove this fallback and strictly use login account only.

Packages.props
cheenamalhotra: We're building x64 arch in ARM64 bits, so there should be no concerns. cc @alanrenmsft

src/Microsoft.SqlTools.Authentication/Authenticator.cs
cheenamalhotra: I'll see what best we can do, this should ideally never ever be hit.
Will throw error for now and trigger authentication in ADS when exception received.

src/Microsoft.SqlTools.Authentication/Authenticator.cs
kburtram: thanks the details!  It's up to you whether it's worth implementing all the way through to here, but I do know that lack of logging has made several AAD related investigations very difficult.  And we added the PII logging to help with that.  It may be a good idea to anticipate getting issues where it is hard to repro the failure and the more detailed logs we can collect the better.

src/Microsoft.SqlTools.Authentication/Authenticator.cs
kburtram: OK, that makes sense and I think I recall you mentioning this case in the design discussion, IIRC.  ADS doesn't currently ship in remote/web scenarios (though we may in the future if needed) but VS Code does, so this wouldn't work in those scenarios.  If there is a way to send the error to the UI client to launch the browser that would be a more maintainable/flexible design IMO.

src/Microsoft.SqlTools.Authentication/AuthenticationParams.cs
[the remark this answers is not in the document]
kburtram: cool, I missed the constructor was initializing all the properties.  Looks good.  Thanks!

src/Microsoft.SqlTools.Authentication/Authenticator.cs
cheenamalhotra: I think we can use `SecurityTokenRequest` to request token from ADS' registered Azure provider (Azure Core) asynchronously and be able to await on it, as done in `AzureActiveDirectoryAuthenticationCallback` in `ConnectionService`. However, the request/response format need a bit of rework to be able to pass necessary information and get back response with expiry date. I can use a `delegate` in `ConnectionService` that can make this request and send it to the `AuthenticationProvider`, to be able to attach Hosting service.

That should work 🤔

cheenamalhotra: @kburtram I've added design changes for this, please take a look.
Meanwhile I'm trying to run some tests by forcing this flow - in normal cases/cache resets I can't get this triggered yet.

cheenamalhotra: I tested this, and you were right, we do hit a deadlock with awaited call - I'm changing back to throwing exceptions so we can catch them when needed. At the moment, our design ensures user account is added and not stale before we even connect/reconnect, and MSAL token cache is reliable, so silent authentication should work every-time we hit STS including cases like OE expand/refresh/dashboard calls where token needs to be refreshed. This new class implementation will be able to always fetch token silently from MSAL cache.

If in any case exception is thrown, it will provide specific information so either we will be able to address the exception by catching and handling it or user will be prompted to take necessary action.

cheenamalhotra:
Proceeding to merge, will follow up validations with insiders/ad-hoc builds

PR #1861, opened 2023-02-10
Adding script move support for SqlProject service

Benjin:


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
```diff
@@ -207,23 +224,40 @@ await service.HandleAddPreDeploymentScriptRequest(new SqlProjectScriptParams()
             }, requestMock.Object);
 
             requestMock.AssertSuccess(nameof(service.HandleAddPreDeploymentScriptRequest));
-            Assert.AreEqual(1, service.Projects[projectUri].PreDeployScripts    .Count, "PreDeployScripts count after re-add");
+            Assert.AreEqual(1, service.Projects[projectUri].PreDeployScripts.Count, "PreDeployScripts count after re-add");
+
+            // Validate moving a pre-deployment object script
+            string movedScriptRelativePath = @"SubPath\\RenamedPreDeploymentScript.sql";
```
kisantia, line 230: can the `\\` be replaced with `\` since this string has `@` in front?

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/SqlObjectScripts/MoveSqlObjectScript.cs
```diff
@@ -0,0 +1,28 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
```
kburtram, line 6: for simple files like this you may consider making them nullable-compliant when adding them to avoid accumulating debt in this area.

kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/SqlObjectScripts/MoveSqlObjectScript.cs
Benjin: This was intentional; I elaborated on why over in this comment in another PR: https://github.com/microsoft/sqltoolsservice/pull/1858/files#r1102060833

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
Benjin: Not just "can" but "should be"!

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/SqlObjectScripts/MoveSqlObjectScript.cs
kburtram: Sounds good, as long as you're making a deliberate choice here.  I assume there will likely always be some code in this project where we'll keep nullable disabled for various reasons.

PR #1864, opened 2023-02-16
Removing migration code and packages from sts

aasimkhan30:


brian-harris (approved):

kburtram (approved):

raymondtruong (approved):

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -27,9 +27,6 @@
 using Microsoft.SqlTools.ServiceLayer.LanguageExtensibility;
 using Microsoft.SqlTools.ServiceLayer.LanguageServices;
 using Microsoft.SqlTools.ServiceLayer.Metadata;
-#if INCLUDE_MIGRATION
```
Charles-Gagnon, line 30: Remove this from the csproj as well and clean up the when condition

corivera: You can remove the entire Choose element here apart from this one PropertyGroup element: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj#L26

Then add `Condition="'$(BUILD_DOTNET_TOOL)' == 'true'"` to the ProertyGroup tag like we do in the Kusto csproj here: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj#L25

Charles-Gagnon (approved):

AkshayMata:
Hi @aasimkhan30,

I had a thought that I wasn't able to get in the [first PR](https://github.com/microsoft/sqltoolsservice/pull/1828) and was hoping to address in this PR. 

I wasn't able to review because this PR was sent out after 7pm EST and merged before 7am. I did mention offline that it was outside business hours and was planning on reviewing in the morning. 

Is there a way to preserve history between the old files and new files? I was thinking we could explore git mv to do so. I know that the history is still saved removing like we did here, but it much harder to navigate to.

PR #1866, opened 2023-02-16
Moving out legacy schemas into their own folder

aasimkhan30:
Moving legacy schemas into their own folder in OE.
The list of legacy schemas was obtained from: https://learn.microsoft.com/en-us/sql/relational-databases/security/authentication-access/ownership-and-user-schema-separation?view=sql-server-ver16#built-in-schemas-for-backward-compatibility
image


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -209,7 +209,7 @@ public override IEnumerable<INodeFilter> Filters
                 {
                     Property = "IsSystemObject",
                     Type = typeof(bool),
-                    Values = new List<object> { 0 },
+                     Values = new List<object> { 0 },
```
lewis-sanchez, line 212: Nit: get rid of the added whitespace

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -387,7 +387,7 @@ public override IEnumerable<INodeFilter> Filters
                 {
                     Property = "IsSystemObject",
                     Type = typeof(bool),
-                    Values = new List<object> { 1 },
+                     Values = new List<object> { 1 },
```
lewis-sanchez, line 390: NIt: get rid of the added whitespace before `values`

alanrenmsft:
where is the requirement coming from? is it appropriate to call them "legacy"?

aasimkhan30:
@alanrenmsft, drew had mentioned about it here.  https://github.com/microsoft/vscode-mssql/pull/17543#issuecomment-1430698289

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -680,6 +680,87 @@ internal partial class DatabaseChildFactory : SmoChildFactoryBase
...
+                });
+                filters.Add(new NodePropertyFilter
+                {
+                    Property = "Name",
+                    Type = typeof(string),
+                    FilterType = FilterType.CONTAINS,
+                    IsNotFilter = true,
+                    Values = new List<object> { "db_securityadmin" },
+                });
+                return filters;
```
lewis-sanchez, line 760: Is there a way to assemble the list of filters outside the getter and just return it immediately without needing to assemble every time?

Charles-Gagnon:
I would go with his suggestion of just putting them at the end instead of in a separate folder to start with then, unless we get specific feedback that people want those hidden. 

The term "legacy" is definitely questionable and would be something we'd want to run by the engine owners first I'd think to make sure they're ok with the term. So for now just moving them to the end is likely fine for our purposes.

alanrenmsft (approved):

Charles-Gagnon:
@dzsquared Thoughts on this?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -708,6 +789,15 @@ protected override void OnExpandPopulateFolders(IList<TreeNode> currentChildren,
                     IsSystemObject = false,
                     ValidFor = ValidForFlag.AllOnPrem|ValidForFlag.AzureV12,
                     SortPriority = SmoTreeNode.NextSortPriority,
+                });
+			}
+			if (WorkspaceService<SqlToolsSettings>.Instance.CurrentSettings.SqlTools.ObjectExplorer.GroupBySchema)
+			{
```
lewis-sanchez, line 795: Formatting looks odd.

lewis-sanchez (approved):

dzsquared:
"Built-in schemas" instead of legacy schemas, matches the docs
I'm ok with them going into a folder or at the end - @erinstellato-ms

Charles-Gagnon:
Wouldn't built-in then also include system schemas like dbo?

Which could be a consideration. Although I think if we're doing this it would be worth seeing about having a different folder icon (folder with schema icon as a badge?) - right now it blends in too well with the other folders IMO.

aasimkhan30:
> Wouldn't built-in then also want to include system schemas like dbo?
> 
> Which could be a consideration. Although I think if we're doing this it would be worth seeing about having a different folder icon (folder with schema icon as a badge?) - right now it blends in too well with the other folders IMO.

@Charles-Gagnon , that's why I did not use 'System schemas' because that would also include dbo, sys, information schema.

aasimkhan30:
Should I use a neutral term like 'Other schemas'?  Putting them towards the end would be a lot of work since our code does not support sorting OE elements as of now.

erinstellato-ms:
We can either put these schemas in their own folder, or hide them altogether.  It's on my list to collect feedback about from users, but official guidance from MS says to not put any objects in these schemas (https://learn.microsoft.com/en-us/sql/relational-databases/security/authentication-access/ownership-and-user-schema-separation?view=sql-server-ver16#built-in-schemas-for-backward-compatibility).  In fact, I've already had a MS PM message me about showing these schemas (and says we should not). 

A reasonable compromise, to start, is to have them in their own folder.  Then we'll see what feedback we get.   I think either Legacy Schemas or Built-In Schemas is fine (I prefer legacy, but I completely get @dzsquared's point about the documentation using Built-It).

The dbo schema should NOT be in there.  It is the default schema where objects are created and must be visible.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodesDefinition.xml
```diff
@@ -64,16 +64,49 @@
   
   <Node Name="Database" LocLabel="string.Empty" Image="Database" BaseClass="ModelBased" NodeType="Database" IsAsyncLoad="" Strategy="CreateModel" TreeNode="ExpandableSchemaTreeNode" PutFoldersAfterNodes="true">
     <ConditionalChildQuerierType Name="SqlSchema" SettingsFlag="GroupBySchema"/>
+    <Filters>
+      <!-- 
+        Excluding built-in schemas for backward compatibility from the database node and moving them to legacy schema folder.
+        The list of schemas was obtained from: 
+        https://learn.microsoft.com/en-us/sql/relational-databases/security/authentication-access/ownership-and-user-schema-separation?view=sql-server-ver16#built-in-schemas-for-backward-compatibility
+      -->
+      <Filter Property="Name" Value="db_accessadmin" Type="string" FilterType="contains" IsNotFilter="true"/>
```
Charles-Gagnon, line 73: You shouldn't be using contains IMO - what if someone had their own schema `db_accessadmin2`? Unlikely, sure. But is there a reason to go with contains over a strict equals?

dzsquared:
> We can either put these schemas in their own folder, or hide them altogether. It's on my list to collect feedback about it from users, but official guidance from MS says to not put any objects in these schemas (https://learn.microsoft.com/en-us/sql/relational-databases/security/authentication-access/ownership-and-user-schema-separation?view=sql-server-ver16#built-in-schemas-for-backward-compatibility). In fact, I've already had a MS PM message me about showing these schemas (and says we should not).
> 
> A reasonable compromise, to start, is to have them in their own folder. Then we'll see what feedback we get. I think either Legacy Schemas or Built-In Schemas is fine (I prefer legacy, but I completely get @dzsquared's point about the documentation using Built-It).
> 
> The dbo schema should NOT be in there. It is the default schema where objects are created and must be visible.

if @erinstellato-ms says Legacy, then we go with Legacy. "Not recommended for your health and wellbeing" doesn't reasonably fit.

erinstellato-ms:
@dzsquared you're funny :)  I'm really good with either legacy or built-in.  Both are appropriate (and yes, something like "do not use these" is too long!)

kburtram (approved):
Looks good.  Agree there is probably a better name than legacy, but the change its self is fine.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodesDefinition.xml
aasimkhan30: Oh yeah, good catch. At first, I was doing name contains "db_" and had thought of a similar scenario.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodesDefinition.xml
```diff
@@ -64,16 +64,41 @@
...
     <ConditionalChildQuerierType Name="SqlSchema" SettingsFlag="GroupBySchema"/>
+    <Filters>
+      <!-- 
+        Excluding built-in schemas for backward compatibility from the database node and moving them to legacy schema folder.
+        The list of schemas was obtained from: 
+        https://learn.microsoft.com/en-us/sql/relational-databases/security/authentication-access/ownership-and-user-schema-separation?view=sql-server-ver16#built-in-schemas-for-backward-compatibility
+      -->
+      <Or>
+        <Filter Property="IsSystemObject" Value="0" Type="bool" />
+        <Filter Property="Name" Value="dbo" Type="string"/>
```
Charles-Gagnon, line 75: For future-proofing I would suggest going back to a hardcoded list of the db_* ones we want to ignore. If a new system schema is added in the future this logic will currently result in it going into the builtin schemas folder which would be annoying to have to notice and remember to fix.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -255,6 +255,46 @@ public async Task GroupBySchemaisEnabled()
...
+                };
+                foreach(var nodes in databaseChildren.Nodes)
+                {
+                    Assert.False(legacySchemas.Contains(nodes.Label), "Legacy schema node should not be found in database node when group by schema is enabled");
+                }
+                var legacySchemasNode = databaseChildren.Nodes.First(t => t.Label == SR.SchemaHierarchy_BuiltInSchema);
+                var legacySchemasChildren = await _service.ExpandNode(session, legacySchemasNode.NodePath);
+                foreach(var nodes in legacySchemasChildren.Nodes)
+                {
+                    Assert.True(legacySchemas.Contains(nodes.Label), "Legacy schema nodes should be found in legacy schemas folder when group by schema is enabled");
```
Charles-Gagnon, line 293: nit: It's better to use NUnit fluent assertions (`Assert.That(legacySchemas, Does.Contain(nodes.Label), ...`)

It's more readable and gives better error messages.

Charles-Gagnon (approved):

PR #1868, opened 2023-02-17
Add login management handlers

caohai:
1. Add handlers for login read/creation/update/delete/dispose
2. Create handler is mostly done (Sql Auth and Win Auth for SQL Server, Sql Auth and AAD Auth for Azure SQL), missing pieces: language setting, server role assignment for Azure
3. Delete handler is done
4. Update handler is still WIP, needs more work and testing after this PR.


src/Microsoft.SqlTools.ServiceLayer/Security/Contracts/LoginInfo.cs
```diff
@@ -12,42 +12,47 @@
 namespace Microsoft.SqlTools.ServiceLayer.Security.Contracts
 {
     [JsonConverter(typeof(StringEnumConverter))]
-    public enum LoginType
+    public enum LoginAuthenticationType
```
kburtram, line 15: I need this in user too, so we may change the name later.  Is fine for now.

src/Microsoft.SqlTools.ServiceLayer/Security/Contracts/LoginRequest.cs
```diff
@@ -39,18 +29,18 @@ public class CreateLoginRequest
         /// Request definition
         /// </summary>
         public static readonly
-            RequestType<CreateLoginParams, CreateLoginResult> Type =
-            RequestType<CreateLoginParams, CreateLoginResult>.Create("security/createlogin");
+            RequestType<CreateLoginParams, object> Type =
```
kburtram, line 32: we should probably use a specific type for the return value, just for readability of code.

src/Microsoft.SqlTools.ServiceLayer/Security/Contracts/LoginRequest.cs
```diff
@@ -62,7 +52,78 @@ public class DeleteLoginRequest
...
+        public static readonly
+            RequestType<UpdateLoginParams, object> Type =
+            RequestType<UpdateLoginParams, object>.Create("objectManagement/updateLogin");
+    }
+
+
+    /// <summary>
+    /// Update Login params
+    /// </summary>
+    public class DisposeLoginViewRequestParams : GeneralRequestDetails
```
kburtram, line 86: probably don't need to derive from `GeneralRequestDetails`.  I think this to get the property bag support.

src/Microsoft.SqlTools.ServiceLayer/Security/LoginData.cs
```diff
@@ -816,6 +914,7 @@ private class LoginPrototypeData : ICloneable
             private Microsoft.SqlServer.Management.Smo.Server server;
             private static string       defaultLanguageDisplay;
             private bool                windowsAuthSupported = true;
+            private bool                aADAuthSupported = false;
```
kburtram, line 917: nit: maybe either `aadAuthSupported` or `isAADAuthSupported` would read a little better.  Private value so no big deal.

src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
```diff
@@ -169,6 +184,152 @@ await requestContext.SendResult(new ResultStatus()
             });
         }
 
+        internal async Task HandleUpdateLoginRequest(UpdateLoginParams parameters, RequestContext<object> requestContext)
+        {
+            ConnectionInfo connInfo;
+            string ownerUri;
+            contextIdToConnectionUriMap.TryGetValue(parameters.ContextId, out ownerUri);
```
kburtram, line 191: the final design will be clone the connect during init and close on dispose.  But this is fine for initial PRs to reuse passed in connection, which is likely from OE or somewhere else.

src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
```diff
@@ -169,6 +184,152 @@ await requestContext.SendResult(new ResultStatus()
...
+            {
+                databases[i] = dataContainer.Server.Databases[i].Name;
+            }
+            string[] languages = new string[dataContainer.Server.Languages.Count];
+            for (int i = 0; i < dataContainer.Server.Languages.Count; i++)
+            {
+                languages[i] = dataContainer.Server.Languages[i].Name;
+            }
+
+            LoginPrototype prototype = parameters.IsNewObject
```
kburtram, line 265: it might make sense to cache the initial prototype between requests for a dialog, since I suspect it takes several DB queries to populate.  We can see if that's worth it later.

kburtram (approved):

kburtram:
@caohai I pushed merge conflict resolutions.  I left some feedback you can consider for follow-up PRs.  Let's merge now so I can pull these changes into my working branch.

caohai:
> @caohai I pushed merge conflict resolutions. I left some feedback you can consider for follow-up PRs. Let's merge now so I can pull these changes into my working branch.

Sure yeah I was looking at the comments and was thinking about adding a few TODO tags for now, I'll just fix those in next PRs.

src/Microsoft.SqlTools.ServiceLayer/Security/LoginData.cs
caohai: Yeah I hate it myself but didn't come up with a good alternative, will update later.

PR #1869, opened 2023-02-17
Build with .NET 7.0.200 + Async improvements

cheenamalhotra:
Placing next tested benchmark to build with .NET 7.0.200

+made some improvements on async waits.

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/HostLoader.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/HostLoader.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Kusto.ServiceLayer
                     // Start the service only after all request handlers are setup. This is vital
                     // as otherwise the Initialize event can be lost - it's processed and discarded before the handler
                     // is hooked up to receive the message
-                    serviceHost.Start().Wait();
+                    serviceHost.Start().GetAwaiter().GetResult();
                     isLoaded = true;
                 }
             }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "7.0.101",
+        "version": "7.0.200",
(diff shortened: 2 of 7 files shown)


kburtram (approved):

src/Microsoft.Kusto.ServiceLayer/HostLoader.cs
```diff
@@ -52,7 +52,7 @@ internal static ServiceHost CreateAndStartServiceHost(SqlToolsContext sqlToolsCo
                     // Start the service only after all request handlers are setup. This is vital
                     // as otherwise the Initialize event can be lost - it's processed and discarded before the handler
                     // is hooked up to receive the message
-                    serviceHost.Start().Wait();
+                    serviceHost.Start().GetAwaiter().GetResult();
```
Charles-Gagnon, line 55: Was this something flagged by the SDK update? 

And what exactly is this change doing? Looking around it seems like it just makes the stack trace a bit cleaner - is that it or is there some other benefit doing it this way brings?

Charles-Gagnon (approved):

alanrenmsft (approved):

src/Microsoft.Kusto.ServiceLayer/HostLoader.cs
cheenamalhotra: It's not flagged, but as a practice this should be followed, the main benefit of `GetAwaiter().GetResult()` is is preseves exception stacktrace and doesn't throw `AggregateException` as `Wait()` or `.Result` do.

Charles-Gagnon: Thanks for the context. Is there some linting/style rule we can enable to have this show up as a warning? Definitely sounds useful but I don't think most people will know about or remember to do this unless we have the build remind us.

PR #1875, opened 2023-02-21
Fix error not getting sent back when generating deploy plan

kisantia:
Fixes https://github.com/microsoft/azuredatastudio/issues/13118. Sometimes the loading spinner would spin forever during the "Generate deploy plan" step in the dacpac extension when deploying to an existing database without any indication if there was an error. This was caused by the error message from STS getting swallowed and not sent to ADS. 

This updates the generate deploy plan request to use the new `RunWithErrorHandling` @Benjin recently added and now the error message gets sent back to ADS and displayed to the user: 
image

Next is to try to make this error message more readable, but at least now the spinner doesn't keep spinning forever when there's actually been an error.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -187,16 +187,19 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
                     out connInfo);
             if (connInfo != null)
             {
-                GenerateDeployPlanOperation operation = new GenerateDeployPlanOperation(parameters, connInfo);
-                operation.Execute(parameters.TaskExecutionMode);
-
-                await requestContext.SendResult(new GenerateDeployPlanRequestResult()
+                await BaseService.RunWithErrorHandling(async () =>
                 {
-                    OperationId = operation.OperationId,
-                    Success = true,
-                    ErrorMessage = string.Empty,
-                    Report = operation.DeployReport
-                });
+                    GenerateDeployPlanOperation operation = new GenerateDeployPlanOperation(parameters, connInfo);
+                    operation.Execute(parameters.TaskExecutionMode);
+
+                    return new GenerateDeployPlanRequestResult()
+                    {
+                        OperationId = operation.OperationId,
+                        Success = true,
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -187,16 +187,19 @@ public async Task HandleGenerateDeployPlanRequest(GenerateDeployPlanParams param
                     out connInfo);
             if (connInfo != null)
             {
-                GenerateDeployPlanOperation operation = new GenerateDeployPlanOperation(parameters, connInfo);
-                operation.Execute(parameters.TaskExecutionMode);
-
-                await requestContext.SendResult(new GenerateDeployPlanRequestResult()
+                await BaseService.RunWithErrorHandling(async () =>
```
Benjin, line 190: I made those methods `public static` precisely for situations like this, but ideally, `DacFxService` would just extend `BaseService`.  My "eventual" hope is to move more of the service boilerplate into that class in order to clean stuff up.

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
Charles-Gagnon: Why is that even a base class? There's nothing it does that requires any actual instance data so you might as well just stick it in a static helper utility class and let people statically import it IMO.

Charles-Gagnon (approved):
For what it's worth - any exceptions thrown will already cause an error event to be sent back and so a pattern I prefer is to not bother trying to catch everything here and instead on the extension side catch those errors and handle them. It's pretty much the same amount of code (since you don't have to check success) and has the benefit of not worrying about whether something like this will "slip" through.

Alternatively if you wanted to keep this pattern I would suggest looking into a specialized "SetRequestHandler" only for requests whose responses extend `ResultStatus` (or possibly override the existing one) that do this whole try/catch logic. That way you could easily add it to any handlers you want to behave this way and save a lot of duplicated code.

kisantia:
Yeah it would definitely be good to go back and cleanup this code since I didn't really understand how things worked when I first added most of this several years ago :)

PR #1878, opened 2023-02-23
Save file structure in scmp file

SakshiS-harma:
This PR has STS changes to fix https://github.com/microsoft/azuredatastudio/issues/20332.


test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1200,6 +1200,41 @@ public async Task SchemaCompareOpenScmpProjectToProjectRequest()
...
         }
 
+        /// <summary>
+        /// Verify folder structure gets stored in scmp file for project endpoint
+        /// </summary>
+        [Test]
+        public async Task VerifyExtractTargetInScmpFile()
+        {
+            SqlTestDb sourceDb = await SqlTestDb.CreateNewAsync(TestServerType.OnPrem, false, null, SourceScript, "SchemaCompareOpenScmpSource");
+            SqlTestDb targetDb = await SqlTestDb.CreateNewAsync(TestServerType.OnPrem, false, null, TargetScript, "SchemaCompareOpenScmpTarget");
```
kisantia, line 1210: why does a target db need to be created if the target in the schema compare is a project?

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareRequest.cs
```diff
@@ -70,6 +71,11 @@ public class SchemaCompareEndpointInfo
         /// Connection details
         /// </summary>
         public ConnectionDetails ConnectionDetails { get; set; }
+
+        /// <summary>
+        /// Extract target of the project used when extracting a database to file system or updating the project from database
+        /// </summary>
+        public DacExtractTarget ExtractTarget { get; set; }
```
Benjin, line 78: Let's get in the habit of being explicit about the nullability of properties :)  This can be `DacExtractTarget? ExtractTarget` since it may be null if it's a source or the target isn't a project, right?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1594,6 +1629,23 @@ public async Task VerifySchemaCompareWarningsBeingExcluded()
             }
         }
 
+        private async Task VerifyContentAndCleanupAsync(string outputFilePath, string patternToMatch)
```
Benjin, line 1632: Consider renaming `patternToMatch` to `textToMatch` since it's not doing a regex pattern match

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
SakshiS-harma: For now, I was just following the pattern used in the test file to create project. Indeed a db is not required to have a project but this is what other tests are doing currently. May be a separate work item should be created to fix it at all the places?

kisantia (approved):

Benjin (approved):

PR #1881, opened 2023-02-27
Add more specific URI strings to Object Explorer

smartguest:
This adds more specific URI to Object Explorer so that sessions can be differentiated between connections that share the same basic connection details but have different properties.

Part of a URI overhaul


kburtram (commented):
@smartguest please don't merge this PR until after we fork for the March release end-of-week.  Thanks!

cheenamalhotra:
@smartguest 

Can you also validate impact on vscode-mssql extension with this PR and update the same to support connection URI similarly?

smartguest:
@cheenamalhotra Will do, although structurally, vscode-mssql is a bit different from ADS and it does not appear at first glance that it uses single string connection IDs (or option keys) on its side, so it will take some time to implement changes to support the new URI

cheenamalhotra:
For now I'd say make sure nothing breaks, so this change should not negatively affect vscode-mssql connections.. (I haven't tested yet) and you can add a new issue/feature request in vscode-mssql to support multiple connections against same target if needed.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -56,7 +56,7 @@ public void InitObjectExplorerServiceTests()
             ConnectedBindingContext connectedBindingContext = new ConnectedBindingContext();
             connectedBindingContext.ServerConnection = new ServerConnection(new SqlConnection(fakeConnectionString));
             connectedBindingQueue = new ConnectedBindingQueue(false);
-            connectedBindingQueue.BindingContextMap.Add($"{details.ServerName}_{details.DatabaseName}_{details.UserName}_NULL", connectedBindingContext);
+            connectedBindingQueue.BindingContextMap.Add($"{details.ServerName}_{details.DatabaseName}_{details.UserName}_NULL_persistSecurityInfo:true", connectedBindingContext);
```
cheenamalhotra, line 59: Any special reason for this change?

smartguest: The PersistSecurityInfo is an option that is added as a result of taking into every property besides password and connection name (as it's not supported by SQL connections) to the URI.

It is considered a non default option when it is true on the ADS side, and since we want to treat non special options consistently, it will have to be added to the string here.

smartguest: The PersistSecurityInfo is an option that is added as a result of taking into every property besides password and connection name (as it's not supported by SQL connections) to the URI.

It is considered a non default option when it is true on the ADS side, and since we want to treat non special options consistently, it will have to be added to the string here.

cheenamalhotra (approved):

PR #1885, opened 2023-03-01
Remove Value in AddSqlCmdVariableParams

kisantia:
This removes `Value` as a parameter for Add and Update SQLCMD variable.  Value is only used for VS, so it doesn't need to be exposed to ADS. It was made optional in DacFx in https://msdata.visualstudio.com/SQLToolsAndLibraries/_git/DacFx/commit/4891e3dc471ceea6b0a8654b0862b431b3ffd91b?refName=refs%2Fheads%2Fmain, so DacFx will calculate the default value to put here if none is provided. 

It's used for mapping the value from a .user file in VS. In the .sqlproj, Value is specified like this for a SqlCmdVariable:
```
         <ItemGroup>
                 <SqlCmdVariable Include="myVar1">
                     <DefaultValue>OtherDatabaseDefaultValue1</DefaultValue>
                     <Value>$(SqlCmdVar__1)</Value>;
                 </SqlCmdVariable>
          </ItemGroup> 
```
          
Looks like this in the .user file:
```
           <PropertyGroup>
                 <SqlCmdVar__1>MyUserValue</SqlCmdVar__1>
           </PropertyGroup>
```

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/SqlCmdVariables/AddSqlCmdVariable.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/SqlCmdVariables/AddSqlCmdVariable.cs
@@ -24,11 +24,6 @@ namespace Microsoft.SqlTools.ServiceLayer.SqlProjects.Contracts
         /// Default value of the SQLCMD variable
         /// </summary>
         public string DefaultValue { get; set; }
-
-        /// <summary>
-        /// Value of the SQLCMD variable, with or without the $()
-        /// </summary>
-        public string Value { get; set; }
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Benjin (approved):

Benjin:
Also remove from vscode-mssql!

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/Contracts/SqlCmdVariables/AddSqlCmdVariable.cs
```diff
@@ -24,11 +24,6 @@ public class AddSqlCmdVariableParams : SqlProjectParams
         /// Default value of the SQLCMD variable
         /// </summary>
         public string DefaultValue { get; set; }
-
-        /// <summary>
-        /// Value of the SQLCMD variable, with or without the $()
-        /// </summary>
-        public string Value { get; set; }
```
Benjin, line 31: Be sure to also remove this from the vscode-mssql extension!

kisantia: yup will do! Thanks for the reminder!

PR #1897, opened 2023-03-03
[User Management] Fix a few bugs in login handlers

caohai:
1. Support server role add/drop for Azure SQL DB
2. Fix language setting not working during login creation

image
image


kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/Security/LoginData.cs
```diff
@@ -864,6 +864,10 @@ private void PopulateServerRolesForAzure()
     internal class LoginPrototype
     {
         private SqlCollationSensitiveStringComparer comparer = null;
+        private static string AZURE_SERVER_ROLE_ALTER_QUERY =
```
alanrenmsft, line 867: no SMO support for this?

caohai: no SMO support for now, I'll need to make a SMO fix to support server role related operations for Azure SQL DB

PR #1916, opened 2023-03-08
Get Application Path from command options

cheenamalhotra:
Addresses https://github.com/microsoft/azuredatastudio/issues/22223 by receiving App Path from command options.

### DIFF
--- a/src/Microsoft.SqlTools.Authentication/Sql/AuthenticationProvider.cs
+++ b/src/Microsoft.SqlTools.Authentication/Sql/AuthenticationProvider.cs
@@ -5,6 +5,7 @@
 
 using Microsoft.Data.SqlClient;
 using Microsoft.SqlTools.Authentication.Utility;
+using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.Authentication.Sql
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.Authentication/Sql/AuthenticationProvider.cs
```diff
@@ -27,13 +28,23 @@ public class AuthenticationProvider : SqlAuthenticationProvider
...
+            if (string.IsNullOrEmpty(applicationName))
+            {
                 applicationName = nameof(SqlTools);
+                Logger.Warning($"Application Name not received with command options, using default application name as: {applicationName}");
             }
-            var cachePath = Path.Combine(Utils.BuildAppDirectoryPath(), applicationName, AzureTokenFolder);
+
+            if (string.IsNullOrEmpty(applicationPath))
+            {
+                applicationPath = Utils.BuildAppDirectoryPath();
```
Charles-Gagnon, line 43: Isn't the issue here that the homedir value you're using doesn't seem to be valid for Linux/MacOS?

`var homedir = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);`

https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix

If you change it to use SpecialFolder.Personal it seems like that should work as well (and match the logic that generates the path being passed from ADS with your current change)

It just seems unnecessary to pass the path like this since the logic on both sides should be the same, and this will still be broken for Linux/MacOS if this parameter isn't passed from ADS it seems like.

cheenamalhotra: In future I intend to use this: https://github.com/microsoft/azuredatastudio/blob/9fddefc8fb11fbbd3bb357e11c0e2809d3868aa5/src/vs/platform/environment/node/userDataPath.js#L49

That's why I decided to pass in via ADS.. so we need not worry about managing this path in backend service layer.

Charles-Gagnon: Yup, good call.

Charles-Gagnon: Coud you still fix the logic here to use Personal folder? It'll just never work if for some reason the path isn't passed in currently so might as well try to avoid that being a potential issue anywhere.

cheenamalhotra: Done, thanks!

Charles-Gagnon (approved):

src/Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs
```diff
@@ -46,6 +46,9 @@ public CommandOptions(string[] args, string serviceName)
                             case "-application-name":
                                 ApplicationName = args[++i];
                                 break;
+                            case "-application-path":
```
Charles-Gagnon, line 49: nit: I'd suggest naming this something like `data-path` to match what VS Code uses

kburtram (approved):

PR #1918, opened 2023-03-08
[User Managment] Handle invalid connection uri

caohai:
https://github.com/microsoft/azuredatastudio/issues/22236


kburtram (approved):

alanrenmsft:
@caohai any idea what is the root cause of invalid uri being passed in?

caohai:
> @caohai any idea what is the root cause of invalid uri being passed in?

My guess is that the connInfo gets disposed somehow by some other services, haven't been able to repro it but I'll try this with a broken connection later

PR #1919, opened 2023-03-08
Expose inner exceptions when sendError

caohai:
https://github.com/microsoft/azuredatastudio/issues/22231

This PR merges inner exceptions into one message then send the error back.
image

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
@@ -176,7 +176,7 @@ namespace Microsoft.SqlTools.Hosting.Protocol
                 e = e.InnerException;
             }
 
-            return res;
+            return res.TrimEnd( Environment.NewLine.ToCharArray());;
         }
 
          public void SetEventHandler<TParams>(
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
@@ -159,10 +159,24 @@ namespace Microsoft.SqlTools.Hosting.Protocol
                     }
                     catch (Exception ex)
                     {
-                        Logger.Error($"{requestType.MethodName} : {ex}");
-                        await requestContext.SendError(ex.Message);
+                        string exceptionMessage = FormatException(ex);
+                        Logger.Error($"{requestType.MethodName} : {exceptionMessage}");
+                        await requestContext.SendError(exceptionMessage);
                     }
                 });
+        }
+
+        private string FormatException(Exception e)
+        {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -159,12 +159,26 @@ public void Stop()
...
                 });
         }
 
+        private string FormatException(Exception e)
+        {
+            string res = string.Empty;
+
+            while (e != null)
+            {
+                res += e.Message + Environment.NewLine;
```
kburtram, line 175: it looks like we're adding an extra newline.  Will this ever matter when we display?  e.g. unexpected linebreaks in logs or UI messages?

kburtram (approved):
Looks fine but maybe better to rewrite the look, to avoid the extra line break.  For example by checking InnerException isn't null before appending.  That would keep output the same in general case with no inner exception.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
caohai: updated, and for exception without inner ex, the message should be the same as the current implementation
image

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -159,12 +159,26 @@ public void Stop()
...
-                        Logger.Error($"{requestType.MethodName} : {ex}");
-                        await requestContext.SendError(ex.Message);
+                        string exceptionMessage = FormatException(ex);
+                        Logger.Error($"{requestType.MethodName} : {exceptionMessage}");
+                        await requestContext.SendError(exceptionMessage);
                     }
                 });
         }
 
+        private string FormatException(Exception e)
```
alanrenmsft, line 169: Rename to GetErrorMessage(Exception e)

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -159,12 +159,26 @@ public void Stop()
...
+        {
+            string res = string.Empty;
```
alanrenmsft, line 171: you can add the messages to an array and then do a join so that you don't need to do the TrimEnd.

also, we probably should also add a space character between the messages since new line is not going to be displayed in the error dialog.

alanrenmsft: String.Join()

caohai: updated
image

alanrenmsft (approved):

PR #1920, opened 2023-03-08
Port invalid uri handling

caohai:
Port #1918 to release

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
@@ -121,7 +121,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Security
 
             if (connInfo == null) 
             {
-                    // raise error here
+                throw new ArgumentException("Invalid ConnectionUri");
             }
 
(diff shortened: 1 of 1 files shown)


kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
```diff
@@ -121,7 +121,7 @@ internal async Task HandleCreateLoginRequest(CreateLoginParams parameters, Reque
 
             if (connInfo == null) 
             {
-                    // raise error here
+                throw new ArgumentException("Invalid ConnectionUri");
```
alanrenmsft, line 124: in main branch, let's localize the string since it is user facing with some tips about how to recover from this, for example, please reconnect or something like that.

caohai: I'll create an issue for this in May

caohai: https://github.com/microsoft/azuredatastudio/issues/22244

PR #1923, opened 2023-03-08
Fix lock out for login update

caohai:
https://github.com/microsoft/azuredatastudio/issues/22235

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
@@ -317,7 +317,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Security
                 SupportWindowsAuthentication = prototype.WindowsAuthSupported,
                 SupportAADAuthentication = prototype.AADAuthSupported,
                 SupportSQLAuthentication = true, // SQL Auth support for login, not necessarily mean SQL Auth support for CONNECT etc.
-                CanEditLockedOutState = true,
+                CanEditLockedOutState = parameters.IsNewObject,
                 Databases = databases,
                 Languages = languages,
                 ServerRoles = prototype.ServerRoles.ServerRoleNames,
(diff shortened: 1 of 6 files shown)


src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
```diff
@@ -317,7 +317,7 @@ await requestContext.SendResult(new LoginViewInfo()
                 SupportWindowsAuthentication = prototype.WindowsAuthSupported,
                 SupportAADAuthentication = prototype.AADAuthSupported,
                 SupportSQLAuthentication = true, // SQL Auth support for login, not necessarily mean SQL Auth support for CONNECT etc.
-                CanEditLockedOutState = true,
+                CanEditLockedOutState = parameters.IsNewObject,
```
alanrenmsft, line 320: I don't think this is how SSMS determines the state. for new login, it is definitely false (I am not even showing it), for existing login, it is only enabled when the login is locked out and current user has permission to unlock the login.

alanrenmsft: for March release, I think I'd simply hide this property to avoid confusion and you can take the time to fix it in May release. sounds good?

src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
```diff
@@ -317,7 +317,7 @@ await requestContext.SendResult(new LoginViewInfo()
                 SupportWindowsAuthentication = prototype.WindowsAuthSupported,
                 SupportAADAuthentication = prototype.AADAuthSupported,
                 SupportSQLAuthentication = true, // SQL Auth support for login, not necessarily mean SQL Auth support for CONNECT etc.
-                CanEditLockedOutState = true,
+                CanEditLockedOutState = !parameters.IsNewObject && prototype.IsLockedOut,
```
alanrenmsft, line 320: looks better, but is this how SSMS determines the state?

alanrenmsft: also, have you tested the unlocking?

caohai: I've tested unlocking with latest iteration and it's working fine.

src/Microsoft.SqlTools.ServiceLayer/Security/SecurityService.cs
caohai: We can probably test again with this PR merged and if it's still not correct we can hide this.

kburtram (approved):
Looks good, we can follow-up to make closer to SSMS post-March if necessary.

kburtram:
I'll merge and take to release branch so this can be in tonight's update.  We can take any other updates in next PR if needed.

PR #1924, opened 2023-03-08
Fix updating users in SQL DB

kburtram:
Fixes updating user accounts on SQL DB.  Previous check for DefaultLanguage support didn't check for Azure.  This is tracked by https://github.com/microsoft/azuredatastudio/issues/22239.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
@@ -12,6 +12,7 @@ using Microsoft.SqlServer.Management.Sdk.Sfc;
 using Microsoft.SqlTools.ServiceLayer.Management;
 using Microsoft.SqlTools.ServiceLayer.Security.Contracts;
 using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlServer.Management.Common;
 
 namespace Microsoft.SqlTools.ServiceLayer.Security
 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
```diff
@@ -237,8 +238,13 @@ private void LoadUserData(CDataContainer context)
...
             {
                 this.authenticationType = existingUser.AuthenticationType;
-                this.defaultLanguageAlias = LanguageUtils.GetLanguageAliasFromName(existingUser.Parent.Parent,
-                                                                existingUser.DefaultLanguage.Name);
+
+                if (context.Server.ServerType != DatabaseEngineType.SqlAzureDatabase)
+                {
+                    this.defaultLanguageAlias = LanguageUtils.GetLanguageAliasFromName(
+                                                                    existingUser.Parent.Parent,
+                                                                    existingUser.DefaultLanguage.Name);
```
kburtram, line 246: `existingUser.DefaultLanguage` is throwing an exception when connected to SQL DB.

src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
```diff
@@ -793,7 +799,8 @@ public bool IsDefaultLanguageSupported
             get
             {
                 //Default Language was not supported before Denali.
-                return SqlMgmtUtils.IsSql11OrLater(this.context.Server.ConnectionContext.ServerVersion);
+                return SqlMgmtUtils.IsSql11OrLater(this.context.Server.ConnectionContext.ServerVersion)
+                    && this.context.Server.ServerType != DatabaseEngineType.SqlAzureDatabase;
```
caohai, line 803: Does `DatabaseEngineType.SqlAzureDatabase` contain MI? default language should be supported with ALTER USER in MI 

https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-user-transact-sql?view=azuresqldb-mi-current&preserve-view=true

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
kburtram: Let's take the MI case post-March since I want to get these changes in today.  I'll test and open a bug to track that if necessary.

caohai (approved):

PR #1925, opened 2023-03-08
Fix sorting for Logins in user initialization request

kburtram:
This moves some code around for better organization, but also sorts the logins as requested in this bug https://github.com/microsoft/azuredatastudio/issues/22246.


src/Microsoft.SqlTools.ServiceLayer/Utility/DatabaseUtils.cs
```diff
@@ -245,5 +246,36 @@ private static void SpecialPostDropActionsForObject(CDataContainer dataContainer
...
+                    string name;
+                    while (reader.Read())
+                    {
+                        // Get the permission name
+                        name = reader.GetString(0);
+                        items.Add(name);
+                    }
+                }
+            }
+            items.Sort();
```
kburtram, line 277: this line is the only functional change.  Rest of PR is moving these methods to a better location.

alanrenmsft (approved):

caohai (approved):

PR #1931, opened 2023-03-09
Only support User with Login users

kburtram:
This is to avoid scenario in bug https://github.com/microsoft/azuredatastudio/issues/22280 for March early preview.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
@@ -1044,17 +1044,17 @@ namespace Microsoft.SqlTools.ServiceLayer.Security
         {
             switch (userType)
             {
-                case ExhaustiveUserTypes.AsymmetricKeyMappedUser:
-                    currentData.userType = UserType.AsymmetricKey;
-                    this.asymmetricKeyMappedUser ??= new UserPrototype(this.context, this.currentData, this.originalData);
-                    this.currentPrototype = asymmetricKeyMappedUser;
-                    break;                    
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
```diff
@@ -1044,46 +1044,47 @@ public UserPrototype GetUserPrototype(ExhaustiveUserTypes userType)
...
+                //     currentData.userType = UserType.SqlUser;
+                //     this.windowsUser ??= new UserPrototypeForWindowsUser(this.context, this.currentData, this.originalData);
+                //     this.currentPrototype = windowsUser;
+                //     break;
                 
                 default:
                     System.Diagnostics.Debug.Assert(false, "Unknown UserType provided.");
                     this.currentPrototype = null;
-                    break;
+                    throw new ApplicationException("Only 'User with Login' user type support in March Preview release");
```
alanrenmsft, line 1086: probably we don't have to call out the release?

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/Security/UserData.cs
kburtram: I'm assuming this won't be the case in the next release :)

kburtram:
The previous implementation wasn't working since those checks were too late (after the type was already hardcoded).  This is an alternate implementation that failure earlier.  Below is the UX.

image

alanrenmsft:
LGTM, if you sync to latest main, the dialog should be automatically closed.

Issue #1938, opened 2023-03-12
Microsoft.SqlTools.ManagedBatchParser does not know anything about SQL Agent escape macros


Matteo-T:
When it comes to T-SQL fragments that contain [SQL Agent escape macros](https://learn.microsoft.com/en-us/sql/ssms/agent/use-tokens-in-job-steps?view=sql-server-ver16#sql-server-agent-escape-macros), the asymmetry between the native BatchParser (used by tools like sqlcmd.exe) and Microsoft.SqlTools.ManagedBatchParser becomes evident.

It would be really nice (the Invoke-Sqlcmd could take advantage of it, for example - and fill another gap it now has when it was moved from the native BatchParser over to the Microsoft.SqlTools.ManagedBatchParser) if we could support them.

# Example 1:

```console
SQLCMD.EXE -S (localdb)\MSSQLLocalDB -q "SELECT N'$(ESCAPE_DQUOTE(SRVR))'" -v SRVR="aa""bb"
```
parses just fine and yields the following:
```console

--------
aa""""bb

(1 rows affected)
```

# Example 2:
```console
SQLCMD.EXE -S (localdb)\MSSQLLocalDB -q "SELECT N'$(ESCAPE_DQUOTE(SRVR))'"
```
parses just fine and yields the following:
```console

'SRVR' scripting variable not defined.
-------
$(SRVR)
```

# Example 3:
```console
SQLCMD.EXE -S (localdb)\MSSQLLocalDB -q "SELECT N'$(ESCAPE_DQUOTE(SRVR))'" -x
```
parses just fine and yields the following:
```console
----------------------
$(ESCAPE_DQUOTE(SRVR))

(1 rows affected)
```

While Example 3 is covered when we add an option to disable the variable substitution/parsing, examples 1 and 2 are just gaps.

From what I see, we would just need to augment the Lexer/Parser to basically consider the stuff after the initial "$(". If we get another "(" while collecting the characters that make up the identifier, it just means that this could be one of the 4 SQL Agent macros. If it is, then we stash it away and continue reading the chars until the "))" and collect the actual variable. The variable would need to be decorated with a possible Agent Macro, so after we call the IVariableResolver.GetVariable() we have another chance to do the escaping.

Issue #1937, opened 2023-03-12
Microsoft.SqlTools.ManagedBatchParser throws NRE when IVariableResolver is not implemented


Matteo-T:
Compile and run the following C# app:

```csharp
using Microsoft.SqlTools.ServiceLayer.BatchParser;
using System;
using System.IO;

class C : ICommandHandler
{
    public bool ResolveVariables = !false;

    public static void Main()
    {
        var c = new C();

        using var parser =
            new Parser(commandHandler: c, variableResolver: null, reader: new StringReader("SELECT N'$(V)'"), name: null)
            {
                ThrowOnUnresolvedVariable = false
            };

        parser.Parse();
    }

    public BatchParserAction ProcessBatch(string str, int num)
    {
        for (var i = 0; i < num; i++) Console.WriteLine($"Processing: {str}");
        return BatchParserAction.Continue;
    }

    public BatchParserAction Go(TextBlock batch, int repeatCount, SqlCmdCommand tokenType)
    {
        batch.GetText(resolveVariables: ResolveVariables, out var text, out _);
        return ProcessBatch(str: text, num: repeatCount);
    }
    public BatchParserAction Include(TextBlock _, out TextReader __, out string ___) => throw new NotImplementedException();

    public BatchParserAction OnError(Token _, OnErrorAction __) => throw new NotImplementedException();
}
```

**Result:**
```console
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.SqlTools.ServiceLayer.BatchParser.Parser.ResolveVariables(Token inputToken, Int32 offset, List`1 variableRefs) in /_/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs:line ...
 ...

**Expected:**
```console
Processing: SELECT N'$(V)'
```

While one could argue that the call to 'batch.GetText()' in `public BatchParserAction Go()` should just pass `resolveVariables = false`, it seems more robust not to throw an NRE.

The fix is trivial, just a couple of elvis-operators:

In `ResolveVariables()`:
string value = variableResolver?.GetVariable(variablePos.Value, reference.VariableName);

In `ParseSetvar()`:
variableResolver?.SetVariable(setvarToken.Begin, variableName, variableValue);

Issue #1936, opened 2023-01-27
Microsoft.SqlTools.ManagedBatchParser does not offer an option to disable variable substitution/parsing [was: Invoke-Sqlcmd does not parse correctly some string literal containing strings that resemble variables]


Matteo-T:
Simple repro (compile for net7.0 after adding a reference to Microsoft.SqlTools.ManagedBatchParser.dll, then run):

```csharp
using Microsoft.SqlTools.ServiceLayer.BatchParser;
using System;
using System.IO;

class C : ICommandHandler
{
    public bool ResolveVariables = false;

    public static void Main()
    {
        var c = new C();

        using var parser =
            new Parser(commandHandler: c, variableResolver: null, reader: new StringReader("SELECT N'$(ESCAPE_SQUOTE(SRVR))'"), name: null)
            {
                ThrowOnUnresolvedVariable = true
            };

        parser.Parse();
    }

    public BatchParserAction ProcessBatch(string str, int num)
    {
        for (var i = 0; i < num; i++) Console.WriteLine($"Processing: {str}");
        return BatchParserAction.Continue;
    }

    public BatchParserAction Go(TextBlock batch, int repeatCount, SqlCmdCommand tokenType)
    {
        batch.GetText(resolveVariables: ResolveVariables, out var text, out _);
        return ProcessBatch(str: text, num: repeatCount);
    }
    public BatchParserAction Include(TextBlock _, out TextReader __, out string ___) => throw new NotImplementedException();

    public BatchParserAction OnError(Token _, OnErrorAction __) => throw new NotImplementedException();
}
```

**Result:**
```console
Unhandled exception. Microsoft.SqlTools.ServiceLayer.BatchParser.BatchParserException: Incorrect syntax was encountered while parsing '$(ESCAPE_SQUOTE('.
 [... callstack ...]
```

**Expected:**

```console
Processing: SELECT N'$(ESCAPE_SQUOTE(SRVR))'
```

------------------------

This repro originated from the https://github.com/microsoft/SQLServerPSModule repo, but in reality, it's a bug in the ManagedBatchParser. I've 

Original title was `Invoke-Sqlcmd does not parse correctly some string literal containing strings that resemble variables`

Run the following cmdlet:

```powershell
invoke-sqlcmd -ServerInstance SQLTOOLS2022-3  -DisableVariables -Query "SELECT N'  ''`$(ESCAPE_SQUOTE(SRVR))'' '"
```

Result:
```console
[...]

PR #1941, opened 2023-03-13
Exposing a DisableVariableSubstitution on the Parser object

Matteo-T:
Addressing issues #1936 and #1937.

With this change, I'm adding a new property on the Parser object `DisableVariableSubstitution` to allow the parsing to completely ignore elements in the T-SQL fragment that may look like Variables, thus allowing consumers of the Parser (e.g. SQL Server Powershell Invoke-Sqlcmd cmdlet) to micking the behavior of `sqlcmd.exe -x` which completely ignores such variables during parsing.

This property has a default value of `false` (for backward compatibility).

This PR also introduced the fix for #1937 is just to allow the Parser to be a little more forgiving and avoid throwing a NRE when the caller is not passing in any IVariableResolver, which most likely means "I do not need/care for any variable substitution, so don't bother".

While combining the 2 fixes, it made sense to change the default value for `DisableVariableSubstitution` to be `true` when the `IVariableResolver` is `null` (this is safe, since no existing code out there allowed it  before... so no existing consumer should be surprised by this change).

> This PR does not mean to address #1938, although it is somehow related to it.

### DIFF
--- a/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
+++ b/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
@@ -34,8 +34,19 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
             this.variableResolver = variableResolver;
             lexer = new Lexer(reader, name);
             tokenBuffer = new List<Token>();
+            DisableVariableSubstitution = variableResolver == null;
         }
 
+        /// <summary>
+        /// Whether to enable or disable the variable substitution.
+        /// A null IVariableResolver implies a true value.
+        /// </summary>
+        public bool DisableVariableSubstitution { get; set; }
+
+        /// <summary>
+        /// Whether to throw an error when a variable to be substituted was not specified.
+        /// Ignored when DisableVariableSubstitution is true.
+        /// </summary>
         public bool ThrowOnUnresolvedVariable { get; set; }
 
         private Token LookaheadToken
@@ -110,9 +121,9 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
 
         private void AddVariableReferences(Token token, int offset, IList<VariableReference> variableRefs)
         {
-            if (lexer.RecognizeSqlCmdSyntax == false)
+            if (!lexer.RecognizeSqlCmdSyntax || DisableVariableSubstitution)
             {
-                // variables are recognized only in sqlcmd mode.
+                // variables are recognized only in sqlcmd mode and the substitution was not explicitly disabled
                 return;
             }
 
@@ -649,11 +660,15 @@ namespace Microsoft.SqlTools.ServiceLayer.BatchParser
                     column: column,
                     offset: reference.Start - offset,
                     filename: inputToken.Filename);
-                string value = variableResolver.GetVariable(variablePos.Value, reference.VariableName);
+                // If the variableResolver is not defined (=null), then we are a little
+                // more robust and avoid throwing a NRE. It may just mean that the caller
+                // did not bother implementing the IVariableResolver interface, presumably
+                // because he/she was not interested in any variable replacement.
+                string value = variableResolver?.GetVariable(variablePos.Value, reference.VariableName);
                 if (value == null)
                 {
                     // Undefined variable
-                    if (ThrowOnUnresolvedVariable == true)
+                    if (ThrowOnUnresolvedVariable)
                     {
                         RaiseError(ErrorCode.VariableNotDefined, inputToken,
                             string.Format(CultureInfo.CurrentCulture, SR.BatchParser_VariableNotDefined, reference.VariableName));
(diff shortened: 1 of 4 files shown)


Matteo-T:
@cheenamalhotra - could you advise on a couple of things? Like, is there a way to link the PR to the issues? Also, are there other things that need to happen before this PR can go in? Like updating labels, milestons, projects, etc? I suppose that whoever is triaging the issues is going to do that? I need to be educated... Thanks!

Charles-Gagnon:
For linking items to a PR the easiest way is to use one of the closing keywords : https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Note that you have to put it in front of each issue to close - e.g.

```
Fixes https://github.com/microsoft/sqltoolsservice/issues/1936
Fixes https://github.com/microsoft/sqltoolsservice/issues/1937
```

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
```diff
@@ -649,11 +660,15 @@ internal string ResolveVariables(Token inputToken, int offset, List<VariableRefe
                     column: column,
                     offset: reference.Start - offset,
                     filename: inputToken.Filename);
-                string value = variableResolver.GetVariable(variablePos.Value, reference.VariableName);
+                // If the variableResolver is not defined (=null), then we are a little
+                // more robust and avoid throwing a NRE. It may just mean that the caller
+                // did not bother implementing the IVariableResolver interface, presumably
+                // because he/she was not interested in any variable replacement.
```
Charles-Gagnon, line 666: super nit:
```suggestion
                // because they were not interested in any variable replacement.
```

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
```diff
@@ -34,8 +34,19 @@ public Parser(ICommandHandler commandHandler, IVariableResolver variableResolver
             this.variableResolver = variableResolver;
             lexer = new Lexer(reader, name);
             tokenBuffer = new List<Token>();
+            DisableVariableSubstitution = variableResolver == null;
         }
 
+        /// <summary>
+        /// Whether to enable or disable the variable substitution.
+        /// A null IVariableResolver implies a true value.
```
Charles-Gagnon, line 42: This isn't directly true though:

1. If `DisableVariableSubstitution` is false, and we don't have a variable resolver (or we have a variable resolver but it resolves some of the variables to null) and `ThrowOnUnresolvedVariable` is true then it will throw
2. But if `DisableVariableSubstitution` is true then it doesn't matter what `ThrowOnUnresolvedVariable` is - it'll parse just fine

Matteo-T: @Charles-Gagnon - what are you suggesting me to do here, besides adding a couple of tests to cover your 2 points there, which is usually a very effective way to answer questions :).

I was a bit thorn around the "should I do try to couple `ThrowOnUnresolvedVariable` and `DisableVariableSubstitution` even more and, essentially, go with  "setting DisableVariableSubstitution=true implies ThrowOnUnresolvedVariable=false". Is that what you are suggesting?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
Matteo-T: Will be fixed in the next commit.

Matteo-T:
I went with the UI (Development): it resembles ADO after all...

---
In reply to: [1466382921](https://github.com/microsoft/sqltoolsservice/pull/1941#issuecomment-1466382921) [](http://example.com/codeflow?ancestors=1466382921,1466382921)

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
Charles-Gagnon: I would say it makes sense to me to keep the existing behavior you have and just update the docs to clarify this. 

1. DisableVariableSubstitution I would just remove the bit about a null IVariableResolver - that doesn't really matter here and it should be clear enough that if a user is specifically disabling substitution it doesn't matter what the resolver is
2. ThrowOnUnresolvedVariable should mention that it only does this if DisableVariableSubstitution is false since that may not be quite as clear to me

And then of course test cases for those are appreciated 😄

Matteo-T: Cool. I'll have a couple of tests coming up + and updated comment.

Matteo-T: Cleaned up documentation. Added 8 little tests to cover the interaction of these properties.

PR #1940, opened 2023-03-13
ManagedBatchParser: cleaning up a few tests; fixing one

Matteo-T:
In preparation for adding some new tests for the upcoming changes I'm making, I am taking the opportunity to cleanup the suite a little. Specifically:
1. I want to make sure all the tests are working (VerifyConnectSqlCmd was failing due to a missing space in the `:Connect ` statement)
2. I removed any reference to PowerShell from some test and changed the test script to be T-SQT and not some PowerShell fragment (I know were it came from, but... really here we are in the business of parsing T-SQL and not PowerShell...)
3. I've renamed some tests to something more descriptive name, thus making the test description redundant in some cases
4. I've added more precise Assert() statements with descriptive message when the test fails

There are no functional changes to either the product or the tests, really.

PR #1943, opened 2023-03-14
Porting changes from main to net6 (#1940 and #1941)

Matteo-T:
Porting changes from main to net6 (#1940 and #1941)

That payload has:
- cleanup of some tests (#1940) 
- the changes for the DisableVariableSubstitution (#1941)


cheenamalhotra (approved):

Charles-Gagnon:
What's this net6 branch? If this is supposed to be an official release branch we should put it under release/ so it has all the branch protections and requirements in place...

Matteo-T:
> What's this net6 branch? If this is supposed to be an official release branch we should put it under release/ so it has all the branch protections and requirements in place...

The SQLServer PowerShell module ships for both net472 and netcoreapp3.1 (at least for now). I should be able to move to net7.0 "soon" (I have a issue logged against the module to do that), but until I stop supporting PowerShell 5.1, I cannot move too far (either netstandard or net472 at most).

It was discussed a while back with @kburtram and @cheenamalhotra, IIRC.

PR #1945, opened 2023-03-15
MSAL encrypted file system cache

cheenamalhotra:
Introduces synchronized encryption algorithm for MSAL file system cache accessed from Azure Data Studio (MSAL.js) and STS (MSAL.NET)

### DIFF
--- a/src/Microsoft.SqlTools.Authentication/MSALEncryptedCacheHelper.cs
+++ b/src/Microsoft.SqlTools.Authentication/MSALEncryptedCacheHelper.cs
@@ -1,0 +1,261 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Text;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Extensions.Msal;
+using Logger = Microsoft.SqlTools.Utility.Logger;
+
+namespace Microsoft.SqlTools.Authentication.Utility
+{
+    /// <summary>
+    /// This class provides capability to register MSAL Token cache and uses the beforeCacheAccess and afterCacheAccess callbacks 
+    /// to read and write cache to file system. This is done as cache encryption/decryption algorithm is shared between NodeJS and .NET.
+    /// Because, we cannot use msal-node-extensions in NodeJS, we also cannot use MSAL Extensions Dotnet NuGet package.
+    /// Ref https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-node-migration#enable-token-caching
+    /// In future we should use msal extensions to not have to maintain encryption logic in our applications, and also introduce support for
+    /// token storage options in system keychain/credential store.
+    /// However - as of now msal-node-extensions does not come with pre-compiled native libraries that causes runtime issues
+    /// Ref https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/3332
+    /// </summary>
+    public class MSALEncryptedCacheHelper
+    {
+        /// <summary>
+        /// Callback delegate to be implemented by Services in Service Host, where authentication is performed. e.g. Connection Service.
+        /// This delegate will be called to retrieve key and IV data if found absent or during instantiation.
+        /// </summary>
+        /// <param name="key">(out) Key used for encryption/decryption</param>
+        /// <param name="iv">(out) IV used for encryption/decryption</param>
+        public delegate void IVKeyReadCallback(out string key, out string iv);
+
+        /// <summary>
+        /// Lock objects for serialization
+        /// </summary>
+        private readonly object _lockObject = new object();
+        private CrossPlatLock? _cacheLock = null;
+
+        private AuthenticatorConfiguration _config;
+        private StorageCreationProperties _storageCreationProperties;
+        private IVKeyReadCallback _ivKeyReadCallback;
+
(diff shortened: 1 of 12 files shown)


src/Microsoft.SqlTools.Authentication/Utility/Encryption/EncryptionUtils.cs
```diff
@@ -0,0 +1,76 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Security.Cryptography;
+
+namespace Microsoft.SqlTools.Authentication.Utility
+{
+    public static class EncryptionUtils
```
kburtram, line 10: I wonder if this utility code is better in the credential assembly in case it's needed somewhere else?  or does this assembly not reference that one?

src/Microsoft.SqlTools.Authentication/Microsoft.SqlTools.Authentication.csproj
```diff
@@ -9,6 +9,7 @@
 		<PackageReference Include="Microsoft.Data.SqlClient" />
 		<PackageReference Include="Microsoft.Identity.Client" />
 		<PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" />
+		<PackageReference Include="Newtonsoft.Json" />
```
kburtram, line 12: couldn't see what this was being added for?

src/Microsoft.SqlTools.Authentication/MSALEncryptedCacheHelper.cs
```diff
@@ -0,0 +1,261 @@
...
+                    {
+                        data = args.TokenCache.SerializeMsalV3();
+                    }
+                    catch (Exception e)
+                    {
+                        Logger.Error($"An exception was encountered while serializing the {nameof(MsalCacheHelper)} : {e}");
+                        Logger.Error($"No data found in the store, clearing the cache in memory.");
+
+                        // The cache is corrupt clear it out
+                        this._cacheStorage.Clear(ignoreExceptions: true);
```
kburtram, line 141: should we propagate failures here (or at least log them), or they're safe to ignore in this case for some reason?

kburtram (approved):

src/Microsoft.SqlTools.Authentication/Utility/Encryption/EncryptionUtils.cs
cheenamalhotra: Yeah, this one doesn't reference any of the service assemblies, only a 'Microsoft.SqlTools.Shared' project for shared constants, logger and utils.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1134,6 +1136,49 @@ public void RegisterOnDisconnectTask(OnDisconnectHandler activity)
...
+                    iv = ivCred.Password;
+                }
+                else
+                {
+                    throw new Exception($"Could not read credential: {AzureAccountProviderCredentials}|{MsalCacheName}-iv");
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex);
```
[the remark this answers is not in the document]
cheenamalhotra: We can't throw exceptions from here, as this is called during service instantiation. Throwing exception here can crash STS on startup. I'll add a comment here. I'm making minor improvements to notify user from NodeJS layer when these are not available in Cred Store.

src/Microsoft.SqlTools.Authentication/Microsoft.SqlTools.Authentication.csproj
cheenamalhotra: Might have left this out after experiments, will clean up !

src/Microsoft.SqlTools.Authentication/MSALEncryptedCacheHelper.cs
cheenamalhotra: These are safe to ignore, as we also do the same from NodeJS layer when cache gets corrupted/intruded.

src/Microsoft.SqlTools.Authentication/MSALEncryptedCacheHelper.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// This class provides capability to register MSAL Token cache and uses the beforeCacheAccess and afterCacheAccess callbacks 
+    /// to read and write cache to file system. This is done as cache encryption/decryption algorithm is shared between NodeJS and .NET.
+    /// Because, we cannot use msal-node-extensions in NodeJS, we also cannot use MSAL Extensions Dotnet NuGet package.
+    /// Ref https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-node-migration#enable-token-caching
+    /// In future we should use msal extensions to not have to maintain encryption logic in our applications, and also introduce support for
+    /// token storage options in system keychain/credential store.
+    /// However - as of now msal-node-extensions does not come with pre-compiled native libraries that causes runtime issues
+    /// Ref https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/3332
+    /// </summary>
+    public class MSALEncryptedCacheHelper
```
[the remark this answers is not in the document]
cheenamalhotra: Capital rush :P

PR #1946, opened 2023-03-15
fix Azure Authentication for renaming

alanrenmsft:
during the testing, I noticed that the AAD auth is not working for renaming, and turns out it will lose the token information during the connection construction.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
@@ -6,7 +6,6 @@
 #nullable disable
 using System;
 using System.Threading.Tasks;
-using Microsoft.Data.SqlClient;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlServer.Management.Sdk.Sfc;
 using Microsoft.SqlServer.Management.Smo;
@@ -66,11 +65,10 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                     requestParams.ConnectionUri,
                     out connInfo))
             {
-                using (SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo, ObjectManagementServiceApplicationName))
+                ServerConnection serverConnection = ConnectionService.OpenServerConnection(connInfo, ObjectManagementServiceApplicationName);
+                using (serverConnection.SqlConnectionObject)
                 {
-
-                    IRenamable renameObject = this.GetRenamable(requestParams, sqlConn);
-
+                    IRenamable renameObject = this.GetRenamable(requestParams, serverConnection);
                     renameObject.Rename(requestParams.NewName);
                 }
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
```diff
@@ -66,11 +65,10 @@ internal async Task HandleRenameRequest(RenameRequestParams requestParams, Reque
                     requestParams.ConnectionUri,
                     out connInfo))
             {
-                using (SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connInfo, ObjectManagementServiceApplicationName))
+                ServerConnection serverConnection = ConnectionService.OpenServerConnection(connInfo, ObjectManagementServiceApplicationName);
```
Charles-Gagnon, line 68: Does this imply there's a bug with `OpenSqlConnection`? Shouldn't that be keeping all the same info regardless?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
```diff
@@ -66,11 +65,10 @@ internal async Task HandleRenameRequest(RenameRequestParams requestParams, Reque
...
+                using (serverConnection.SqlConnectionObject)
                 {
-
-                    IRenamable renameObject = this.GetRenamable(requestParams, sqlConn);
-
+                    IRenamable renameObject = this.GetRenamable(requestParams, serverConnection);
```
Charles-Gagnon, line 71: While you're looking at this stuff - right now this is going to throw an ugly error if the SMO object isn't an IRenamable (it'll fail to cast it). I'd hope we don't ever hit that - but just in case it wouldn't hurt to use `as IRenamable` and then if it's null throw a more friendly error.

alanrenmsft: Not a concern as far as ADS is concerned because from the UI side we specify which node types will have the rename support.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
alanrenmsft: no, OpenSqlConnection is fine, the problem is with how we construct the ServerConnection using SqlConnection in this service. we didn't include the token information like the `OpenServerConnection()`

image

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
```diff
@@ -87,12 +85,11 @@ internal async Task HandleRenameRequest(RenameRequestParams requestParams, Reque
         /// Method to get the sql object, which should be renamed
         /// </summary>
         /// <param name="requestParams">parameters which are required for the rename operation</param>
-        /// <param name="connection">the sqlconnection on the server to search for the sqlobject</param>
+        /// <param name="connection">the server connection on the server to search for the sqlobject</param>
         /// <returns>the sql object if implements the interface IRenamable, so they can be renamed</returns>
-        private IRenamable GetRenamable(RenameRequestParams requestParams, SqlConnection connection)
+        private IRenamable GetRenamable(RenameRequestParams requestParams, ServerConnection connection)
         {
-            ServerConnection serverConnection = new ServerConnection(connection);
```
alanrenmsft, line 94: @Charles-Gagnon, here, we are not including the access token information when constructing the server connection.

caohai (approved):

kburtram (approved):

PR #1949, opened 2023-03-17
Expose parent node path of tree node

alanrenmsft:
For [#22342 ](https://github.com/microsoft/azuredatastudio/issues/22342), together with these changes in ADS: https://github.com/microsoft/azuredatastudio/pull/22356

In ADS side, we need to find the parent node path, but there is no reliable way of determining the path of the parent node. so exposing it directly.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/NodeInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/NodeInfo.cs
@@ -26,7 +26,8 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.Contracts
         public string NodePath { get; set; }
 
         /// <summary>
-        /// The path of the parent node.
+        /// The path of the parent node. This is going to be used by the client side to determine the parent node.
+        /// We are not referencing the parent node directly because the information needs to be passed between processes.
         /// </summary>
         public string ParentNodePath { get; set; }
 
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/NodeInfo.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/NodeInfo.cs
@@ -26,6 +26,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.Contracts
         public string NodePath { get; set; }
 
         /// <summary>
+        /// The path of the parent node.
+        /// </summary>
+        public string ParentNodePath { get; set; }
+
+        /// <summary>
         /// The type of the node - for example Server, Database, Folder, Table
         /// </summary>
         public string NodeType { get; set; }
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
```diff
@@ -229,6 +229,7 @@ public NodeInfo ToNodeInfo()
                 IsLeaf = this.IsAlwaysLeaf,
                 Label = this.Label,
                 NodePath = this.GetNodePath(),
+                ParentNodePath = this.Parent?.GetNodePath() ?? "",
```
Charles-Gagnon, line 232: Why not get it set up to actually have the Parent link in the NodeInfo? That seems a lot more useful

alanrenmsft: this is designed to be consumed by ADS side.

Charles-Gagnon: Sure, and it would still be useful there. Having to duplicate any parent properties in children nodes as well isn't exactly a great pattern. And then what happens if we want to walk up the tree to find the root node or something for example?

alanrenmsft: The node path is just like the id of the node, ADS core (OE service) has a collection of all the nodes, if we need to walk up the tree, we can call getParent function (the one I modified in the ADS PR.) until it reaches the root of tree. I don't think it is a good idea to include all the ancester nodes in every node, not necessary and not efficient.

Charles-Gagnon: You definitely wouldn't include the node objects themselves. You would pass a parentId or something and then on the ADS side convert it into a reference to the actual parent object. 

Although if you have a getParent method already then why wouldn't you just call node.getParent().nodePath on the ADS side?

alanrenmsft: the getParent() needs to use the parentNodePath to find the parent node.

Charles-Gagnon: Oh ok, so this is for doing what I was suggesting - you're just using the path for the "id". That makes sense then.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/NodeInfo.cs
```diff
@@ -25,6 +25,11 @@ public class NodeInfo
         /// </summary>
         public string NodePath { get; set; }
 
+        /// <summary>
+        /// The path of the parent node.
```
Charles-Gagnon, line 29: I would suggest putting a remark here though that this intended to allow nodes to look up their parents - just to make it clear that we aren't trying to store all the parent node info here.

Charles-Gagnon (approved):

PR #1951, opened 2023-03-17
.editorconfig Cleanup for cs1711

chlafreniere:
Just took a gander at the .editorconfig file and wanted to make a little dent in cleaning up some of these issues here.

Started with the end, and found one that ended up being quite simple to fix. Doc here: https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1711.

Was unsure whether we wanted to mark these warnings as errors (as I've done here), or just remove this entry from the .editorconfig file. I went with error here to prevent further occurrences from becoming non-blocking warnings in the future, but please let me know if you disagree 😄.


.editorconfig
```diff
@@ -146,7 +146,7 @@ dotnet_diagnostic.CS1587.severity = none
 dotnet_diagnostic.CS1574.severity = none
 dotnet_diagnostic.CS0419.severity = none
 dotnet_diagnostic.CS1723.severity = none
-dotnet_diagnostic.CS1711.severity = none
+dotnet_diagnostic.CS1711.severity = error
```
Charles-Gagnon, line 149: Just remove this, the default is error
```suggestion
```

Charles-Gagnon (approved):

.editorconfig
chlafreniere: That was the first thing that I tried, but it didn't result in an error (before I made the change to remove the offending line in source):
image

Perhaps because this is a compiler warning and not an analyzer?

By the way, setting dotnet_analyzer_diagnostic.severity = error doesn't turn all of the analyzers on and have them error out; it only affects those that are turned on by default: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options#scope

Charles-Gagnon: Ah, good to know. I do recall this being not very clear when I first added this stuff. Maybe we can move these into a separate section at least so it's easy to see which ones still need to be fixed?

chlafreniere: Happy to move it if you think it's unclear on those marked as "none" vs. "error" (or put a comment)

Charles-Gagnon: I just don't want us to get to the point where we have like 5 left and have to scan through a huge list of "error" ones to try and pick them out.

Benjin (approved):

kburtram (approved):

PR #1953, opened 2023-03-20
simplify drop object request handler

alanrenmsft:
1. Use a single endpoint for drop object request
2. Moved the new request handler to object management service folder

Login 
rename-delete-login

Rename and Drop user (with error)
drop-rename-user

Drop user
drop-user

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
@@ -11,7 +11,9 @@ using Microsoft.SqlServer.Management.Sdk.Sfc;
 using Microsoft.SqlServer.Management.Smo;
 using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.Connection;
+using Microsoft.SqlTools.ServiceLayer.Management;
 using Microsoft.SqlTools.ServiceLayer.ObjectManagement.Contracts;
+using Microsoft.SqlTools.ServiceLayer.Utility;
 using Microsoft.SqlTools.Utility;
 
 namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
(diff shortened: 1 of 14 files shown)


src/Microsoft.SqlTools.ServiceLayer/Security/UserActions.cs
```diff
@@ -282,42 +282,10 @@ await requestContext.SendResult(new ResultStatus()
...
-                throw new ArgumentException("Invalid null parameter");
-            }
-
-            CDataContainer dataContainer = CDataContainer.CreateDataContainer(connInfo, databaseExists: true);
-            string dbUrn = "Server/Database[@Name='" + Urn.EscapeString(parameters.Database) + "']";
-            Database? parent = dataContainer.Server.GetSmoObject(new Urn(dbUrn)) as Database;
-            User user = parent.Users[parameters.Name];
-            dataContainer.SqlDialogSubject = user;
-
-            CheckForSchemaOwnerships(parent, user);
```
alanrenmsft, line 308: SQL Engine will perform the check and throw, so deleting this is safe.

caohai (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
```diff
@@ -58,40 +59,71 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+        /// <summary>
+        /// Method to handle the delete object request
+        /// </summary>
+        /// <param name="requestParams">parameters which are needed to execute deletion operation</param>
+        /// <param name="requestContext">Request Context</param>
+        /// <returns></returns>
+        internal async Task HandleDropRequest(DropRequestParams requestParams, RequestContext<bool> requestContext)
+        {
+            Logger.Verbose("Handle Request in HandleDeleteRequest()");
+            ExecuteActionOnObject(requestParams.ConnectionUri, requestParams.ObjectUrn, (dbObject) =>
```
kburtram, line 87: should we use `DatabaseUtils.DoDropObject` to drop objects or is all the extra process that's doing unnecessary?  For context that code was ported directly from the "Drop Object" dialog in SSMS.

kburtram (approved):
Looks fine, but the drop logic you're doing isn't including all the logic ported from SSMS.  Do we not need all those extra checks?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
alanrenmsft: for the object types we current support, what I have is good enough, but I ended up using what you suggested as it is more comprehensive.

PR #1972, opened 2023-03-27
make the NodeTypeId nullable

alanrenmsft:
This PR fixes https://github.com/microsoft/azuredatastudio/issues/22476

Not all nodes have the NodeTypeId set, previously the first value (AggregateFunction) of the enum is used as the default value, which is wrong, I am changing this to nullable to reflect the actual value.

we could also fix it by adding enum values for all node types, but at least it is not needed for my use case since those the node type and actual object type are the same.  @aasimkhan30 what do you think.

before
image

after
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/TreeNode.cs
@@ -103,7 +103,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.Nodes
         /// <summary>
         /// Enum defining the type of the node - for example Server, Database, Folder, Table
         /// </summary>
-        public NodeTypes NodeTypeId { get; set; }
+        public NodeTypes? NodeTypeId { get; set; }
 
         /// <summary>
         /// Node Sub type - for example a key can have type as "Key" and sub type as "PrimaryKey"


aasimkhan30 (approved):

Charles-Gagnon (commented):
When would node type and object type ever differ? (assuming we have both) They seem like they should be used for the same thing so not sure why we have both in the first place.

alanrenmsft:
@Charles-Gagnon an example is Folder nodes, the NodeType will be Folder and NodeTypeId will be the actual object type of it, for example: Databases, Tables.

Charles-Gagnon:
Couldn't we use nodeSubType for that?

Charles-Gagnon:
And wait, your example didn't even include objectType which is the thing I'm questioning even needing. So where does objectType fit into that scenario (or any other scenario that nodeType/nodeTypeId can't fit already)

alanrenmsft:
The ObjectType is mapped to NodeTypeId, are you recommending to reuse the NodeSubType? NodeType  = Folder and NodeSubType = Databases?

alanrenmsft:
I also don't know why those properties are introduced in the first place, but renaming is definitely needed to make it clear.

Charles-Gagnon (approved):
Anyways, guess this is fine to fix the immediate issue but I think it'd be worth spending a bit of time seeing if you can get rid of the objectType field. That should be low risk to breaking existing stuff since it's a new addition so unlikely any other providers would be using it.

PR #1977, opened 2023-03-30
Add OE messages to docs

Charles-Gagnon:
Working on some doc updates and noticed that we don't have anything for the OE messages here, so adding them. 

I did them in a different format than the other examples (direct links to classes/types instead of copying it here) because keeping this updated is going to be annoying otherwise, but for now just going to leave the existing ones as they are since it'll be a lot of manual work to update them all and doesn't provide a huge amount of benefit. Going forward I think we should do direct links though.


Charles-Gagnon:
Merging in so I can continue to work on the other docs, but if you see anything to fix just leave a comment and I'll come back in and clean it up.

PR #1991, opened 2023-04-05
Fixing OE database expansion read only datamarts databases

aasimkhan30:
This error is caused due the check we have here:
https://github.com/microsoft/sqltoolsservice/blob/f288bee29418fd205a9d14e3c3a4cac360caa11d/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SqlServerType.cs#L44
The check executes this query on the readonly database causing a failure as tables cannot be created. 
https://msdata.visualstudio.com/DefaultCollection/SQLToolsAndLibraries/_git/SqlManagementObjects?path=src/Microsoft/SqlServer/Management/SqlEnum/xml/Database.xml&version=GC458bf2b107818c3b85fd58589921c31d612b4baf&line=95&lineStartColumn=102&lineEndColumn=102&_a=contents

Here is the link to the original issue in the ads issue:
https://github.com/microsoft/azuredatastudio/issues/22500

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -105,10 +105,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
                 && db.ServerVersion.Major == 12;
         }
 
-        private bool isUnknownDatabaseEdition(Database db){
+        private bool isUnknownDatabaseEdition(Database db)
+        {
             // If the database engine edition is not defined in the enum, it is an unknown edition
-            return db!= null
-                && !Enum.IsDefined(typeof(DatabaseEngineEdition),db.DatabaseEngineEdition);
+            return db != null
+                && !Enum.IsDefined(typeof(DatabaseEngineEdition), db.DatabaseEngineEdition);
         }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -80,9 +80,9 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             catch (Exception ex)
             {
                 // IsAccessible is not set of DW Gen3 and Dataverse so exception is expected in this case
-                // Incase of dataverses, isAccessible creates a temp table to check if the database is accessible, however dataverse 
-                // don't support ddl statements and therefore this check fails.
-                if (IsDWGen3(context?.Database) || IsDataverse(context?.Database))
+                // Incase of dataverses and other TDS endpoints isAccessible creates a temp table to check if the database  
+                // is accessible, however these endpoints may not support ddl statements and therefore the check fails.
+                if (IsDWGen3(context?.Database) || isUnknownDatabaseEdition(context?.Database))
                 {
                     return true;
                 }
@@ -105,11 +105,10 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
                 && db.ServerVersion.Major == 12;
         }
 
-        private bool IsDataverse(Database db){
-            // Dataverses have a DatabaseEngineEdition of 1000. We have to cast it to int as DatabaseEngineEdition enum does not have a value for dataverse.
+        private bool isUnknownDatabaseEdition(Database db){
+            // If the database engine edition is not defined in the enum, it is an unknown edition
             return db!= null
-                && (int)db.DatabaseEngineEdition == 1000 
-                && db.ServerVersion.Major == 12;
+                && !Enum.IsDefined(typeof(DatabaseEngineEdition),db.DatabaseEngineEdition);
         }
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -80,7 +80,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             catch (Exception ex)
             {
                 // IsAccessible is not set of DW Gen3 so exception is expected in this case
-                if (IsDWGen3(context?.Database))
+                if (IsDWGen3(context?.Database) || IsDataMart(context?.Database))
                 {
                     return true;
                 }
@@ -100,6 +100,13 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
         {
             return db != null
                 && db.DatabaseEngineEdition == DatabaseEngineEdition.SqlDataWarehouse
+                && db.ServerVersion.Major == 12;
+        }
+
+        private bool IsDataMart(Database db){
+            // Datamarts have a DatabaseEngineEdition of 1000. We have to cast it to int as DatabaseEngineEdition enum does not have a value for Datamarts.
+            return db!= null
+                && (int)db.DatabaseEngineEdition == 1000 
                 && db.ServerVersion.Major == 12;
         }
     }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
@@ -79,8 +79,10 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
             }
             catch (Exception ex)
             {
-                // IsAccessible is not set of DW Gen3 so exception is expected in this case
-                if (IsDWGen3(context?.Database) || IsDataMart(context?.Database))
+                // IsAccessible is not set of DW Gen3 and Dataverse so exception is expected in this case
+                // Incase of dataverses, isAccessible creates a temp table to check if the database is accessible, however dataverse 
+                // don't support ddl statements and therefore this check fails.
+                if (IsDWGen3(context?.Database) || IsDataverse(context?.Database))
                 {
                     return true;
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SqlServerType.cs
```diff
@@ -41,7 +42,17 @@ public class ServerVersionHelper
         /// </summary>
         public static ValidForFlag GetValidForFlag(SqlServerType serverType, Database database = null)
         {
-            return GetValidForFlag(serverType, database != null && database.IsSqlDw);
+            var isSqlDw = false;
+            try
+            {
+                isSqlDw = database.IsSqlDw;
+            }
+            catch (Exception e)
```
alanrenmsft, line 50: what exceptions are we expecting here? if recoverable, we should log it as information and add explanation says this is expected so that it won't cause false alarms.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -80,7 +80,7 @@ public bool IsAccessible(SmoQueryContext context)
             catch (Exception ex)
             {
                 // IsAccessible is not set of DW Gen3 so exception is expected in this case
-                if (IsDWGen3(context?.Database))
+                if (IsDWGen3(context?.Database) || IsDataMart(context?.Database))
```
alanrenmsft, line 83: isn't this supposed to be the opposite? if DataMart then not accessible?

kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -102,5 +102,12 @@ private bool IsDWGen3(Database db)
                 && db.DatabaseEngineEdition == DatabaseEngineEdition.SqlDataWarehouse
                 && db.ServerVersion.Major == 12;
         }
+
+        private bool IsDataMart(Database db){
+            // Datamarts have a DatabaseEngineEdition of 1000. We have to cast it to int as DatabaseEngineEdition enum does not have a value for Datamarts.
```
alanrenmsft, line 107: Time to update the enum definition?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
aasimkhan30: isAccessible throws an exception in case of datamarts.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
aasimkhan30: Yup. This probably has to be updated in smo

alanrenmsft (approved):

alanrenmsft:
please add more comments as we discussed.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
Charles-Gagnon: @shueybubbles FYI

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -102,5 +104,12 @@ private bool IsDWGen3(Database db)
                 && db.DatabaseEngineEdition == DatabaseEngineEdition.SqlDataWarehouse
                 && db.ServerVersion.Major == 12;
         }
+
+        private bool IsDataverse(Database db){
+            // Dataverses have a DatabaseEngineEdition of 1000. We have to cast it to int as DatabaseEngineEdition enum does not have a value for dataverse.
+            return db!= null
+                && (int)db.DatabaseEngineEdition == 1000
```
shueybubbles, line 111: >nt)db.DatabaseEngineEdition == 1000  [](http://example.com/codeflow?start=21&length=36)

can we instead have a list of known valid editions and treat anything else the same way we treat datamart?

shueybubbles: you could use Enum.TryParse and if it doesn't have a value just assume it's data mart.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
shueybubbles: SMO should probably report this as "Unknown" instead of accepting random values

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -102,5 +104,12 @@ private bool IsDWGen3(Database db)
                 && db.DatabaseEngineEdition == DatabaseEngineEdition.SqlDataWarehouse
                 && db.ServerVersion.Major == 12;
         }
+
+        private bool IsDataverse(Database db){
+            // Dataverses have a DatabaseEngineEdition of 1000. We have to cast it to int as DatabaseEngineEdition enum does not have a value for dataverse.
+            return db!= null
+                && (int)db.DatabaseEngineEdition == 1000 
+                && db.ServerVersion.Major == 12;
```
shueybubbles, line 112: > 12; [](http://example.com/codeflow?start=44&length=4)

they could upgrade their version number at any time. Does it really matter what version they report?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
Charles-Gagnon: Are you saying that datamart doesn't have a static engine edition? Why would we assume no mapped value == datamart otherwise? And what happens when a new edition is added later on - that could cause problems if we just assume it's data mart by default.

shueybubbles: My logic is this - we would add this value to the SMO enumeration only after we add real support for it. Once it shows up there presumably stuff would "just work" without the special cases in UI code.

shueybubbles: I'm saying that datamart is just one example of a TDS endpoint that isn't a real SQL Server. There could be others. The goal of the UI would be to provide a minimal OE for any such endpoint, correct?

Charles-Gagnon: I could see that - where we basically say "if we don't know what type of server this is we default to some basic set of functionality that we know should work". So instead of having all this stuff be datamart-specific just have it check if it's a known edition and if not then skip certain logic. 

The SMO change to have an "Unknown" enum value would definitely help with that.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -79,8 +79,10 @@ public bool IsAccessible(SmoQueryContext context)
             }
             catch (Exception ex)
             {
-                // IsAccessible is not set of DW Gen3 so exception is expected in this case
-                if (IsDWGen3(context?.Database))
+                // IsAccessible is not set of DW Gen3 and Dataverse so exception is expected in this case
+                // Incase of dataverses, isAccessible creates a temp table to check if the database is accessible, however dataverse 
+                // don't support ddl statements and therefore this check fails.
+                if (IsDWGen3(context?.Database) || IsDataverse(context?.Database))
```
aasimkhan30, line 85: @Charles-Gagnon and @shueybubbles if I change this check to unknown, won't it make else branch useless?

shueybubbles:
Does IsAccessible have any meaning in the cloud? I thought you could only login if you have at least public access to it.
Should it be hard coded to true in SMO for anything other than on premises instances?

aasimkhan30:
> Does IsAccessible have any meaning in the cloud? I thought you could only login if you have at least public access to it. Should it be hard coded to true in SMO for anything other than on premises instances?

@shueybubbles, I wonder if there are special permissions that let you connect to the database but not access database objects. Won't isAccessible be false in those cases?

shueybubbles:
HAS_DBACCESS returns 1 if the user has access to the database, 0 if the user has no access to the database, and NULL if the database name is not valid.

HAS_DBACCESS returns 0 if the database is offline or suspect.

HAS_DBACCESS returns 0 if the database is in single-user mode and the database is in use by another user.

---
In reply to: [1498256364](https://github.com/microsoft/sqltoolsservice/pull/1991#issuecomment-1498256364) [](http://example.com/codeflow?ancestors=1498256364,1498256364)

shueybubbles:
if you connect to a cloud user database successfully, then it seems evident none of the 3 cases is true


---
In reply to: [1498258019](https://github.com/microsoft/sqltoolsservice/pull/1991#issuecomment-1498258019) [](http://example.com/codeflow?ancestors=1498258019,1498256364,1498256364)

shueybubbles:
And I don't see how it could be accurate to query has_dbaccess('somedatabase') in the logical master, because the logical master doesn't know which users have been added to specific databases

---
In reply to: [1498259616](https://github.com/microsoft/sqltoolsservice/pull/1991#issuecomment-1498259616) [](http://example.com/codeflow?ancestors=1498259616,1498258019,1498256364,1498256364)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs
```diff
@@ -79,8 +79,10 @@ public bool IsAccessible(SmoQueryContext context)
             }
             catch (Exception ex)
             {
-                // IsAccessible is not set of DW Gen3 so exception is expected in this case
-                if (IsDWGen3(context?.Database))
+                // IsAccessible is not set of DW Gen3 and Dataverse so exception is expected in this case
+                // Incase of dataverses and other TDS endpoints isAccessible creates a temp table to check if the database  
+                // is accessible, however these endpoints may not support ddl statements and therefore the check fails.
+                if (IsDWGen3(context?.Database) || isUnknownDatabaseEdition(context?.Database))
```
cheenamalhotra, line 85: If this Enum starts including DataVerse in future, we might start getting issues.. as it won't be DW Gen3 and neither an unknown.. is there no other identification available for identifying whether `db` isDataVerse?

cc @shueybubbles do you have any thoughts?

cheenamalhotra: Since Dataverse is officially supported by SSMS ([MS Docs](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/dataverse-sql-query)), I wonder if we could do better to identify dataverse database from SMO.

shueybubbles: uh I don't think our team stated SSMS supports it, their team is saying it will work with SSMS.

cheenamalhotra (approved):

PR #1993, opened 2023-04-05
Enabling DacFx telemetry

llali:
- Updated DacFx nuget to 162 to get telemetry feature.
- Requires this ADS PR to get telemetry config changes https://github.com/microsoft/azuredatastudio/pull/22644
- If telemetry config is not set to OFF in ADS, enables telemetry in DacFx to send events


src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -62,6 +65,13 @@ public void InitializeService(ServiceHost serviceHost)
             serviceHost.SetRequestHandler(GenerateTSqlModelRequest.Type, this.HandleGenerateTSqlModelRequest, true);
             serviceHost.SetRequestHandler(GetObjectsFromTSqlModelRequest.Type, this.HandleGetObjectsFromTSqlModelRequest, true);
             serviceHost.SetRequestHandler(SavePublishProfileRequest.Type, this.HandleSavePublishProfileRequest, true);
+            Workspace.WorkspaceService<SqlToolsSettings>.Instance.RegisterConfigChangeCallback(UpdateSettings);
+        }
+
+        internal Task UpdateSettings(SqlToolsSettings newSettings, SqlToolsSettings oldSettings, EventContext eventContext)
+        {
+            TelemetryEnabled = newSettings.TelemetrySettings.Telemetry != TelemetryLevel.Off;
+            return Task.FromResult(0);
```
Charles-Gagnon, line 74:
```suggestion
            return Task.CompletedTask;
```

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
```diff
@@ -35,15 +39,18 @@ abstract class DacFxOperation : ITaskOperation
 
         protected DacServices DacServices { get; private set; }
 
+        protected bool TelemetryEnabled { get; private set; }
+
         protected ConnectionInfo ConnInfo { get; private set; }
 
-        protected DacFxOperation(ConnectionInfo connInfo)
+        protected DacFxOperation(ConnectionInfo connInfo, bool telemetryEnabled = false)
```
Charles-Gagnon, line 46: Instead of passing it in why not just update the library directly when the config changes? Right now you won't change the telemetry state until you start the next operation which isn't really what we want.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
```diff
@@ -118,5 +127,32 @@ protected DacDeployOptions GetDefaultDeployOptions()
...
+
+        /// <summary>
+        /// Changes telemetry status 
+        /// </summary>
+        private void UpdateTelemetryStatus()
+        {
+            try
+            {
+                DacExtensions.SetServiceSetting(this.DacServices, "TelemetryEnabled", TelemetryEnabled);
+                DacExtensions.SetServiceSetting(this.DacServices, "TelemetryApplicationName", TelemetryApplicationName);
```
Charles-Gagnon, line 139: You should only set the app name and version once on startup - no reason to reset it every time. Only the enabled state should change and even that only when the config changes.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
llali: hmm yes you are right and I'm wondering now why it was working ...

llali: does a static field work?

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
llali: DacFx settings need to set for each operation. this is by design for DacFx

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
Charles-Gagnon: Not really I think - you need each operation itself to be updated separately since they're each creating their own services instance. So instead I think you'd have DacFxService watch for the changes and then loop through each operation and tell it to update the telemetry enabled state for each one.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
Charles-Gagnon: Why did you make it per services instance anyways though? Telemetry should be enabled/disabled state I would think - is there a scenario where you specifically wanted to have it enabled for one operation but not for another or something?

llali: yes, user can do an import with telemetry then disable the telemetry and do export ...

llali: for each DacFx operation we create a new DacSerive and telemetry will be enabled at the service level

Charles-Gagnon: Is there a specific scenario where you thought that would be useful? It just seems easier and more expected to have the telemetry toggle be at the global level. I can't think of any tools that have it more fine grained than that.

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
llali: I already added the watch in the DacFxService class. and that service doesn't have access the the operations. operations are created for new requests

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxOperation.cs
Charles-Gagnon: Anyways it is what it is for now, but I would strongly suggest going back and thinking about whether you really want to keep it on a per-service level. It's not too late to change it IMO and would make it a lot easier and safer to ensure you're not sending events that you don't intend to.

llali: sure I can add a new API in DacFx to enable telemetry globally so user doesn't have to do it eveytime creating dac service. I think that make sense

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -417,5 +432,38 @@ internal void PerformOperation(DacFxOperation operation, TaskExecutionMode taskE
...
+                }
+                else
+                {
+                    DacServices.DisableTelemetry();
+                }
+            
+            }
+            catch (Exception ex)
+            {
+                Logger.Write(TraceEventType.Warning, $"Failed to update DacFx telemetry status. telemetry enable: {telemetryEnabled}, error: {ex.Message}");
```
Charles-Gagnon, line 455: FYI - I added extension methods to make this easier (e.g. `Logger.Warning("Message"...`)

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -72,7 +72,7 @@ public void InitializeService(ServiceHost serviceHost)
 
         internal Task UpdateSettings(SqlToolsSettings newSettings, SqlToolsSettings oldSettings, EventContext eventContext)
         {
-            Settings.PreloadDatabaseModel = newSettings.MssqlTools.TableDesigner.PreloadDatabaseModel;
+            Settings.PreloadDatabaseModel = newSettings.MssqlTools.TableDesigner != null ? newSettings.MssqlTools.TableDesigner.PreloadDatabaseModel : false;
```
Charles-Gagnon, line 75: Were you seeing an issue here?

FYI @alanrenmsft 
```suggestion
            Settings.PreloadDatabaseModel = newSettings.MssqlTools.TableDesigner?.PreloadDatabaseModel ?? false;
```

llali: when I was debugging getting the requests from mssql-vscode

Charles-Gagnon: Ah yes, that'd do it. Definitely something to keep in mind when adding other settings so we don't end up accidently breaking VS Code MSSQL. Thanks for fixing!

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -417,5 +432,46 @@ internal void PerformOperation(DacFxOperation operation, TaskExecutionMode taskE
...
+            try
+            {
+                string fileVersion = FileVersionInfo.GetVersionInfo(System.Reflection.Assembly.GetExecutingAssembly().Location).FileVersion;
+                if (Version.TryParse(fileVersion, out Version version))
+                {
+                    return version;
+                }
+                return null;
+            }
+            catch(Exception ex)
```
Charles-Gagnon, line 470: nit: Format this file

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -417,5 +432,46 @@ internal void PerformOperation(DacFxOperation operation, TaskExecutionMode taskE
...
+                }
+            
+            }
+            catch (Exception ex)
+            {
+                Logger.Warning($"Failed to update DacFx telemetry status. telemetry enable: {telemetryEnabled}, error: {ex.Message}");
+            }
+        }
+
+        private static Version LoadServiceVersion()
```
Charles-Gagnon, line 459: Please make sure to use proper nullable types here - this should be `Version?` for example

I would suggest removing the `#nullable disable` command from this file and seeing how many there are to fix - if it's just a couple then it'd be nice to fix them now. Otherwise you can at least use that to easily see if the new code you added is using proper nullable types.

Charles-Gagnon (approved):

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
llali: I made the change but I get warning now because of "#nullable disable". should I still keep it?

Charles-Gagnon: No, and in fact it would be great for everyone working in STS to be slowly trying to remove these by fixing up the nullable stuff bit by bit. So if you've done that for this file then you can remove the pragma and we're one step closer to having that complete!

llali: I enabled it back and fix the only nullable issue I saw. @Benjin @kisantia please let me know if you have any concern enabling nullable in this file

PR #1844, opened 2023-02-03
Add support to save publish profile

SakshiS-harma:
This PR adds support to save publish profile. Currently it doesn't set deployment options to be sent to DacFx.

PR #1997, opened 2023-04-07
Send deploymentOptions to DacFx to save in the publish profile xml file

SakshiS-harma:
This PR completes the backend functionality for saving publish profile (first part: https://github.com/microsoft/sqltoolsservice/pull/1844).


SakshiS-harma:
Packages.props will be updated once the DacFx nuget is released. Ignore the conflicts because of that for now.

kisantia (approved):

Packages.props
```diff
@@ -22,7 +22,7 @@
 
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
 		<PackageReference Update="Microsoft.SqlServer.Management.SmoMetadataProvider" Version="170.12.0" />
-		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="161.8406.0-preview" />
+		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="162.0.14" />
```
llali, line 25: the version I put in nuget is [162.0.15-preview](https://www.nuget.org/packages/Microsoft.SqlServer.DacFx/162.0.15-preview) not sure where you are getting 162.0.14 from

SakshiS-harma: So that one was from the testing I was doing. I have updated the version to 162.0.15-preview now.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -882,6 +882,17 @@ public async Task ValidateSavePublishProfile()
             string profileFilePath = Path.Combine(folderPath, fileName);
             string expectedFile = Path.Combine(publishProfileFolder, fileName);
 
+            DeploymentOptions deploymentOptions = DeploymentOptions.GetDefaultPublishOptions();
+
+            // a few extra options get added in the publish.xml file, because of the defaults that are set in STS for maintaining compatibility with ADS and SSDT 
+            // these defaults are defined in ..\sqltoolsservice\src\Microsoft.SqlTools.ServiceLayer\DacFx\Contracts\DeploymentOptions.cs
+            deploymentOptions.ExcludeObjectTypes = new DeploymentOptionProperty<string[]>(new[] { Enum.GetName(ObjectType.Views) });
```
Benjin, line 889: `nameof(ObjectType.Views)` occurs at compile-time, so it'll be more performant overall.

More blathering: I remembered some issue with `Enum.GetName` where it would return the wrong name if there were multiple enums with overlapping values (i.e. if an enum is being used as a flag/set), so I throw a search to the ether on whether `Enum.GetName()` or `myEnum.ToString()` is better, and found [this post](https://stackoverflow.com/questions/483794/convert-enum-to-string) that told me that this third option is best!

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -329,7 +330,7 @@ public async Task HandleSavePublishProfileRequest(SavePublishProfileParams param
                     DacProfile profile = new DacProfile();
                     profile.TargetDatabaseName = parameters.DatabaseName;
                     profile.TargetConnectionString = parameters.ConnectionString;
-                    //TODO: Set deploy options to pass on to DacFx
+                    profile.DeployOptions = SchemaCompareUtils.CreateSchemaCompareOptions(parameters.DeploymentOptions);
```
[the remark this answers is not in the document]
SakshiS-harma: I had the same thought while working on the changes, because the name and location of the method should be generic instead of being specific to Schema Compare. Updated it appropriately.

PR #2002, opened 2023-04-13
Support scripting for Users

kburtram:


src/Microsoft.SqlTools.ServiceLayer/Security/UserActions.cs
```diff
@@ -296,11 +299,42 @@ internal async Task HandleUpdateUserRequest(UpdateUserParams parameters, Request
 
             await requestContext.SendResult(new ResultStatus()
             {
-                Success = result.Item1,
-                ErrorMessage = result.Item2
+                Success = true,
```
kburtram, line 302: we don't use the Success & ErrorMessage properties in the front-end.  I'll remove them from object definitions in subsequent PR.

caohai (approved):

PR #2004, opened 2023-04-14
Print commands on build failure

Charles-Gagnon:
Was working on something else and was getting an error but couldn't easily find out which one of the many commands that are ran during the build/publish process were failing - so to help with that I'm modifying the build script to output the command that failed in addition to the error. 

I also switched the RunRestore task to use the Run command instead of the Cake built-in StartAndReturnProcess since that returns a specialized Cake version of the Process and isn't directly compatible with the Run command values (which use System.Diagnostic.Process). I originally changed it so I could pass the Process object directly to the ExitStatus but then I remembered you can't access all properties of the Process after it's exited so went back on that change (and just pass the cmd/args directly). But making the type of Process object used in the script be consistent throughout is a good idea anyways so leaving that change.

### DIFF
--- a/scripts/runhelpers.cake
+++ b/scripts/runhelpers.cake
@@ -31,25 +31,20 @@ public class RunOptions
 /// </summary>
 public struct ExitStatus
 {
-    private int _code;
+    private string _cmd;
+    private string _args;
+    private int _exitCode;
     private bool _timeOut;
-    /// <summary>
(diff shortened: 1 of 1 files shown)


alanrenmsft (approved):

kburtram (approved):

Charles-Gagnon:
Test failure unrelated to these changes (we see it occasionally in normal run)

PR #2007, opened 2023-04-14
Add SourceLink to build + fix assembly versioning

Charles-Gagnon:
This originally started as just adding the SourceLink to the build - which will cause the build to append the commit hash to the version number which will help make it easy to identify the STS version during debugging, as well as add other helpful features documented here : https://github.com/dotnet/sourcelink

In order to do this though I had to remove the AssemblyInfo.cs files - since those were hardcoding the version info to 1.0.0.0. But in general using AssemblyInfo.cs isn't something we should be doing anyways - all of the properties in it can be generated by the build itself and is easier to maintain/share values when they're build time props instead of hardcoded values. 

Changes resulting from this PR:

1. Assembly/File versions will now actually match the version of the release instead of being hardcoded to 1.0.0.0
2. The InformationalVersion will have the commit has appended to it
3. Remove the GUID/ComVisible assembly properties - we don't deal with COM so I didn't see any reason to add extra stuff just for that given the very specific usage of this library
4. Deleted AssemblyInfo.cs files

NOTE: The version is currently mismatched since we're using different version numbers for the release than what gets passed to the build. I'll plan on following up on fixing that in a separate PR.

Old info:

image

image

New info:
image

image


scripts/runhelpers.cake
```diff
@@ -38,7 +38,9 @@ public struct ExitStatus
     /// <summary>
     ///  Default constructor when the execution potentially timed out.
     /// </summary>
-    /// <param name="process">The process this status is for</param>
+    /// <param name="cmd">The cmd/file being executed</param>
```
Charles-Gagnon, line 41: Missed updating these in my last PR so just sneaking them in now...

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -70,13 +71,11 @@
 	<ItemGroup>
 		<EmbeddedResource Include=".\Agent\NotebookResources\NotebookJobScript.ps1" />
 	</ItemGroup>
-	<!-- this target enables dependency files to be copied as part of the output of ProjectReference.
-	https://github.com/dotnet/sdk/issues/1675
-	 -->
-	<Target Name="AddRuntimeDependenciesToContent" Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'" BeforeTargets="GetCopyToOutputDirectoryItems" DependsOnTargets="GenerateBuildDependencyFile;&#xD;&#xA;							GenerateBuildRuntimeConfigurationFiles">
```
Charles-Gagnon, line 76: This doesn't seem to be necessary when adding the SourceLink package - this was causing an error on publish since the build itself was also copying the deps.json file. So removing this since the SourceLink package is handling doing the copying for us.

src/Directory.Build.props
```diff
@@ -1,23 +1,21 @@
...
 		<RepositoryType>git</RepositoryType>
 		<PackageTags>SQL XPLAT</PackageTags>
-	</PropertyGroup>
-	<PropertyGroup>
 		<DebugType>portable</DebugType>
+		<!-- add Source Link info to package/pdbs-->
+		<PublishRepositoryUrl>true</PublishRepositoryUrl>
 	</PropertyGroup>
+	<ItemGroup>
+		<PackageReference Include="Microsoft.SourceLink.GitHub" />
```
shueybubbles, line 19: >	<PackageReference Include="Microsoft.SourceLink.GitHub" /> [](http://example.com/codeflow?start=1&length=59)

this could be a GlobalPackageReference in packages.props instead
If you leave it here you should set PrivateAssets=All at least.

shueybubbles (approved):
:shipit:

src/Directory.Build.props
shueybubbles: and do you need to pass `/p:ContinuousIntegrationBuild=true` in your official build to get the full benefit?

cheenamalhotra:
Can we have the assembly file version 3.0.xx.x match the current STS versioning scheme (4.7.x.x) too?

Charles-Gagnon:
@cheenamalhotra - I'll be doing that as a follow-up PR to avoid stuffing too many changes in this one.

src/Directory.Build.props
Charles-Gagnon: Updated. Decided to go with property instead since we just have a single build script and I didn't feel like it was worth the effort to plumb through a whole new parameter just for this (and it also ensures we always have the property set in the pipelines)

Directory.Build.props
```diff
@@ -21,13 +27,19 @@
...
-
     <ToolsServiceTargetRuntimes>win-x64;win-x86;win-arm64;ubuntu.14.04-x64;ubuntu.16.04-x64;centos.7-x64;rhel.7.2-x64;debian.8-x64;fedora.23-x64;opensuse.13.2-x64;osx.10.11-x64;osx-x64;osx-arm64;linux-x64;linux-arm64</ToolsServiceTargetRuntimes>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <!-- This is required for IDE0005 to fail the build https://github.com/dotnet/roslyn/issues/41640 -->
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <Nullable>enable</Nullable>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TF_BUILD)' == 'true' Or '$(GITHUB_ACTIONS)' == 'true'">
```
Matteo-T, line 42: ><PropertyGroup Condition="'$(TF_BUILD)' == 'true' Or '$(GITHUB_ACTIONS)' == 'true'"> [](http://example.com/codeflow?start=2&length=84)



Probably obvious for people who are in this repo all day long; but for newcomers it is black magic.

src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj
```diff
@@ -8,14 +8,15 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
     <EnableDefaultNoneItems>false</EnableDefaultNoneItems>
     <EmbeddedResourceUseDependentUponConvention>false</EmbeddedResourceUseDependentUponConvention>
-    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <DefineConstants>$(DefineConstants);NETCOREAPP1_0;TRACE</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PreserveCompilationContext>true</PreserveCompilationContext>
     <DebugType>portable</DebugType>
     <RuntimeIdentifiers>$(ToolsServiceTargetRuntimes)</RuntimeIdentifiers>
 	<!-- TODO FIX THESE WARNINGS ASAP -->
```
Matteo-T, line 16: >ASAP [](http://example.com/codeflow?start=30&length=4)

promises... promises... :)

Matteo-T (approved):
:shipit:

src/Microsoft.SqlTools.Hosting/Microsoft.SqlTools.Hosting.csproj
```diff
@@ -26,10 +26,10 @@
 		<InternalsVisibleTo Include="Microsoft.SqlTools.ServiceLayer.UnitTests" />
 		<InternalsVisibleTo Include="Microsoft.SqlTools.ServiceLayer.IntegrationTests" />
 		<InternalsVisibleTo Include="Microsoft.SqlTools.ServiceLayer.Test.Common" />
-        <!-- Workaround for SRGen issues -->
+		<!-- Workaround for SRGen issues -->
```
shueybubbles, line 29: what srgen issues? It could be easy to fix.

brian-harris (approved):

src/Microsoft.SqlTools.Hosting/Microsoft.SqlTools.Hosting.csproj
Charles-Gagnon: *shrug* Been here forever https://github.com/microsoft/sqltoolsservice/pull/321

Probably something we can remove, I'll do a quick check after this PR to see if there's any issues.

src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj
Charles-Gagnon: Right? This is why I suggest people just file issues instead of leaving comments...

PR #2008, opened 2023-04-17
Error on unused private properties

Charles-Gagnon:
Add build error for unused private properties. This will help ensure that things aren't mistakenly being missed and that old/unused code is cleaned up. I noticed these as warnings in the build when I was adding the SourceLink stuff, so in an attempt to lower the number of warnings being spit out during a build promoting these to a full error seemed to make sense.

https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0169
https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0414

@caohai FYI, I'm seeing a lot of UI-based code/comments in the stuff you brought over from SSMS. It would be great to try and clean up anything you don't actually need when you're porting code so we don't end up with a bunch of unnecessary stuff cluttering up the code. Just something to try and keep in mind for the future.


src/Microsoft.SqlTools.ServiceLayer/Security/AppRoleGeneral.cs
```diff
@@ -38,40 +36,10 @@ internal class AppRoleGeneral
         //property by the initialization code
         private ServerConnection serverConnection;
 
+        #endregion
 
-        /// <summary>
-        /// execution mode by default for now is success
-        /// </summary>
-        private ExecutionMode m_executionMode = ExecutionMode.Success;
```
kburtram, line 45: I think @caohai is currently working on Roles, so it's possible he neither needs something here or has it already deleted in a feature branch.  Though from a quick pass these look unneeded to me.

kburtram (approved):
LGTM though it is cleaning up some code that is being actively worked on as part of a new feature so may conflict with what @caohai is already doing.

Charles-Gagnon:
Deleted a few commented-out code blocks you brought up @kburtram, but there's still lots more. I'll leave that for you two to clean up when you get a chance 😄

caohai:
`AppRoleGeneral.cs` and `DatabaseRoleGeneral.cs` are currently being worked on as we are trying to finish the related user management features for May release. I was planning to clean up the unused code in my next PR that will use ported code. I think it's okay if you remove some of those in this one.

src/Microsoft.SqlTools.ServiceLayer/Security/AppRoleGeneral.cs
caohai: Thanks @kburtram ! Yeah I'm working on this part and I was planning to remove unused code with my next PR.

PR #2012, opened 2023-04-17
Use major + minor from pipeline for assembly versioning

Charles-Gagnon:
Use the same version we use for tagging the release for the build version.

The versioning stuff was originally added for ManagedBatchParser to use, and since it doesn't really make sense to tie the version of that to the STS version I created two separate version numbers. 

If anyone knows offhand a cleaner way to do this I'm open to suggestions - I'm sure there's a better way. But I didn't want to spend a bunch of time refactoring this for what is a pretty minor thing so I'm happy enough with at least having the major + minor version match. The build version is generated by the release script so would require more work there to pull out and use for the build as well.

In addition I changed the default version to 99.99.99.99 - this way it's easy to tell when a "dev" version of the library is being used. 

Assembly attributes after : 

image

image

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -12,9 +12,9 @@
     <RootDir>$(MSBuildThisFileDirectory)</RootDir>
 
     <!-- Defaults-->
-    <Major Condition="$(Major) == ''">1</Major>
-    <Minor Condition="$(Minor) == ''">0</Minor>
-    <Build_BuildNumber Condition="'$(Build_BuildNumber)' == ''">0.0</Build_BuildNumber>
+    <Major Condition="$(Major) == ''">99</Major>
+    <Minor Condition="$(Minor) == ''">99</Minor>
+    <Build_BuildNumber Condition="'$(Build_BuildNumber)' == ''">99.99</Build_BuildNumber>
     <StableRelease Condition="$(StableRelease) == ''">false</StableRelease>
 
     <!-- AssemblyVersion should not change for non-major releases. -->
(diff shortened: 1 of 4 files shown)


Charles-Gagnon:
@erpett FYI in case you need to update docs, the location of the version number to bump post-release has changed slightly. It'll be the numbers in build-and-release.yml now.

Directory.Build.props
```diff
@@ -12,9 +12,9 @@
     <RootDir>$(MSBuildThisFileDirectory)</RootDir>
 
     <!-- Defaults-->
-    <Major Condition="$(Major) == ''">1</Major>
-    <Minor Condition="$(Minor) == ''">0</Minor>
-    <Build_BuildNumber Condition="'$(Build_BuildNumber)' == ''">0.0</Build_BuildNumber>
+    <Major Condition="$(Major) == ''">99</Major>
+    <Minor Condition="$(Minor) == ''">99</Minor>
+    <Build_BuildNumber Condition="'$(Build_BuildNumber)' == ''">99.99</Build_BuildNumber>
```
cheenamalhotra, line 17: I'd prefer to keep it 1.0.0.0 by default unless overridden.. otherwise local build produces 99.99.99.99 version of nuget packages, which is unusual... I think industry-wide default is usually 1.0.0.0

Charles-Gagnon: I don't think there's an industry standard for dev/local versions. And 1.0.0.0 is too easily confused with actual released versions - especially given that prior to my changes our actual official released versions all had an assembly version of 1.0.0.0. So I'd prefer to keep it as 99.99.99.99 since that's extremely clear that it's not a "real" released version.

Charles-Gagnon:
I'll be merging this in by EOD unless there's any blocking issues brought up.

PR #2015, opened 2023-04-19
Simplify Object Management APIs

alanrenmsft:
Use one set of APIs for all object types.
1. Initialize View
2. Save
3. Drop
4. Rename
5. Dispose View
6. Script

ADS side changes: https://github.com/microsoft/azuredatastudio/pull/22781


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Credential/CredentialHandler.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.SqlTools.ServiceLayer.Management;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// Credential object type handler
+    /// </summary>
+    public class CredentialHandler : ObjectTypeHandler
+    {
+        public CredentialHandler(ConnectionService connectionService) : base(connectionService)
```
kburtram, line 20: This class was built a while back to a different protocol as user/login, so not sure if it will work with updates, etc.  We can consider deleting if we aren't going to build UI for this object type in the near future.

kburtram (approved):
I'm not sure I'd call this a "simplification" 😄 but looks like an improved design.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Credential/CredentialHandler.cs
alanrenmsft: for credential object type, I only implemented the create/drop operations that are needed for the integration tests.

alanrenmsft:
> I'm not sure I'd call this a "simplification" 😄 but looks like an improved design.

It takes a lot of changes to simplify things :)

alanrenmsft:
@Charles-Gagnon  @corivera  FYI.

PR #2017, opened 2023-04-19
Move pipeline version vars

Charles-Gagnon:
Move up the version vars to the global level so they actually get passed to the Release stage and can be used there. 

I did an ad-hoc test where I modified the release stage to just print out the values and confirmed it seemed to work there - but this isn't something we'll be able to fully verify until we run an actual release. I considered adding a "dry run" parameter to the release scripts, but given how infrequently we'd need to actually use such a feature it didn't seem worth the effort currently.

### DIFF
--- a/azure-pipelines/build-and-release.yml
+++ b/azure-pipelines/build-and-release.yml
@@ -10,44 +10,44 @@ variables:
   value: '7'
 
 stages:
-# - stage: Build
-#   variables:
-#   - name: buildConfiguration
-#     value: 'Release'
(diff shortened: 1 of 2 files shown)


alanrenmsft (approved):

Charles-Gagnon:
@erpett I updated the instructions in the release pipeline with the new steps. It's mostly the same - although I did remove the part to bump up the patch (3rd) version number. Since we bump the minor version every time we release there shouldn't be need to have the patch be bumped too.

PR #2020, opened 2023-04-20
Pass in patch version from pipeline

Charles-Gagnon:
Currently the release script expects that the versions for each branch are unique for the first 3 numbers (x.x.x). So the main branch will have a version like 4.7.0 while the release branch that is cut off needs to be bumped to 4.7.1 - otherwise it will reset back to a revision number of 0 (4.7.0.0) and cause conflicts with the already-released versions from main.

In order to avoid having to vBump two separate numbers during a release - while still also keeping the assembly version number in line with this - I've added a variable for the patch number and pass that in to the build. 

So this will bring this back in line with the current process of updating the minor version (4.x) in the main branch after a release version is cut off, and updating the patch version (4.7.x) in the release branch.


Directory.Build.props
```diff
@@ -23,7 +30,7 @@
     <!-- AssemblyFileVersion should change for every build. -->
     <!-- For preview releases, sample Version = 3.0.20221104.1-preview -->
     <!-- For stable releases, sample Version = 3.0.0 -->
-    <VersionPrefix>$(Major).$(Minor).$(Build_BuildNumber)</VersionPrefix>
+    <VersionPrefix>$(Major).$(Minor).$(Patch).$(Revision)</VersionPrefix>
```
Charles-Gagnon, line 33: Using the ADO build number like this was originally done for generating the ManagedBatchParser nuget package - but that number has nothing to do with the STS builds we generate so I'm changing that here. I'll be following up with a separate PR for getting the ManagedBatchParser back to using the ADO build number itself since that doesn't need to follow the STS versioning scheme.

erpett (approved):

Charles-Gagnon:
@erpett Release pipeline steps updated as well

PR #2025, opened 2023-04-20
Removed advanced options from table designer title

smartguest:
In response to feedback and reflection, I have decided to remove the advanced options from the table designer as it's generally unnecessary in the context it is used in. (Edit, fixed it so that null advanced options won't throw an error anymore)


kburtram (approved):
We shouldn't be showing advanced options in the UI unless they are needed to disambiguate similar connections.  User can review the advanced options in the Connection Dialog, if they need to get that context for some reason.

PR #2030, opened 2023-04-24
[User Management] Implement handlers for app role, db role and server role

caohai:
1. Implement handlers for roles
2. Securable search and permission setting are not in this PR.

I'll leaning towards merging this bulk one first then do more e2e testing later and send fixes for any issues.


alanrenmsft (approved):
add integration test?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/ServerRole/ServerRoleInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// a class for storing various server role properties
+    /// </summary>
+    public class ServerRoleInfo : SqlObject
+    {
+        public string? Owner { get; set; }
```
alanrenmsft, line 13: this is missing the owner type information, it could be server role or login

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/ServerRole/ServerRoleInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+        public string[]? Members { get; set; }
```
alanrenmsft, line 14: same here, the members could be login or server role

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/ServerRole/ServerRoleViewContext.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    public class ServerRoleViewContext : SqlObjectViewContext
+    {
+        public ServerRoleViewContext(Contracts.InitializeViewRequestParams parameters) : base(parameters)
+        {
+        }
+
+        public override void Dispose()
+        {
```
alanrenmsft, line 15: nothing to be disposed?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/ServerRole/ServerRoleInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+        public string[]? Memberships { get; set; }
+        public bool IsFixedRole { get; set; }
```
alanrenmsft, line 16: you can only add/remove members on fixed role, right?

alanrenmsft: I'd recommend to move this to the view info since we are never going to set it in the UI.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/ServerRole/ServerRoleInfo.cs
caohai: Yes that's correct, something like this will just work
```
    var serverRole = new ServerRole(srv, "cliRole");
    serverRole.Owner = "login name or server role name";
    serverRole.Create();
```

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/ServerRole/ServerRoleViewContext.cs
caohai: For `ServerRole`, I didn't create a new connection (same behavior as Login) so I didn't dispose it here.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/ServerRole/ServerRoleInfo.cs
caohai: only add/remove members on fixed role - Yes
Makes sense I'll move it to view info.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/DatabaseRole/DatabaseRoleInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// a class for storing various database role properties
+    /// </summary>
+    public class DatabaseRoleInfo : SqlObject
+    {
+        public string? Owner { get; set; }
+        public string[]? SchemasOwned { get; set; }
```
alanrenmsft, line 14: rename to OwnedSchemas to be consistent with the same property in User.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/AppRole/AppRoleInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// a class for storing various application role properties
+    /// </summary>
+    public class AppRoleInfo : SqlObject
+    {
+        public string? DefaultSchema { get; set; }
+        public string? Password { get; set; }
+        public string[]? SchemasOwned { get; set; }
```
alanrenmsft, line 15: OwnedSchemas

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/DatabaseRole/DatabaseRoleViewInfo.cs
```diff
@@ -0,0 +1,15 @@
...
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// a class for storing various database role view properties
+    /// </summary>
+    public class DatabaseRoleViewInfo : SqlObjectViewInfo
+    {
+        public string[]? Schemas { get; set; }
```
alanrenmsft, line 13: we should also have an IsFixedRole property, right?

alanrenmsft: also need following properties for the owner picking experience to work.
1. Users
2. Database Roles
3. Application Roles

alanrenmsft: actually, I'll use the object search API to get these objects.

caohai: IsFixedRole - Yes, I missed this one, will add.
Yes I was also planning to do that via the obj search API

kburtram (approved):

alanrenmsft:
@caohai I've fixed a bunch of issues encountered during the integration testing, please follow up on the following things:
1. performance - please review and make sure things are optimized.
2. update password for application role - I didn't add the compare with fake password logic.

caohai:
> @caohai I've fixed a bunch of issues encountered during the integration testing, please follow up on the following things:
> 
> 1. performance - please review and make sure things are optimized.
> 2. update password for application role - I didn't add the compare with fake password logic.

@alanrenmsft Got it, thanks for picking this up!

1. I'll take a look at this today, I saw Karl fixed connection related issue in User, I'll go over the change and follow.
2. IIRC the fake pwd comparing logic is already there but I can double check.

PR #2034, opened 2023-04-28
Fix application names to respect connection's appname property

cheenamalhotra:


src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -1798,7 +1799,7 @@ private Dac.TableDesigner CreateTableDesigner(TableInfo tableInfo)
             {
                 var connectionStringBuilder = new SqlConnectionStringBuilder(tableInfo.ConnectionString);
                 connectionStringBuilder.InitialCatalog = tableInfo.Database;
-                connectionStringBuilder.ApplicationName = TableDesignerService.TableDesignerApplicationName;
```
alanrenmsft, line 1801: I wasn't sure why do we need to have different app name for different features, since you are working on this, maybe we should make it consistent across STS?
1. if user specified the app name, then use it
2. if not, we use a default app name: azdata.

cheenamalhotra: We still need to append 'azdata' at least otherwise we lose ADS telemetry from server side connection activity.

Feature identifiers are debatable - they do help track feature based connection activity, but we'll have to take them away eventually when we enable connection pooling to truly experience benefits of pooled connections.

alanrenmsft (approved):

PR #2036, opened 2023-04-28
change user type implementation

alanrenmsft:
Remove Authentication type from user and update user type

user-type


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Security/UserActions.cs
```diff
@@ -70,23 +70,19 @@ internal static ExhaustiveUserTypes GetUserTypeForUserInfo(UserInfo user)
...
-                case DatabaseUserType.Contained:
-                    if (user.AuthenticationType == ServerAuthenticationType.AzureActiveDirectory)
-                    {
-                        userType = ExhaustiveUserTypes.ExternalUser;
-                    }
-                    else
-                    {
-                        userType = ExhaustiveUserTypes.SqlUserWithPassword;
-                    }
+                case DatabaseUserType.SqlAuthentication:
```
kburtram, line 79: I think this is cleaner than the previous mappings.

kburtram (approved):

PR #2037, opened 2023-04-29
Remove extra connection in app role and db role

caohai:
Similar to #2033


alanrenmsft:
please take a look at the build failures. @caohai

alanrenmsft (approved):

kburtram (approved):
LGTM, the previous pattern was opening a "general" connection then immediately opening a "data container" connection.  That was causing a bit of the delay we were seeing opening the user dialog.

PR #2038, opened 2023-05-02
Query logins on system database for Azure connections

kburtram:
This fixes issue https://github.com/microsoft/azuredatastudio/issues/22897.  The query on sys.sql_logins fails for user databases in SQL DB, so this query needs to run on a system DB in that context.  The connection to master is only done if needed to avoid the perf cost of the extra connection.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/User/UserHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/User/UserHandler.cs
@@ -177,6 +177,30 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
             }
             supportedUserTypes.Add(DatabaseUserType.NoLoginAccess);
 
+            string[] logins = DatabaseUtils.LoadSqlLogins(dataContainer.ServerConnection);
+
+            // If we couldn't load logins on current connection and this is a SQL DB connection 
+			// not to master then try to connect to master.  The "sys.sql_logins" DMV is not visible to user databases.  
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/User/UserHandler.cs
```diff
@@ -177,6 +177,30 @@ public override async Task<InitializeViewResult> InitializeObjectView(Contracts.
...
             supportedUserTypes.Add(DatabaseUserType.NoLoginAccess);
 
+            string[] logins = DatabaseUtils.LoadSqlLogins(dataContainer.ServerConnection);
+
+            // If we couldn't load logins on current connection and this is a SQL DB connection 
+			// not to master then try to connect to master.  The "sys.sql_logins" DMV is not visible to user databases.  
+            if (logins.Length == 0 && isSqlAzure 
+                && string.Compare(parameters.Database, "master", true) != 0)
+            {
+                ServerConnection orgServerConnection = null;
```
alanrenmsft, line 187: `orgServerConnection`, do you mean `masterDBConnection`?

alanrenmsft (approved):

caohai (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/User/UserHandler.cs
kburtram: yeah, I was just connecting to the originally provided DB but later switched to use master to be more explict.  I'll change the name.

PR #2043, opened 2023-05-05
Fixing durability type filter in OE

aasimkhan30:
Now, I am passing both display name and value. As the value used in SFC filter is different from display value for the filter dropdown.


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
```diff
@@ -335,12 +335,16 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectExplorer.SmoModel
...
+																		foreach (XmlElement choice in choiceValues)
 																		{
-																				var enumValueName = enumValue.GetAttribute("Name");
-																				WriteLine("							\"{0}\",", enumValueName);
+																				var choiceName = choice.GetAttribute("LocLabel");
+																				var choiceValue = choice.GetAttribute("Value");
+																				WriteLine("							new NodeFilterPropertyChoice {");
+																				WriteLine("								DisplayName = {0},", choiceName);
+																				WriteLine("								Value = \"{0}\",", choiceValue);
+																				WriteLine("							},");
```
lewis-sanchez, line 347: Are .tt files auto generated? This formatting is awful.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.cs
```diff
@@ -1361,13 +1379,19 @@ protected override void OnExpandPopulateFolders(IList<TreeNode> currentChildren,
...
 					{
-						Name = "DurabilityType",
+						Name = "Durability",
 						DisplayName = SR.FilterDurabilityType,
 						Type = NodeFilterPropertyDataType.Choice,
 						Description = SR.FilterDurabilityTypeDescription,
-						Choices = new string[] {
-							"SchemaAndData",
-							"SchemaOnly",
+						Choices = new NodeFilterPropertyChoice[] {
```
lewis-sanchez, line 1386: Should we move this array instantiation to a separate method and just reference that to avoid having to change things in 4 different locations if choices are added or removed in the future?

aasimkhan30: This is generated code. So repetitions don't matter much.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoTreeNodes.tt
aasimkhan30: Tt files are not auto generated. This is the file that generates the source in other file.

lewis-sanchez (approved):

cheenamalhotra (approved):

PR #2045, opened 2023-05-07
Updating DacFx.Projects to bring in UpdateSqlCmdVariable fix

Benjin:


src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
```diff
@@ -531,9 +531,8 @@ internal async Task HandleUpdateSqlCmdVariableRequest(AddSqlCmdVariableParams re
         {
             await RunWithErrorHandling(() =>
             {
-                SqlProject project = GetProject(requestParams.ProjectUri, onlyLoadProperties: true);
-                project.SqlCmdVariables.Delete(requestParams.Name); // idempotent (won't throw if doesn't exist)
-                project.SqlCmdVariables.Add(new SqlCmdVariable(requestParams.Name, requestParams.DefaultValue));
+                SqlProject project = GetProject(requestParams.ProjectUri!, onlyLoadProperties: true);
+                project.SqlCmdVariables.Update(requestParams.Name, requestParams.DefaultValue); // won't throw if doesn't exist
```
Benjin, line 535: `Delete(); Add();` broke when the target variable was in use by a database reference because `Delete()` would throw to prevent a dangling reference.  New dedicated `Update()` function handles that case.

kisantia (approved):

PR #2054, opened 2023-05-10
Remove Localization from Empty Strings

smartguest:
This code fixes the part of the codeql check that fails because a couple of the projects do not have localized strings which leads to issues with the task. 

However, now there is a failure with the buildtests which I'm not sure what is causing it (there is no issue when building locally). More investigation is needed as to what is causing it to fail.


smartguest:
By the way this isn't caused by the Lego Check-In Specifically, it has always been this way since Code Scanning was first run. Something about our code isn't suited for Code Scanning.

PR #2059, opened 2023-05-11
Adding filtering for group by schema

aasimkhan30:


cheenamalhotra (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/NodeInfo.cs
```diff
@@ -142,7 +142,7 @@ public class NodeFilter
         /// <summary>
         /// The name of the filter property
         /// </summary>
-        public string DisplayName { get; set; }
+        public string Name { get; set; }
```
Charles-Gagnon, line 145: Make sure you call out additional changes like this in the PR description as well

Charles-Gagnon (approved):

PR #2061, opened 2023-05-12
Fix effective permission column field

caohai:
Fix effective permission column field


kburtram (approved):
Looks fine but confirm with @alanrenmsft if it will break the UI contracts.

caohai:
> Looks fine but confirm with @alanrenmsft if it will break the UI contracts.

@kburtram I chatted with Alan and we are going to hide the column info part for May.

alanrenmsft (approved):

PR #2071, opened 2023-05-18
Add object management handler for creating a database

corivera:
These are the backend changes required to support the Create Database dialog in ADS. Currently this handler supports the following properties when configuring a database:
1. Name
2. Owner
3. Collation
4. Compatibility Level
5. Containment Type

Further options will be added in subsequent PRs. Next I'll be working on the Configure SLO options for Azure SQL DBs.

To validate, I exported the sql scripts from the ADS Create Database dialog and compared them with the exported script in SSMS' Create Database dialog, and both scripts were always the same with whatever options I tried.


src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -2662,3 +2662,143 @@ Permission_ImpersonateAnyLogin = Impersonate Any Login
 
 ServiceProviderNotSet = SetServiceProvider() was not called to establish the required service provider
 ServiceNotFound(string serviceName) = Service {0} was not found in the service provider
+
+############################################################################
+# Create Database
```
corivera, line 2667: A bunch of localized strings were missing from the DatabasePrototype code, so I copied over every Create Database string from SSMS just to be safe. Previously we wouldn't even get messages for exceptions because the localized strings weren't defined.

alanrenmsft (approved):

caohai (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -26,17 +87,507 @@ public override bool CanHandleType(SqlObjectType objectType)
...
+                    ServerConnection serverConnection = dataContainer?.Server?.ConnectionContext;
+                    if (serverConnection != null && serverConnection.IsOpen)
+                    {
+                        serverConnection.Disconnect();
+                    }
+                }
+            }
+        }
+
+        private static string SanitizeFileName(string name)
```
Charles-Gagnon, line 293: Doc comments for helper functions are useful - there's a lot of different things "sanitize" could mean, so having the comment explain what it's sanitizing it for use for would be helpful.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -26,17 +87,507 @@ public override bool CanHandleType(SqlObjectType objectType)
...
+        private string[] PopulateCollationDropdownWithPrototypeCollation(DatabasePrototype prototype)
+        {
+            return new string[] { prototype.Collation };
+        }
+
+        private string[] PopulateContainmentTypeDropdown(CDataContainer dataContainer, DatabasePrototype prototype)
+        {
+            if (!(SqlMgmtUtils.IsSql11OrLater(dataContainer.Server.ServerVersion)) || IsAnyManagedInstance(dataContainer.Server))
+            {
+                return null;
```
Charles-Gagnon, line 381: Why null instead of empty array?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -26,17 +87,507 @@ public override bool CanHandleType(SqlObjectType objectType)
...
+            foreach (var level in this.displayCompatLevels.Keys)
+            {
+                if (level == prototype.DatabaseCompatibilityLevel)
+                {
+                    // Azure can't change the compat level so we only populate the current version
+                    return new string[] { this.displayCompatLevels[level] };
+                }
+            }
+
+            System.Diagnostics.Debug.Fail(string.Format(CultureInfo.InvariantCulture, "Unknown compat version '{0}'", prototype.DatabaseCompatibilityLevel));
```
Charles-Gagnon, line 494: This seems like it should be just a throw, not sure having Debug.Fail calls is actually useful (it likely just means that if there's ever an issue during release it'll just fail silently or elsewhere)

Charles-Gagnon (commented):
Please add some tests - if there's difficulties in writing tests for this stuff let's start a discussion how we can make that better.

@alanrenmsft @caohai You've been doing tests for your features too, right? Any suggestions or pointers for how best to approach testing Object Management functionality?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: A bunch of these methods were ported over from SSMS, so I was trying to maintain the same behavior.

Charles-Gagnon: That's fine to start with, but SSMS and STS are very different applications - so we should be doing what makes sense for STS where we can. Debug.Fail is only marginally useful even normally (especially in a situation like this where it's clearly indicating a runtime error has occurred and there isn't really a way to recover from it) so seems like an easy thing to start with replacing.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -109,9 +108,9 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
                         // azure sql db doesn't have a sysadmin fixed role
                         var compatibilityLevelEnabled = !isDw &&
                                                         (dataContainer.LoggedInUserIsSysadmin ||
-                                                        dataContainer.Server.ServerType ==
+                                                        dataContainer.Server!.ServerType ==
```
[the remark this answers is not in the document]
corivera: I added a null check in the method we create the data container in, so it'll throw if Server is null before hitting these non-null assertions.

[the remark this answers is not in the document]
corivera: Added null checks and exception throws instead of using the non-null operator.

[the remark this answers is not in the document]
Charles-Gagnon: Works for me, thanks

corivera:
@Charles-Gagnon added some test cases and re-opened the PR

ssreerama:
A small request, if you already have a sample UI generated in your local with your ADS code changes, can you add the screenshot to the description, please? It would be helpful to picture how the UI and these changes are related. Thanks

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
```diff
@@ -11341,4950 +11341,6457 @@ public static string ServiceProviderNotSet
...
         {
-            return Keys.GetString(Keys.ConnectionServiceListDbErrorNotConnected, uri);
+            get
+            {
+                return Keys.GetString(Keys.autogrowth_dialog_title);
+            }
         }
 
-        public static string ConnectionServiceDbErrorDefaultNotConnected(string uri)
+        public static string autogrowth_dialog_filestreamtitle
```
ssreerama, line 11352: Is there a reason for the name to be all in small letters? Instead of using a camel case? Thanks

corivera: This file is auto generated (all the SR ones are), so I think git gets a bit confused when rendering it. This string still exists lower in the file.

corivera:
@ssreerama Here's a screenshot of the ADS UI, which was already merged in a while back. It just needs to be enabled after merging these changes.

NewDatabaseDialog

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -5,8 +5,19 @@
 
 using System;
```
ssreerama, line 6: Just a suggestion, in next visit of these files, could we arrange the namespaces alphabetically starting with 's'.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
corivera: I imported all of these strings from SSMS, so that's the convention that gets used there. I kept the casing for the string names to prevent issues when we import SSMS code in the future, so that we wouldn't need to convert every resource string name. Especially since a lot of resource manager calls in other SQL Tools files use a string literal for the resource name rather than an SR property, so there wouldn't be any compiler errors if a name was changed and not updated where it was used.

barbaravaldez (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -26,17 +82,512 @@ public override bool CanHandleType(SqlObjectType objectType)
...
+            }
+            return collationItems.ToArray();
+        }
+
+        private string[] GetCollationsWithPrototypeCollation(DatabasePrototype prototype)
+        {
+            return new string[] { prototype.Collation };
+        }
+
+        private string[] GetContainmentTypes(Server server, DatabasePrototype prototype)
```
Charles-Gagnon, line 378: It would be great to add comments for methods like this - both as top level xml doc and within the method itself

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -26,17 +82,512 @@ public override bool CanHandleType(SqlObjectType objectType)
...
+        {
+            if (!(SqlMgmtUtils.IsSql11OrLater(server.ServerVersion)) || IsAnyManagedInstance(server))
```
Charles-Gagnon, line 380: Would be good to add comments for why stuff like this is done

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -0,0 +1,193 @@
...
+                try
+                {
+                    var parametersForCreation = ObjectManagementTestUtils.GetInitializeViewRequestParams(connectionResult.ConnectionInfo.OwnerUri, "master", true, SqlObjectType.Database, "", "");
+                    var script = await ObjectManagementTestUtils.ScriptObject(parametersForCreation, testDatabase);
+                    Assert.True(!databaseExists(testDatabase.Name!, server), $"Database should not have been created for scripting operation");
+                    Assert.True(script.ToLowerInvariant().Contains($"create database [{testDatabase.Name!.ToLowerInvariant()}]"));
+                }
+                finally
+                {
+                    // Cleanup database on the off-chance that scripting somehow created the database
```
Charles-Gagnon, line 100: Shouldn't that be a part of the test to verify that it doesn't? That'd be really bad if it was somehow actually creating the database.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -0,0 +1,193 @@
...
+                    Assert.True(ex.InnerException.InnerException is SqlException, $"Received unexpected inner-inner exception type: {ex.InnerException.InnerException!.GetType()}");
+                }
+                finally
+                {
+                    dropDatabase(server, testDatabase.Name!);
+                }
+            }
+        }
+
+        private bool databaseExists(string dbName, Server server)
```
Charles-Gagnon, line 165: nit: Method names should be PascalCased (here and below)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
corivera: I already have a verify check for that in the try block above.

Charles-Gagnon: Ah, ok I missed that. Maybe update the comment a bit to make it clear that this would only happen if the test fails and the database is created. Not a huge deal though.

PR #2074, opened 2023-05-19
Load TrustServerCertificate value from connection settings when running integration tests.

corivera:
I was seeing some connection failures against a local sql express instance because we weren't loading the TrustServerCertificate value from the connection profile.


Charles-Gagnon (approved):
@cheenamalhotra Do you know why this seems to be working just fine in the integration tests pipeline? That's using a docker image so I'd expect that it would require this to succeed.

PR #2080, opened 2023-05-31
Enable connection pooling for languageService

cheenamalhotra:
Discussed offline, allowing connection pooling for languageService feature addresses https://github.com/microsoft/azuredatastudio/issues/22970 

_(exec sp_reset_connection before Login is an indication of a pooled connection reset for use)_

### DIFF
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/Constants.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/Constants.cs
@@ -16,6 +16,9 @@ namespace Microsoft.SqlTools.Hosting.Protocol
         public static readonly string DefaultApplicationName = "azdata";
         public static readonly string SqlLoginAuthenticationType = "SqlLogin";
 
+        // Feature names
+        public static readonly string languageServiceFeature = "languageService";
+
         static Constants()
         {
             JsonSerializerSettings = new JsonSerializerSettings();
--- a/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/Constants.cs
+++ b/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/Constants.cs
@@ -17,7 +17,7 @@ namespace Microsoft.SqlTools.Hosting.Protocol
         public static readonly string SqlLoginAuthenticationType = "SqlLogin";
 
         // Feature names
-        public static readonly string languageServiceFeature = "languageService";
+        public static readonly string LanguageServiceFeature = "languageService";
 
         static Constants()
         {
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -1329,7 +1329,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         /// <summary>
         /// Build a connection string from a connection details instance
         /// </summary>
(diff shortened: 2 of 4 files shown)


src/Microsoft.SqlTools.Hosting/Hosting/Protocol/Constants.cs
```diff
@@ -16,6 +16,9 @@ public static class Constants
         public static readonly string DefaultApplicationName = "azdata";
         public static readonly string SqlLoginAuthenticationType = "SqlLogin";
 
+        // Feature names
+        public static readonly string languageServiceFeature = "languageService";
```
kburtram, line 20: nit: following naming convention of other public constants this should be `LanguageServiceFeature`.

kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1330,16 +1330,16 @@ public static bool IsDedicatedAdminConnection(ConnectionDetails connectionDetail
         /// Build a connection string from a connection details instance
         /// </summary>
         /// <param name="connectionDetails"></param>
-        public static string BuildConnectionString(ConnectionDetails connectionDetails)
+        public static string BuildConnectionString(ConnectionDetails connectionDetails, bool forceDisablePooling = true)
         {
```
shueybubbles, line 1334: do you not having any rules about breaking changes and preserving existing contracts?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1833,17 +1836,23 @@ public static SqlConnection OpenSqlConnection(ConnectionInfo connInfo, string fe
                 bool? originalPersistSecurityInfo = connInfo.ConnectionDetails.PersistSecurityInfo;
                 bool? originalPooling = connInfo.ConnectionDetails.Pooling;
 
+                // allow pooling connections for language service feature to improve intellisense connection retention and performance.
+                bool shouldForceDisablePooling = featureName != Constants.LanguageServiceFeature;
```
shueybubbles, line 1840: >bool [](http://example.com/codeflow?start=16&length=4)

`var` is your friend! Maybe other folks in ADS disagree though

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
kburtram: probably this method should be `internal` but we really don't have API-type scenarios for these services, so typically if the project compiles than it is backward compatible with all its consumers.  Though this isn't always the case so some caution is warranted.

shueybubbles (approved):
:shipit:

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: If type is known, I'd rather use it - strong typings help will nullables, var doesn't :)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: Not in here, since this is internal only. This is however a non-breaking API change.

PR #2083, opened 2023-06-01
Stops OE from breaking when paused DBs are expanded

lewis-sanchez:
This PR fixes: https://github.com/microsoft/azuredatastudio/issues/22211

Previously the root server node in Object Explorer would disconnect and collapse when an attempt to expand a paused DB was expanded. The new behavior renders an error node explaining to the user that the paused DB wasn't able to be expanded.
image


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -453,7 +453,24 @@ internal ExpandResponse QueueExpandNodeRequest(ObjectExplorerSession session, st
                                else
                                {
                                    Logger.Verbose($"Expanding {nodePath}");
-                                   nodes = node.Expand(cancelToken, securityToken?.Token, filters).Select(x => x.ToNodeInfo()).ToArray();
+                                   try
+                                   {
+                                       nodes = node.Expand(cancelToken, securityToken?.Token, filters).Select(x => x.ToNodeInfo()).ToArray();
+                                   }
+                                   catch (ConnectionFailureException ex)
```
Charles-Gagnon, line 460: Any specific reason you're only catching ConnectionFailureExceptions? Seems like we should be catching any exception and displaying that to the user.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -453,7 +453,24 @@ internal ExpandResponse QueueExpandNodeRequest(ObjectExplorerSession session, st
...
                                {
                                    Logger.Verbose($"Expanding {nodePath}");
-                                   nodes = node.Expand(cancelToken, securityToken?.Token, filters).Select(x => x.ToNodeInfo()).ToArray();
+                                   try
+                                   {
+                                       nodes = node.Expand(cancelToken, securityToken?.Token, filters).Select(x => x.ToNodeInfo()).ToArray();
+                                   }
+                                   catch (ConnectionFailureException ex)
+                                   {
+                                       Logger.Error($"Failed to expand node: {ex.InnerException.Message}");
```
Charles-Gagnon, line 462: Are we guaranteed to have an InnerException?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -453,7 +453,24 @@ internal ExpandResponse QueueExpandNodeRequest(ObjectExplorerSession session, st
...
+                                       var errorNode = new NodeInfo()
```
Charles-Gagnon, line 463: I would suggest making a helper function for creating an error node to avoid copy/paste errors if we add more in the future. Put it as a static method on the NodeInfo class.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
lewis-sanchez: No, it's hard to guarantee that we're always going to have an inner exception since the exception is being thrown from external code. That external exception is caught in the snippet linked below and a second one is thrown when the `context?.Database` is accessed.

https://github.com/microsoft/sqltoolsservice/blob/a12244f3d0c1377df8ee30fc36a12781ac547677/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/DatabaseTreeNode.cs#LL87C1-L87C1

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -491,6 +501,27 @@ internal ExpandResponse QueueExpandNodeRequest(ObjectExplorerSession session, st
             return response;
         }
 
+        /// <summary>
+        /// Creates an error node
+        /// </summary>
+        /// /// <param name="parentNodePath">Path to parent node that triggered the error.</param>
+        /// <param name="errorMessage">Message shown to the user</param>
+        /// <returns>NodeInfo instance with the specified parent path and error message</returns>
+        private NodeInfo CreateErrorNode(string parentNodePath, string errorMessage)
```
Charles-Gagnon, line 510: I'd suggest just making an ErrorNodeInfo subclass - easier to find/reuse

cheenamalhotra (approved):

PR #2084, opened 2023-06-01
Fix schema level search and refactor search code for user management

caohai:
https://github.com/microsoft/azuredatastudio/issues/23241
1. Fix schema level object search
2. Refactor search code path to make it more logical

image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
@@ -12,6 +12,8 @@ using Microsoft.SqlTools.ServiceLayer.ObjectManagement.Contracts;
 using System.Collections.Generic;
 using System.Collections.Concurrent;
 using Microsoft.SqlTools.ServiceLayer.Management;
+using System.Linq;
+using Microsoft.SqlTools.ServiceLayer.ObjectManagement.PermissionsData;
 
 namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
```diff
@@ -142,30 +145,43 @@ internal async Task HandleSearchRequest(SearchRequestParams requestParams, Reque
...
+                // only schema-Scoped securableTypes support schema level search
+                if (!string.IsNullOrEmpty(requestParams.Schema) && !schemaTypes.Contains(searchableObjectType))
+                {
+                    continue;
+                }
 
-                if (requestParams.SearchText != null)
+                SearchableObjectTypeDescription desc = SearchableObjectTypeDescription.GetDescription(searchableObjectType);
+                
+                if (desc.IsDatabaseObject)
```
alanrenmsft, line 156: there are so many branches here, I think this can be refactored to just one SearchableObject.Search() call.

alanrenmsft (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
caohai: These branches are by nature, and we are still going to have similar branches in the refactored function (ultimately we need to construct different URN strings for each branch). If we want to keep code in objectmanagementservice clean I can probably move this chunk of code to utils class later.

PR #2086, opened 2023-06-02
Replace <default> values in Owner and Collation fields for Create Database dialog

corivera:
We had a discussion about how the "\<default\>" values in the Owner and Collation dropdowns were not very clear. These changes add actual values in place of that default label, using the current connection's SQL login for Owner and the server's default collation for Collation as the defaults if the database doesn't already exist.

Fixes https://github.com/microsoft/azuredatastudio/issues/23306
Fixes https://github.com/microsoft/azuredatastudio/issues/23307


Charles-Gagnon (approved):
FYI, usually it's good not to close out the issue until you actually update STS in ADS (although you should still mention it so it gets linked properly)

barbaravaldez (approved):

PR #1064, opened 2020-08-31
Add "open xel file" support to profiler in sqltoolsservice

shueybubbles:
This update will enable ADS to implement XEL file open support. 
Key design tenets:

- An XEL file is treated as a regular XE session with a finite lifespan. When reading the file completes, the service sends a SessionStopped notification. 
- Don't break existing contracts for using ring buffer target.
- Lay the foundation for using the live target instead of the ring buffer for regular XE sessions in a future release.

PR #2091, opened 2023-06-07
Add "Open XEL file" support to profiler in sqltoolsservice

SakshiS-harma:
This update will enable ADS to implement XEL file open support.
Key design tenets:

An XEL file is treated as a regular XE session with a finite lifespan. When reading the file completes, the service sends a SessionStopped notification.
Don't break existing contracts for using ring buffer target.
Lay the foundation for using the live target instead of the ring buffer for regular XE sessions in a future release.

[Leveraging implementation from #1064 ]


kisantia:
Looks like the test `VerifyImportTableShouldReturnTrueGivenTableCreatdByTheService` is failing (and also has a typo in the name "Creatd")

src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// Start could be called again, but only new subscribers will see the data.
+        /// </summary>
+        public void Close()
+        {
+            cancellationTokenSource.Cancel();
+            var currentObservers = CurrentObservers;
+            currentObservers.ForEach(o => o.OnCompleted());
+            lock (syncObj)
+            {
+                currentObservers.ForEach(o => observers.Remove(o));
```
shueybubbles, line 90: > currentObservers.ForEach(o => observers.Remove(o)); [](http://example.com/codeflow?start=15&length=52)

does this work?
```C#
currentObservers.ForEach(observers.Remove);
```

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -1,420 +1,505 @@
...
-
-            // get session names from the session list
-            List<string> results = store.Sessions.Aggregate(new List<string>(), (result, next) =>
+        /// <summary>
+        /// Handle request to stop a profiling session
+        /// </summary>
+        internal async Task HandleStopProfilingRequest(StopProfilingParams parameters, RequestContext<StopProfilingResult> requestContext)
+        {
+            await Task.Run(async () =>
             {
```
shueybubbles, line 225: Pls remind me why these tasks need to use Task.Run explicitly. Why is the use of `await` in the body of the method insufficient?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -1,420 +1,505 @@
...
+                                break;
+                            }
+                            catch (InvalidOperationException)
+                            {
+                                remainingAttempts--;
+                                if (remainingAttempts == 0)
+                                {
+                                    throw;
+                                }
+                                Thread.Sleep(500);
```
shueybubbles, line 251: > Thread.Sleep(500); [](http://example.com/codeflow?start=31&length=19)

maybe the Sleeps should be replaced with `await Task.Delay(500)` now

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -1,201 +1,336 @@
...
+                try
+                {
+                    var targetXml = XEventSession.GetTargetXml();
+
+                    XmlDocument xmlDoc = new XmlDocument();
+                    xmlDoc.LoadXml(targetXml);
+
+                    var nodes = xmlDoc.DocumentElement.GetElementsByTagName("event");
+                    var rawEvents = nodes.Cast<XmlNode>().Select(ParseProfilerEvent).ToList();
+                    FilterOldEvents(rawEvents);
```
shueybubbles, line 257: >     FilterOldEvents(rawEvents); [](http://example.com/codeflow?start=15&length=32)

something to consider in a future improvement - find a way to filter the events without copying them all into a list and rewriting the list. It feels like we should be able to keep some minimal state that is updated as new events are read from the target. Maybe implement an IQueryable so it can be used as a Linq expression.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
```diff
@@ -1,201 +1,336 @@
...
+            if ((currentEvent.Name.Equals("sql_batch_completed")
+                || currentEvent.Name.Equals("sql_batch_starting"))
+                && currentEvent.Values.TryGetValue("batch_text", out string value))
+            {
+                return value.Contains("SELECT target_data FROM sys.dm_xe_session_targets")
+                    || currentEvent.Values["batch_text"].Contains("SELECT target_data FROM sys.dm_xe_database_session_targets");
+            }
+
+            return false;
+        }
```
shueybubbles, line 135: it'd be nice to remove this event filter eventually. The SSMS model is to use whatever events the session has defined; SSMS XEvent profiler will create the initial session, but the user is free to modify it as they like. We can't really presume that this limited set of events is adequate for a very broad range of investigations.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Profiler/ProfilerServiceTests.cs
```diff
@@ -14,72 +14,154 @@
...
+            var requestContext = new Mock<RequestContext<CreateXEventSessionResult>>();
+            requestContext.Setup(c => c.SendResult(It.IsAny<CreateXEventSessionResult>()))
+              .Returns<CreateXEventSessionResult>((result) => { return Task.FromResult(0); });
+            await profilerService.HandleCreateXEventSessionRequest(createParams, requestContext.Object);
+            return sessionName;
+        }
+
+
+
+#pragma warning disable JSON002 // Probable JSON string detected
```
shueybubbles, line 137: > JSON002 // Prob [](http://example.com/codeflow?start=23&length=16)

why does the compiler warn about a json string literal?

src/Microsoft.SqlTools.ServiceLayer/Profiler/IXEventSession.cs
```diff
@@ -1,35 +1,79 @@
...
-        /// Reads XEvent XML from the default session target
-        /// </summary>
-        string GetTargetXml();
+    public class SessionId
+    {
+        private readonly string sessionId;
+        // SQL Server starts session counters at around 64k, so it's unlikely that this process-scoped counter would collide with a real session id
+        // Eventually the profiler extension in ADS will use the string instead of the number.
+        private static int numericIdCurrent = -1;
+        /// <summary>
```
Benjin, line 54: Nit: leave blank lines between fields and the next field's comment

src/Microsoft.SqlTools.ServiceLayer/Profiler/IXEventSession.cs
```diff
@@ -1,35 +1,79 @@
...
-
-        /// <summary>
-        /// Reads XEvent XML from the default session target
-        /// </summary>
-        string GetTargetXml();
+    public class SessionId
+    {
+        private readonly string sessionId;
+        // SQL Server starts session counters at around 64k, so it's unlikely that this process-scoped counter would collide with a real session id
+        // Eventually the profiler extension in ADS will use the string instead of the number.
```
Benjin, line 52: How "eventually" is this "eventually"?  Should this change be made now?

README.md
```diff
@@ -104,13 +108,11 @@ so that your commits provide a good history of the changes you are making.  To b
 
 ### Add Unit Tests for New Code
 
-If you're adding a new feature to the project, please make sure to include adequate [xUnit](http://xunit.github.io/)
-tests with your change.  In this project, we have chosen write out unit tests in a way that uses the
-actual PowerShell environment rather than extensive interface mocking.  This allows us to be sure that
-our features will work in practice.  
+If you're adding a new feature to the project, please make sure to include adequate [nUnit](http://nunit.org/)
```
kisantia, line 111: did this change from xUnit to nUnit for the whole repo?

src/Microsoft.SqlTools.ServiceLayer/Profiler/Contracts/ProfilerSessionStoppedNotification.cs
```diff
@@ -1,25 +1,38 @@
...
+
+#nullable disable
+
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Profiler.Contracts
+{
+    public class ProfilerSessionStoppedParams
+    {
+        public string OwnerUri { get; set; }
```
kisantia, line 14: nit: doc comment for this one too

README.md
shueybubbles: When I wrote this code however many years ago I was annoyed with the crappy xunit asserts and was trying to push people to nunit.
I don't know if it took root when I dropped this PR.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
SakshiS-harma: This is merge mess up. Working on it.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSession.cs
SakshiS-harma: Yes, that is the next big item ticket that will be worked on (switching to live target).

README.md
SakshiS-harma: So, most of the services are using nUnit now, and just a few that are still using xUnit framework.  Going by numbers, it is about 2% of test cases using xUnit framework and rest is on nUnit.

src/Microsoft.SqlTools.ServiceLayer/Profiler/IXEventSession.cs
SakshiS-harma: My intention is to setup a basic workflow first for opening XEL files, and once the framework is there, work on the improvements. 
So, to your question this change should hopefully follow soon.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
SakshiS-harma: Yes, it works as expected.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
SakshiS-harma: Fixed this in next iteration.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
SakshiS-harma: Fixed.

src/Microsoft.SqlTools.ServiceLayer/Profiler/Contracts/ProfilerSessionStoppedNotification.cs
SakshiS-harma: Done.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
```diff
@@ -0,0 +1,186 @@
...
+    {
+        private IXEventFetcher iXEventFetcher;
+        private string filePath;
+        public XEventFetcher(string filePath)
+        {
+            this.filePath = filePath;
+        }
+        public Task ReadEventStream(HandleXEvent xeEventHandler, CancellationToken cancellationToken)
+        {
+            iXEventFetcher ??= new XEFileEventStreamer(filePath);
```
shueybubbles, line 176: >w XEFileEventStreamer(filePath); [](http://example.com/codeflow?start=33&length=32)

this class is coupled to files while ObservableXEventSession is meant to handle any source of extended events, not just files.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/XeStreamObservableTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                Assert.That(observer.ProfilerEvents.Last().Name, Is.EqualTo("sql_batch_completed"), "Last event in the file");
+
+            });
+        }
+
+        [Test]
+        public void XeStreamObservable_calls_OnError_when_the_fetcher_fails()
+        {
+            var observer = InitializeFileObserver("thispathdoesnotexist.xel");
+            Assert.Throws<FileNotFoundException>(observer.Observable.Start);    // exception handled in StartLocalFileSession task (ProfilerServic.cs)
```
shueybubbles, line 50: >tion>(observer.Observable.Start);  [](http://example.com/codeflow?start=43&length=34)

this behavior is not aligned with the Observable contract.
`Start` should always return an IObservable instance and not throw .

src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
shueybubbles, line 176: > new XEFileEventStreamer(filePath); [](http://example.com/codeflow?start=30&length=35)

when this constructor throws, catch it and return Task.FromException

shueybubbles (changes requested):
:clock1:

src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
```diff
@@ -0,0 +1,186 @@
...
+using Microsoft.SqlTools.ServiceLayer.Profiler.Contracts;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.SqlTools.ServiceLayer.Profiler
+{
+    /// <summary>
+    /// Wrapper XEventSession for XEventFetcher instances
```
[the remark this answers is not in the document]
SakshiS-harma: Addressed in latest iteration.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/XeStreamObservableTests.cs
SakshiS-harma: Addressed in latest iteration.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
SakshiS-harma: Addressed in latest iteration.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Profiler/ProfilerServiceTests.cs
SakshiS-harma: I am not sure why the compiler is warning against the use of json string. But it looked harmless, hence suppressed the warning.

Charles-Gagnon: We should always understand why we disable compiler warnings - just blindly disabling stuff in files doesn't help anyone. Either we should fix them or disable it in the editorconfig if it isn't useful. And when disabling please always include a reason why so we can have that context later on.  

In this case as far as I've been able to tell, this is warning that you don't have a comment like this telling the compiler that it's a JSON string

`/*lang=json,strict*/`

e.g. https://github.com/Azure/azure-functions-sql-extension/blob/main/test/Integration/SqlInputBindingIntegrationTests.cs#L128

Which lets it verify that the JSON is valid in the editor/at compile time. It's useful to have so I'd suggest adding it instead of disabling. 

(annoyingly I can't find any actual documentation on this whole lang comment stuff. The best I could find with a brief search was https://github.com/dotnet/docs/issues/30361...)

SakshiS-harma: I had the same problem, I couldn't find anything related to JSON002 error. But point noted.
Updated the code appropriately.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/XeStreamObservableTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                Assert.That(observer.ProfilerEvents.Count, Is.EqualTo(149), "Number of events read");
+                Assert.That(observer.ProfilerEvents[0].Name, Is.EqualTo("rpc_completed"), "First event in the file");
+                Assert.That(observer.ProfilerEvents.Last().Name, Is.EqualTo("sql_batch_completed"), "Last event in the file");
+
+            });
+        }
+
+        [Test]
+        public void XeStreamObservable_calls_OnError_when_the_fetcher_fails()
```
shueybubbles, line 50: >ble_calls_OnError_when_the_fetcher_fails( [](http://example.com/codeflow?start=35&length=41)

this test name doesn't match the code

shueybubbles: isn't the goal to make the original test work? observer.Completed should be true and observer.Error should be non-null

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -1,420 +1,464 @@
...
+        public IXEventSession OpenLocalFileSession(string filePath)
+        {
+            return new ObservableXEventSession(() =>
+            {
+                Task<IXEventFetcher> task1 = initIXEventFetcher(filePath);
+
+                var iXEventFetcher = task1.GetAwaiter().GetResult();
+                return iXEventFetcher;
+            }, new SessionId(filePath));
+        }
```
shueybubbles, line 208: this feels like too much `Task` management at this level.
Can we not just pass `() => new XEFileEventStreamer(filePath)` as the func, and have `XeStreamObservable.Start` handle the exception from the func, and then use `Task.FromException.ContinueWith` instead of calling `ReadEventStream.ContinueWith`?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/Profiler/XeStreamObservableTests.cs
SakshiS-harma: Updated the code to the original test.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
SakshiS-harma: Updated it appropriately.

shueybubbles (approved):
:shipit:

kisantia (approved):

PR #2093, opened 2023-06-08
Enabling database properties general tab with real time values from SMO

ssreerama:
This PR adds the ability to get the real time values of a database from smo and sending them back to ADS database properties view.

Images/SS:
generalTab


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +231,39 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
         {
-            ConnectionInfo connectionInfo = this.GetConnectionInfo(connectionUri);
+            ConnectionInfo connectionInfo = this.GetConnectionInfo(requestParams.ConnectionUri);
             CDataContainer dataContainer = CDataContainer.CreateDataContainer(connectionInfo, databaseExists: configAction != ConfigAction.Create);
             if (dataContainer.Server == null)
             {
                 throw new InvalidOperationException(serverNotExistsError);
             }
-            string objectUrn = (configAction != ConfigAction.Create && database != null)
+            string objectUrn = requestParams.IsNewObject ? ((configAction != ConfigAction.Create && database != null)
```
corivera, line 249: Since we're already using a ternary operator here, could this extra IsNewObject ternary operator be turned into a plain if case around the existing code?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -680,5 +701,15 @@ private static AzureEditionDetails[] GetAzureMaxSizes(IEnumerable<AzureEdition>
             }
             return sizes.ToArray();
         }
+
+        /// <summary>
+        /// Converts value in KBs to MBs with two decimal places
+        /// </summary>
+        /// <param name="valueInKb"></param>
```
corivera, line 708: Missing a parameter description here

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -120,6 +120,27 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
+                        // Collect the Database properties information
+                        if (!requestParams.IsNewObject)
+                        {
+                            var smoDatabaseProperties = dataContainer.SqlDialogSubject as Database;
+                            databaseViewInfo.ObjectInfo = new DatabaseInfo()
+                            {
+                                Name = smoDatabaseProperties.Name,
+                                CollationName = smoDatabaseProperties.Collation,
+                                DateCreated = smoDatabaseProperties.CreateDate.ToString(),
+                                LastDatabaseBackup = smoDatabaseProperties.LastBackupDate == DateTime.MinValue ? "None" : smoDatabaseProperties.LastBackupDate.ToString(),
```
corivera, line 132: Do any of the user facing strings like "None" need to be localized?

ssreerama: Need to check with other language installers, but do we have any localized files maintained in STS, or do declaring 'const string' variables automatically take care of localization?

corivera: Yes, new strings get added in sr.strings here: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings

When you do a full build the localization files get generated (and have to be checked in). You can then use localized strings with the SR class, like SR.BasicAzureEdition for the basic edition name for an Azure database, for example.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +231,43 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
-                    "Server");
+            string objectUrn = requestParams.ObjectUrn;
+            if (!requestParams.IsNewObject)
+            {
+                objectUrn = ((configAction != ConfigAction.Create && database != null)
+                    ? string.Format(System.Globalization.CultureInfo.InvariantCulture,
+                        "Server/Database[@Name='{0}']",
+                        Urn.EscapeString(database.Name))
+                    : string.Format(System.Globalization.CultureInfo.InvariantCulture,
+                        "Server"));
```
corivera, line 257: Still need to remove the extra outer layer of parentheses left over from the ternary operator.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +231,43 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
                 throw new InvalidOperationException(serverNotExistsError);
             }
-            string objectUrn = (configAction != ConfigAction.Create && database != null)
-                ? string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                    "Server/Database[@Name='{0}']",
-                    Urn.EscapeString(database.Name))
-                : string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                    "Server");
+            string objectUrn = requestParams.ObjectUrn;
+            if (!requestParams.IsNewObject)
```
corivera, line 250: Does Create Database still work with this change? We only do New Objects in that case.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -120,6 +120,27 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
+                                Name = smoDatabase.Name,
+                                CollationName = smoDatabase.Collation,
+                                DateCreated = smoDatabase.CreateDate.ToString(),
+                                LastDatabaseBackup = smoDatabase.LastBackupDate == DateTime.MinValue ? "None" : smoDatabase.LastBackupDate.ToString(),
+                                LastDatabaseLogBackup = smoDatabase.LastLogBackupDate == DateTime.MinValue ? "None" : smoDatabase.LastLogBackupDate.ToString(),
+                                MemoryAllocatedToMemoryOptimizedObjectsInMb = ConvertKbtoMbString(smoDatabase.MemoryAllocatedToMemoryOptimizedObjectsInKB),
+                                MemoryUsedByMemoryOptimizedObjectsInMb = ConvertKbtoMbString(smoDatabase.MemoryUsedByMemoryOptimizedObjectsInKB),
+                                NumberOfUsers = smoDatabase.Users.Count.ToString(),
+                                Owner = smoDatabase.Owner.ToString(),
+                                SizeInMb = smoDatabase.Size.ToString("0.00") + " MB",
```
Charles-Gagnon, line 138: You shouldn't add the units here - that can be added on the UI side as well

(and TBH this should probably be a number, not sure why it would need to be a string)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -120,6 +120,27 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
+                                SpaceAvailableInMb = ConvertKbtoMbString(smoDatabase.SpaceAvailable),
```
Charles-Gagnon, line 139: Does this have the units on it already from SMO?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -680,5 +705,15 @@ private static AzureEditionDetails[] GetAzureMaxSizes(IEnumerable<AzureEdition>
...
         }
+
+        /// <summary>
+        /// Converts value in KBs to MBs with two decimal places
+        /// </summary>
+        /// <param name="valueInKb">value in kilo bytes</param>
+        /// <returns>Returns as String</returns>
+        private string ConvertKbtoMbString(double valueInKb)
+        {
+            return (Math.Round(valueInKb / 1000, 2)).ToString("0.00") + " MB";
```
Charles-Gagnon, line 716: 1. This should just be a static function - preferably in a general utils class
2. Same as above - we shouldn't pass the units and instead add those in on the UI side

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -120,6 +120,27 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
+                        // Collect the Database properties information
+                        if (!requestParams.IsNewObject)
+                        {
+                            var smoDatabase = dataContainer.SqlDialogSubject as Database;
+                            databaseViewInfo.ObjectInfo = new DatabaseInfo()
+                            {
+                                Name = smoDatabase.Name,
+                                CollationName = smoDatabase.Collation,
+                                DateCreated = smoDatabase.CreateDate.ToString(),
+                                LastDatabaseBackup = smoDatabase.LastBackupDate == DateTime.MinValue ? "None" : smoDatabase.LastBackupDate.ToString(),
```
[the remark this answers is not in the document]
corivera: I've been using nulls or Empty to represent something not being supported at all, so it would be left out of the UI. Like leaving out the Owner field in Azure dialogs.

Also, localizing None in the service side let's us more easily check it during Update operations later. So if you send a set of properties with None to the frontend and get the same None value back, then you skip updating that value.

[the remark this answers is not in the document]
Charles-Gagnon: This is actually a case where null vs undefined would be useful...

Can you verify that STS is picking up the correct display language? We have so few features in STS that give UI strings that I'm not actually sure, and I know loc is a big gap for a lot of our stuff.

In the end I still don't think that STS should be doing loc stuff in general since it just makes the code here more complex (let the UI be concerned about UI stuff) - but if we can't find a simple solution for handling this then I'm ok with this as long as STS is at least handling the language stuff correctly.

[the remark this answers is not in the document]
corivera: We talked offline about it, and it makes more sense to check if a property is present in the DatabaseViewInfo object, rather than checking if it's undefined or empty, to determine if it should be rendered in the frontend. Charles will take a look at our RPC serializer to see if properties don't get defined if we receive a null value.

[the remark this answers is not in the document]
Charles-Gagnon: Chatted offline with Cory about this - you can leave this as it is for now. I need to investigate whether we can use another pattern for detecting "unsupported" options vs. empty/default options.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: Oops, this should be newObject case.

But how did i test it successfully? Hmm! could be missed building the STS! will revert and test!

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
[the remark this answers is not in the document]
ssreerama: Hmm, I didn't think this way, as current experience of general tab would be a read-only! I'll take a look at your asks! Thanks :)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: The reason that I'm adding units here is to have all conversions on STS side and to display on ADS UI directly with the prepared values. 

But, I agree with you, adding units on UI side would be good!

ssreerama: Since, we are displaying the values as 16.00 MB, so considered it as string here, I can change it to its original type/int and process on UI side.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Also looks like we divide database sizes by 1024 in other code, like DatabasePrototype or DbSize, when converting to larger units

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
Charles-Gagnon, line 132: This should be localized

@corivera Did you get a chance to verify whether STS respects the display language setting in ADS?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +231,43 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
                 throw new InvalidOperationException(serverNotExistsError);
             }
-            string objectUrn = (configAction != ConfigAction.Create && database != null)
-                ? string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                    "Server/Database[@Name='{0}']",
-                    Urn.EscapeString(database.Name))
-                : string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                    "Server");
+            string objectUrn = requestParams.ObjectUrn;
+            if (requestParams.IsNewObject)
```
Charles-Gagnon, line 250: Why is this check being added? Do you not need an object URN for existing objects as well?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
Charles-Gagnon: See the note here : https://github.com/microsoft/sqltoolsservice/wiki#building

It would be great to add more explicit documentation around the process of adding localized strings - so if this is something you two could work on that'd be fantastic.

ssreerama: I was just able to add a localized string and will be adding the process to the doc! @corivera, please update it with your insights too!

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Ideally you should be able to use the existing case as is, since it includes the database name if it was provided. Was it not working for the properties dialog?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Yeah, the dialog switches to Spanish values when I set ADS' language mode to the same.

corivera: Although it doesn't look like all fields are returning localized values. I'll look into it

corivera: Made an issue for that. https://github.com/microsoft/azuredatastudio/issues/23392

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: No, ConfigAction.Create is stopping to use the case, objectUrn should be "Server[@Name='SQLTools2022-3']/Database[@Name='keep_AdventureWorksDW']" but we get "Server" instead, I can make modifications and use it

ssreerama: Passing configAction(Update) and DatabaseInfo works fine to use the existing case to get the smo object. reverted the change too! Thanks

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +233,37 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
                 throw new InvalidOperationException(serverNotExistsError);
             }
-            string objectUrn = (configAction != ConfigAction.Create && database != null)
-                ? string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                    "Server/Database[@Name='{0}']",
-                    Urn.EscapeString(database.Name))
-                : string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                    "Server");
+            string objectUrn = configAction != ConfigAction.Create && database != null 
+                ? string.Format(System.Globalization.CultureInfo.InvariantCulture,"Server/Database[@Name='{0}']",Urn.EscapeString(database.Name))
```
corivera, line 252: Looks like some parts didn't quite get reverted here. Some args ended up on the same lines

corivera (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: Oh this is actually because I formatted the logic in a readable way. Do we use any auto formatted shortcuts for cs files?

corivera: You can use Command Palette -> Format Document

Charles-Gagnon (changes requested):
Let's get some tests added as part of this too. Sync up with @corivera if you're unsure how to best approach this.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -260,6 +260,50 @@ public void GetAzureMaxSizesTest()
             }
         }
 
+        [Test]
+        /// This test validates the newly created database properties and verifies with some default values
+        public async Task VerifyDatabaseProperties()
```
corivera, line 265: Small style thing, could you add Test to the end of this method name?

corivera (approved):

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -260,6 +260,50 @@ public void GetAzureMaxSizesTest()
...
+        /// This test validates the newly created database properties and verifies with some default values
+        public async Task VerifyDatabasePropertiesTest()
+        {
+            // setup, drop database if exists.
+            var connectionResult = await LiveConnectionHelper.InitLiveConnectionInfoAsync("master", serverType: TestServerType.OnPrem);
+            using (SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connectionResult.ConnectionInfo))
+            {
+                var server = new Server(new ServerConnection(sqlConn));
+
+                var testDatabase = ObjectManagementTestUtils.GetTestDatabaseInfo();
```
Charles-Gagnon, line 273: Maybe `testDatabaseInfo` instead.

But also I'd be careful with trying to use actual "default" values like 4 users and size. Those values are not something we control so it makes this test inherently unstable. I think it's fine for this test just to verify that we have some value for the properties like that which we can't control.

Charles-Gagnon (approved):

ssreerama:
https://github.com/microsoft/azuredatastudio/issues/23336

PR #2094, opened 2023-06-08
Retrieve Azure SLO details in Database Handler

corivera:
These SLO details are used for the Create Database dialog in ADS when creating an Azure SQL DB. These changes add the ability to select the DB's Edition, Backup Storage Redundancy level, Service Level Objective, and Max Size.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototypeAzure.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototypeAzure.cs
@@ -91,58 +91,58 @@ namespace Microsoft.SqlTools.ServiceLayer.Admin
             {
                 return AzureSqlDbHelper.GetAzureEditionDisplayName(this.currentState.azureEdition);
             }
-            // set
-            // {
-            //     AzureEdition edition;
-            //     if (AzureSqlDbHelper.TryGetAzureEditionFromDisplayName(value, out edition))
-            //     {
-            //         //Try to get the ServiceLevelObjective from the api,if not the default hardcoded service level objectives will be retrieved.
-            //         string serverLevelObjective = AzureServiceLevelObjectiveProvider.TryGetAzureServiceLevelObjective(value, AzureServiceLocation);
+            set
+            {
+                AzureEdition edition;
+                if (AzureSqlDbHelper.TryGetAzureEditionFromDisplayName(value, out edition))
+                {
+                    //Try to get the ServiceLevelObjective from the api,if not the default hardcoded service level objectives will be retrieved.
+                    // string serverLevelObjective = AzureServiceLevelObjectiveProvider.TryGetAzureServiceLevelObjective(value, AzureServiceLocation);
 
-            //         if (!string.IsNullOrEmpty(serverLevelObjective))
-            //         {
-            //             this.currentState.azureEdition = edition;
-            //             this.currentState.currentServiceLevelObjective = serverLevelObjective;
-            //             // Instead of creating db instance with default Edition, update EditionToCreate while selecting Edition from the UI.
-            //             this.EditionToCreate = MapAzureEditionToDbEngineEdition(edition);
-            //             string storageAccountType = AzureServiceLevelObjectiveProvider.TryGetAzureStorageType(value, AzureServiceLocation);
-            //             if (!string.IsNullOrEmpty(storageAccountType))
-            //             {
-            //                 this.currentState.backupStorageRedundancy = storageAccountType;
-            //             }
+                    // if (!string.IsNullOrEmpty(serverLevelObjective))
+                    // {
+                    //     this.currentState.azureEdition = edition;
+                    //     this.currentState.currentServiceLevelObjective = serverLevelObjective;
+                    //     // Instead of creating db instance with default Edition, update EditionToCreate while selecting Edition from the UI.
+                    //     this.EditionToCreate = MapAzureEditionToDbEngineEdition(edition);
+                    //     string storageAccountType = AzureServiceLevelObjectiveProvider.TryGetAzureStorageType(value, AzureServiceLocation);
+                    //     if (!string.IsNullOrEmpty(storageAccountType))
+                    //     {
+                    //         this.currentState.backupStorageRedundancy = storageAccountType;
+                    //     }
 
-            //             // Try to get the azure maxsize from the api,if not the default hardcoded maxsize will be retrieved.
-            //             DbSize dbSize = AzureServiceLevelObjectiveProvider.TryGetAzureMaxSize(value, serverLevelObjective, AzureServiceLocation);
-            //             if (!string.IsNullOrEmpty(dbSize.ToString()))
-            //             {
-            //                 this.currentState.maxSize = new DbSize(dbSize.Size, dbSize.SizeUnit);
-            //             }
-            //         }
-            //         else
-            //         {
-            //             if (edition == this.currentState.azureEdition)
-            //             { //No changes, return early since we don't need to do any of the changes below
-            //                 return;
-            //             }
+                    //     // Try to get the azure maxsize from the api,if not the default hardcoded maxsize will be retrieved.
+                    //     DbSize dbSize = AzureServiceLevelObjectiveProvider.TryGetAzureMaxSize(value, serverLevelObjective, AzureServiceLocation);
+                    //     if (!string.IsNullOrEmpty(dbSize.ToString()))
+                    //     {
+                    //         this.currentState.maxSize = new DbSize(dbSize.Size, dbSize.SizeUnit);
+                    //     }
+                    // }
+                    // else
+                    // {
+                    if (edition == this.currentState.azureEdition)
+                    { // No changes, return early since we don't need to do any of the changes below
+                        return;
+                    }
 
-            //             this.currentState.azureEdition = edition;                        
-            //             this.EditionToCreate = MapAzureEditionToDbEngineEdition(edition);
-            //             this.CurrentServiceLevelObjective = AzureSqlDbHelper.GetDefaultServiceObjective(edition);
-            //             this.BackupStorageRedundancy = AzureSqlDbHelper.GetDefaultBackupStorageRedundancy(edition);
-            //             var defaultSize = AzureSqlDbHelper.GetDatabaseDefaultSize(edition);
+                    this.currentState.azureEdition = edition;                        
+                    this.EditionToCreate = MapAzureEditionToDbEngineEdition(edition);
+                    this.CurrentServiceLevelObjective = AzureSqlDbHelper.GetDefaultServiceObjective(edition);
+                    this.BackupStorageRedundancy = AzureSqlDbHelper.GetDefaultBackupStorageRedundancy(edition);
+                    var defaultSize = AzureSqlDbHelper.GetDatabaseDefaultSize(edition);
 
-            //             this.MaxSize = defaultSize == null ? String.Empty : defaultSize.ToString();
-            //         }
-            //         this.NotifyObservers();
-            //     }
-            //     else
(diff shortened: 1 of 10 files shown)


src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototypeAzure.cs
```diff
@@ -91,58 +91,58 @@ public string AzureEditionDisplay
...
-            //             string.Format(CultureInfo.InvariantCulture,
-            //                 "Failed to parse edition display name '{0}' back into AzureEdition", value));
-            //     }
-            // }
+            set
+            {
+                AzureEdition edition;
+                if (AzureSqlDbHelper.TryGetAzureEditionFromDisplayName(value, out edition))
+                {
+                    //Try to get the ServiceLevelObjective from the api,if not the default hardcoded service level objectives will be retrieved.
```
corivera, line 99: This section deals with retrieving the SLO details dynamically. I'll look into enabling it after merging this PR.

corivera: We don't have this provider file in STS yet, so I'd have to port it over from SSMS.

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototypeAzure.cs
```diff
@@ -91,58 +91,58 @@ public string AzureEditionDisplay
...
+                    var defaultSize = AzureSqlDbHelper.GetDatabaseDefaultSize(edition);
+
+                    this.MaxSize = defaultSize == null ? String.Empty : defaultSize.ToString();
+                    // }
+                    this.NotifyObservers();
+                }
+                else
+                {
+                    //Can't really do much if we fail to parse the display name so just leave it as is and log a message
+                    System.Diagnostics.Debug.Assert(false,
```
Charles-Gagnon, line 141: I'm going to keep bugging you about this every time I see these - there is 0 good reason to keep a debug assert in this code vs having an exception or just default to some known good value. This one just seems like it should be logged and move on.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -583,5 +618,79 @@ private string[] GetCompatibilityLevels(int sqlServerVersion, DatabasePrototype
...
+
+        /// <summary>
+        /// Get supported service level objectives for this Azure server.
+        /// </summary>
+        private static AzureEditionDetails[] GetAzureServiceLevels(string[] editionDisplayNames)
+        {
+            var levels = new List<AzureEditionDetails>();
+            foreach (string displayName in editionDisplayNames)
+            {
+                if (AzureSqlDbHelper.TryGetAzureEditionFromDisplayName(displayName, out var azureEdition))
```
Charles-Gagnon, line 638: Log error if unknown display name is passed?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -583,5 +618,79 @@ private string[] GetCompatibilityLevels(int sqlServerVersion, DatabasePrototype
...
+                {
+                    if (AzureSqlDbHelper.TryGetServiceObjectiveInfo(azureEdition, out var serviceInfoPair))
```
Charles-Gagnon, line 640: Log error if no SLO?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -583,5 +618,79 @@ private string[] GetCompatibilityLevels(int sqlServerVersion, DatabasePrototype
...
+                    }
+                }
+            }
+            return levels.ToArray();
+        }
+
+        /// <summary>
+        /// Get supported editions for this Azure server.
+        /// </summary>
+        private static string[] GetAzureEditions()
```
Charles-Gagnon, line 662: `GetAzureEditionDisplayNames`? (and update doc comment too)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -583,5 +618,79 @@ private string[] GetCompatibilityLevels(int sqlServerVersion, DatabasePrototype
...
+
+        /// <summary>
+        /// Get supported maximum sizes for this Azure server.
+        /// </summary>
+        private static AzureEditionDetails[] GetAzureMaxSizes(string[] editionDisplayNames)
+        {
+            var sizes = new List<AzureEditionDetails>();
+            foreach (string displayName in editionDisplayNames)
+            {
+                if (AzureSqlDbHelper.TryGetAzureEditionFromDisplayName(displayName, out var azureEdition))
```
Charles-Gagnon, line 675: Log error if no matching edition?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -583,5 +618,79 @@ private string[] GetCompatibilityLevels(int sqlServerVersion, DatabasePrototype
...
+                {
+                    if (AzureSqlDbHelper.TryGetDatabaseSizeInfo(azureEdition, out var sizeInfoPair))
```
Charles-Gagnon, line 677: Log error if no matching size info?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -71,6 +76,12 @@ static DatabaseHandler()
             {
                 recoveryModelEnums.Add(displayRecoveryModels[key], key);
             }
+
+            // Azure SLO info is invariant of server information, so set up static objects we can return later
+            AzureEditions = GetAzureEditions();
+            AzureBackupLevels = GetAzureBackupLevels();
+            AzureMaxSizes = GetAzureMaxSizes(AzureEditions);
```
Charles-Gagnon, line 83: Instead of passing in the display names, it seems better to have GetAzureEditions actually return the AzureEditions and passing that around instead.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -583,5 +618,79 @@ private string[] GetCompatibilityLevels(int sqlServerVersion, DatabasePrototype
...
+                        // Move default value to the front of the list
+                        var sizeInfoList = new List<DbSize>(sizeInfoPair.Value);
+                        var defaultIndex = sizeInfoPair.Key;
+                        if (defaultIndex >= 0 && defaultIndex < sizeInfoList.Count)
+                        {
+                            var defaultSizeInfo = sizeInfoList[defaultIndex];
+                            sizeInfoList.RemoveAt(defaultIndex);
+                            sizeInfoList.Insert(0, defaultSizeInfo);
+                        }
+                        var details = new AzureEditionDetails() { EditionDisplayName = displayName, Details = sizeInfoList.Select(info => info.ToString()).ToArray() };
```
Charles-Gagnon, line 688: Is there a reason this stuff is being calculated dynamically each time? The edition -> available size mappings seems like another value that could be made static.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -162,6 +166,111 @@ public async Task DatabaseAlreadyExistsErrorTest()
...
+            {
+                AzureEdition.Basic.DisplayName,
+                AzureEdition.Standard.DisplayName,
+                AzureEdition.Premium.DisplayName,
+                AzureEdition.DataWarehouse.DisplayName,
+                AzureEdition.BusinessCritical.DisplayName,
+                AzureEdition.GeneralPurpose.DisplayName,
+                AzureEdition.Hyperscale.DisplayName
+            };
+            Assert.That(actualEditions, Is.EquivalentTo(expectedEditions));
```
Charles-Gagnon, line 183: What exactly is the point of this test? Seems like to actually be useful it should be verifying that all the editions listed in AzureEdition are returned by AzureEditions. Right now if I add a new edition to AzureEdition but forget to update that helper method then this test won't catch that...

Same question applies to the other similar tests below.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -162,6 +166,111 @@ public async Task DatabaseAlreadyExistsErrorTest()
             }
         }
 
+        [Test]
```
Charles-Gagnon, line 169: It'd be great to see some tests for the helper method functionality (such as GetAzureServiceLevels) too

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: This is the builder method that gets called from the static constructor, so it only gets called once.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
corivera: We already do that indirectly, since the AzureServiceLevels property is a statically cached result of that method.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Removed this method since it was just passing a single property back

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -583,5 +621,65 @@ private string[] GetCompatibilityLevels(int sqlServerVersion, DatabasePrototype
...
+                        var defaultServiceObjective = serviceLevelsList[defaultIndex];
+                        serviceLevelsList.RemoveAt(defaultIndex);
+                        serviceLevelsList.Insert(0, defaultServiceObjective);
+                    }
+                    var details = new AzureEditionDetails() { EditionDisplayName = edition.DisplayName, Details = serviceLevelsList.ToArray() };
+                    levels.Add(details);
+                }
+                else
+                {
+                    Logger.Write(TraceEventType.Error, $"Failed to get service level objective info for edition '{edition.Name}'");
```
Charles-Gagnon, line 649: There's extension methods to make this easier - e.g. `Logger.Error`

Charles-Gagnon (approved):

PR #2096, opened 2023-06-09
support copy result in STS

alanrenmsft:
Handle the copy results to clipboard in STS for optimized performance.  I tested all the selection patterns and also with large result set.

Test data: 50k rows with 40 columns
Before (ADS loads all data to UI side and then copy) - ~2 minutes with UI response delay.
Now: less than 10 seconds.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
@@ -25,7 +25,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
         /// <summary>
         /// Whether to remove the line break from cell values.
         /// </summary>
-        public bool RemoveNewLine { get; set; }
+        public bool RemoveNewLines { get; set; }
 
         /// <summary>
         /// Whether to include the column headers.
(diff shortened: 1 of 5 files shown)


alanrenmsft:
@kburtram there is no dotnet built-in support to interact with clipboard, and this is the most popular package out there, any concerns?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        /// <summary>
+        /// Whether to remove the line break from cell values.
+        /// </summary>
+        public bool RemoveNewLines { get; set; }
+
+        /// <summary>
+        /// Whether to include the column headers.
+        /// </summary>
+        public bool IncludeHeaders { get; set; }
```
kburtram, line 33: is this going to support "Copy only headers" like discussed in this issue https://github.com/microsoft/azuredatastudio/issues/8902?

kburtram (approved):
The TextCopy package looks popular/broadly used & is actively maintained.  Since it's MIT it is a pre-approved license for use in this project.  We should update the 3rd party notices file in ADS & vscode-mssql to include the license for this package though.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
alanrenmsft: this property is to control whether the column names should be included along with the selected cells. the copy header action is a separate thing and we can already get it from the result metadata directly in ADS.

image

PR #2097, opened 2023-06-09
Remove Linked Server Login node from OE

caohai:
https://github.com/microsoft/azuredatastudio/issues/23365
This PR removes the Linked Server Login under Security node (server level) for OE.

Before:
image

After:
image

SSMS:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
@@ -1141,14 +1141,6 @@ namespace Microsoft.SqlTools.ServiceLayer
             }
         }
 
-        public static string SchemaHierarchy_LinkedServerLogins
-        {
-            get
-            {
-                return Keys.GetString(Keys.SchemaHierarchy_LinkedServerLogins);
-            }
-        }
-
         public static string SchemaHierarchy_Logins
         {
             get
@@ -13394,9 +13386,6 @@ namespace Microsoft.SqlTools.ServiceLayer
 
 
             public const string SchemaHierarchy_LinkedServers = "SchemaHierarchy_LinkedServers";
-
-
-            public const string SchemaHierarchy_LinkedServerLogins = "SchemaHierarchy_LinkedServerLogins";
 
 
             public const string SchemaHierarchy_Logins = "SchemaHierarchy_Logins";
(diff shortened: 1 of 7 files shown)


aasimkhan30 (approved):

alanrenmsft:
@aasimkhan30 looking at the original issue, is it possible that the linked server logins are supposed to be displayed under this folder, but due to a bug it is not?

aasimkhan30:
https://github.com/microsoft/sqltoolsservice/blob/c26a2aea142524c9dec4625448fe0f1ff2b2fcde/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/SmoModel/SmoQueryModelDefinition.xml#L27

It seems like we are fetching the wrong property for linkedserverlogins

aasimkhan30:
Based on smo docs, linked server logins is property of Linkedservers. 
https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.linkedserver?view=sql-smo-160#properties

Server => LinkedServers => LinkedServer => Security => LinkedServerLogins. 

@erinstellato-ms , where do you want to see LinkedServerLogins?

erinstellato-ms:
> Based on smo docs, linked server logins is property of Linkedservers. https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.linkedserver?view=sql-smo-160#properties
> 
> Server => LinkedServers => LinkedServer => Security => LinkedServerLogins.
> 
> @erinstellato-ms , where do you want to see LinkedServerLogins?

@aasimkhan30 I can't find that we surface it anywhere in SSMS.  @shueybubbles - can you confirm I'm not missing anything please? We don't have a LinkedServerLogins node/folder in SSMS, correct?

shueybubbles:
I think we only show them on linked server property sheet.

caohai:
> I think we only show them on linked server property sheet.

@erinstellato-ms @aasimkhan30 @alanrenmsft If we don't show them in OE for SSMS, should I go ahead and merge this PR to remove them from ADS OE as well?

alanrenmsft (approved):

aasimkhan30:
Yeah, I think it is safe to merge this PR. If we want to add support for linked servers, it can be a follow up PR.

PR #2099, opened 2023-06-09
Fix DacFx nullable warnings

kisantia:
Fixes the nullable warnings in the DacFx folder. This brings the number of warnings down from 275 to 264 when building the ServiceLayer. 

I also did a quick verification of these operations in ADS to make sure they are still working as expected.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
@@ -178,7 +178,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
                 metadata.DatabaseName = parameters.DatabaseName;
                 metadata.Name = SR.GenerateScriptTaskName;
 
-                SqlTask sqlTask = SqlTaskManagerInstance.CreateAndRun<SqlTask>(metadata);
+                operation.SqlTask = SqlTaskManagerInstance.CreateAndRun<SqlTask>(metadata);
 
                 await requestContext.SendResult(new DacFxResult()
                 {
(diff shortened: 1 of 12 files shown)


src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -26,8 +26,8 @@ namespace Microsoft.SqlTools.ServiceLayer.DacFx
     /// </summary>
     class DacFxService
     {
-        private static ConnectionService connectionService = null;
-        private SqlTaskManager sqlTaskManagerInstance = null;
+        private static ConnectionService? connectionService = null;
```
Charles-Gagnon, line 29: Is it actually expected that these will be null? Shouldn't that be something that's set up when this is created?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
kisantia: removed them since they weren't really needed. Not sure why I added them 4 years ago haha

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -397,12 +395,7 @@ private SqlTaskManager SqlTaskManagerInstance
         {
             get
             {
-                sqlTaskManagerInstance ??= SqlTaskManager.Instance;
-                return sqlTaskManagerInstance;
-            }
-            set
-            {
-                sqlTaskManagerInstance = value;
```
kisantia, line 405: these setters were never being used, so removed them

SakshiS-harma (approved):

PR #2100, opened 2023-06-13
Exclude Data Warehouse as an option when creating an Azure database

corivera:
Creating a new Data Warehouse instance is disabled in Portal, and trying to create one directly with a TSQL statement returns a deprecation message like this: "Azure SQL Data Warehouse Gen1 has been deprecated in this region. Please use SQL Analytics in Azure Synapse." This change excludes Data Warehouse from the list of Editions that gets displayed in the New Database dialog in ADS.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/AzureSqlDbHelper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/AzureSqlDbHelper.cs
@@ -527,10 +527,10 @@ namespace Microsoft.SqlTools.ServiceLayer.Admin
             yield return AzureEdition.Basic;
             yield return AzureEdition.Standard;
             yield return AzureEdition.Premium;
-            yield return AzureEdition.DataWarehouse;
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Admin/Database/AzureSqlDbHelper.cs
```diff
@@ -527,10 +527,10 @@ public static IEnumerable<AzureEdition> GetValidAzureEditionOptions()
             yield return AzureEdition.Basic;
             yield return AzureEdition.Standard;
             yield return AzureEdition.Premium;
-            yield return AzureEdition.DataWarehouse;
-            yield return AzureEdition.BusinessCritical;
             yield return AzureEdition.GeneralPurpose;
             yield return AzureEdition.Hyperscale;
+            yield return AzureEdition.BusinessCritical;
```
corivera, line 532: Moved this one down to match the ordering in portal

Charles-Gagnon (approved):

PR #2105, opened 2023-06-15
Adding like filter and removing contains to OE filtering

aasimkhan30:


src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Contracts/NodeInfo.cs
```diff
@@ -157,6 +157,10 @@ public enum NodeFilterOperator
         NotBetween = 7,
         Contains = 8,
         NotContains = 9,
+        IsNull = 10,
+		IsNotNull = 11,
+        StartsWith = 12,
+        Like = 13
```
Charles-Gagnon, line 163: What about "NotStartsWith" and "NotLike"?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +610,45 @@ public void TextNumericFilters()
             };
             Assert.Throws<IndexOutOfRangeException>(() => INodeFilter.GetPropertyFilter(filterList, typeof(SqlHistoryTableQuerier), ValidForFlag.All), "Error not thrown when the array contains single value for between operator");
         }
+
+        [Test]
+        public void TestStartsWithFilter()
```
Charles-Gagnon, line 615: Can you add tests for testing the filtering logic itself to show that you're constructing the clauses correctly?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +610,45 @@ public void TextNumericFilters()
...
+        {
+            // Testing text filter with like operator
+            var filterList = new List<NodePropertyFilter>
+            {
+                new NodePropertyFilter()
+                {
+                    Property = "Name",
+                    Type = typeof(string),
+                    ValidFor = ValidForFlag.All,
+                    Values = new List<object> { "test%" },
```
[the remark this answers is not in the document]
aasimkhan30: I am letting the users query however they want to. so, if users want
test = test
testobject = test%
mytestobject = %test%
objecttest = %test.

Should I hardcode it to %{string}%?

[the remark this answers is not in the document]
Charles-Gagnon: Hmm, I guess the question is how much do we think people we care about LIKE versus just using CONTAINS.

Not sure how much we want to have T-SQL specific syntax for filtering either (such as LIKE and %). This theoretically should be a generic thing for any provider, right?

It's definitely kind of weird to have all other filtering in the app be either just simple string-based searching or regex, but then this is its own completely separate thing...

@erinstellato-ms Any thoughts here?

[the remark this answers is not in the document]
erinstellato-ms: I'm not clear on CONTAINS vs LIKE.

If I look at the query executed in SSMS when you filter on table name CONTAINS order, this is in the query predicate:

tbl.name like N'%order%'

That is what I expected - CONTAINS is basically LIKE with wildcards around it.

So what does LIKE do in ADS?  Can the user specify where the wildcard is?  Such as:

tbl.name LIKE N'%order%'
tbl.name LIKE N'order%'
tbl.name LIKE N'%order'

If users can put the wildcard anywhere, then we do not need starts with or ends with...

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +613,211 @@ public void TextNumericFilters()
...
+                        Property = "Name",
+                        Type = typeof(string),
+                        Values = new List<object> { "test" },
+                        IsNotFilter = true,
+                        FilterType = FilterType.EQUALS,
+                        IsDateTime = false
+                    }
+                );
+
+                // Test case for contains operator with string value
```
Charles-Gagnon, line 689:
```suggestion
                // Test case for like operator with string value 
```

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +613,211 @@ public void TextNumericFilters()
...
+                        Property = "Name",
+                        Type = typeof(string),
+                        Values = new List<object> { "test" },
+                        IsNotFilter = false,
+                        FilterType = FilterType.LIKE,
+                        IsDateTime = false
+                    }
+                );
+
+                // Test case for not contains operator with string value
```
Charles-Gagnon, line 713:
```suggestion
                // Test case for not like operator with string value
```

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +613,211 @@ public void TextNumericFilters()
...
+                    Property = "Name",
+                    Type = typeof(string),
+                    ValidFor = ValidForFlag.All,
+                    Values = new List<object> { "%test" },
+                    FilterType = FilterType.LIKE,
+                    IsDateTime = false
+                }
+            };
+            var filterString = INodeFilter.GetPropertyFilter(filterList, typeof(SqlHistoryTableQuerier), ValidForFlag.All);
+            Assert.AreEqual("[(like(@Name, '%test'))]", filterString, "Error parsing text filter with like operator");
```
Charles-Gagnon, line 634: What about values entered by the user that have % in them?

Do you have tests that actually run the filter to verify that the generated filter string is correct?

aasimkhan30: The like operator by default will have no wildcards. We take the string from the user and use it as it is.

Charles-Gagnon: Yes, but did you test what happens if the user enters a string with a % in it? That's a valid character to have in object names.

Charles-Gagnon: See https://github.com/microsoft/azuredatastudio/issues/23391 for a very similar issue

erinstellato-ms: The LIKE should mean (open to discussion) that the user would enter the wildcard.  So they could do

Order%
%Order%
%Order

Charles-Gagnon: Hm, I still feel like this an awfully T-SQL specific way of doing this for something that's currently in the general extensibility layer. Will every provider then be expected to handle the same % syntax? Or is it up to them how they parse like statements?

Even for this - does the filtering correctly handle escaping the % then? e.g. if I enter `%[%]%` will it find all objects with % in their names?

Charles-Gagnon: (I realize that would be a SMO issue if it didn't work if we're assuming the user has to know to enter valid syntax, but we should test it)

Charles-Gagnon: Looking into it a bit more, it does at least seem that LIKE with % is a general SQL thing. So better. But then we're saying in ADS that every single other filter/search uses wildcards/regex but this one place uses SQL like? That seems weird and not very user friendly to me...could we not make it similar to the other filter widgets in the app (maybe only supporting wildcards though since regex isn't supported AFAIK) and at least use `*`?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs
```diff
@@ -182,7 +184,19 @@ public string ToPropertyFilterString(Type type, ValidForFlag validForFlag)
                     object propertyValue = value;
                     if (Type == typeof(string))
                     {
-                        propertyValue = $"'{propertyValue}'";
+                        var escapedString = CUtils.EscapeStringSQuote((string)propertyValue);
```
Charles-Gagnon, line 187: Please make sure you call out when you do other fixes like this so they can be tracked and verified separately

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs
```diff
@@ -276,6 +293,9 @@ public enum FilterType
         LESSTHANOREQUAL,
         GREATERTHANOREQUAL,
         BETWEEN,
-        NOTBETWEEN
+        NOTBETWEEN,
```
Charles-Gagnon, line 296: Shouldn't there just be between? The others don't have their NOT versions here

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +613,307 @@ public void TextNumericFilters()
...
+                    Property = "Name",
+                    Type = typeof(string),
+                    ValidFor = ValidForFlag.All,
+                    Values = new List<object> { "test" },
+                    FilterType = FilterType.CONTAINS,
+                    IsDateTime = false
+                }
+            };
+            var filterString = INodeFilter.GetPropertyFilter(filterList, typeof(SqlHistoryTableQuerier), ValidForFlag.All);
+            Assert.AreEqual("[(contains(@Name, 'test'))]", filterString, "Error parsing text filter with contains operator");
```
Charles-Gagnon, line 634: We should be using fluent assertions for all STS tests (Assert.That)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +613,307 @@ public void TextNumericFilters()
             };
             Assert.Throws<IndexOutOfRangeException>(() => INodeFilter.GetPropertyFilter(filterList, typeof(SqlHistoryTableQuerier), ValidForFlag.All), "Error not thrown when the array contains single value for between operator");
         }
+
+        [Test]
+        public void TestContainsFilter()
+        {
+            // Testing text filter with ends with operator
+            var filterList = new List<NodePropertyFilter>
```
Charles-Gagnon, line 621: I don't think I've ever seen a response to this - do you have any plans to add tests that actually run the filter through STS to verify it's being built correctly? Right now this doesn't actually tell us whether the queries generated are valid or not.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs
```diff
@@ -182,7 +184,19 @@ public string ToPropertyFilterString(Type type, ValidForFlag validForFlag)
                     object propertyValue = value;
                     if (Type == typeof(string))
                     {
-                        propertyValue = $"'{propertyValue}'";
+                        var escapedString = CUtils.EscapeStringSQuote((string)propertyValue);
+                        switch (FilterType)
+                        {
+                            case FilterType.STARTSWITH:
+                                propertyValue = $"'{escapedString}%'";
```
Charles-Gagnon, line 191: You also need to escape % (assuming here that SMO supports that syntax)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +613,307 @@ public void TextNumericFilters()
             };
             Assert.Throws<IndexOutOfRangeException>(() => INodeFilter.GetPropertyFilter(filterList, typeof(SqlHistoryTableQuerier), ValidForFlag.All), "Error not thrown when the array contains single value for between operator");
         }
+
+        [Test]
```
Charles-Gagnon, line 617: It would be good to add tests for special cases - such as strings with reserved characters (`'` and `%`)

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs
aasimkhan30: That's because not between works in a different way than other not versions.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
aasimkhan30: You mean Integration tests?

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -15,7 +15,7 @@
 		<AssemblyTitle>SqlTools Editor Services Host Protocol Library</AssemblyTitle>
 		<Description>Provides message types and client/server APIs for the SqlTools Editor Services JSON protocol.</Description>
 		<!-- False alerts, disabled due to issue: https://github.com/dotnet/roslyn/issues/65850 -->
-		<NoWarn>$(NoWarn);CS8795</NoWarn>
+		<NoWarn>$(NoWarn);CS8795;SYSLIB1045</NoWarn>
```
Charles-Gagnon, line 18: 1. Why are you disabling this warning?
2. Whenever we disable a warning we should be adding a comment on why it's disabled (such as what is already there for the CS8795 warning)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
Charles-Gagnon: That'd probably be required - I don't believe SMO has a way to filter without a live connection.

@shueybubbles Any thoughts here on what the best way to get this kind of test added would be? (in case it's not clear, these filter strings are added to the SMO URN used to fetch the objects)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
```diff
@@ -610,5 +613,352 @@ public void TextNumericFilters()
...
+            INodeFilter actualParsedFilter = ObjectExplorerUtils.ConvertExpandNodeFilterToNodeFilter(filter, prop);
+            Assert.That(
+                JsonConvert.SerializeObject(actualParsedFilter),
+                Is.EqualTo(JsonConvert.SerializeObject(expectedParsedFilter)),
+                $"Error parsing node filter '{prop.Name} ({prop.Type.ToString()})' with operator '{filter.Operator.ToString()}' and value '{filter.Value}'"
+            );
+        }
+
+        [Test]
+        [TestCase("test''test", "[(@Name = 'test''''test')]", "[(like(@Name, 'test''''test%'))]")]
```
Charles-Gagnon, line 920: Did you verify that these filter strings are verified manually? Ideally the integration tests I asked about above would cover this, but at the very least let's make sure we're verifying there aren't any bugs in the SMO stuff

kburtram (approved):

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
Charles-Gagnon: We have ways to get a live connection - I just wanted to ask if you knew of a way to do an offline test of the filtering functionality. IIRC though these filter queries are turned into actual T-SQL queries so that's probably not possible. Is that correct?

shueybubbles:
>                 return string.Empty;

Consider moving a lot of this functionality into SMO itself. It'd be great to have these programmatic ways of building a URN filter there instead of relying on so much manual code. There are a fair number of SMO APIs that build such filters.
EG I don't think SMO code is doing anything to ensure that parameters to the like function are properly escaped like you do here.


---
Refers to: src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs:89 in f03f443. [](commit_id = f03f443345663dfbd0c15eee301fe1ec1a5faa8d, deletion_comment = False)

test/Microsoft.SqlTools.ServiceLayer.UnitTests/ObjectExplorer/NodeFilterTests.cs
shueybubbles: yeah there's a bunch of code in SFC that converts the URN snippets to TSQL snippets. I think that area is only tested indirectly not exhaustively in the SMO suites. You might consider putting your tests for correct SQL query generation in SMO itself to prevent regressions there.

shueybubbles:
>                 return string.Empty;

There are strongly type classes for filters in [https://github.com/microsoft/sqlmanagementobjects/blob/main/src/Microsoft/SqlServer/Management/Sdk/Sfc/Enumerator/FilterNode.cs](https://github.com/microsoft/sqlmanagementobjects/blob/main/src/Microsoft/SqlServer/Management/Sdk/Sfc/Enumerator/FilterNode.cs)

I don't know if they are useful for your scenario. They probably need more extensive testing.

---
In reply to: [1601234945](https://github.com/microsoft/sqltoolsservice/pull/2105#issuecomment-1601234945) [](http://example.com/codeflow?ancestors=1601234945)

---
Refers to: src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/Nodes/NodePropertyFilter.cs:89 in f03f443. [](commit_id = f03f443345663dfbd0c15eee301fe1ec1a5faa8d, deletion_comment = False)

Charles-Gagnon:
@aasimkhan30 If we could start by getting some tests added here that'd be great. The suggestions David made are good, but those are probably best done as follow up items I think

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
aasimkhan30: Moved to GeneratedRegexAttribute

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectExplorer/ObjectExplorerServiceTests.cs
```diff
@@ -267,32 +267,121 @@ public async Task GroupBySchemaHidesLegacySchemas()
...
+
+                // Testing contains operator
+                NodeFilter filter = new NodeFilter()
+                {
+                    Name = NameProperty.Name,
+                    Value = "table",
+                    Operator = NodeFilterOperator.Contains
+                };
+                var tablesChildren = await _service.ExpandNode(session, tablesNode.NodePath, true, null, new NodeFilter[] { filter });
+                Assert.True(tablesChildren.Nodes.Any(t => t.Label == "dbo.testTable"), "testTable node should be found in tables node");
```
Charles-Gagnon, line 324: Fluent assertions please

Charles-Gagnon (approved):

PR #2067, opened 2023-05-15
Restore new connection string for multiple similar connections

smartguest:
This reverts commit 1b5f7747417cd17dd3be7dfbc0459d0ccbb748f9. (Prelude to reintroducing the connection string to ADS)

Companion PR to https://github.com/microsoft/azuredatastudio/pull/23142

Also introduces fix for connections disconnecting with the same info but different connection names (they are not differentiated in the DBConnection connection string class.)

PR #2109, opened 2023-06-19
fix delay in query execution

alanrenmsft:
for issue: https://github.com/microsoft/azuredatastudio/issues/23145

1. Properly set the connection purpose so that it won't be automatically closed later in the connection handler: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs#L420
2. Currently, the connection that was made when connecting the query editor has the connection type set to default, so when we execute the query a new connection will need to be made, I am update the query service to use the default connection type so that we don't have to make a new connection.

I've spoken to @cheenamalhotra to review the current design and work with the area owners if changes need to be made.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
@@ -448,7 +448,7 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                 }
 
                 // Locate and setup the connection
-                queryConnection = await ConnectionService.Instance.GetOrOpenConnection(editorConnection.OwnerUri, ConnectionType.Query);
+                queryConnection = await ConnectionService.Instance.GetOrOpenConnection(editorConnection.OwnerUri, ConnectionType.Default);
                 onErrorAction = OnErrorAction.Ignore;
                 sqlConn = queryConnection as ReliableSqlConnection;
                 if (sqlConn != null)
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -491,13 +491,18 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         {
             if (connectionParams != null && connectionParams.Type == ConnectionType.Default && !string.IsNullOrWhiteSpace(connectionParams.OwnerUri))
             {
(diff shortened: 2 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -448,7 +448,7 @@ private async Task ExecuteInternal()
                 }
 
                 // Locate and setup the connection
-                queryConnection = await ConnectionService.Instance.GetOrOpenConnection(editorConnection.OwnerUri, ConnectionType.Query);
```
Charles-Gagnon, line 451: It would be great to go and try to make sense of all this connection type stuff. Why do we even have it if we just ignore/override it on the STS side in some circumstances? And what do each of the values mean? It'll be too easy for any changes to this stuff to end up breaking some other completely unrelated thing if we aren't having a strict set of rules around this stuff...

(and then please document this in the wiki or somewhere so we actually have something to refer back to in the future)

Charles-Gagnon (approved):

cheenamalhotra:
This is actually causing connection leak @alanrenmsft - try opening multiple query editor windows and close them.
Run `sp_who2` on a new editor and you'll find leaked connections identified with `azdata-query` app name.

image

alanrenmsft:
@cheenamalhotra I did verify the closing the editor will close the underlying connection. let me take another look.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
cheenamalhotra: As of now, it seems the connections are always momentarily created for most scenarios (this is why I was interested in introducing connection pooling), however I'd need to thoroughly test all features to ensure that statement is true. I have briefly looked into primary touch areas as of now.

If connections are not falling in the 'closeable' category, features requesting connection need to own the lifecycle of the connection, as QE intends to do now.

alanrenmsft:
@cheenamalhotra the leak is introduced by this change here: https://github.com/microsoft/sqltoolsservice/pull/2067/files#diff-343925ee5b368e81e1a3adad48a135ee26e2796d6577be04b0c844c13ad340afL1047, I'll leave it to you and Alex to fix the problem.

cheenamalhotra:
Discussed offline, yes, the change is related with new URI changes, omitting that change I too don't see any concerns.

PR #2110, opened 2023-06-20
Enable Detach Database in database handler

corivera:
These changes are needed for the Detach Database dialog in ADS. Currently working on adding some more integration tests in addition to the scripting ones I have already.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DetachDatabase.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement.Contracts
+{
+    public class DetachDatabaseRequestParams : GeneralRequestDetails
+    {
+        /// <summary>
+        /// SFC (SMO) URN identifying the object  
+        /// </summary>
+        public string? ObjectUrn { get; set; }
```
Charles-Gagnon, line 16: Why is this optional? When will we ever not have this?

(same for the connection URI)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DetachDatabase.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class DetachDatabaseRequestResponse { }
+
+    public class DetachDatabaseRequest
+    {
+        public static readonly RequestType<DetachDatabaseRequestParams, DetachDatabaseRequestResponse> Type = RequestType<DetachDatabaseRequestParams, DetachDatabaseRequestResponse>.Create("objectManagement/detachDatabase");
+    }
+
+    public class ScriptDetachDatabaseRequest
+    {
+        public static readonly RequestType<DetachDatabaseRequestParams, string> Type = RequestType<DetachDatabaseRequestParams, string>.Create("objectManagement/scriptDetach");
```
Charles-Gagnon, line 44:
```suggestion
        public static readonly RequestType<DetachDatabaseRequestParams, string> Type = RequestType<DetachDatabaseRequestParams, string>.Create("objectManagement/scriptDetachDatabase");
```

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DetachDatabase.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// Whether to drop active connections to this database
+        /// </summary>
+        public bool DropConnections { get; set; }
+        /// <summary>
+        /// Whether to update the optimization statistics related to this database
+        /// </summary>
+        public bool UpdateStatistics { get; set; }
+        /// <summary>
+        /// Whether to throw an error if the object does not exist. The default value is false.
```
Charles-Gagnon, line 30: When would we ever want this to be false?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +219,141 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
+        /// <param name="objectUrn">URN of the database to detach</param>
+        /// <param name="dropConnections">Whether to drop active connections to the database before detaching it</param>
+        /// <param name="updateStatistics">Whether to update the query optimization statistics related to the database</param>
+        /// <param name="throwIfNotExist">Whether to throw an exception if the specified database doesn't exist</param>
+        public Task Detach(string connectionUri, string objectUrn, bool dropConnections, bool updateStatistics, bool throwIfNotExist)
+        {
+            ConnectionInfo connectionInfo = this.GetConnectionInfo(connectionUri);
+            using (var dataContainer = CreateDatabaseDataContainer(connectionUri, objectUrn))
+            {
+                var smoDatabase = dataContainer.SqlDialogSubject as Database;
```
Charles-Gagnon, line 242: Add comments here for these blocks explaining what all this extra stuff is doing

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +219,141 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
+                try
+                {
+                    var smoDatabase = dataContainer.SqlDialogSubject as Database;
+                    if (smoDatabase != null)
+                    {
+                        builder.AppendLine("USE [master]");
+                        builder.AppendLine("GO");
+                        if (dropConnections)
+                        {
+                            builder.AppendLine($"ALTER DATABASE [{smoDatabase.Name}] SET SINGLE_USER WITH ROLLBACK IMMEDIATE");
```
Charles-Gagnon, line 302: You need to escape the database name here

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +219,141 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
+                            builder.AppendLine("GO");
+                        }
+                        builder.Append($"EXEC master.dbo.sp_detach_db @dbname = N'{smoDatabase.Name}'");
```
Charles-Gagnon, line 305: Can't you just set the execution mode on SMO to script and then do it through there? We should avoid writing our own scripts (especially if they may differ from SMO) whenever possible

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +219,141 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
+                    {
+                        // Revert to previous level of user access if we changed it before encountering the exception
+                        if (originalAccess != smoDatabase.DatabaseOptions.UserAccess)
+                        {
+                            smoDatabase.DatabaseOptions.UserAccess = originalAccess;
+                            smoDatabase.Alter(TerminationClause.RollbackTransactionsImmediately);
+                        }
+                        if (!(ex.InnerException is MissingObjectException) || (ex.InnerException is MissingObjectException && throwIfNotExist))
+                        {
+                            throw;
```
Charles-Gagnon, line 266: Would suggest still logging if we aren't throwing when it's missing

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -210,39 +219,141 @@ public override Task Save(DatabaseViewContext context, DatabaseInfo obj)
...
+                        builder.AppendLine("GO");
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"Provided URN '{objectUrn}' did not correspond to an existing database.");
+                    }
+                }
+                catch (FailedOperationException ex)
+                {
+                    if (!(ex.InnerException is MissingObjectException) || (ex.InnerException is MissingObjectException && throwIfNotExist))
```
Charles-Gagnon, line 320: Would suggest still logging if we aren't throwing when it's missing

corivera:
Looking into test failures. Looks like some of the changes are causing create operations to instead produce rename database scripts.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Even if it comes from the SMO Database object?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DetachDatabase.cs
corivera: I removed "nullable disable" from the top of the file, and then compiler wanted these strings to be optional.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DetachDatabase.cs
corivera: This was copy pasted from the drop database section. I'll remove it.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: I retrieve the Database object using the object URN, rather than using any arguments directly.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DetachDatabase.cs
Charles-Gagnon: You can leave nullable disabled for these files - we haven't found a good way to handle that for "contract" classes like this that isn't just clutter. 

But these should match the typings on the ADS side - where I assume you won't have these properties be optional.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Where can I find the docs for that? I was only seeing scripting options for Create statements

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
Charles-Gagnon: That name isn't going to be escape...so yes. A database named `MyTest]Database` will break the query currently.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Never mind, found the capture mode docs

Charles-Gagnon: Yup. It's what we do for scripting in general.

https://learn.microsoft.com/en-us/sql/relational-databases/server-management-objects-smo/create-program/using-capture-mode?view=sql-server-ver16 for others' reference

corivera: I tried the SMO scripting method, but apparently there's a bug with Detach in particular where it marks the database as dropped in the SMO code in scripting mode, even though the database still exists, which causes failures when trying to use any methods on the Database object afterwards. I'll stick with the manual script creation until that bug eventually gets fixed.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Removed the throwIfExists option

barbaravaldez (approved):

PR #2117, opened 2023-06-22
Get server general properties 

barbaravaldez:
Create request for retrieving the general properties for server.


src/Microsoft.SqlTools.ServiceLayer/ServerConfigurations/ServerConfigService.cs
```diff
@@ -239,5 +239,27 @@ private ConfigProperty GetSmoConfig(Server server, int configNumber)
...
+                        serverConfig = configProperty;
+                        break;
+                    }
+                }
+
+                return serverConfig != null ? ServerConfigProperty.ToServerConfigProperty(serverConfig) : null;
+            }
+            catch (Exception ex)
+            {
+                throw new ServerConfigException($"Failed to get config. config name: ${configName}", ex);
```
corivera, line 261: Could you add the server name to this message too? Maybe something like: $"Failed to get config property '{configName}' for server '{server.Name}'." ?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+                    MaxServerMemory = configService.GetConfigByName(this.server, "max server memory (MB)").ConfigValue
+            };
+
+                return Task.FromResult(new InitializeViewResult { ViewInfo = this.serverViewInfo, Context = context });
+            }
+        }
+
+        public override Task Save(ServerViewContext context, ServerInfo serverInfo)
+        {
+            throw new NotImplementedException();
```
corivera, line 75: Could this be a NotSupportedException instead? Same with Script below.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+        public Task<Server> CreateServerSmoObject(ServerViewContext context)
+        {
+            using (context.Connection.SqlConnectionObject)
+            {
+                this.server = new Server(context.Connection);
+                string objectUrn = string.Format(System.Globalization.CultureInfo.InvariantCulture, "Server");
+                var serverSmo = server.GetSmoObject(new Urn(objectUrn)) as Server ?? throw new InvalidOperationException(serverNotExistsError);
```
corivera, line 88: Would GetSmoObject already throw a FailedOperationException if the specified server doesn't exist? I've seen errors like that when getting non-existent database objects with the same method

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.SqlTools.ServiceLayer.ServerConfigurations;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// Server object type handler
+    /// </summary>
+    public class ServerHandler : ObjectTypeHandler<ServerInfo, ServerViewContext>
+    {
+        private const string serverNotExistsError = "Server was not created for data container";
```
corivera, line 23: Could you create a localized string for this if you're planning on keeping the error? You can add new localized strings to this file: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
Then you can regenerate the associated SR files by running build.cmd from the project root.

corivera (approved):
Looks good, pending my other comments.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+                    };
+                }
+
+                return Task.FromResult(new InitializeViewResult { ViewInfo = this.serverViewInfo, Context = context });
+            }
+        }
+
+        public override Task Save(ServerViewContext context, ServerInfo serverInfo)
+        {
+            throw new NotSupportedException();
```
corivera, line 77: Could you add some more descriptive errors for this exception and the one in Script? Something that would say this exception is coming from the server handler, so we don't get cryptic errors being reported to the frontend if someone happens to call either of these for a server.

corivera:
@barbaravaldez Would you be able to add some integration tests for the server handler?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// A class for storing various properties needed for Saving & Scripting a server
+    /// </summary>
+    public class ServerInfo : SqlObject
+    {
+        public string? HardwareGeneration { get; set; }
```
Charles-Gagnon, line 13: Are these really all going to be possibly null? We should only use nullable where it's expected to have a possibly null value.

(if you're doing this because of the nullable warnings - it's fine to disable that for contract files like this)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerViewContext.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public override void Dispose()
+        {
+            try
+            {
+                this.Connection.Disconnect();
+            }
+            catch
+            {
+                // ignore
```
Charles-Gagnon, line 26: We should at least log a warning message here I think

src/Microsoft.SqlTools.ServiceLayer/ServerConfigurations/ServerConfigService.cs
```diff
@@ -239,5 +239,27 @@ private ConfigProperty GetSmoConfig(Server server, int configNumber)
...
         }
+
+        public ServerConfigProperty GetConfigByName(Server server, string configName)
+        {
+            try
+            {
+                ConfigProperty serverConfig = null;
+                foreach (ConfigProperty configProperty in server.Configuration.Properties)
+                {
+                    if (configProperty.DisplayName == configName)
```
Charles-Gagnon, line 250: Why are we using display name for this instead of some sort of non-localized value?

Charles-Gagnon (commented):
Please also add a basic test to verify the properties are populated correctly

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+
+                return Task.FromResult(new InitializeViewResult { ViewInfo = this.serverViewInfo, Context = context });
+            }
+        }
+
+        public override Task Save(ServerViewContext context, ServerInfo serverInfo)
+        {
+            var ex = new NotSupportedException();
+            ex.Data["UserMessage"] += "ServerHandler does not support Save method";
```
Charles-Gagnon, line 79: Any reason you're not using the string constructor for the message instead?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public override Task<string> Script(ServerViewContext context, ServerInfo obj)
+        {
+            var ex = new NotSupportedException();
+            ex.Data["UserMessage"] += "ServerHandler does not support Script method";
+            throw ex;
+        }
+        public Task<Server?> CreateServerSmoObject(ServerViewContext context)
+        {
+            using (context.Connection.SqlConnectionObject)
```
Charles-Gagnon, line 91: Are you sure you should be disposing the connection after this method is done? Usually the caller should be the one responsible for the lifetime of the connection.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+            throw ex;
+        }
+
+        public override Task<string> Script(ServerViewContext context, ServerInfo obj)
+        {
+            var ex = new NotSupportedException();
+            ex.Data["UserMessage"] += "ServerHandler does not support Script method";
+            throw ex;
+        }
+        public Task<Server?> CreateServerSmoObject(ServerViewContext context)
```
Charles-Gagnon, line 89: Why is this public? And TBH, this doesn't even really seem necessary to create a whole new method for. Especially if you remove the GetSmoObject call.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            using (context.Connection.SqlConnectionObject)
+            {
+                this.server = new Server(context.Connection);
+                string objectUrn = string.Format(System.Globalization.CultureInfo.InvariantCulture, "Server");
+                var serverSmo = server.GetSmoObject(new Urn(objectUrn)) as Server;
```
Charles-Gagnon, line 95: You already create the server on line 93, why do you need to call GetSmoObject?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/ServerHandlerTests.cs
```diff
@@ -0,0 +1,53 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
```
Charles-Gagnon, line 6: Please don't disable nullable except in contract files.

src/Microsoft.SqlTools.ServiceLayer/ServerConfigurations/ServerConfigService.cs
barbaravaldez: Change it to use the property number instead of the displayName

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
barbaravaldez: I reverted this change

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRequestParams requestParams)
+        {
+            ConnectionInfo connInfo = this.GetConnectionInfo(requestParams.ConnectionUri);
+            ServerConnection serverConnection = ConnectionService.OpenServerConnection(connInfo, ObjectManagementService.ApplicationName);
+
+            using (var context = new ServerViewContext(requestParams, serverConnection))
+            {
+                this.server = new Server(context.Connection); ;
+                if(this.server != null)
```
Charles-Gagnon, line 44: nit: please format these files

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        public override Task<string> Script(ServerViewContext context, ServerInfo obj)
+        {
+            throw new NotSupportedException("ServerHandler does not support Script method");
+        }
+
+        private int GetServerMaxMemory()
+        {
+            return configService.GetSmoConfig(server, 1544).ConfigValue;
```
Charles-Gagnon, line 87: Please make constants for these numbers (I'd add them in the config service file)

Charles-Gagnon, line 87: Could you update this method to be `GetSmoServerConfig`? That's a bit more descriptive.

Charles-Gagnon (approved):

PR #2120, opened 2023-06-23
Fix for schema compare dropping constraints

kisantia:
Fixes https://github.com/microsoft/azuredatastudio/issues/21378. There were a few changes that had to be made in DacFx to fully fix this bug. The first problem was that the method for retrieving the scripts of the differences when publishing changes didn't support scripting inline constraints for databases. This was fixed by switching to using a different method, the same one extract to file uses, for generating the script of elements from a database, which does support scripting inline constraints.

The second problem was that same method was also used in STS for retrieving the scripts to display in schema compare in ADS. Even if the actual publish dropping constraints in problem 1 was fixed, the schema compare differences in ADS still looked like the constraints were getting dropped. New helper functions `SchemaComparisonResult.GetDiffEntrySourceScript()` and `SchemaComparisonResult.GetDiffEntryTargetScript()` needed to be added in DacFx to expose support for scripting inline constraints for schema compare with databases.  

### Example of comparing sql project to db with one column added and one column removed
Before - incorrectly showing constraints are shown as dropped, when there should only be one new and one removed column:
image

Fixed - only column changes are shown:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
@@ -58,9 +58,11 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
                 {
                     string sourceScript = schemaComparisonResult.GetDiffEntrySourceScript(difference);
 
-                    // don't add scripts that start with alter because those are handled by a top level element's create
-                    // ex: if a column changes, then the parent table's script will show an alter, but GetDiffEntrySourceScript()
-                    // will return an alter table script for that column when getting the child script
+                    // Child scripts that do not use alter need to be added if they are being changed, ex: "EXECUTE sp_addextendedproperty...".
+                    // Don't add scripts that start with alter because those are handled by a top level element's create
+                    // ex: if a column changes, then the parent table's script will have the column updated, but GetDiffEntrySourceScript() on the child
+                    // will return an alter table statement for updating that column when getting the child script. The child's alter script is unecessary
+                    // for displaying the script in schema compare because the comparison displays the create scripts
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -56,23 +56,39 @@ internal static DiffEntry CreateDiffEntry(SchemaDifference difference, DiffEntry
...
-                    difference.SourceObject.TryGetScript(out sourceScript);
-                    diffEntry.SourceScript = FormatScript(sourceScript);
+                    string sourceScript = schemaComparisonResult.GetDiffEntrySourceScript(difference);
+
+                    // Child scripts that do not use alter need to be added if they are being changed, ex: "EXECUTE sp_addextendedproperty...".
+                    // Don't add scripts that start with alter because those are handled by a top level element's create
+                    // ex: if a column changes, then the parent table's script will have the column updated, but GetDiffEntrySourceScript() on the child
+                    // will return an alter table statement for updating that column when getting the child script. The child's alter script is unecessary
+                    // for displaying the script in schema compare because the comparison displays the create scripts
+                    if (!sourceScript.ToLowerInvariant().StartsWith("alter"))
```
kisantia, line 66: This check wasn't needed before because `TryGetScript()` didn't return `Alter` scripts for changes like columns that are covered in the parent's script. Now `GetDiffEntrySourceScript()` uses 2 different ways to get scripts for the differences
- database - uses `ElementScripter.Script()` which returns `Alter` scripts for some child differences
- dacpac and project - uses `service.TryGetScript() ` which doesn't return anything for those child differences

I'm not sure why they're different, but I'm guessing it has something to do with the different internal and public models in DacFx and them not being consistent, similar to inline constraints only being supported for databases with one and not the other.

SakshiS-harma (approved):

PR #2121, opened 2023-06-26
Sai/db properties options tab

ssreerama:
This PR adds ability to view/modify the database properties of options tab.

image


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -259,7 +293,7 @@ public string Detach(DetachDatabaseRequestParams detachParams)
         {
             var sqlScript = string.Empty;
             ConnectionInfo connectionInfo = this.GetConnectionInfo(detachParams.ConnectionUri);
-            using (var dataContainer = CreateDatabaseDataContainer(detachParams.ConnectionUri, detachParams.ObjectUrn, false))
+            using (var dataContainer = CreateDatabaseDataContainer(detachParams.ConnectionUri, detachParams.ObjectUrn, false, ""))
```
corivera, line 296: You could use null for the database name here to avoid creating a new string since you already have a nullable string argument.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -401,10 +440,48 @@ private string ConfigureDatabase(InitializeViewRequestParams viewParams, Databas
...
+                        }
+                        if (prototype is DatabasePrototype160 db160)
+                        {
+                            if (database.IsLedgerDatabase != null)
+                            {
+                                db160.IsLedger = (bool)database.IsLedgerDatabase;
+                            }
+                        }
+
+                        // AutoCreateStatisticsIncremental can only be set when AutoCreateStatistics is enabled
```
corivera, line 479: Do you need to add an if case here to account for this condition?

ssreerama: We do not show any message or error when the AutoCreateStatisticsIncremental is set to true when the other one is set to false, we are just closing the window when saved and doing no updates to this property. I found these are interrelated and added a comment here. Also, wouldn't cause anything if we add a condition here!

corivera (approved):

PR #2124, opened 2023-06-27
Use config instead of parameter for query editor result copy and fix some wording

caohai:


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -789,7 +789,7 @@ internal async Task HandleCopyResultsRequest(CopyResultsRequestParams requestPar
                             }
                         }
                         // Add line break if this is not the last row in all selections.
-                        if (rowIndex + pageStartRowIndex != lastRowIndex && (!builder.ToString().EndsWith(Environment.NewLine) || !requestParams.AvoidNewLineAfterTailingLineBreak))
+                        if (rowIndex + pageStartRowIndex != lastRowIndex && (!builder.ToString().EndsWith(Environment.NewLine) || !requestParams.SkipNewLineAfterTrailingLineBreak))
```
Charles-Gagnon, line 792: Why is this even a request param anyways? It's a global setting so you should probably just be getting it from the general config IMO

Charles-Gagnon: See https://github.com/microsoft/sqltoolsservice/blob/a37093a773ab92d646768005be1b834a217c66a8/src/Microsoft.SqlTools.ServiceLayer/SqlContext/SqlToolsSettings.cs for examples of these settings. STS will get notified whenever a config value is updated so that reduces the load on our end to send this stuff ourselves.

caohai: fixed

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
```diff
@@ -22,21 +22,6 @@ public class TableSelectionRange
     /// </summary>
     public class CopyResultsRequestParams : SubsetParams
     {
-        /// <summary>
```
caohai, line 25: @alanrenmsft FYI I'm updating this part and switching to read from config for these 3 settings. I've tested all 3 settings and they are working as expected.

alanrenmsft (approved):

Charles-Gagnon (approved):
Awesome, thanks for taking care of the other ones as well!

PR #2126, opened 2023-06-28
Update memory settings and implement ServerPrototype

barbaravaldez:
Add method for updating max and min memory settings.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -91,5 +92,23 @@ private int GetServerMinMemory()
...
             return configService.GetServerSmoConfig(server, configService.MinServerMemoryPropertyNumber).ConfigValue;
         }
+
+        private void UpdateServerProperties(InitializeViewRequestParams viewParams, ServerInfo serverObj)
+        {
+            if (viewParams != null)
+            {
+                ConnectionInfo connInfo = this.GetConnectionInfo(viewParams.ConnectionUri);
+                ServerConnection serverConnection = ConnectionService.OpenServerConnection(connInfo, ObjectManagementService.ApplicationName);
+                this.server = new Server(serverConnection);
```
Charles-Gagnon, line 102: Missed this earlier - we shouldn't be storing state like this. This is a singleton instance for handling ALL requests so the server will change from request to request. And while it's unlikely to have two overlapping given how the UI works it's still something we should write this code in a functional way.

Look at how the other handlers are doing this for an example

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -91,5 +92,23 @@ private int GetServerMinMemory()
...
+                if (GetServerMaxMemory() != serverObj.MaxServerMemory)
+                {
+                    configService.UpdateConfig(serverConnection, configService.MaxServerMemoryPropertyNumber, serverObj.MaxServerMemory);
```
Charles-Gagnon, line 105: Normally we should be updating the server SMO object and then saving that - is there a reason you decided to do it this way? If doing separate queries it feels like this should be done in a transaction too so we rollback all the changes if any errors occur...

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -91,5 +92,23 @@ private int GetServerMinMemory()
...
+                }
+                else if (GetServerMinMemory() != serverObj.MinServerMemory)
```
corivera, line 107: Are we able to update both min memory and max memory in the same operation?  With this else statement we always skip min memory updates if there's a max memory update.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
barbaravaldez: I decided to go this route because we already have the updateConfig method in the serverConfig but I agree that when modifying multiple properties at the same time then it's not the best approach.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerActions.cs
```diff
@@ -0,0 +1,45 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
```
Charles-Gagnon, line 6: Please remove this and fix the nullable issues (only the "contract" files should have this left in)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,805 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
```
Charles-Gagnon, line 6: Remove and fix issues

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,805 @@
...
+using System;
+using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlTools.ServiceLayer.Management;
+using Microsoft.SqlServer.Management.Common;
+using Microsoft.SqlTools.ServiceLayer.ServerConfigurations;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    /// <summary>
+    /// ServerGeneral - main app server page
```
Charles-Gagnon, line 17: Update - there shouldn't be any UI terms used in STS

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,805 @@
...
+        private ServerConnection sqlConnection = null;
+        private ServerConfigService configService = null;
+
+
+        private ServerPrototypeData currentState;
+        private ServerPrototypeData originalState;
+        #endregion
+
+        #region Trace support
+        private const string componentName = "ServerGeneral";
```
Charles-Gagnon, line 38: Why ServerGeneral?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,805 @@
...
+            }
+
+        }
+        #endregion
+
+        public void ApplyInfoToPrototype(ServerInfo serverInfo)
+        {
+            this.Name = serverInfo.Name;
+            this.Language = serverInfo.Language;
+            this.Memory = serverInfo.MemoryInMB;
```
Charles-Gagnon, line 343: Should this be `MemoryInMB` as well?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,805 @@
...
+                    {
+                        LoadData();
+                    }
+
+                    return this.serverName;
+                }
+
+                set
+                {
+                    System.Diagnostics.Debug.Assert(this.initialized, "unexpected property set before initialization");
```
Charles-Gagnon, line 407: Remove all instances of Debug.Assert. We should either be throwing an exception or logging it and providing some appropriate default behavior.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,805 @@
...
+            {
+            }
+
+
+            /// <summary>
+            /// constructor
+            /// </summary>
+            /// <param name="context">The context in which we are modifying an existing serverRole</param>
+            /// <param name="serverRole">The serverRole we are modifying</param>
+            public ServerPrototypeData(CDataContainer context, Server server, ServerConfigService service)
```
Charles-Gagnon, line 749: Server, not ServerRole

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -91,5 +95,33 @@ private int GetServerMinMemory()
...
+                ServerConnection serverConnection = ConnectionService.OpenServerConnection(connInfo, ObjectManagementService.ApplicationName);
+                CDataContainer dataContainer = CDataContainer.CreateDataContainer(connInfo);
+
+                ServerPrototype prototype = new ServerPrototype(dataContainer);
+                prototype.ApplyInfoToPrototype(serverInfo);
+                ConfigureServerRole(dataContainer, ConfigAction.Update, RunType.RunNow, prototype);
+            }
+        }
+
+        private void ConfigureServerRole(CDataContainer dataContainer, ConfigAction configAction, RunType runType, ServerPrototype prototype)
```
Charles-Gagnon, line 113: `ConfigureServer`?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -91,5 +95,33 @@ private int GetServerMinMemory()
...
+        {
+            string sqlScript = string.Empty;
+            using (var actions = new ServerActions(dataContainer, prototype, configAction))
+            {
+                var executionHandler = new ExecutonHandler(actions);
+                executionHandler.RunNow(runType, this);
+                if (executionHandler.ExecutionResult == ExecutionMode.Failure)
+                {
+                    throw executionHandler.ExecutionFailureException;
```
Charles-Gagnon, line 122: Is this going to have information about why it failed?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -91,5 +95,33 @@ private int GetServerMinMemory()
         {
             return configService.GetServerSmoConfig(server, configService.MinServerMemoryPropertyNumber).ConfigValue;
         }
+
+        private void UpdateServerProperties(InitializeViewRequestParams viewParams, ServerInfo serverInfo)
+        {
+            if (viewParams != null)
+            {
+                ConnectionInfo connInfo = this.GetConnectionInfo(viewParams.ConnectionUri);
+                ServerConnection serverConnection = ConnectionService.OpenServerConnection(connInfo, ObjectManagementService.ApplicationName);
```
Charles-Gagnon, line 104: Make sure this connection is closed

Charles-Gagnon (dismissed):
Please add tests as well

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
barbaravaldez: It was for general properties but I'm including memory settings as well. So, I removed it.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
barbaravaldez: I checked and this is the same implementation that ServerRoleHandler and DatabaseHandler follow. Should I append a more descriptive error to the exception data property or create a new exception?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerActions.cs
```diff
@@ -0,0 +1,44 @@
...
+    internal class ServerActions : ManagementActionBase
+    {
+        #region Variables
+        private ServerPrototype serverData = null;
+        private ConfigAction configAction;
+        #endregion
+
+        #region Constructor
+        /// <summary>
+        /// required when loading from Object Explorer context
```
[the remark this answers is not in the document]
barbaravaldez: I used another class and then modified. Forgot to remove this comments

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,1005 @@
...
+
+            if (this.currentState.MinMemory > this.currentState.MaxMemory)
+            {
+                throw new InvalidOperationException(minMemoryCannotBeGreaterThanMaxMemory);
+            }
+
+            if (this.currentState.MinMemory != this.originalState.MinMemory)
+            {
+                changesMade = true;
+                if (this.currentState.MinMemory > this.serverMaxMemoryProperty.ConfigValue)
```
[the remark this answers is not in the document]
barbaravaldez: Yup, that's something that I just started considering. I'll probably send another pr tomorrow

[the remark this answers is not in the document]
barbaravaldez: I'll remove them from the server and send a different pr tomorrow. You are right after giving it a little bit more thought this should be done in the UI

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
corivera: Looks like this property is just a cached version of the last thrown exception, so it will still have any relevant error messages.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/ServerHandlerTests.cs
```diff
@@ -47,5 +49,59 @@ public async Task GetServerProperties()
...
+                    Version = result.Version,
+                    MinServerMemory = result.MinServerMemory,
+                    MaxServerMemory = result.MaxServerMemory
+                };
+
+                // Change memory settings
+                serverInfo.MinServerMemory = 10;
+                serverInfo.MaxServerMemory = 500;
+
+                Assert.AreNotEqual(result.MinServerMemory, serverInfo.MinServerMemory, "Server property should not be equal after update");
```
corivera, line 96: We've started using the NUnit constraints for our newer C# tests. It helps make things a bit more readable if you want to use them with these tests. An example of the constraints: https://docs.nunit.org/articles/nunit/writing-tests/constraints/AndConstraint.html#examples-of-use

barbaravaldez: I'll send out another pr tomorrow to add minimum and maximum values of the memory properties. I'll make sure to update the tests :)

barbaravaldez: I just want to make sure this change gets in for the bug bash

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -0,0 +1,971 @@
...
+using System;
+using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlTools.ServiceLayer.Management;
+using Microsoft.SqlServer.Management.Common;
+using Microsoft.SqlTools.ServiceLayer.ServerConfigurations;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
+{
+    internal class ServerPrototype
```
corivera, line 15: Could this file be renamed to ServerPrototype.cs to better reflect the contents?

corivera (approved):
Approved pending my other comments.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
barbaravaldez: Addressing this in next pr

PR #2127, opened 2023-07-05
Support enabling connection pooling

cheenamalhotra:
Primary benefit:
- Reduces connection opening time by re-using physical server connections where possible.

Enable connection pooling with setting:
image

_P.S. I will be working on adding tests separately to capture event counters for when connections are opened/closed and various other events, tracking it as a separate activity._

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -173,6 +173,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         public bool EnableSqlAuthenticationProvider { get; set; }
 
         /// <summary>
+        /// Enables connection pooling for all SQL connections, removing feature name identifier from application name to prevent unwanted connection pools.
+        /// </summary>
+        public static bool EnableConnectionPooling { get; set; }
+
+        /// <summary>
         /// Returns a connection queue for given type
         /// </summary>
         /// <param name="type"></param>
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -474,8 +479,8 @@ private void TryCloseConnectionTemporaryConnection(ConnectParams connectionParam
         internal static string GetApplicationNameWithFeature(string applicationName, string featureName)
         {
             string appNameWithFeature = applicationName;
-
-            if (!string.IsNullOrWhiteSpace(applicationName) && !string.IsNullOrWhiteSpace(featureName) && !applicationName.EndsWith(featureName))
+            // Connection Service will not set custom application name if connection pooling is enabled on service.
```
kisantia, line 482: do different features in ADS currently have their own custom application name for their connections? It's not like all the connections from ADS have ADS as the custom application name?

cheenamalhotra: Yeah currently we have feature names appended e.g. azdata-languageService, azdata-TableDesigner, azdata-GeneralConnection, azdata-ObjectExplorer and so on... 

But when connection pooling is enabled we need to make sure there is no change of connection string for one connection (as configured by user) - as otherwise the more different connection strings we send to driver - it will lead to multiple connection pools - which is not desired.

kisantia: Makes sense. And just checking, it doesn't make sense/add any value when connection pooling is enabled to have the application name be azdata or another name to distinguish that these connections are from ADS? I'm assuming that the connection pooling here is meant to only share the connections ADS makes. 

Just wondering if it could help in any debugging scenarios to have the application name set to something that identifies that it's coming from ADS (one time someone from the connectivity team asked me what the application name from connections from ADS was when they were investigating an issue).

cheenamalhotra: > I'm assuming that the connection pooling here is meant to only share the connections ADS makes.

Connection pooling is client side, so it happens within the scope of sqltoolsservice.exe. 'azdata' is our identifier, so all connections containing 'azdata' in application name are used for tracking our connection activity.

> Just wondering if it could help in any debugging scenarios to have the application name set to something that identifies that it's coming from ADS

This doesn't change anything for debugging purposes, just that now we won't track feature wise connections - but only 'azdata' connections to maintain healthy connection pools for optimum performance.

Do you mean to say we should change 'azdata' to something else?

kisantia: got it, thanks for explaining! My confusion was that I had thought the whole application name was no longer getting set, but now I see that it's just the feature part of it (was reading too quickly earlier). Nope not suggesting changing 'azdata' to something else.

kisantia (approved):

PR #2128, opened 2023-07-06
Add explicit CREATE TABLE permission check for table designer

caohai:
Part of fix for https://github.com/microsoft/azuredatastudio/issues/23519
1. Not doing `ALTER` check here after some experiments as it can get complicated.
2. The model loading error is thrown from DacFx and I'll need to dig into that code path later to see if it's possible to give the actual error instead of throwing a generic error.

image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
@@ -1805,15 +1805,21 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
                 connectionStringBuilder.InitialCatalog = tableInfo.Database;
                 connectionStringBuilder.ApplicationName = ConnectionService.GetApplicationNameWithFeature(connectionStringBuilder.ApplicationName, TableDesignerService.TableDesignerApplicationNameSuffix);
                 var connectionString = connectionStringBuilder.ToString();
-
-                // Sanity checks, passing these checks doesn't guarantee that the user can open the table designer.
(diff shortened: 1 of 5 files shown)


caohai:
> @caohai can you elaborate a bit about the alter scenario?

I initially thought some of the operations (like CREATE INDEX) when designing an existing table doesn't need ALTER TABLE permission which is not true. I've added back the ALTER TABLE permission check. Can you take a look at that query?

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -1802,6 +1805,9 @@ private Dac.TableDesigner CreateTableDesigner(TableInfo tableInfo)
                 connectionStringBuilder.InitialCatalog = tableInfo.Database;
                 connectionStringBuilder.ApplicationName = ConnectionService.GetApplicationNameWithFeature(connectionStringBuilder.ApplicationName, TableDesignerService.TableDesignerApplicationNameSuffix);
                 var connectionString = connectionStringBuilder.ToString();
+
+                // Sanity checks, passing these checks doesn't guarantee that the user can open the table designer.
+                CheckPermissions(tableInfo, connectionString);
```
alanrenmsft, line 1810: unfortunately, this is going to have negative impact on the loading time, I am wondering whether we can do the check only after an exception is caught, so that the happy path won't be impacted.

caohai: makes sense, I've moved the permission check into the catch part

alanrenmsft (approved):

PR #2129, opened 2023-07-07
Close connection explicitly on rename

cheenamalhotra:
Addresses https://github.com/microsoft/azuredatastudio/issues/23633

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypeHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypeHandler.cs
@@ -74,6 +74,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                     throw new Exception(SR.ObjectNotRenamable(objectUrn));
                 }
             }
+            // If connection is not instantiated in the using statement, it can be leaked here, therefore close explicitly.
+            if(serverConnection.SqlConnectionObject.State == System.Data.ConnectionState.Open)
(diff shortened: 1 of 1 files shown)


aasimkhan30 (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypeHandler.cs
```diff
@@ -74,6 +74,11 @@ public virtual Task Rename(string connectionUri, string objectUrn, string newNam
                     throw new Exception(SR.ObjectNotRenamable(objectUrn));
                 }
             }
+            // If connection is not instantiated in the using statement, it can be leaked here, therefore close explicitly.
```
Charles-Gagnon, line 77: I'm not sure I understand this - using should always call Dispose on the object. Is there something special about these connections that make it so they don't actually get closed even when Disposed?

Sample I was playing around with here : https://dotnetfiddle.net/A8LWSE

cheenamalhotra: Here's a quick repro, and on debug I found this was happening too..

image

Charles-Gagnon: So you're saying somewhere in the using statement the actual underlying connection object is being swapped out? With the result that the old one is disposed (and closed) correctly, but the new one isn't?

PR #2130, opened 2023-07-07
Send min and max values for server properties

barbaravaldez:
- Send min and max values for memory properties
- Address remaining comments from yesterday's PR


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerInfo.cs
```diff
@@ -28,7 +28,14 @@ public class ServerInfo : SqlObject
         public string? ServiceTier { get; set; }
         public int? StorageSpaceUsageInMB { get; set; }
         public string Version { get; set; }
-        public int MaxServerMemory { get; set; }
-        public int MinServerMemory { get; set; }
+        public ServerProperty MaxServerMemory { get; set; }
+        public ServerProperty MinServerMemory { get; set; }
+    }
+
+    public class ServerProperty
```
Charles-Gagnon, line 35: Maybe `NumericServerProperty`? Since we can have other types of server properties too

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -68,43 +68,44 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
-    }
+        public override Task<string> Script(ServerViewContext context, ServerInfo obj)
+        {
+            throw new NotSupportedException("ServerHandler does not support Script method");
+        }
 
-    private void UpdateServerProperties(InitializeViewRequestParams viewParams, ServerInfo serverInfo)
-    {
-        if (viewParams != null)
+
```
corivera, line 84: Extra space here

corivera (approved):

PR #2143, opened 2023-07-12
[Release/4.8] Only set Restrict Access on a database if a non-null value is provided (#2140)

corivera:
Fixes https://github.com/microsoft/azuredatastudio/issues/23729

Porting https://github.com/microsoft/sqltoolsservice/commit/88ec8713d0149f512f7a8d74c40816ec2628fa95 to release branch.


Charles-Gagnon (approved):

Charles-Gagnon:
Note you'll have to manually kick off an STS build once this is merged in targeting this branch. That'll generate the new STS release - after that the update process is the same as always.

kisantia (approved):

PR #2144, opened 2023-07-12
Fix for Database prototype file is not log file error

ssreerama:
This PR fixes the error that is causing when updating the database properties.

Error: "Database prototype's second file was not a Log file"

ADS Issue: https://github.com/microsoft/azuredatastudio/issues/23721 
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
@@ -401,8 +401,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                         prototype.Name = database.Name;
 
                         // Update database file names now that we have a database name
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -401,8 +401,7 @@ private string ConfigureDatabase(InitializeViewRequestParams viewParams, Databas
                         prototype.Name = database.Name;
 
                         // Update database file names now that we have a database name
-                        // Modifying logical file name is not supported in SQL Database Managed Instance.
-                        if (!prototype.HideFileSettings && dataContainer.Server.DatabaseEngineEdition != DatabaseEngineEdition.SqlManagedInstance)
```
ssreerama, line 405: Removing MI condition, as it was added to fix the same issue for MI db properties, but the IsNewObject check fixes all cases.

corivera (approved):

PR #2145, opened 2023-07-13
Removed advanced options from Table Designer tooltip

smartguest:
companion PR for https://github.com/microsoft/azuredatastudio/pull/23873

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
@@ -1845,12 +1845,6 @@ namespace Microsoft.SqlTools.ServiceLayer.TableDesigner
         private void UpdateTableTitleInfo(TableInfo tableInfo)
         {
             var td = GetTableDesigner(tableInfo);
-            // Disabled due to issues with title generation.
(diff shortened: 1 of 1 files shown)


aasimkhan30 (approved):

src/Microsoft.SqlTools.ServiceLayer/TableDesigner/TableDesignerService.cs
```diff
@@ -1845,14 +1845,15 @@ private Dac.TableDesigner GetTableDesigner(TableInfo tableInfo)
         private void UpdateTableTitleInfo(TableInfo tableInfo)
         {
             var td = GetTableDesigner(tableInfo);
-            var advancedOpsIndex = tableInfo.Tooltip.LastIndexOf('[');
-            var advancedOps = "";
-            if(advancedOpsIndex > -1){
-                advancedOps = " " + tableInfo.Tooltip.Substring(advancedOpsIndex);
-            }
+            // Disabled due to issues with title generation.
```
Charles-Gagnon, line 1848: Just delete this stuff IMO - as Cheena pointed out in the other one we can always revert a commit or refer back to the original code if/when we add this back in.

Charles-Gagnon (approved):

cheenamalhotra (approved):

PR #2147, opened 2023-07-14
Enabling Database scoped configurations tab to DB Properties

ssreerama:
This PR adds the ability to view database scoped configurations options of a database in a separate DSC tab of database properties. 

These changes are send the real values for the dsc options to the UI from SMO for a specific supported dsc properties and saving/scripting the changes.


src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                if (db.DatabaseScopedConfigurations[i].Value != this.currentState.databaseScopedConfigurations[i].Value)
+                {
+                    db.DatabaseScopedConfigurations[i].Value = this.currentState.databaseScopedConfigurations[i].Value;
+                }
+
+                // Below propertiesoption does not allow updates for the secondaries replica while this option is only allowed to be set for the primary.
+                // ELEVATE_ONLINE(11), ELEVATE_RESUMABLE(12), GLOBAL_TEMPORARY_TABLE_AUTO_DROP(21), IDENTITY_CACHE(6), PAUSED_RESUMABLE_INDEX_ABORT_DURATION_MINUTES(25)
+                if (db.DatabaseScopedConfigurations[i].ValueForSecondary != this.currentState.databaseScopedConfigurations[i].ValueForSecondary
+                    && !new[] { 6, 11, 12, 21, 25 }.Contains(db.DatabaseScopedConfigurations[i].Id))
```
Charles-Gagnon, line 55: Define this one time and reuse, no reason to create a new array each time

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -0,0 +1,62 @@
...
+            base.SaveProperties(db);
+
+            for(int i = 0; i < db.DatabaseScopedConfigurations.Count; i++)
+            {
+                if (db.DatabaseScopedConfigurations[i].Value != this.currentState.databaseScopedConfigurations[i].Value)
+                {
+                    db.DatabaseScopedConfigurations[i].Value = this.currentState.databaseScopedConfigurations[i].Value;
+                }
+
+                // Below propertiesoption does not allow updates for the secondaries replica while this option is only allowed to be set for the primary.
```
Charles-Gagnon, line 52: Are there docs or something you could link to for explaining this more thoroughly?

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -0,0 +1,62 @@
...
+                // ELEVATE_ONLINE(11), ELEVATE_RESUMABLE(12), GLOBAL_TEMPORARY_TABLE_AUTO_DROP(21), IDENTITY_CACHE(6), PAUSED_RESUMABLE_INDEX_ABORT_DURATION_MINUTES(25)
```
Charles-Gagnon, line 53: Reworded to read slightly clearer based on my understanding of why this is being done, feel free to make any corrections if I got something wrong

```suggestion
                // The below configurations do not allow updates for the secondary replica - they are only applied to the primary
                // IDENTITY_CACHE(6)
                // ELEVATE_ONLINE(11)
                // ELEVATE_RESUMABLE(12)
                // GLOBAL_TEMPORARY_TABLE_AUTO_DROP(21)
                // PAUSED_RESUMABLE_INDEX_ABORT_DURATION_MINUTES(25)
```

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
```diff
@@ -2787,6 +2787,10 @@ prototype_db_prop_encryptionEnabled = Encryption Enabled
 prototype_db_prop_databasescopedconfig_value_off = OFF
 prototype_db_prop_databasescopedconfig_value_on = ON
 prototype_db_prop_databasescopedconfig_value_primary = PRIMARY
+prototype_db_prop_databasescopedconfig_value_when_supported = WHEN_SUPPORTED
```
Charles-Gagnon, line 2790: These shouldn't actually be localized, right?

And actually, same for the above (OFF/ON/PRIMARY) - generally we shouldn't be localizing the actual values if they map directly to T-SQL options (and instead rely on descriptions to tell the user what they mean with a localized string if needed)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -463,6 +493,35 @@ private string ConfigureDatabase(InitializeViewRequestParams viewParams, Databas
...
+                        {
+                            if (databaseScopedConfigurationsCollection != null)
+                            {
+                                foreach (DatabaseScopedConfigurationsInfo dsc in database.DatabaseScopedConfigurations)
+                                {
+                                    foreach (DatabaseScopedConfiguration smoDscCollection in databaseScopedConfigurationsCollection)
+                                    {
+                                        if (smoDscCollection.Name == dsc.Name)
+                                        {
+                                            smoDscCollection.Value = dsc.ValueForPrimary == SR.prototype_db_prop_databasescopedconfig_value_fail_enabled
```
Charles-Gagnon, line 506: Isn't this going to be a problem with the localized strings? Yet another reason to not use localized strings for the option values IMO

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -463,6 +493,35 @@ private string ConfigureDatabase(InitializeViewRequestParams viewParams, Databas
...
+                                                ? "1" : dsc.ValueForPrimary == SR.prototype_db_prop_databasescopedconfig_value_fail_disabled
+                                                ? "0" : dsc.ValueForPrimary;
+
+                                            // seconday value needs to set back to 'PRIMARY' to avoid unnecessary script generation for few properties
```
Charles-Gagnon, line 510: "for few properties" - what does this mean? If any of the values are changed shouldn't we be generating the script?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -883,5 +942,28 @@ private static AzureEditionDetails[] GetAzureMaxSizes(IEnumerable<AzureEdition>
...
+        /// Prepares database scoped configurations list
+        /// </summary>
+        /// <param name="smoDSCMetaData"></param>
+        /// <returns></returns>
+        private DatabaseScopedConfigurationsInfo[] GetDSCMetaData(DatabaseScopedConfigurationCollection smoDSCMetaData)
+        {
+            var dscMetaData = new List<DatabaseScopedConfigurationsInfo>();
+            foreach (DatabaseScopedConfiguration dsc in smoDSCMetaData)
+            {
+                // Options MAXDOP(Id = 1) and PAUSED_RESUMABLE_INDEX_ABORT_DURATION_MINUTES(Id = 25) are integer numbers but coming as string value type, need exclusion from the below conversion
```
Charles-Gagnon, line 956: Would suggest creating a helper method for this - makes it easier to update and reuse. I'd also suggest using a switch statement - bit easier to read and update later on.

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
ssreerama: We have a MSDoc and linking in the comment. Thanks

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: I'll update the comment more meaningful!

FYI- We display primaryValue to the secondary when it has the 'PRIMARY' value. When we are sending back to SMO, we must update back to PRIMARY if not changed, so that smo would not generate alter dsc for secondary script, otherwise it generates the below unnecessary syntax. 

ALTER DATABASE SCOPED CONFIGURATION FOR SECONDARY SET **DSC_PROPERTY_HERE** = OFF

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
ssreerama: We are localizing the options values in STS and sending them to ADS UI.

@corivera correct me if I'm wrong! thanks

corivera: I would agree with Charles' opinion here. Were these strings that were already being localized in SSMS?

corivera: I also took a look at translated versions of the TSQL page for ALTER DATABASE SCOPED CONFIGURATION and the TSQL arguments are not translated either.

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                // The below configurations do not allow updates for the secondary replica - they are only applied to the primary.
+                // The excludedPropertiesArray containst the configurtation Ids of the below properties
+                // IDENTITY_CACHE(6)
+                // ELEVATE_ONLINE(11)
+                // ELEVATE_RESUMABLE(12)
+                // GLOBAL_TEMPORARY_TABLE_AUTO_DROP(21)
+                // PAUSED_RESUMABLE_INDEX_ABORT_DURATION_MINUTES(25)
+                if (db.DatabaseScopedConfigurations[i].ValueForSecondary != this.currentState.databaseScopedConfigurations[i].ValueForSecondary
+                    && !excludedPropertiesArray.Contains(db.DatabaseScopedConfigurations[i].Id))
```
corivera, line 64: Since the array is only being used for Contains, you could replace it with a static HashSet for better performance

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -300,6 +300,8 @@ public async Task VerifyDatabasePropertiesTest()
                     Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).AutoUpdateStatisticsAsynchronously, Is.False, $"AutoUpdateStatisticsAsynchronously should match with testdata");
                     Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).PageVerify, Is.EqualTo(testDatabase.PageVerify), $"PageVerify should match with testdata");
                     Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).RestrictAccess, Is.EqualTo(testDatabase.RestrictAccess), $"RestrictAccess should match with testdata");
+                    Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).DatabaseScopedConfigurations, Is.Not.Null, $"DatabaseScopedConfigurations is not null");
+                    Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).DatabaseScopedConfigurations.Count, Is.GreaterThan(0), $"DatabaseScopedConfigurations should have atleast few properties");
```
corivera, line 304:
```suggestion
                    Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).DatabaseScopedConfigurations.Count, Is.GreaterThan(0), $"DatabaseScopedConfigurations should have at least a few properties");
```

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.strings
corivera: No, I was looking at the docs page. Here's the page in spanish, for example: https://learn.microsoft.com/es-es/sql/t-sql/statements/alter-database-scoped-configuration-transact-sql?view=sql-server-ver16

You can change the locale ID after the ".com" to see other variations

ssreerama: Good to know, Verified the PageVerify and restrictAccess for Spanish and the Page Verify options are not translated there too! Will send a new PR of Un localizing the options.

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+
+        protected override void SaveProperties(Database db)
+        {
+            base.SaveProperties(db);
+
+            // Properties that doen't support secondary value updates
+            // More info here: https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-database-scoped-configuration-transact-sql?view=sql-server-ver16
+            HashSet<int> excludedPropertiesSet = new HashSet<int> { 6, 11, 12, 21, 25 };
```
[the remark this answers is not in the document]
Charles-Gagnon: And give it a more descriptive var name - "excluded properties" is a bit too general (the comment is good though, thank you!)

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -6702,10 +6702,6 @@ The Query Processor estimates that implementing the following index could improv
     <value>ON</value>
```
Charles-Gagnon, line 6702: There's a number of other values here that look like they shouldn't be localized. In general - if the value is passed as an option to the T-SQL (which most options are) then it should NOT be localized. 

Could you make a work item to follow up on verifying that you're handling these correctly? An easy way to check is by changing the name here in the resx and then trying to update options using these values. That should still work fine since these should be display names only.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -883,5 +938,51 @@ private static AzureEditionDetails[] GetAzureMaxSizes(IEnumerable<AzureEdition>
             }
             return sizes.ToArray();
         }
+
+        /// <summary>
+        /// Prepares database scoped configurations list
+        /// </summary>
+        /// <param name="smoDSCMetaData"></param>
+        /// <returns></returns>
+        private DatabaseScopedConfigurationsInfo[] GetDSCMetaData(DatabaseScopedConfigurationCollection smoDSCMetaData)
```
Charles-Gagnon, line 947: These look like they could be static

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -300,6 +300,8 @@ public async Task VerifyDatabasePropertiesTest()
                     Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).AutoUpdateStatisticsAsynchronously, Is.False, $"AutoUpdateStatisticsAsynchronously should match with testdata");
                     Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).PageVerify, Is.EqualTo(testDatabase.PageVerify), $"PageVerify should match with testdata");
                     Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).RestrictAccess, Is.EqualTo(testDatabase.RestrictAccess), $"RestrictAccess should match with testdata");
+                    Assert.That(((DatabaseInfo)databaseViewInfo.ObjectInfo).DatabaseScopedConfigurations, Is.Not.Null, $"DatabaseScopedConfigurations is not null");
```
Charles-Gagnon, line 303: Please add some tests for updating the values as well.

src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
ssreerama: Created a task https://github.com/microsoft/azuredatastudio/issues/24084!

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -0,0 +1,71 @@
...
+                    db.DatabaseScopedConfigurations[i].Value = this.currentState.databaseScopedConfigurations[i].Value;
+                }
+
+                // The below configurations do not allow updates for the secondary replica - they are only applied to the primary.
+                // The secondaryValUnsupportedPropsSet containst the configurtation Ids of the below properties
+                // IDENTITY_CACHE(6)
+                // ELEVATE_ONLINE(11)
+                // ELEVATE_RESUMABLE(12)
+                // GLOBAL_TEMPORARY_TABLE_AUTO_DROP(21)
+                // PAUSED_RESUMABLE_INDEX_ABORT_DURATION_MINUTES(25)
```
Charles-Gagnon, line 62: Consider moving this comment up to where you define the set

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            for (int i = 0; i < db.DatabaseScopedConfigurations.Count; i++)
+            {
+                if (db.DatabaseScopedConfigurations[i].Value != this.currentState.databaseScopedConfigurations[i].Value)
+                {
+                    db.DatabaseScopedConfigurations[i].Value = this.currentState.databaseScopedConfigurations[i].Value;
+                }
+
+                // The below configurations do not allow updates for the secondary replica - they are only applied to the primary.
+                // The secondaryValUnsupportedPropsSet containst the configurtation Ids of the below properties
```
Charles-Gagnon, line 57:
```suggestion
                // The secondaryValUnsupportedPropsSet containst the configuration Ids of the below properties
```

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype100.cs
```diff
@@ -199,51 +199,6 @@ protected override void SaveProperties(Database db)
             {
                 db.EncryptionEnabled = this.currentState.encryptionEnabled;
             }
-
-            // Check if we support database scoped configurations in this database. Since these were all added at the same time,
-            // only check if MaxDop is supported rather than each individual property.
-            if (db.IsSupportedProperty("MaxDop"))
```
corivera, line 205: Why did this code get removed? I'm not seeing it moved to anywhere else.

ssreerama: Since databaseScopedConfigurations are listed from the databse.DatabaseScopedCofigurationCollections, having these here has no use, as I don't see these are using anywhere. And when we have new dsc props, these old 4 properties are being overriding the new big list of dsc properties which causes an issue. So, removed them from saving properties.

Charles-Gagnon: If these existed in 100 shouldn't they stay here though?

Charles-Gagnon: Gotcha, that explanation makes sense to me. Thank you!

PR #2151, opened 2023-07-18
Send user as empty string (default value)

cheenamalhotra:
Changes to address https://github.com/microsoft/azuredatastudio/issues/23921 - allow empty string to be treated as a default for "user".

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -1732,7 +1732,9 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
                 ServerName = defaultBuilder.DataSource != builder.DataSource ? builder.DataSource : null,
                 TrustServerCertificate = defaultBuilder.TrustServerCertificate != builder.TrustServerCertificate ? builder.TrustServerCertificate : false,
                 TypeSystemVersion = defaultBuilder.TypeSystemVersion != builder.TypeSystemVersion ? builder.TypeSystemVersion : null,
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1732,7 +1732,9 @@ public ConnectionDetails ParseConnectionString(string connectionString)
                 ServerName = defaultBuilder.DataSource != builder.DataSource ? builder.DataSource : null,
                 TrustServerCertificate = defaultBuilder.TrustServerCertificate != builder.TrustServerCertificate ? builder.TrustServerCertificate : false,
                 TypeSystemVersion = defaultBuilder.TypeSystemVersion != builder.TypeSystemVersion ? builder.TypeSystemVersion : null,
-                UserName = defaultBuilder.UserID != builder.UserID ? builder.UserID : null,
+                // !!! ALERT - DO NOT CHANGE USER !!!
+                // SSMS 19 treats "user" as mandatory, always set it to value from connection string builder, even if it's an empty string.
+                UserName = builder.UserID,
```
Charles-Gagnon, line 1737: Did you verify that there aren't any other properties that were changed to return null instead of some default value?

cheenamalhotra: We did match defaults with MSSQL extension's connection options defaults (as in package.json) in PR https://github.com/microsoft/sqltoolsservice/pull/2132 as otherwise defaults don't match and we end up populating a lot of connection options in profile, even though user did not provide them as advanced options.

But other properties are not of concern, as 'user' was used by SSMS to generate connection name where it crashed.

Charles-Gagnon (approved):

PR #2155, opened 2023-07-25
Added retry policy for sleeping serverless error for SqlConnections

smartguest:
Tested to work with two different serverless databases. The time intervals and amount of retries chosen were selected based on what worked consistently. Justification for adding an extended timeout for this error specifically is found here: 

https://github.com/dotnet/SqlClient/issues/1957

Fixes #https://github.com/microsoft/azuredatastudio/issues/22311

And #https://github.com/microsoft/azuredatastudio/issues/22445


src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableSqlConnection.cs
```diff
@@ -63,6 +63,9 @@ public sealed partial class ReliableSqlConnection : DbConnection, IDisposable
         public ReliableSqlConnection(string connectionString, RetryPolicy connectionRetryPolicy, RetryPolicy commandRetryPolicy, string azureAccountToken)
         {
             _underlyingConnection = new SqlConnection(connectionString);
+
+            _underlyingConnection.RetryLogicProvider = RetryPolicyUtils.ServerlessWaitRetryLogicProvider();
```
cheenamalhotra, line 67: I wonder if someone would want to customize this retry logic provider applying on SqlConnection.

@Matteo-T any thoughts? Since it affects Managed Batch parser, what is your thought in specifying this provider by default. It seems to me it's a necessity to have a retry logic provider for serverless connections, but would you need to customize the retry attempt and intervals?

cheenamalhotra:
FYI @smartguest I found another use-case not yet covered by this PR, https://github.com/microsoft/azuredatastudio/issues/23999 
Can you take a look into that too?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -51,6 +51,10 @@ public class ConnectionService
         private const int DoesNotMeetPWReqs = 18466; // Password does not meet complexity requirements.
         private const int PWCannotBeUsed = 18463; // Password cannot be used at this time.
 
+        // Default SQL constants (required to ensure connections such as serverless are able to connect and retry properly).
+        private const string DefaultApplicationName = "azdata";
```
cheenamalhotra, line 55: We already have ApplicationName on line 177 that stores the default application name so this is not necessary.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1895,9 +1887,8 @@ public static SqlConnection OpenSqlConnection(ConnectionInfo connInfo, string fe
                 // allow pooling connections for language service feature to improve intellisense connection retention and performance.
                 bool shouldForceDisablePooling = !EnableConnectionPooling && featureName != Constants.LanguageServiceFeature;
 
-                // increase the connection and command timeout to at least 30 seconds and and build connection string
-                connInfo.ConnectionDetails.ConnectTimeout = Math.Max(30, connInfo.ConnectionDetails.ConnectTimeout ?? 0);
-                connInfo.ConnectionDetails.CommandTimeout = Math.Max(30, connInfo.ConnectionDetails.CommandTimeout ?? 0);
+                // increase the connection and command timeout to at least 30 seconds and set application name.
+                FillInDefaultDetailsForConnections(connInfo.ConnectionDetails, featureName);
```
cheenamalhotra, line 1891: Reassign connInfo.ConnectionDetails here - by returning updated connection details from the method.

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ConnectSqlCmdCommand.cs
```diff
@@ -71,6 +71,7 @@ private DbConnection AttemptToEstablishCurConnection(SqlConnectionInfo ci)
                 connString += ";Pooling=false"; //turn off connection pooling (this is done in other tools so following the same pattern)
 
                 conn = new SqlConnection(connString);
+                (conn as SqlConnection).RetryLogicProvider = Connection.ReliableConnection.RetryPolicyUtils.ServerlessWaitRetryLogicProvider();
```
Matteo-T, line 74: An `as` without a check for null is almost always a mistake: either cast (if you know it's a SqlConnection) or check for null.

Matteo-T: Considering that at line 73 you new'ed the SqlConnection, I fail to see why you'd want the `as`. Could you explain?

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
```diff
@@ -26,9 +30,25 @@ public static class RetryPolicyUtils
         /// Blocklist of non-transient errors that should stop retry during data transfer operations
         /// </summary>
         private static readonly HashSet<int> _nonRetryableDataTransferErrors;
+        /// <summary>
+        /// Max intervals between retries to wake up serverless instances.
+        /// </summary>
+        private const int _serverlessMaxIntervalTime = 30;
```
Matteo-T, line 36: Add the unit of measure (in the name and/or the XML comment), e.g. `_serverlessMaxIntervalTimeInSeconds`.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -668,6 +646,20 @@ private string MapServerEdition(ReliableConnectionHelper.ServerInfo serverInfo)
             return serverEdition;
         }
 
+        private static void FillInDefaultDetailsForConnections(ConnectionDetails ConnectionDetails, string featureName) { 
+
+            if(string.IsNullOrEmpty(ConnectionDetails.ApplicationName)) {
+                ConnectionDetails.ApplicationName = DefaultApplicationName;
+            }
+            else {
```
Matteo-T, line 654: General comment to be applied everywhere: try to follow the conventions adopted (at least) in this files, which is to align matching curly braces under the same column.

src/Microsoft.SqlTools.ServiceLayer/Scripting/SmoScriptingOperation.cs
```diff
@@ -79,6 +79,7 @@ protected string GetServerNameFromLiveInstance(string connectionString, string a
             string serverName = null;
             using (SqlConnection connection = new SqlConnection(connectionString))
             {
+                connection.RetryLogicProvider = Connection.ReliableConnection.RetryPolicyUtils.ServerlessWaitRetryLogicProvider();
```
Matteo-T, line 82: If you ask me, the choice of name for this this class is weird. I did not see anything in there that suggested "Serverless" stuff. But what do I know? :)

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/ReliableSqlConnection.cs
Matteo-T: @cheenamalhotra - I cannot really tell the impact of this Provider overall (it's probably negligible for the way I use it), but... the idea of being able to either customize this "retry provider" (as I wrote elsewhere, seeing a "Serverless" in the name confuses me), reset to "null" (=no provider, which seems to be the default in MDS), or even pass my own would be kind of cool. Maybe it's not a lot of work to make it more general-purpose without over-engineering it?

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
```diff
@@ -357,6 +377,38 @@ public static void RaiseSchemaAmbientRetryMessage(RetryState retryState, int err
...
+                DeltaTime = TimeSpan.FromSeconds(1),
+                TransientErrors = _retryableExtendedNetworkConnectivityError
+            };
+
+            var provider = SqlConfigurableRetryFactory.CreateFixedRetryProvider(serverlessRetryLogic);
+
+            provider.Retrying += (object s, SqlRetryingEventArgs e) =>
+            {
+                int attempts = e.RetryCount + 1;
+                Logger.Write(TraceEventType.Information, $"attempt {attempts} - current delay time:{e.Delay} \n");
```
Matteo-T, line 398: Perhaps use System.Environment.NewLine which does "the right thing" in a more portable way?

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
```diff
@@ -357,6 +377,38 @@ public static void RaiseSchemaAmbientRetryMessage(RetryState retryState, int err
...
+                MaxTimeInterval = TimeSpan.FromSeconds(_serverlessMaxIntervalTime),
+                DeltaTime = TimeSpan.FromSeconds(1),
+                TransientErrors = _retryableExtendedNetworkConnectivityError
+            };
+
+            var provider = SqlConfigurableRetryFactory.CreateFixedRetryProvider(serverlessRetryLogic);
+
+            provider.Retrying += (object s, SqlRetryingEventArgs e) =>
+            {
+                int attempts = e.RetryCount + 1;
```
Matteo-T, line 397: I do not know what the convention for this repo (or file) is, but... since you used `var` above (which I like), keep using it :-)

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
```diff
@@ -357,6 +377,38 @@ public static void RaiseSchemaAmbientRetryMessage(RetryState retryState, int err
...
+                Logger.Write(TraceEventType.Information, $"attempt {attempts} - current delay time:{e.Delay} \n");
+                if (e.Exceptions[e.Exceptions.Count - 1] is SqlException ex)
```
Matteo-T, line 399: What do you think of 

```csharp
Logger.Write(
    TraceEventType.Information,
    (e.Exceptions[e.Exceptions.Count - 1] is SqlException ex)
    : $"{ex.Number}-{ex.Message}{Environment.NewLine}"
    ? $"{e.Exceptions[e.Exceptions.Count - 1].Message}{Environment.NewLine}");
```
which avoids duplicating some code?

Matteo-T (changes requested):

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
Charles-Gagnon: Why do we even need a newline? Generally log messages would each be on their own new line anyways (or if not, a version that adds it such as Console.Write vs Console.WriteLine would be available)

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
Charles-Gagnon: We also have overloads for each of the logger message types, so you can just do `Logger.Information(...)` instead of Logger.Write

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
```diff
@@ -26,9 +30,25 @@ public static class RetryPolicyUtils
...
+        /// </summary>
+        private const int _serverlessMaxIntervalTime = 30;
+        /// <summary>
+        /// Maximum number of retries to wake up serverless instances.
+        /// </summary>
+        private const int _serverlessMaxRetries = 4;
 
         static RetryPolicyUtils()
         {
+            _retryableExtendedNetworkConnectivityError = new HashSet<int>
```
Charles-Gagnon, line 44: How did you come up with this name? I'm not really sure how it relates to the error inside. And given that it is only used for the "serverless" retry shouldn't it say something about that?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ConnectSqlCmdCommand.cs
smartguest: @Charles-Gagnon Yeah that's right, thought it was needed there for a reason, but if that's safe to do so, I'm going with the change.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicyUtils.cs
smartguest: This was modified from some code (Example for retry provider), so I'm going to fix this based on feedback

smartguest:
Added boolean enable disable for the new retrylogicprovider, if it is preferred to pass in the provider instead and to move the retrylogicprovider outside of ManagedBatchParser, it can be done. although I think the retrylogicprovider belongs in retrypolicyutils personally and it's easier to pass in a boolean value.

cheenamalhotra:
> Added boolean enable disable for the new retrylogicprovider for ReliableSqlConnection, if it is preferred to pass in the provider instead and to move the retrylogicprovider outside of ManagedBatchParser, it can be done. although I think the retrylogicprovider belongs in retrypolicyutils personally and it's easier to pass in a boolean value. Change was tested to still work.

HI Alex, I think as Matteo mentioned above, being able to pass a provider implementation should be a better design. Please update the method to accept retry logic provider for SqlConnection.

smartguest:
@Charles-Gagnon I'm not sure if the current RetryLogicProvider class supports having multiple providers and is able to run through them sequentially as the retry option class (used for determining the retry time and intervals) and the provider class that takes in the option are defined outside of STS but I'll have to ask if this is possible without having to alter the RetryLogicProvider class itself.

Charles-Gagnon:
Looking more at the retry provider logic, it seems like even just exposing the error codes for "serverless" errors should be fine. Since we don't really implement any of the actual retry logic ourself (aside from a bit of logging).

But then yeah, allowing users to pass in their own providers seems the best to give full control for specialized circumstances.

smartguest:
Added RetryLogicProvider as parameter to function, the implementation of the provider remains in RetryPolicyUtils as its accessible in all the areas that it is needed in, plus it makes sense to keep retry code in the same location.

cheenamalhotra (approved):

PR #2158, opened 2023-07-31
Close connection used for detaching a database after operation completes

corivera:
We weren't cleaning up the connection used for detach operations, so trying to delete the database afterwards would hang. This connection logic is the same as that used in the create database code.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -287,48 +287,58 @@ public override Task<string> Script(DatabaseViewContext context, DatabaseInfo ob
         public string Detach(DetachDatabaseRequestParams detachParams)
         {
             var sqlScript = string.Empty;
-            ConnectionInfo connectionInfo = this.GetConnectionInfo(detachParams.ConnectionUri);
             using (var dataContainer = CreateDatabaseDataContainer(detachParams.ConnectionUri, detachParams.ObjectUrn, false, null))
```
Charles-Gagnon, line 290: What about having the data container close the connection when it's disposed? That seems like a better pattern to avoid these kinds of mistakes from happening in the future.

corivera: CDataContainer.Dispose already has a bunch of connection closing logic, but still seems to miss this one. I'll look into it.

src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
```diff
@@ -1081,6 +1081,12 @@ private void Dispose(bool disposing)
 
                     if (disposing)
                     {
+                        // Disconnect our own connection, since disposing sqlCiWithConnection may not catch it
+                        if (ownConnection && this.serverConnection != null)
+                        {
+                            this.serverConnection.Disconnect();
```
Charles-Gagnon, line 1087: Any reason not to use dispose instead? And should probably null this out afterwards like the others do.

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1933,9 +1933,8 @@ public static SqlConnection OpenSqlConnection(ConnectionInfo connInfo, string fe
                     "Failed opening a SqlConnection: error:{0} inner:{1} stacktrace:{2}",
                     ex.Message, ex.InnerException != null ? ex.InnerException.Message : string.Empty, ex.StackTrace);
                 Logger.Write(TraceEventType.Error, error);
+                throw;
             }
-
-            return null;
```
corivera, line 1938: There wasn't any null handling after calling this method, so it made more sense to just throw the exception above.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -358,10 +346,11 @@ private string CreateDetachScript(DetachDatabaseRequestParams detachParams, stri
...
         {
             ConnectionInfo connectionInfo = this.GetConnectionInfo(connectionUri);
-            if (!isNewDatabase && !string.IsNullOrEmpty(databaseName))
-            {
-                connectionInfo.ConnectionDetails.DatabaseName = databaseName;
-            }
+            // Update the database name in case the associated connection with the URI is for a different database like master.
+            // This also ensures we'll switch back to the original database name for this connection in subsequent calls so that it
+            // doesn't cause problems if we delete a database and this connection is still associated with that dropped database.
+            connectionInfo.ConnectionDetails.DatabaseName = databaseName ?? "master";
```
corivera, line 352: I was running into issues where trying to open the New Database dialog would fail after Scripting a Detach Database operation and then dropping that same database. This is because we pass the ConnectionInfo by reference here, so that Detach operation would update the database name here. After dropping the database, trying to re-use the same connection would fail due to the associated database being dropped and the actual backing connection now getting cleaned up after each operation.

corivera: @ssreerama any concerns about this change? I tested updates with the database properties dialog and didn't see any issues.

corivera:
Fixed the connection cleanup more cleanly in a separate PR

PR #2159, opened 2023-08-01
Implements Contextualization API into Azure Data Studio to get better query recommendations from extensions like Copilot

lewis-sanchez:
This PR implements the endpoints to obtain metadata, in the form of create table statements, for a server connection. The statements obtained from a server will be used by the Copilot extension, so it can provide accurate query suggestions back to a user.


src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterFactory.cs
```diff
@@ -0,0 +1,191 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
```
Charles-Gagnon, line 6: It would be great to not disable nullable warnings in new files like this

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptCacher.cs
```diff
@@ -0,0 +1,108 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using Logger = Microsoft.SqlTools.Utility.Logger;
+
+namespace Microsoft.SqlTools.ServiceLayer.Metadata
+{
+    public static class MetadataScriptCacher
```
Charles-Gagnon, line 15: It's not clear to me how this is supposed to be used - there doesn't seem to be any sort of cache expiration or validation that the cache isn't valid so right now it seems completely up to the client side to determine when to refresh the cache which doesn't seem like an easy thing to do

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterFactory.cs
```diff
@@ -0,0 +1,191 @@
...
+            };
+
+            var scripter = new Scripter(server);
+            scripter.Options = scriptingOptions;
+            var generatedScripts = scripter.Script(urns);
+
+            var scripts = new List<string>();
+            foreach (var s in generatedScripts)
+            {
+                if (s.StartsWith("CREATE TABLE"))
```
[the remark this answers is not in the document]
lewis-sanchez: It's including things like `SET ANSI_NULLS ON` and `SET QUOTED_IDENTIFIER ON` for every create table script. We don't need any of that since the only thing needed is the table name and column names and types.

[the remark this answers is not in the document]
Charles-Gagnon: Is there any harm in having it?

If you're going to make this into a very generic sounding service then I'd say let's try to avoid doing extra unexpected things like removing parts of the generated data. Otherwise if someone else tries to use this and doesn't realize it's been made specifically for copilot use that could be problematic.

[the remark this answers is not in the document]
Charles-Gagnon: Also keep in mind there's other statements that can be applied, such as `ALTER TABLE...ADD PRIMARY KEY` which would probably be useful for copilot to know about.

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Linq;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Metadata
+{
+    public static class MetadataScriptTempFileStream
+    {
+        private const string DirectoryName = "TableAndViewScripts";
+
+        public static void Write(string serverName, IEnumerable<string> scripts)
```
kisantia, line 18: please add doc comments

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            var tempFileName = $"{DirectoryName}/{serverName}.tmp";
```
kisantia, line 20: use path.combine here to avoid any slash direction problems on different platforms

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,99 @@
...
+                {
+                    return scripts;
+                }
+
+                using (StreamReader sr = new StreamReader(tempFilePath))
+                {
+                    while (!sr.EndOfStream)
+                    {
+                        var line = sr.ReadLine();
+                        if (line != null)
```
kisantia, line 59: should whitespace be added or can the check be String.IsNullOrWhitespace()?

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                var tempFilePath = Path.Combine(Path.GetTempPath(), tempFileName);
+                if (!File.Exists(tempFilePath))
+                {
+                    return false;
+                }
+                else
+                {
+                    var lastWriteTime = File.GetLastWriteTime(tempFilePath);
+                    return (DateTime.Now - lastWriteTime).TotalDays < 30;
```
kisantia, line 89: use a const instead of magic number so it's more clear what 30 means

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -115,6 +119,90 @@ public void InitializeService(ServiceHost serviceHost)
...
+
+            if (connectionInfo != null)
+            {
+                using (SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connectionInfo, "metadata"))
+                {
+                    try
+                    {
+                        if (MetadataScriptTempFileStream.IsScriptTempFileValid(connectionInfo.ConnectionDetails.ServerName))
+                        {
+                            await requestContext.SendResult(true);
```
kisantia, line 138: is this true result sent here used by ADS? How does ADS differentiate it from the result sent on line 147?

lewis-sanchez: ADS doesn't use the returned bool value. I couldn't find a way to specify `RequestContext<void>` or just `RequestContext`  since what's being returned isn't overly useful, so ended up going with `RequestContext<bool>`.

Charles-Gagnon: If you're not waiting on a result then this should probably be a notification, not a request (which doesn't have a response)

kisantia: or can you just return the file path with the second result? And make a new result type that has the string path and success

lewis-sanchez: @Charles-Gagnon, to return a notification I would need to swap out `SetRequestHandler` for `SetEventHandler`, and then I wouldn't need to return anything or would I still need to invoke `sendEvent` on the `serviceHost`?

@kisantia, would the success flag really be needed since sendError is being used in the handle request endpoint when there is an error? That part seems kind of redundant, but the file path string makes sense to me.

Charles-Gagnon: Correct and correct

> @Charles-Gagnon, to return a notification I would need to swap out `SetRequestHandler` for `SetEventHandler`, and then I wouldn't need to return anything or would I still need to invoke `sendEvent` on the `serviceHost`?

src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/GetServerTableMetadataRequest.cs
```diff
@@ -9,11 +9,17 @@ namespace Microsoft.SqlTools.ServiceLayer.Metadata.Contracts
...
+        /// <summary>
+        /// The URI of the connection to generate scripts for.
+        /// </summary>
         public string OwnerUri { get; set; }
     }
 
     public class GetServerTableMetadataResult
     {
+        /// <summary>
+        /// An array containing generated table and view scripts,
```
kisantia, line 21: nit: remove trailing comma

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -135,45 +136,40 @@ public void InitializeService(ServiceHost serviceHost)
...
                         {
                             try
                             {
                                 MetadataScriptTempFileStream.Write(connectionInfo.ConnectionDetails.ServerName, scripts);
-                                await requestContext.SendResult(true);
                             }
                             catch (Exception ex)
                             {
                                 Logger.Error($"An error was encountered while writing to the cache. Error: {ex.Message}");
-                                await requestContext.SendError(ex);
```
kisantia: sending these errors back isn't needed anymore?

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -135,45 +136,40 @@ public void InitializeService(ServiceHost serviceHost)
                     {
                         if (MetadataScriptTempFileStream.IsScriptTempFileValid(connectionInfo.ConnectionDetails.ServerName))
```
kisantia, line 137: add comment explaining that if the script temp file is valid, then the scripts don't need to be regenerated (at least that's how I was understanding it)

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
lewis-sanchez: There currently is no mechanism to return errors from event handlers.

Charles-Gagnon: So if you're expecting to have errors and handle them in some way on the client side you should be using a request. Notifications are for things that don't expect a response of any sort back. Does your design have any plans for how to handle issues happening when initializing the metadata?

lewis-sanchez: I'm not expecting any major errors. In the worst case, if STS is not able to write scripts to the temp directory, then we wouldn't be able to supply Copilot with any context and Copilot would work as it currently does. To mitigate that, the generated scripts could then be held in memory, but then we would probably run into high memory consumption issues depending on how many connections a user has and the size of those connections.

Charles-Gagnon: Sounds reasonable to me, you can always change it later if needed. Keeping it simpler for now seems like a good idea.

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterFactory.cs
```diff
@@ -0,0 +1,191 @@
...
+            }
+
+            return scripts;
+        }
+
+        private static UrnCollection GetAllServerTableAndViewUrns(Server server)
+        {
+            UrnCollection urnCollection = new UrnCollection();
+
+            foreach (Database db in server.Databases)
```
Charles-Gagnon, line 175: If anything goes wrong with querying the info for a database (such as not having access) then this whole thing will fail. Would suggest either scoping this to just the database the connection is for or at least putting all operations into try/catches so that it will at least get some of the server info back even if it can't get it all.

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterFactory.cs
```diff
@@ -0,0 +1,191 @@
...
+using Microsoft.Data.SqlClient;
+using Microsoft.SqlServer.Management.Common;
+using Microsoft.SqlServer.Management.Smo;
+using Microsoft.SqlTools.SqlCore.Connection;
+using Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection;
+
+
+namespace Microsoft.SqlTools.ServiceLayer.Metadata
+{
+    internal static class SmoScripterFactory
```
Charles-Gagnon, line 18: nit: But I don't really see how this is a "Scripter Factory" at all (it's not making a script). Something like `SmoScripterHelpers` would be better IMO

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterFactory.cs
```diff
@@ -0,0 +1,191 @@
...
+    {
+        public static async Task<IEnumerable<string>> GenerateAllServerTableScripts(DbConnection connection)
+        {
+            var serverConnection = SmoScripterFactory.GetServerConnection(connection);
+            if (serverConnection == null)
+            {
+                return null;
```
Charles-Gagnon, line 25: You don't want to return an empty enumerable instead? If you do return this it'd be nice to type this correctly (`Task<IEnumerable<string>?>`

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -116,6 +120,89 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    try
+                    {
+                        // If scripts have been generated within the last 30 days then there isn't a need to go through the process
+                        // of generating scripts again.
+                        if (MetadataScriptTempFileStream.IsScriptTempFileUpdateNeeded(connectionInfo.ConnectionDetails.ServerName))
+                        {
+                            return;
+                        }
+
+                        var scripts = await SmoScripterFactory.GenerateAllServerTableScripts(sqlConn);
```
Charles-Gagnon, line 145: We shouldn't block the message thread while generating scripts since that could take a while. Please sync up with @cheenamalhotra on how to handle this better

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Metadata/MetadataServiceTests.cs
```diff
@@ -125,11 +127,106 @@ public async Task GetTableInfoReturnsValidResults()
...
+            {
+                var tempFilePath = Path.Combine(Path.GetTempPath(), tempFileName);
+                if (File.Exists(tempFilePath))
+                {
+                    File.Delete(tempFilePath);
+                }
+            }
+            catch
+            {
+                throw;
```
Charles-Gagnon, line 226: If you're just rethrowing you can just remove the try/catch completely

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            return scripts;
+        }
+
+        /// <summary>
+        /// Checks to see if the temporary file containing the scripts for a server needs to be updated.
+        /// </summary>
+        /// <param name="serverName">The name of the file associated with the given server name.</param>
+        /// <returns>Flag indicating that the script file is valid.</returns>
```
Charles-Gagnon, line 92: This doesn't seem accurate - there isn't any "validity" being checked that I can see

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,117 @@
...
+            try
+            {
+                var tempFilePath = Path.Combine(Path.GetTempPath(), tempFileName);
+                if (!File.Exists(tempFilePath))
+                {
+                    return false;
+                }
+                else
+                {
+                    var lastWriteTime = File.GetLastWriteTime(tempFilePath);
```
Charles-Gagnon, line 106: Would be good to put a comment here on why you chose 30 days and any other info that would be useful for anyone trying to understand how this whole thing works.

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,117 @@
...
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.Metadata
+{
+    /// <summary>
+    /// This class is responsible for reading, writing, and checking the validity of script files.
+    /// </summary>
+    public static class MetadataScriptTempFileStream
+    {
+        private const short NumOfDaysSinceLastWrite = 30;
```
Charles-Gagnon, line 19: Something like `ScriptFileExpirationInDays` would be clearer IMO

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,117 @@
...
+                }
+                else
+                {
+                    var lastWriteTime = File.GetLastWriteTime(tempFilePath);
+                    return (DateTime.Now - lastWriteTime).TotalDays < NumOfDaysSinceLastWrite;
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.Warning($"Unable to determine if the script file is older than 30 days. Error: {ex.Message}");
```
Charles-Gagnon, line 112: This should use the const you defined above

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,123 @@
...
+        private const short ScriptFileExpirationInDays = 30;
+
+        /// <summary>
+        /// This method writes the passed in scripts to a temporary file.
+        /// </summary>
+        /// <param name="serverName">The name of the server which will go on to become the name of the file.</param>
+        /// <param name="scripts">The generated scripts that will be written to the temporary file.</param>
+        public static void Write(string serverName, IEnumerable<string> scripts)
+        {
+            var tempFileName = $"{serverName}.tmp";
```
Charles-Gagnon, line 28: Server names can contain invalid filename characters (like `/`), so you should probably be sanitizing this before reading/writing. Probably easiest to just convert to base64 or something.

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataScriptTempFileStream.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            return scripts;
+        }
+
+        /// <summary>
+        /// Determines if the script file for a server is too old.
+        /// </summary>
+        /// <param name="serverName">The name of the file associated with the given server name.</param>
+        /// <returns>True: The file was created within the expiration period; False: The script file needs to be created
```
Charles-Gagnon, line 92: Isn't this reversed? True would imply that it needs to be updated according to the method name, but you're returning false for when that should occur.

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -116,6 +120,89 @@ public void InitializeService(ServiceHost serviceHost)
...
+                    try
+                    {
+                        // If scripts have been generated within the last 30 days then there isn't a need to go through the process
+                        // of generating scripts again.
+                        if (MetadataScriptTempFileStream.IsScriptTempFileUpdateNeeded(connectionInfo.ConnectionDetails.ServerName))
+                        {
+                            return;
+                        }
+
+                        var scripts = await SmoScripterHelpers.GenerateAllServerTableScripts(sqlConn);
```
Charles-Gagnon, line 145: We shouldn't block the message thread while generating scripts since that could take a while. Please sync up with @cheenamalhotra on how to handle this better

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
```diff
@@ -0,0 +1,206 @@
...
+            if (sqlConnection == null)
+            {
+                return null;
+            }
+
+            var serverConnection = SmoScripterHelpers.ConnectToServerWithSmo(sqlConnection);
+            return serverConnection;
+        }
+
+        private static SqlConnection? TryFindingReliableSqlConnection(ReliableSqlConnection reliableSqlConnection)
```
Charles-Gagnon, line 48: Why do we need all this extra logic here anyways to pull out the actual connection? Shouldn't the connection we receive always be the same type? And if not - it seems like it'd be better to have the place that gets the connection pull out the sqlconnection to use instead (since it doesn't appear like we actually really care about whether it's a "ReliableSqlConnection" or not...)

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
```diff
@@ -0,0 +1,206 @@
...
+            {
+                // If we don't have connection we can use with SMO, just give up on using SMO
+                return null;
+            }
+
+            // We have a reliable connection, use the underlying connection
+            return reliableSqlConnection.GetUnderlyingConnection();
+        }
+
+        private static ServerConnection ConnectToServerWithSmo(SqlConnection connection)
```
Charles-Gagnon, line 61: I've seen stuff like this around before, if there isn't a common helper method already to do this then it seems useful to make one.

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
```diff
@@ -0,0 +1,206 @@
...
+            };
+
+            var scripter = new Scripter(server);
+            scripter.Options = scriptingOptions;
+            var generatedScripts = await Task.Run(() => scripter.Script(urns));
+
+            var scripts = new List<string>();
+            foreach (var s in generatedScripts)
+            {
+                var script = s.Replace("\r", string.Empty).Replace("\n", string.Empty);
```
Charles-Gagnon, line 165: Explain why you're removing these

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
lewis-sanchez: I broke this down for my own understanding, which is why I moved it into separate methods. Others seem to just have a block of code to get a connection. Examples:

https://github.com/microsoft/sqltoolsservice/blob/16c7e323567023b08200ab35b2bfb2a5042385e0/src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoMetadataFactory.cs#L49C1-L63C14

https://github.com/microsoft/sqltoolsservice/blob/16c7e323567023b08200ab35b2bfb2a5042385e0/src/Microsoft.SqlTools.ServiceLayer/EditData/SmoEditMetadataFactory.cs#L45C13-L59C14

Charles-Gagnon: I would love to know why this is necessary...definitely seems weird (and worrisome if this is just a general thing that can happen).

At the very least it would be great to just make a single helper method in the connection area of the code that all these places call so we're at least not duplicating the same logic over and over again.

lewis-sanchez: I agree, but I also want to unblock the PostgreSQL engineer, which is why I'm trying to get this merged in and addressing feedback quickly.

Could moving this be a separate PR? I'm going to need to alter this implementation in the near future too, but I've been told this implementation is fine for now as a proof of concept that generating and sending context will make suggestions better for Copilot users.

Charles-Gagnon: Sure, we can focus on the blocking issues (like the thread blocking one I brought up)

But curious how this affects PG SQL? They're going to need to write their own implementation here, so the work in STS isn't going to affect that, right?

lewis-sanchez: No, Postgres is not going to use any of the code here, but your comment in the ADS repo mentioned that the ADS PR would only be merged once this is merged in. Postgres is waiting on core ADS changes, and Jordan is waiting on this PR and the ADS PR, so that he can test his fork of Copilot.

Charles-Gagnon: Gotcha. So yeah, that side can go in earlier as long as you make sure that you disable the functionality on the MSSQL side (likely by disabling the provider registration in that PR) until the STS changes are ready so it doesn't throw an error when it tries to send a message the STS there isn't hooked up to understand yet.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Metadata/MetadataServiceTests.cs
```diff
@@ -125,15 +129,108 @@ public async Task GetTableInfoReturnsValidResults()
...
+            CreateTestTable(sqlConn, this.testTableSchema, this.testTableName);
+            CreateTestTable(sqlConn, this.testTableSchema, this.testTableName2);
+
+            var generateServerContextualizationParams = new GenerateServerContextualizationParams
+            {
+                OwnerUri = connectionResult.ConnectionInfo.OwnerUri
+            };
+
+            var service = new MetadataService();
+            var methodInfo = typeof(MetadataService).GetMethod("GenerateServerContextualization", BindingFlags.NonPublic | BindingFlags.Static);
```
Charles-Gagnon, line 155: You shouldn't need to do this - the test assembly is already a friend to the servicelayer assembly so you should be able to call it directly. (we should avoid using reflection wherever possible)

https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj#L78

lewis-sanchez: It wasn't letting me access the method in the integration test, so I resorted to reflection. I tried to add the `InternalsVisisbleTo` attribute to the `GenerateServerContextualization` method too, but it's not recognizing `Microsoft.SqlTools.ServiceLayer.IntegrationTests` as an assembly name.

lewis-sanchez: Oh... I needed to switch the access specifier from `private` to `internal`

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
lewis-sanchez: This is the only other spot I could find that does something similar, but it's getting the server name and not the server itself.

https://github.com/microsoft/sqltoolsservice/blob/16c7e323567023b08200ab35b2bfb2a5042385e0/src/Microsoft.SqlTools.ServiceLayer/Scripting/SmoScriptingOperation.cs#L90C16-L98C21

lewis-sanchez:
@Charles-Gagnon, am I good to merge this in after addressing the thread blocking issues?

lewis-sanchez:
@cheenamalhotra, could you review the new endpoints, specifically how I avoid blocking the message thread by starting a separate background thread in the MetadataService.cs file?

cheenamalhotra (approved):

PR #2161, opened 2023-08-02
Improve Query editor Read performance + cancel timely for large data

cheenamalhotra:
Reading results with `ReadAsync()` is terribly slow with SqlClient (known issues) and also doesn't respect cancellation token when reading large data - major contributor to hung queries/unresponsive cancel. Switching to `Read()` brings massive performance improvement in large data scenarios (also used by SSMS for same reason).

Perf difference is quite significant when reading 4 rows of large strings (10 million characters each):

- **Insiders**: _9:40 mins_ (still only manages to read 1 row)
- **Dev**: _1 sec_ ❤️ (reads all 4 rows)

image

Addresses https://github.com/microsoft/azuredatastudio/issues/3231 and many other complaints about speed issues with query editor when it comes to large data for enterprise users.

_Note: This is not a complete solution to Query Editor performance issues, as we need to resolve many more scenarios._

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -429,7 +429,6 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                         {
                             do
                             {
-                                // Verify that the cancellation token hasn't been canceled
                                 cancellationToken.ThrowIfCancellationRequested();
                                 columnSchemas.Add(reader.GetColumnSchema().ToArray());
                             } while (reader.NextResult());
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
@@ -429,6 +429,8 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
                         {
                             do
                             {
+                                // Verify that the cancellation token hasn't been canceled
+                                cancellationToken.ThrowIfCancellationRequested();
                                 columnSchemas.Add(reader.GetColumnSchema().ToArray());
                             } while (reader.NextResult());
                         }
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/StorageDataReader.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/StorageDataReader.cs
@@ -94,6 +94,15 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.DataStorage
         }
 
         /// <summary>
+        /// Pass-through to DbDataReader.Read()
+        /// </summary>
+        /// <returns></returns>
+        public bool Read()
+        {
+            return DbDataReader.Read();
+        }
+
+        /// <summary>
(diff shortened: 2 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Batch.cs
```diff
@@ -429,6 +429,8 @@ private async Task ExecuteOnce(DbConnection conn, CancellationToken cancellation
                         {
                             do
                             {
+                                // Verify that the cancellation token hasn't been canceled
```
Charles-Gagnon, line 432: nit: I think the code is self-documenting in this case, but that may just be me 😄

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -396,8 +396,10 @@ public async Task ReadResultToEnd(DbDataReader dbDataReader, CancellationToken c
                     //
                     availableTask = SendCurrentResults();
 
-                    while (await dataReader.ReadAsync(cancellationToken))
+                    while (dataReader.Read())
```
Charles-Gagnon, line 399: Please add a comment explaining why you're using read vs readasync. is there an issue/doc you can link to about the perf issue? That is not intuitive to me at all (generally using async calls is preferred for async code like STS) so it's important to make sure future people know why a particular choice was made.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/StorageDataReader.cs
```diff
@@ -93,6 +93,15 @@ public Task<bool> ReadAsync(CancellationToken cancellationToken)
             return DbDataReader.ReadAsync(cancellationToken);
         }
 
+        /// <summary>
+        /// Pass-through to DbDataReader.Read()
+        /// </summary>
+        /// <returns></returns>
+        public bool Read()
```
Charles-Gagnon, line 100: Would suggest adding a big warning to ReadAsync with information about the performance issues (and same as below, links to an issue/docs would be preferred)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -396,8 +396,10 @@ public async Task ReadResultToEnd(DbDataReader dbDataReader, CancellationToken c
...
                     {
+                        // check for cancellation token before actually making connection
```
Charles-Gagnon, line 401: This doesn't seem accurate as the connection has already been made...and also same comment about this being self-documenting.

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/StorageDataReader.cs
cheenamalhotra: I'm gonna delete it eventually - need to move it away from EditData area as well - planned for another PR.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
cheenamalhotra: I probably just copied over comments in excitement 😄

PR #2168, opened 2023-08-09
Add extra delete database functionality to DatabaseHandler

corivera:
We already have a drop() method in the base object handler class, but delete has two extra options for closing active connections and deleting backup/restore history.


corivera:
Still need to add integration tests

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -164,14 +163,23 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
                                     EncryptionEnabled = smoDatabase.EncryptionEnabled
                                 };
 
+                                if (!isAzureDB) {
```
Charles-Gagnon, line 166: This doesn't seem related to the description, is this just an additional fix or something?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Oh, forgot to note that. Yes, it's an additional fix. We don't have the properties dialog enabled for Azure yet, so some of these properties haven't been excluded yet. An error gets thrown when trying to retrieve these properties against an Azure SQL DB.

corivera: Delete _is_ enabled for Azure though, which is why I was hitting it.

corivera: Note: We had a version of delete before for on-prem and azure DBs, but now that it uses a dialog it goes through this initialize view step.

corivera: Already checked with @ssreerama and he aggreed.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DeleteDatabase.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement.Contracts
+{
+    public class DeleteDatabaseRequestParams : GeneralRequestDetails
```
Charles-Gagnon, line 12: nit: but I still think drop makes sense to use here to keep the same term usage across STS. So it'd just be `DropDatabaseRequestParams`

But not a huge deal if you'd prefer to keep it like this

corivera: We already have a general Drop method for all object handlers, but I can add an overload for it for databases

Charles-Gagnon: That would ensure that the same logic is always used even if drop is called too which isn't a bad thing.

PR #2169, opened 2023-08-15
Enabling Files Tab to the database properties

ssreerama:
This PR adds the ability to get the database files from SMO and modify/remove/add new files to the collection. Also, returns the TSql script for the changes to the files.


src/Microsoft.SqlTools.ServiceLayer/Admin/Database/CreateDatabaseObjects.cs
```diff
@@ -1001,6 +1006,20 @@ public FileData Clone()
 
         #region properties
 
+        /// <summary>
+        /// The logical name of the file, without extension
```
Charles-Gagnon, line 1010: This doesn't seem to match? Or if it is accurate, seems more like an implementation detail rather than a description of what "ID" is in this context.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseInfo.cs
```diff
@@ -49,4 +51,20 @@ public class DatabaseScopedConfigurationsInfo
...
+        public string Name { get; set; }
+        public string Type { get; set; }
+        public string Path { get; set; }
+        public string FileGroup { get; set; }
+        public string FileNameWithExtension { get; set; }
+        public double SizeInMb { get; set; }
+        public bool IsAutoGrowthEnabled { get; set; }
+        public double AutoFileGrowth { get; set; }
+        public string AutoFileGrowthType { get; set; }
+        public double MaxSizeLimit { get; set; }
```
Charles-Gagnon, line 67: units?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -626,6 +641,61 @@ private string ConfigureDatabase(InitializeViewRequestParams viewParams, Databas
...
+
+                                    // Add newFile to the prototype files
+                                    prototype.Files.Add(newFile);
+                                }
+                                // Edit file properties: updating the existed files with modified data
+                                else
+                                {
+                                    foreach(var existedFile in prototype.Files)
+                                    {
+                                        if(existedFile.ID == file.Id)
```
Charles-Gagnon, line 677: You can simplify this using LINQ `FirstOrDefault` to get the file with the matching ID.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -374,6 +378,90 @@ public async Task VerifyDatabaseScopedConfigurationsTest()
...
+        {
+            // setup, drop database if exists.
+            var connectionResult = await LiveConnectionHelper.InitLiveConnectionInfoAsync("master", serverType: TestServerType.OnPrem);
+            using (SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connectionResult.ConnectionInfo))
+            {
+                var server = new Server(new ServerConnection(sqlConn));
+
+                var testDatabase = ObjectManagementTestUtils.GetTestDatabaseInfo();
+                var objUrn = ObjectManagementTestUtils.GetDatabaseURN(testDatabase.Name);
+                await ObjectManagementTestUtils.DropObject(connectionResult.ConnectionInfo.OwnerUri, objUrn);
```
Charles-Gagnon, line 396: Why would the DB exist? We should be making a new DB for each test run, so that should never happen (and doing this could hide other problems if we are accidently sharing test objects)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
```diff
@@ -374,6 +378,90 @@ public async Task VerifyDatabaseScopedConfigurationsTest()
...
+                    Assert.That(((DatabaseInfo)databaseViewInfoAfterFileDelete.ObjectInfo).Files.Count, Is.EqualTo(3), $"Should have only 3 files as we removed one");
+
+                    // cleanup
+                    await ObjectManagementTestUtils.DropObject(connectionResult.ConnectionInfo.OwnerUri, objUrn, throwIfNotExist: true);
+                    Assert.That(DatabaseExists(testDatabase.Name!, server), Is.False, $"Database '{testDatabase.Name}' was not dropped succesfully");
+                }
+                finally
+                {
+                    // Cleanup using SMO if Drop didn't work
+                    DropDatabase(server, testDatabase.Name!);
```
Charles-Gagnon, line 460: It seems redundant to have two cleanups - we should just have one and then fail the test if it can't drop it.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseInfo.cs
ssreerama: Sorry! didn't get this comment, anything wrong here?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -38,10 +38,12 @@ public class DatabaseHandler : ObjectTypeHandler<DatabaseInfo, DatabaseViewConte
         private static readonly Dictionary<RecoveryModel, string> displayRecoveryModels = new Dictionary<RecoveryModel, string>();
         private static readonly Dictionary<PageVerify, string> displayPageVerifyOptions = new Dictionary<PageVerify, string>();
         private static readonly Dictionary<DatabaseUserAccess, string> displayRestrictAccessOptions = new Dictionary<DatabaseUserAccess, string>();
+        private static readonly Dictionary<FileType, string> displayFileTypes = new Dictionary<FileType, string>();
```
cheenamalhotra, line 41: Please use ConcurrentDictionary where possible.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseInfo.cs
Charles-Gagnon: This doesn't say what units this should be...is it KB? MB? Something else? Any vars that store numeric values should generically have their units appended to the name for easier understanding.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/ObjectManagement/DatabaseHandlerTests.cs
ssreerama: Is it ok to leave the database on server during the test failure? It is file till the STS tests pipelines are not scheduled daily.

Charles-Gagnon: No, which is why I would say this is the one that should stay. 

Although for tests like this where we aren't testing the actual creating of the database itself have you considered using C:\src\sqltoolsservice\test\Microsoft.SqlTools.ServiceLayer.Test.Common\SqlTestDb.cs ? That will auto-drop once it's disposed which can make things simpler.

ssreerama: Verified the SqlTestDb class and using it to create the database for files and dsc tests. Thanks

corivera (approved):

PR #2170, opened 2023-08-15
Update DacFx version to 162.1.124-preview

zijchen:


chlafreniere (approved):

Charles-Gagnon:
Looks like a DeployWithOptions test failure that seems related

zijchen:
> Looks like a DeployWithOptions test failure that seems related

Yeah thanks for the heads up, it uncovered a regression that only showed up in the release build. Fixed and updated to latest.

PR #2171, opened 2023-08-16
Query Store Service

Benjin:
Adds the Query Store service, which generates queries needed to fetch the data for all query store reports


Benjin:
Just a draft for now to take a look at how things are working as I finish adding the tests.

At @cheenamalhotra's recommendation, I may end up swapping to instead construct and execute a SqlCommand, and return an ID that ADS can then use to retrieve the result set, as this would remove the need to roll my own .NET -> TSQL datatype conversion code.

src/Microsoft.SqlTools.ServiceLayer/QueryStore/QueryStoreService.cs
```diff
@@ -0,0 +1,545 @@
...
+        /// <summary>
+        /// Converts an object (that would otherwise be set as a SqlParameter value) to an entirely TSQL representation.
+        /// Only handles the same subset of object types that Query Store query generators use:
+        /// int, long, string, and DateTimeOffset
+        /// </summary>
+        /// <param name="paramValue"></param>
+        /// <returns>data type and value portions of a parameter declaration, in the form "INT = 999"</returns>
+        internal static string GetTSqlRepresentation(object paramValue)
+        {
+            types.Add(paramValue.GetType());
```
kisantia, line 525: do the types added to this HashSet get used somewhere later? I couldn't find where when I did a quick search

src/Microsoft.SqlTools.ServiceLayer/QueryStore/Contracts/QueryStoreReportParams.cs
```diff
@@ -0,0 +1,55 @@
...
+    public abstract class QueryConfigurationParams<T> : TypedQueryStoreReportParams<T> where T : QueryConfigurationBase, new()
+    {
+        public Metric SelectedMetric;
+        public Statistic SelectedStatistic;
+        public int TopQueriesReturned;
+        public bool ReturnAllQueries;
+        public int MinNumberOfQueryPlans;
+
+        /// <summary>
+        /// Column name by which to order, if any.  Not all query generators involve ordering.
```
kisantia, line 32: is this comment supposed to go with something else?

Benjin: Yup, something I moved to implementing classes

src/Microsoft.SqlTools.ServiceLayer/QueryStore/QueryStoreService.cs
```diff
@@ -0,0 +1,545 @@
...
+        /// Instance of the connection service, used to get the connection info for a given owner URI
+        /// </summary>
+        private ConnectionService ConnectionService { get; }
+
+        private QueryStoreService()
+        {
+            ConnectionService = ConnectionService.Instance;
+        }
+
+        internal QueryStoreService(ConnectionService connService)
```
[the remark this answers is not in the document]
Benjin: It was originally for testing, but I changed how I was doing that (mocking out the call that needed to run a query instead).

src/Microsoft.SqlTools.ServiceLayer/QueryStore/QueryStoreService.cs
Benjin: Nope; leftover from making sure I'd collected all the different data types that needed to be handled with this code.

src/Microsoft.SqlTools.ServiceLayer/QueryStore/Contracts/GetForcedPlanQueriesReport.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using Microsoft.SqlServer.Management.QueryStoreModel.Common;
+using Microsoft.SqlServer.Management.QueryStoreModel.ForcedPlanQueries;
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+#nullable disable
+
+namespace Microsoft.SqlTools.ServiceLayer.QueryStore.Contracts
+{
+    public class GetForcedPlanQueriesReportParams : QueryConfigurationParams<ForcedPlanQueriesConfiguration>, IOrderableQueryParams
```
Benjin, line 14: TODO: docstring comments for all properties and requests, `get;set;`ers for properties

kisantia:
looks like the other tests also need the line ending change made for the ForcedPlanQueries and TrackedQueries tests, now that they're getting past the timezone failure

kisantia (approved):

PR #2173, opened 2023-08-17
Publishing SQL Core library in nuget

aasimkhan30:
This PR does following things.

- Sets the target framework of SQL core and hosting to be netstandard 2.0 so that it can be used in projects that have target frameworks of 472 and net core 7.0.
- Publishing sql core in the build pipeline to internal feed. 
- Made necessary changes in sql core project to make it compatible netstandard 2.0


publishSqlCoreProject.yml
```diff
@@ -0,0 +1,46 @@
+# ASP.NET
+# Build and test ASP.NET projects.
+# Add steps that publish symbols, save build artifacts, deploy, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/apps/aspnet/build-aspnet-4
+
+trigger:
```
Charles-Gagnon, line 6: This should probably be a part of the existing build/release pipeline - https://github.com/microsoft/sqltoolsservice/blob/main/azure-pipelines/build-and-release.yml

aasimkhan30: I will integrate it in the main pipeline after I am done finalizing all the steps. Converting this to draft PR for now.

Charles-Gagnon:
@aasimkhan30 Are these test failures due the issue you're fixing in the other PR?

src/Microsoft.SqlTools.Hosting/Utility/SqlClientEventListener.cs
```diff
@@ -41,7 +42,7 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
             {
                 if (payload != null)
                 {
-                    Logger.Verbose($"eventTID:{eventData.OSThreadId} {payload.ToString()}");
+                    Logger.Verbose($"eventTID:{Environment.CurrentManagedThreadId} {payload.ToString()}");
```
Charles-Gagnon, line 45: I did this because the thread the listener is running on isn't necessarily the same one that the event was logged on. So doing it this way could end up with the wrong thread ID.

Can you at least have an IFDEF that will still use the OSThreadId in netstandard2.1+ builds?

src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
```diff
@@ -12,6 +12,12 @@
 		<PreserveCompilationContext>true</PreserveCompilationContext>
 		<AssemblyTitle>SqlTools SqlCore Library</AssemblyTitle>
 		<Description>Provides core sql functionality for SQL server editors like Object explorer, Query Execution and Scripting</Description>
+		<TargetFramework>netstandard2.0</TargetFramework>
+		<LangVersion>8.0</LangVersion>
+		<AssemblyName>Microsoft.SqlTools.SqlCore</AssemblyName>
+		<Authors>Microsoft</Authors>
+		<Version>1.0.0</Version>
+		<Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
```
Charles-Gagnon, line 20: We already define this in https://github.com/microsoft/sqltoolsservice/blob/main/Directory.Build.props#L8, so you shouldn't need to redefine these common props

src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
```diff
@@ -36,4 +42,23 @@
 		<InternalsVisibleTo Include="Microsoft.SqlTools.ServiceLayer.Test.Common" />
 		<InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
 	</ItemGroup>
+
+	<PropertyGroup>
+		<TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);CopyProjectReferencesToPackage</TargetsForTfmSpecificBuildOutput>
```
Charles-Gagnon, line 47: What is all this for? What do you need to filter, and could you not use a nuspec instead?

src/Microsoft.SqlTools.SqlCore/ObjectExplorer/Nodes/INodeFilter.cs
```diff
@@ -13,23 +13,23 @@ namespace Microsoft.SqlTools.SqlCore.ObjectExplorer.Nodes
     /// <summary>
     /// Has information for filtering a SMO object by properties 
     /// </summary>
-    public interface INodeFilter
+    public abstract class INodeFilter
```
Charles-Gagnon, line 16: Why is this an abstract class? I don't see any implementation here.

aasimkhan30: It does implement static methods which are not allowed for an interface in lang version 8

Charles-Gagnon: Ah, that's unfortunate.

Almost feels like it would have been easier/better to just update our workload to net6...going forward we should be trying to avoid using netfx whenever possible. Anyways too late for that now I guess.

azure-pipelines/build.yml
```diff
@@ -281,13 +293,22 @@ steps:
...
   condition: true
 
-- task: NuGetCommand@2
+- task: DotNetCoreCLI@2
   displayName: 'NuGet push'
   condition: eq(variables['PUSH_SYMBOLS'], 'true')
   inputs:
     command: push
     packagesToPush: '$(Build.SourcesDirectory)/artifacts/nugetPackages/**/*.nupkg;!$(Build.ArtifactStagingDirectory)/**/*.symbols.nupkg'
-    publishVstsFeed: '2191dd5f-4aec-491b-ac50-568bbc331c8a'
```
cheenamalhotra, line 290: What does this change for existing packages being published? Or is it just a task switch?

Directory.Build.props
```diff
@@ -8,7 +8,6 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <HighEntropyVA>true</HighEntropyVA>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
-    <TargetFramework>net7.0</TargetFramework>
```
cheenamalhotra, line 11: What benefit did you get from removing this?

cheenamalhotra: Nvm, I found my answer

src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
```diff
@@ -12,6 +12,11 @@
 		<PreserveCompilationContext>true</PreserveCompilationContext>
 		<AssemblyTitle>SqlTools SqlCore Library</AssemblyTitle>
 		<Description>Provides core sql functionality for SQL server editors like Object explorer, Query Execution and Scripting</Description>
+		<TargetFramework>netstandard2.0</TargetFramework>
+		<LangVersion>8.0</LangVersion>
```
cheenamalhotra, line 16: Could you move this to Directory.Build.Props?

src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
```diff
@@ -12,6 +12,11 @@
...
+		<AssemblyName>Microsoft.SqlTools.SqlCore</AssemblyName>
+		<Authors>Microsoft</Authors>
```
cheenamalhotra, line 18: Move to Directory.Build.Props

src/Directory.Build.props
```diff
@@ -12,6 +12,7 @@
 		<RepositoryType>git</RepositoryType>
 		<PackageTags>SQL XPLAT</PackageTags>
 		<DebugType>portable</DebugType>
+		<Authors>Microsoft</Authors>
```
Charles-Gagnon, line 15: We already have this at the root build props. You shouldn't need to re-add it

https://github.com/microsoft/sqltoolsservice/pull/2173/files#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16eL7

src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
aasimkhan30: I think the rest of the projects are using lang version 11 (default in net core 7). The default for net standard 2.0 is 7.3 which does not support nullable types so I am forcing that here.

azure-pipelines/build.yml
aasimkhan30: Restored the task to original state.

src/Microsoft.SqlTools.SqlCore/ObjectExplorer/Nodes/INodeFilter.cs
aasimkhan30: I agree. We should move other projects to net core 7/8. However, that will be a bigger task.

Charles-Gagnon (approved):

PR #2175, opened 2023-08-18
Fixes BindingQueue, GetSignatureHelp and RefreshIntellisenseCache to be truly async

cheenamalhotra:
### Description

The core change is to make `parseAndBind` an Async method to allow continuations to happen, along with binding queue to process queue items asynchronously. This allows us to call `parseAndBind` without awaiting on method completion (as it can take as long as 40 seconds to complete), such that completion notification can be sent to ADS when results are ready, without blocking main thread.

This allows us to unblock main thread of STS notification queue while parse and bind happens in background.

### User impact:
- Fixes unwanted query hangs due to "signature help" notification & "refresh intellisense cache" command action.
- ❗Significant impact in time taken to open a new connection, as earlier the metadata queries were run alongwith opening connection synchronously.

#### Addresses below bugs:
- https://github.com/microsoft/azuredatastudio/issues/24045
- https://github.com/microsoft/azuredatastudio/issues/9170
- https://github.com/microsoft/azuredatastudio/issues/23400
- Major contribution to fixing https://github.com/microsoft/azuredatastudio/issues/7440
- Many more Bugs related with speed and performance of Connection time, Query Editor and Intellisense!

### Before (Insiders):

queryHang

### After (Dev):

queryHangFixed


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -370,49 +370,44 @@ private void ProcessQueue()
...
                                 {
-                                    try
-                                    {
-                                        // check if the binding tasks completed within the binding timeout                           
-                                        if (bindTask.Wait(bindTimeout))
-                                        {
-                                            queueItem.Result = result;
-                                        }
-                                        else
+                                    await bindTask.ContinueWithOnFaulted(t => Logger.Error("Binding queue threw exception " + t.Exception.ToString()))
```
Charles-Gagnon, line 379: ContinueWithOnFaulted already logs the error so do we really need to log it again? Might be useful to update that to take in a custom error message though since it's pretty generic right now...

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -370,49 +370,44 @@ private void ProcessQueue()
...
+                                        .ContinueWith(t =>
                                         {
-                                            cancelToken.Cancel();
+                                            if (t.IsCanceled)
```
Charles-Gagnon, line 382: When would the task itself be cancelled outside of the cancelToken already being cancelled?

cheenamalhotra: Do you want me to update `cancelToken` to cancel on `bindTimeoutInMs` instead?
Currently cancelToken doesn't have a timeout attached.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -551,18 +534,20 @@ private static TelemetryProperties CreatePeekTelemetryProps(bool succeeded, bool
...
-                    await requestContext.SendResult(new SignatureHelp());
+                    // Start task asynchronously without blocking main thread - by design.
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+                    GetSignatureHelp(textDocumentPosition, scriptFile)
+                        .ContinueWith(async task =>
+                    {
+                        var result = await task;
+                        if (result != null)
+                        {
+                            await requestContext.SendResult(result);
```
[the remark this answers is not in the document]
cheenamalhotra: I probably thought it wasn't necessary - saw no consequences on testing.. 
But I've added it back if you think we should always be sending a result.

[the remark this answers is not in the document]
Charles-Gagnon: If you don't then whatever sent that message will be waiting forever for a response. I haven't looked specifically into how the VS Code language server stuff is implemented so maybe that isn't a problem there, but it's better practice to just always ensure we send a response when we're done handling a message. 

Did you specifically test the scenario where GetSignatureHelp returns null? Is that even a normal/expected scenario?

Charles-Gagnon (approved):

PR #2181, opened 2023-08-22
Adding missing directories needed for integration tests

aasimkhan30:
https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=209765&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=6278c56e-b9dd-5471-9c3c-f3062039459c

Based on this build I realized I was not creating directories needed for "pt-BR". This pr fixes that.


Charles-Gagnon:
Did you look to see if there was something we could do to catch this issue earlier? It'd be great not to have to rely on digging through logs to see that there was an issue in the future.

Charles-Gagnon:
Nevermind, I see that this is already set as a PR validation run. Did you just not notice it with your previous PRs? Or did something happen so it didn't show up?

azure-pipelines/createBuildDirectories.sh
```diff
@@ -58,6 +58,7 @@ requiredLocDirectories=(
     "pl"
     "pl-PL"
     "pt-br"
+    "pt-BR"
```
Charles-Gagnon, line 61: Can you remove the `pt-br`? If not, do you know why this only affects some of the langs (for example, `ja-JP` doesn't seem to be affected by this?)

aasimkhan30: Finally figured out what's causing this issue. This problem happens with pt-BR, zh-Hant and zh-Hans.

- Net 472 stores these langs as        pt-BR, zh-Hant and zh-Hans 
- While net 7 stores these langs as   pt-br, zh-hant and zh-hans

While this is not a problem in windows as dirs are case insensitive, this causes a problem in linux as some projects which only target net7.0 won't have net472 required dirs created in their bin folder when we build those projects. This causes unsuccessful copy attempts and eventually fails the build. 

It does not happen in case of ja-JP as both net 472 and net 7 uses ja folder to store the loc files.

aasimkhan30:
@Charles-Gagnon , what's weird is that my previous builds did pass the checks.
https://github.com/microsoft/sqltoolsservice/pull/2178
https://github.com/microsoft/sqltoolsservice/pull/2176

azure-pipelines/createBuildDirectories.sh
Charles-Gagnon: Thanks for the investigation. Can you add this as a comment here for future reference?

Charles-Gagnon (approved):

azure-pipelines/createBuildDirectories.sh
```diff
@@ -39,35 +42,12 @@ dotnetProjectArray=(
 framework7="/bin/Debug/net7.0/"
 
 requiredLocDirectories=(
-    "cs"
-    "cs-CZ"
```
cheenamalhotra, line 43: Why are we removing others?

aasimkhan30: Those do not have the casing problem. Only the 3 langs have it pr-BR, zh-hans, zh-hant

PR #2182, opened 2023-08-22
Enabling FILEGROUPS tab experience to the Database Properties

ssreerama:
This PR enables the experiences of Filegroups tab to the database properties, where user can edit, create, remove the filegroups.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -753,6 +751,50 @@ private string ConfigureDatabase(InitializeViewRequestParams viewParams, Databas
...
+                                    newfileGroup.FileGroupType = fg.Type;
+                                    newfileGroup.Name = fg.Name;
+                                    newfileGroup.IsReadOnly = fg.IsReadOnly;
+                                    newfileGroup.IsDefault = fg.IsDefault;
+                                    newfileGroup.IsAutogrowAllFiles = fg.IsDefault;
+                                    prototype.Filegroups.Add(newfileGroup);
+                                }
+                                else
+                                {
+                                    var existedFilegroup = prototype.Filegroups.FirstOrDefault(x => x.Name == fg.Name);
```
corivera, line 772: ~To avoid n^2 loop iterations here, you could have a separate HashSet that contains all the existing names in the prototype.~

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -753,6 +751,50 @@ private string ConfigureDatabase(InitializeViewRequestParams viewParams, Databas
                             }
                         }
 
+                        if (!viewParams.IsNewObject && database.Filegroups != null)
+                        {
+                            HashSet<string> fileGroupToRemove = new HashSet<string>(prototype.Filegroups.Select(file => file.Name));
```
corivera, line 756: Since we aren't calling `Contains` on this Set until after we've trimmed its elements, could you start with this HashSet being empty and then add items to it as necessary rather than removing them? That would reduce our overall memory usage with this HashSet. For example, if you only ended up needing to track one FileGroup here to set `Removed = true` on, but the HashSet started out containing a hundred others that end up getting removed from the HashSet, then that was memory usage you didn't need.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
corivera: Disregard that, we need to get the actual FileGroup object from the prototype. Doing a Dictionary of filegroup Name to FileGroup object would work better there.

corivera: For example:
```
var groupNameMap = new Dictionary<string, Smo.FileGroup>();
foreach (FileGroup group in prototype.FileGroups)
{
    groupNameMap.Add(group.Name, group)
}

foreach (FileGroups fg in database.Filegroups)
{
    if (groupNameMap.TryGetValue(fg.Name, out existingFileGroup))
    {
        ...
    }
}
```

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseInfo.cs
```diff
@@ -69,4 +70,14 @@ public class DatabaseFile
         public double MaxSizeLimitInMb { get; set; }
     }
 
+    public class FileGroups
```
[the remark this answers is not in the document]
corivera: Although SMO has its own FileGroup class that gets used in the same context as this one, so maybe something like `FileGroupSummary` would be better? I definitely got confused looking through the `ConfigureDatabase` changes due to the similar naming

corivera (changes requested):
Left some initial comments

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: Achieved it without using any extra set, and modified the Files save logic too. Thanks

corivera (approved):

PR #2187, opened 2023-08-23
Add security tab

barbaravaldez:


Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerPrototype.cs
```diff
@@ -425,6 +453,35 @@ public bool UpdateMemoryValues(Microsoft.SqlServer.Management.Smo.Server server)
...
+                server.Settings.AuditLevel = this.currentState.LoginAuditing;
+                alterServer = true;
+            }
+
+            if (needServerRestart)
+            {
+                // user is applying the changes right now (not scripting/scheduling)
+                // and changes require server restart in order to be effective so
+                // we warn the user about the required restart action (see also SQLBU# 355718)
+                Logger.Information("Changes require server restart in order to be effective");
```
corivera, line 480: Does this need to be localized?

corivera (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerPrototype.cs
Charles-Gagnon: @barbaravaldez we should be doing something on the ADS side so it shows a toast notification. Can you move this check over there instead?

Charles-Gagnon (dismissed):
Need to figure out a better way to notify the user that a server restart is required

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerPrototype.cs
Charles-Gagnon: Or maybe it would be better as a message in the dialog...should chat with @erinstellato-ms and come up with a design and we can review it.

erinstellato-ms: I voted for message (toast notifications are easily missed by folks not familiar with ADS/VS Code)

PR #2188, opened 2023-08-23
Fixing sqlcore nuget dlls not being signed.

aasimkhan30:
The fix was to reorder the steps in pipeline so that we are packing sqlcore just after signing it. 

https://github.com/microsoft/sqltoolsservice/blob/1eba98384a4100151160d68ac4b1a4f39338dce1/azure-pipelines/build.yml#L286
This step was rebuilding some of the projects in the directory thereby overwriting the files we signed in the esrp step for sqlcore. By packaging sqlcore before the rebuild, we are making sure we get the signed files.

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -295,7 +295,7 @@ steps:
   inputs:
     command: custom
     custom: pack
-    arguments: '--output $(Build.SourcesDirectory)\artifacts\nugetPackages --no-build --no-restore src\Microsoft.SqlTools.SqlCore\Microsoft.SqlTools.SqlCore.csproj'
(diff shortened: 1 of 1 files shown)


Charles-Gagnon (approved):
@corivera I noticed that the Tools binaries aren't being signed either. Do you have time to look into fixing that?

Charles-Gagnon (approved):

corivera:
@Charles-Gagnon Sure

corivera:
Although I remember doing certificate checks before uploading them to nuget. Is this a recent bug?

corivera:
Maybe that was just the package itself getting signed

PR #2189, opened 2023-08-23
Removing script as feature from service layer to sqlcore

aasimkhan30:
Since the rest of the scripting code is heavily tied to sts classes, for the first iteration, I am just moving 'Script As' feature from sts to sql core. 

I will create a follow up PRs for moving out necessary tests from service layer to sqlcore for both scripting and OE.


src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
```diff
@@ -0,0 +1,38 @@
...
+                if (args.HasError)
+                {
+                    scriptAsTask.SetException(new Exception(args.ErrorMessage));
+                }
+                scriptAsTask.SetResult(scriptAsOperation.ScriptText);
+            };
+
+            scriptAsOperation.ProgressNotification += (sender, args) =>
+            {
+                scriptAsTask.SetException(new Exception(args.ErrorMessage));
```
Charles-Gagnon, line 31: Did you copy this from somewhere? Not sure why a "progress notification" is always being handled as an error.

src/Microsoft.SqlTools.SqlCore/Connection/SqlRetryProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+//
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Data.SqlClient;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.SqlCore.Connection.ReliableConnection
+{
+    public static class SqlRetryProviders
```
Charles-Gagnon, line 13: It would be great not to duplicate this...

It seems like having a small connection library that is used by this, STS and ManagedBatchParser wouldn't be a bad idea (and move all the "reliable connection" stuff from ManagedBatchParser to that). But something like that should be a separate PR - so maybe for this at least just move this class over completely and have ManagedBatchParser reference this instead?

@cheenamalhotra

kburtram (approved):

src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceProvider.cs
```diff
@@ -77,9 +74,7 @@ public static ExtensionServiceProvider CreateFromAssembliesInDirectory(IEnumerab
 
                 try
                 {
-                    assemblies.Add(
-                        context.LoadFromAssemblyName(
-                            AssemblyLoadContext.GetAssemblyName(path)));
+                    assemblies.Add(Assembly.LoadFrom(path));
```
Charles-Gagnon, line 77: What are the changes in this file for? It's unclear how they relate to the overall PR.

Charles-Gagnon (approved):

src/Microsoft.SqlTools.Hosting/Extensibility/ExtensionServiceProvider.cs
aasimkhan30: Sorry, I had to remove dependency on System.Runtime.Loader as it is not available in net 472. These changes are related to that.

Charles-Gagnon: Ah, and I assume this didn't show up as a problem before because we were never using this in a net472 environment? Interesting that it didn't cause a compile error.

aasimkhan30: Probably because we are using netstandard2.0 for compilation which has the required apis for it.

PR #2190, opened 2023-08-23
Throw exception on connection failure to be captured by frontend

cheenamalhotra:
Addresses https://github.com/microsoft/azuredatastudio/issues/24199

I was able to surround methods with try-catch and handle exceptions with best solutions available in respective services.
Let me know if anything needs fixing.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
@@ -153,25 +153,40 @@ namespace Microsoft.SqlTools.ServiceLayer.Agent
                 parameters.OwnerUri,
                 out connInfo);
 
-            if (connInfo != null)
+            try
             {
-                var serverConnection = ConnectionService.OpenServerConnection(connInfo);
-                var fetcher = new JobFetcher(serverConnection);
-                var filter = new JobActivityFilter();
-                var jobs = fetcher.FetchJobs(filter);
-                var agentJobs = new List<AgentJobInfo>();
-                if (jobs != null)
+                if (connInfo != null)
                 {
-                    foreach (var job in jobs.Values)
+                    var serverConnection = ConnectionService.OpenServerConnection(connInfo);
+                    var fetcher = new JobFetcher(serverConnection);
+                    var filter = new JobActivityFilter();
+                    var jobs = fetcher.FetchJobs(filter);
+                    var agentJobs = new List<AgentJobInfo>();
+                    if (jobs != null)
                     {
-                        agentJobs.Add(AgentUtilities.ConvertToAgentJobInfo(job));
+                        foreach (var job in jobs.Values)
+                        {
+                            agentJobs.Add(AgentUtilities.ConvertToAgentJobInfo(job));
+                        }
                     }
+                    result.Success = true;
+                    result.Jobs = agentJobs.ToArray();
+                    serverConnection.SqlConnectionObject.Close();
                 }
-                result.Success = true;
-                result.Jobs = agentJobs.ToArray();
-                serverConnection.SqlConnectionObject.Close();
+                await requestContext.SendResult(result);
             }
-            await requestContext.SendResult(result);
+            catch (Exception ex)
+            {
+                result.Success = false;
+                result.ErrorMessage = ex.Message;
+                Exception exception = ex.InnerException;
+                while (exception != null)
+                {
+                    result.ErrorMessage += Environment.NewLine + "\t" + exception.Message;
+                    exception = exception.InnerException;
+                }
+                await requestContext.SendResult(result);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


corivera (approved):

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1231,6 +1267,12 @@ internal async Task HandleAgentNotebooksRequest(AgentNotebooksParams parameters,
             {
                 result.Success = false;
                 result.ErrorMessage = e.ToString();
+                Exception exception = e.InnerException;
+                while (exception != null)
```
Charles-Gagnon, line 1271: Would be great to make a helper extension method for this (if one doesn't exist already...I know I've seen similar logic in various places already)

lewis-sanchez (approved):

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
cheenamalhotra: Sure, will follow up in a separate PR for this. Need to review we're logging errors from everywhere as well.

src/Microsoft.SqlTools.ServiceLayer/AzureBlob/BlobService.cs
```diff
@@ -60,16 +60,24 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
-                await requestContext.SendResult(response);
+                    SharedAccessSignatureCreator sharedAccessSignatureCreator = new SharedAccessSignatureCreator(sqlServer);
+                    string sharedAccessSignature = sharedAccessSignatureCreator.CreateSqlSASCredential(optionsParams.StorageAccountName, optionsParams.BlobContainerKey, optionsParams.BlobContainerUri, optionsParams.ExpirationDate);
+                    response.SharedAccessSignature = sharedAccessSignature;
+                    await requestContext.SendResult(response);
+                }
+            }
+            catch (Exception ex)
+            {
+                await requestContext.SendError(ex);
```
Charles-Gagnon, line 79: Don't need to do this - we already catch any uncaught exceptions for requests and send an error back in the MessageDispatcher

https://github.com/microsoft/sqltoolsservice/blob/38f33fa211086d0def50b016dd6940b942ac9255/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs#L162

cheenamalhotra: That's good to know!

cheenamalhotra: It now makes me think, should we always let this be handled by dispatcher then? 🤔

Charles-Gagnon: That's what I prefer for messages that are set up to expect errors to be thrown. But some messages have a "success" property that's used to determine if failures happen (like [agent](https://github.com/microsoft/sqltoolsservice/pull/2190/files#diff-81a70bd053fe2c2d1385f7bad2a4213e86b421ea8d13a35a151b6c8d77423e8cR180) and so we should follow the pattern there for those that use that unless you verify that they can also handle errors being thrown in a clean manner.

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -307,44 +315,51 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+                            {
+                                sqlConn.Dispose();
+                            }
+                        };
+                        response.Result = true;
+                    }
+                    catch
+                    {
+                        // Ensure that the connection is closed if any error occurs while starting up the task
+                        sqlConn.Dispose();
```
Charles-Gagnon, line 355: using would be better here

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -307,44 +315,51 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+                        throw;
+                    }
                 }
-
             }
-            else
+            catch (Exception ex)
             {
-                response.Result = false;
+                Logger.Error(ex);
```
Charles-Gagnon, line 362: Maybe explain here that the result will be sent back as false if an error occurs, so we just need to log for additional information.

src/Microsoft.SqlTools.ServiceLayer/LanguageExtensibility/ExternalLanguageService.cs
```diff
@@ -90,12 +90,20 @@ public async Task HandleExternalLanguageDeleteRequest(ExternalLanguageDeleteRequ
...
+                        ExternalLanguageOperations.DeleteLanguage(dbConnection, parameters.LanguageName);
+                    }
 
-                await requestContext.SendResult(response);
+                    await requestContext.SendResult(response);
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error(ex);
+                    await requestContext.SendError(ex);
```
Charles-Gagnon, line 105: unneeded

src/Microsoft.SqlTools.ServiceLayer/LanguageExtensibility/ExternalLanguageService.cs
```diff
@@ -122,12 +130,20 @@ public async Task HandleExternalLanguageUpdateRequest(ExternalLanguageUpdateRequ
...
+                    {
+                        ExternalLanguageOperations.UpdateLanguage(dbConnection, parameters.Language);
+                    }
 
-                await requestContext.SendResult(response);
+                    await requestContext.SendResult(response);
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error(ex);
```
Charles-Gagnon, line 144: unneeded

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -129,28 +129,36 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
                     }
                 }
-            }
 
-            await requestContext.SendResult(response);
+                await requestContext.SendResult(response);
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex);
```
Charles-Gagnon, line 159: unneeded

src/Microsoft.SqlTools.ServiceLayer/LanguageExtensibility/ExternalLanguageService.cs
```diff
@@ -155,12 +171,20 @@ public async Task HandleExternalLanguageStatusRequest(ExternalLanguageStatusRequ
...
+                        response.Status = ExternalLanguageOperations.GetLanguageStatus(dbConnection, parameters.LanguageName);
+                    }
 
-                await requestContext.SendResult(response);
+                    await requestContext.SendResult(response);
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error(ex);
+                    await requestContext.SendError(ex);
```
Charles-Gagnon, line 186: unneeded

src/Microsoft.SqlTools.ServiceLayer/LanguageExtensibility/ExternalLanguageService.cs
```diff
@@ -187,12 +211,20 @@ public async Task HandleExternalLanguageListRequest(ExternalLanguageListRequestP
...
+                        response.Languages = ExternalLanguageOperations.GetLanguages(dbConnection);
+                    }
 
-                await requestContext.SendResult(response);
+                    await requestContext.SendResult(response);
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error(ex);
+                    await requestContext.SendError(ex);
```
Charles-Gagnon, line 226: unneeded

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -94,6 +94,7 @@ public void InitializeService(ServiceHost serviceHost)
 
             Task task = Task.Run(async () => await requestHandler()).ContinueWithOnFaulted(async t =>
             {
+                Logger.Error(t.Exception);
```
Charles-Gagnon, line 97: unneeded

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -242,21 +250,30 @@ internal static async Task GetServerContextualization(GetServerContextualization
...
+                await requestContext.SendResult(new TableMetadataResult
+                {
+                    Columns = metadata
+                });
+            }
+            catch (Exception ex)
             {
-                Columns = metadata
-            });
+                Logger.Error(ex.ToString());
```
Charles-Gagnon, line 274: unneeded

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
```diff
@@ -75,143 +76,216 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
-            await requestContext.SendResult(new RenameRequestResponse());
+            try
+            {
+                var handler = this.GetObjectTypeHandler(requestParams.ObjectType);
+                await handler.Rename(requestParams.ConnectionUri, requestParams.ObjectUrn, requestParams.NewName);
+                await requestContext.SendResult(new RenameRequestResponse());
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex);
```
Charles-Gagnon, line 87: unneeded....

(you know what, you get the point 😜 )

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -372,8 +387,9 @@ private bool IsBackupRestoreOperationSupported(string ownerUri, out ConnectionIn
                     }
                 }
             }
-            catch
```
lewis-sanchez, line 375: Usually these kinds of catches are used to catch exceptions from unmanaged code, which is something to keep in mind if that is still a potential risk.

Charles-Gagnon: In this case we should just be using a `using` anyways to ensure the connection is closed even if an error happens

cheenamalhotra: All exceptions are derivatives of Exception class, so I don't expect any change in behavior.

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
```diff
@@ -1508,6 +1534,22 @@ string targetDatabase
             return result;
         }
 
+        /// <summary>
+        /// Helper method to create result object from exception
+        /// </summary>
+        /// <param name="ex"></param>
+        private string GetErrorMessageFromException(Exception e)
```
Charles-Gagnon, line 1541: Would suggest putting this in a more common location (and maybe combining with https://github.com/microsoft/sqltoolsservice/blob/38f33fa211086d0def50b016dd6940b942ac9255/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs#L167)

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
cheenamalhotra: Addressed

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/Agent/AgentService.cs
cheenamalhotra: Made an extension

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
Charles-Gagnon: Was curious, so looked it up and turns out that is not necessarily true - although you have to specifically disable wrapped exceptions.

https://learn.microsoft.com/en-us/dotnet/csharp/how-to/how-to-catch-a-non-cls-exception

That being said - this shouldn't be an issue in STS.

SakshiS-harma (approved):

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
cheenamalhotra: It seems to cause test failure, I'll revert this part as it's not critical.

PR #2192, opened 2023-08-24
Adding project references in Sql core nuget package

aasimkhan30:


azure-pipelines/build.yml
```diff
@@ -59,16 +59,16 @@ steps:
     useGlobalJson: true
 
 - task: PowerShell@2
-  displayName: 'Add version number to SqlCore project for packaging'
+  displayName: 'Add version number to SqlCore nuspec for packaging'
```
Charles-Gagnon, line 62: Make sure the DLL has the correct version too

azure-pipelines/build.yml
```diff
@@ -59,16 +59,16 @@ steps:
...
   inputs:
     targetType: 'inline'
     script: |
-      $projectFile = '$(Build.SourcesDirectory)\src\Microsoft.SqlTools.SqlCore\Microsoft.SqlTools.SqlCore.csproj'
+      $nuspecFile = '$(Build.SourcesDirectory)\src\Microsoft.SqlTools.SqlCore\Microsoft.SqlTools.SqlCore.nuspec'
```
Charles-Gagnon, line 66: FYI - you can also just pass in properties directly 

https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-using-a-nuspec-file

src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.nuspec
```diff
@@ -0,0 +1,17 @@
...
+    <id>Microsoft.SqlTools.SqlCore</id>
+    <version>1.0.0</version>
+    <authors>Microsoft</authors>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <license type="expression">MIT</license>
+    <projectUrl>https://github.com/Microsoft/sqltoolsservice/</projectUrl>
+    <description>Provides core SQL functionality for SQL server editors like Object explorer, Query Execution and Scripting</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <dependencies>
+      <dependency id="Microsoft.SqlServer.SqlManagementObjects" version="170.12.0" />
```
Charles-Gagnon, line 13: This should be using the versions from the Packages.props so we don't risk forgetting to update this. You'll probably need to make a Property for these versions and then reference that in the Packages.props though.

kburtram (approved):

PR #2193, opened 2023-08-25
Add Attach Database functionality to Object Management Service

corivera:
This is needed to enable the Attach Database dialog in ADS


src/Microsoft.SqlTools.ServiceLayer/Admin/Contracts/GetAssociatedFilesRequest.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Microsoft.SqlTools.ServiceLayer.Admin.Contracts
+{
+    /// <summary>
+    /// Params for a get database info request
+    /// </summar>
+    public class GetAssociatedFilesParams
+    {
+        /// <summary>
+        /// URI identifier for the connection to get the server folder info for
```
ssreerama, line 18: Is this the whole description? or anything after for?

corivera: Rephrased it a bit

PR #2194, opened 2023-08-25
Return context after it has been generated

lewis-sanchez:
This PR returns the generated context back to the client, so that the client doesn't need to wait for it to be written out to a file before it can be read.


src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/GenerateServerContextualizationCompleteNotification.cs
```diff
@@ -0,0 +1,36 @@
...
+        public string OwnerUri { get; set; }
+        /// <summary>
+        /// Indicates if generating context completed successfully
+        /// </summary>
+        public bool CompletedGeneratingContext { get; set; }
+        /// <summary>
+        /// Holds any error messages, if errors were encountered while
+        /// generating context
+        /// </summary>
+        public string Messages { get; set; }
```
kisantia, line 24: have Error in the name so it's clear this has error messages without needing to look at the comments

src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/GenerateServerContextualizationCompleteNotification.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Metadata.Contracts
+{
+    public class GenerateServerContextualizationCompleteParams
+    {
+        /// <summary>
+        /// The URI identifying the owner of the connection
+        /// </summary>
+        public string OwnerUri { get; set; }
```
kisantia, line 15: this isn't getting set when there's an error. Should this be nullable?

src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/GenerateServerContextualizationCompleteNotification.cs
lewis-sanchez: I was thinking that it would default to an empty string, but I'm thinking that setting the owner URI is a better way to go.

src/Microsoft.SqlTools.ServiceLayer/Metadata/Contracts/GenerateServerContextualizationCompleteNotification.cs
lewis-sanchez: I was thinking that it would default to an empty string, but I'm thinking that setting the owner URI is a better way to go.

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -153,27 +153,47 @@ internal static void GenerateServerContextualization(GenerateServerContextualiza
...
+                                MetadataScriptTempFileStream.Write(connectionInfo.ConnectionDetails.ServerName, scripts);
+                            }
+                            catch (Exception ex)
+                            {
+                                Logger.Error($"An error was encountered while writing to the cache. Error: {ex.Message}");
+
+                                var contextFailedToCompleteParams = new GenerateServerContextualizationCompleteParams()
+                                {
+                                    OwnerUri = contextualizationParams.OwnerUri,
+                                    ErrorMessage = ex.Message
```
kisantia, line 172: is `CompletedGeneratingContext` expected to be false here? It would be clearer when reading the code to explicitly set it

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
```diff
@@ -153,27 +153,47 @@ internal static void GenerateServerContextualization(GenerateServerContextualiza
...
-                        catch (Exception ex)
+                        else
                         {
-                            Logger.Error($"An error was encountered while writing to the cache. Error: {ex.Message}");
+                            Logger.Error("Failed to generate server scripts");
+
+                            var contextFailedToCompleteParams = new GenerateServerContextualizationCompleteParams()
+                            {
+                                OwnerUri = contextualizationParams.OwnerUri,
+                                ErrorMessage = "Failed to generate server scripts"
```
kisantia, line 184: error messages that'll show in the UI in ADS should get localized

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/Metadata/MetadataService.cs
lewis-sanchez: Localized strings in latest commit.

cheenamalhotra (approved):

Charles-Gagnon:
Do you actually need this? The context will only be retrieved when an editor is opened right? If so couldn't you just have the request to get the context not return until the generation is done? (all done in background threads of course to be non-blocking)

PR #2199, opened 2023-08-29
Simplify connection cleanup and clear pools to prevent lingering connections

corivera:
After opening an object management dialog on an existing database, there's an issue where trying to then detach or drop the database fails due to the database still being in use. Looks like this is due to connection pooling holding open connections to that database, so I added a step where we clear the connection pool before doing a drop or detach operation. I also cleaned up our connection disconnect logic to reduce redundancy


Charles-Gagnon:
@cheenamalhotra Let's make sure that the pooling you enabled isn't causing any other similar issues

@corivera Did you test out whether you still have hangs if you do other operations on the same server connection?

https://github.com/microsoft/sqltoolsservice/pull/2127 for reference

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerHandler.cs
```diff
@@ -32,44 +32,41 @@ public override bool CanHandleType(SqlObjectType objectType)
...
+                    StorageSpaceUsageInMB = prototype.StorageSpaceUsageInMB,
+                    Version = prototype.Version,
+                    MinServerMemory = prototype.MinServerMemory,
+                    MaxServerMemory = prototype.MaxServerMemory,
+                    AutoProcessorAffinityMaskForAll = prototype.AutoProcessorAffinityMaskForAll,
+                    AutoProcessorAffinityIOMaskForAll = prototype.AutoProcessorAffinityIOMaskForAll,
+                    NumaNodes = prototype.NumaNodes
+                };
             }
+            var context = new ServerViewContext(requestParams);
```
corivera, line 68: The connection for this context wasn't being used, and it also already gets disposed when the view itself is disposed.

corivera:
@Charles-Gagnon the hangs were happening after the ServerConnection object had already been disconnected, so it wasn't related to any pending operation that I could tell. The hangs disappear when re-running the Drop command after waiting about 10 minutes, which further points to it being a connection pooling issue, since pooled connections persists for 4-8 minutes according to the docs: https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-connection-pooling#removing-connections

Charles-Gagnon:
The PR I linked to has pooling enabled for all connections that are opened through the connection service (as long as the config is enabled, which it is by default for dev).

So if that's the case, isn't that going to cause problems where everything else is opening up pooled connections? It shouldn't just be affecting this one place (unless I'm missing something)

corivera:
This only impacted drop and detach database, since those are blocked if the database is still in use

corivera:
These are also direct connections to the database, rather than the server itself if that makes any difference

corivera:
Discussed offline with @cheenamalhotra and we agreed that leaving connections open from pooling was more acceptable than taking a perf hit from new connections every time, like if users are on slow networks or using a VPN.

corivera:
Scratch that, using ClearPool after opening a connection for just doing Drops & Detaches may be an acceptable workaround

cheenamalhotra:
Firstly, pooled connections are not the only ones to blame here. There are connection leaks in STS in multiple paths/features that should close "ideally" but they don't. So even if you disable pooling that doesn't make things 100% work.

Pooling is highly important for performance reasons, we've had multiple reports of slow connection time, and all operations slow - due to the same reason that everytime a new connection was opened it was slow being non-pooled. And when working in slow/VPN networks, making 1 connection time can be as slow as 10-12 seconds.

What I am working on is supporting "Clearing All Pooled connections" command in ADS - that will clear all pooled connections and users can proceed as needed. And secondly, fixing all connection leaks one by one which is more painful but absolutely needed for reliability.

corivera:
I'll try out the ClearPool option and re-open this PR if that works. Otherwise, I'll remove the pooling changes here and just merge the simple connection cleanup ones

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/DetachDatabaseRequest.cs
```diff
@@ -16,6 +16,10 @@ public class DetachDatabaseRequestParams : GeneralRequestDetails
         /// </summary>
         public string ObjectUrn { get; set; }
         /// <summary>
+        /// The target database name.
+        /// </summary>
+        public string Database { get; set; }
```
corivera, line 21: Have to pass in the database name for these operations so that we generate the same connection string as the pool with the lingering db connections.

Charles-Gagnon (approved):

PR #2200, opened 2023-08-31
Enable QueryStore tab to the database properties

### LINKED ISSUE #2212: Provide SMO database properties experience to the QueryStore Tab

ssreerama:
This PR adds the ability to view/update query store options to the database proeprties.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -89,6 +98,33 @@ static DatabaseHandler()
             displayFileTypes.TryAdd(FileType.Log, SR.prototype_file_logFile);
             displayFileTypes.TryAdd(FileType.FileStream, SR.prototype_file_filestreamFile);
 
+            displayOperationModeOptions.TryAdd(QueryStoreOperationMode.Off, CommonConstants.QueryStoreOperationMode_Off);
```
Charles-Gagnon, line 101: I don't think we should be localizing option values like OFF and READ_ONLY - do we do that in SSMS?

@erinstellato-ms what do you think?

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
```diff
@@ -66,6 +82,64 @@ protected override void SaveProperties(Database db)
...
+            if (db.IsSupportedObject<QueryStoreOptions>())
+            {
+                try
+                {
+                    // Querystore propeties do not initialized at the time of create database
+                    var desiredState = this.currentState.queryStoreOptions.DesiredState;
+                }
+                catch (Exception ex)
+                {
+                    areQueryStorePropertiesInitialized = false;
```
Charles-Gagnon, line 95: What exception are you expecting to be thrown here? I'm not sure what "initialized" would mean in this case, or why we'd want to skip setting the values completely in that case.

corivera (approved):
Looks good to me pending other people's comments.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: Other couple of dropdown options(All,Auto,None,Custom) are translated in the MS quey store help doc, so localized these too.

erinstellato-ms: What's our general practice on localizing options in ADS?  Specifically, options in a drop down?  Example: Backup functionality (in preview), we have Set backup compression as a drop down with three options.  Are those localized?

Fundamentally...first, let's be consistent across ADS.  Second, I will have to install another language of SSMS to see what we do there :)

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype130.cs
ssreerama: While create database the value this.currentState.queryStoreOptions is not null but its properties (ex: DesiredState...etc) throws "'this.currentState.queryStoreOptions.DesiredState' threw an exception of type 'System.NullReferenceException'"

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: I have created a task to verify and add/remove all such dropdown options:
https://github.com/microsoft/sqltoolsservice/issues/2213

erinstellato-ms (approved):

corivera (approved):
Approved pending my other comments

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
```diff
@@ -562,6 +564,21 @@ public DatabaseData(CDataContainer context, string databaseName)
...
+                // Supported from Sql Server 2016 and higher
+                // QueryStore properties are not initialized at the time of database creation
+                if (db.IsSupportedObject<QueryStoreOptions>())
+                {
+                    try
+                    {
+                        var queryStoreActualStateProperty = db.QueryStoreOptions.ActualState;
+                        this.queryStoreOptions = db.QueryStoreOptions;
+                    }
+                    catch (NullReferenceException)
```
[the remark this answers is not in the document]
ssreerama: db.QueryStoreOptions is not null, but the properties(actualState ...) inside it are Not initialized and when accessing them it throws the null reference exception.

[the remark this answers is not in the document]
corivera: Could you add a comment here for that?

PR #2202, opened 2023-08-31
Only delete database backup and restore history after the database has been successfully dropped

corivera:
For this issue: https://github.com/microsoft/azuredatastudio/issues/24194

I'll close that after doing an STS version bump in ADS. SSMS has the same issue with deleting backup history before doing the drop, so I'm not sure if that's an intentional behavior or an old bug.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -495,11 +495,11 @@ public string Drop(DropDatabaseRequestParams dropParams)
                             var sqlConn = dataContainer.ServerConnection.SqlConnectionObject;
                             SqlConnection.ClearPool(sqlConn);
                         }
+                        smoDatabase.Drop();
```
Charles-Gagnon, line 498: I put this in the other one, but we should consider having these be a transactional operation so that if either of them fails the whole thing is rolled back.

PR #2204, opened 2023-09-01
update projects nuget to 0.1.12-preview

kisantia:
Updating the Microsoft.SqlServer.DacFx.Projects nuget version used to 0.1.12-preview, which updates the default SDK used when creating new SDK-style projects to be version 0.1.12-preview (coincidence that the version numbers are the same), which has bug fixes and support for the new target platform.


zijchen (commented):
Sorry, clicked too fast, meant to approve instead of merging this PR. Can't add review anymore, so this comment is my approval...

PR #2216, opened 2023-09-07
Optimize parallel message processing to be enabled by default

cheenamalhotra:
This PR makes necessary changes + some grooming to enable parallel message processing by default.
_Enabling this feature by default applies to all feature services where handlers have called in for enabling this property._
 
Noteworthy changes:
- The limit on number of parallel operations that can be queued is now configurable by user. It's set to 100 by default, which should be optimal.
- PMP is enabled on all Object Explorer operations.
- PMP is disabled for `SubsetRequest` since the execution of this request should be ordered and not in parallel. Disabling PMP on this one will always queue this request on the main dispatcher thread and not as a detached task.

image


alanrenmsft (dismissed):

Charles-Gagnon:
One thing that would probably be nice is to add further explanation of what "detached background" vs "parallel processing" is, since it can be confusing to have a "background" task that still blocks the message dispatcher from handling other messages.

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
```diff
@@ -153,8 +154,19 @@ public void Stop()
                             }
                         }
 
-                        await requestHandler(typedParams, requestContext);
-                        Logger.Verbose($"Finished processing message with id[{requestMessage.Id}], of type[{requestMessage.MessageType}] and method[{requestMessage.Method}]");
+                        if (isBackgroundTask)
+                        {
+                            // Start a detached task that runs in background, unblocking dispatcher thread.
+                            await Task.Factory.StartNew(() => Task.Run(async () =>
```
Charles-Gagnon, line 160: Why bother nesting the calls here vs just calling Task.Run and not awaiting? These should be doing the same thing so it seems unnecessary.

https://devblogs.microsoft.com/pfxteam/task-run-vs-task-factory-startnew/

Charles-Gagnon:
I realize now I misread what the code is doing (another reason why it's super important to fully explain what a "detached background thread" is). Since, unless I'm misreading this again, you're basically forcing parallel message processing for these messages.

A couple of other things: 

1. I'm definitely concerned about running into issues with forcing parallel processing on messages without really thorough testing. What's your plan for ensuring we don't regress in any areas?(https://github.com/microsoft/azuredatastudio/issues/24095 being an example)
2. It's really confusing to have both parallel message processing AND this given that they're basically doing the same thing. I'd suggest either removing the parallel message processing completely - or maybe even better is to leave it but repurpose it for this and have it default enabled as true. That would let customers disable it if there turns out to be issues while we work on a fix.

cheenamalhotra:
> 1. I'm definitely concerned about running into issues with forcing parallel processing on messages without really thorough testing. What's your plan for ensuring we don't regress in any areas?([Inconsistent result order in Notebooks azuredatastudio#24095](https://github.com/microsoft/azuredatastudio/issues/24095) being an example)
> 2. It's really confusing to have both parallel message processing AND this given that they're basically doing the same thing. I'd suggest either removing the parallel message processing completely - or maybe even better is to leave it but repurpose it for this and have it default enabled as true. That would let customers disable it if there turns out to be issues while we work on a fix.

Agreed, parallel message processing is likely similar and I also realized that after preparing this change. The issues arose with that option due to it being enabled in lot more request handlers, where there was lack of synchronization or results weren't sent in the order of how requests were received. So if we should repurpose that, I would be disabling it from handlers that are not ready to work in parallel, to prevent issues like [azuredatastudio#24095](https://github.com/microsoft/azuredatastudio/issues/24095)

Before enabling this API on handlers, we need to be very selective with a design review of whether internal jobs are meant to be synchronized with other handlers and if it can support in-parallel execution. All scenarios need to be considered before making handlers use this option. It will include we ensure:
- All collections used are concurrent
- Resource locking is in place where necessary
- Responses are ordered where multiple requests need to be processed/responded to in the order of how they're received.

I will update the API docs with more detailed comments as well :)

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
cheenamalhotra: Not really, if you run a simple test, you'll notice Task.Run doesn't detach the thread from the main thread, whereas Task.Factory.StartNew does.

Link to example: https://gist.github.com/cheenamalhotra/b47770f5fbc7330aa79968a8999c8884

Charles-Gagnon: You have a bug in your example, `Task.Factory.StartNew` returns a `Task<TResult>`. In your case the TResult is a task itself, so the return type is `Task<Task>`, but since you only await once it never awaits the inner task (the delay).

If you add a second await (or call `.Unwrap()` then both of them behave the same : 

`await Run("Task.Factory.StartNew", async () => await await Task.Factory.StartNew(async () => await PrintAndDelayAsync("Task.Factory.StartNew", 1)));`

This blob explains it better : https://code-maze.com/csharp-task-run-vs-task-factory-startnew/ (the `The async Awareness
` section)

> On top of this, StartNew wraps this in a Task construct. Eventually, we end up with a Task<Task<string>> instance which is not what we desire.



So what you have is essentially the same as calling Task.Run directly and not awaiting the task it returns.

Charles-Gagnon: And actually that blog clearly states that Task.Run should be preferred : 

> Now, let’s talk about the difference from the TaskScheduler context. Task.Run(action) internally uses the default TaskScheduler, which means it always offloads a task to the thread pool.  StartNew(action), on the other hand, uses the scheduler of the current thread which may not use thread pool at all! 

> This can be a matter of concern particularly when we work with the UI thread! If we initiate a task by StartNew(action) within a UI thread, it will utilize the scheduler of the UI thread and no offloading happens. That means, if the task is a long-running one, UI will soon become irresponsive. Task.Run is free from this risk as it will always offload work to the thread pool no matter in which thread it has been initiated. So, Task.Run is the safer option in such cases.

Charles-Gagnon:
Good summary. I would say that it'd be fine to revert all handlers back to synchronous by default and then do what you suggested - we originally were very optimistic that the parallel message processing wouldn't cause any problems but clearly that isn't the case so a slower approach makes sense to me!

src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs
cheenamalhotra: > And actually that blog clearly states that Task.Run should be preferred :

I wasn't aware of the thread scheduler defaults being a concern, so thanks for sharing that. In that case we can explicitly specify the `TaskScheduler` when calling `StartNew` to address that.

Using `Task.Run` is an alternative but we'll have to disable/re-enable C# Pragma warning to not await on it to make it truly non-blocking which I'm trying to avoid. I'll update the defaults in `Task.Factory.StartNew` to ensure we're doing the right thing.

Charles-Gagnon: You can also assign it to a discard var 

e.g. 

`_ = Task.Run(async () =>` 

https://github.com/microsoft/sqltoolsservice/blob/16c7e323567023b08200ab35b2bfb2a5042385e0/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs#L353

Charles-Gagnon: I have a very slight preference to just always use Task.Run since it's simpler and will help avoid someone copying it and forgetting to add the correct set of defaults. But not a blocking concern if you'd prefer to use StartNew

alanrenmsft (approved):

brian-harris (approved):

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1195,6 +1193,7 @@ private IAuthenticator GetAuthenticator(CommandOptions commandOptions)
 
         private Task HandleEncryptionKeysNotificationEvent(EncryptionKeysChangeParams @params, EventContext context)
         {
+            Logger.Verbose("EncryptionKeysNotificationEvent");
```
Charles-Gagnon, line 1196: We already log the incoming messages 

https://github.com/microsoft/sqltoolsservice/blob/16c7e323567023b08200ab35b2bfb2a5042385e0/src/Microsoft.SqlTools.Hosting/Hosting/Protocol/MessageDispatcher.cs#L211

Do we really need another log that's basically doing the same thing?

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
```diff
@@ -123,11 +123,11 @@ public override void InitializeService(IProtocolEndpoint serviceHost)
 
             // Not enabling parallel processing for OEService as it might cause OE to hang.
```
Charles-Gagnon, line 124: Remove this, and have you verified that whatever was causing this issue is definitely not going to be a problem anymore?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -176,7 +176,7 @@ public void InitializeService(ServiceHost serviceHost)
             serviceHost.SetRequestHandler(ExecuteDocumentSelectionRequest.Type, HandleExecuteRequest, true);
             serviceHost.SetRequestHandler(ExecuteDocumentStatementRequest.Type, HandleExecuteRequest, true);
             serviceHost.SetRequestHandler(ExecuteStringRequest.Type, HandleExecuteRequest, true);
-            serviceHost.SetRequestHandler(SubsetRequest.Type, HandleResultSubsetRequest, true);
+            serviceHost.SetRequestHandler(SubsetRequest.Type, HandleResultSubsetRequest);
```
Charles-Gagnon, line 179: Add comment here explaining why this needs to be false

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
cheenamalhotra: The timing of logs can differ based on when the handler task begins (which is based on semaphore) so I see no harm in having both logs. It can help us investigate issues if parallel message processing blocks tasks for longer periods of time.

src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs
cheenamalhotra: I haven't noticed any hangs with this and there would have been other reasons why it happened before, as we noticed it happening even with disabled PMP - will remove comment.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
Charles-Gagnon: What about adding the code to the message dispatcher instead? Saves having to add this line to every single request/event

PR #2220, opened 2023-09-08
Fix missing index include column list

lewis-sanchez:
This PR is the necessary backend change to fix https://github.com/microsoft/azuredatastudio/issues/20249

This PR fixes an issue around index column names appearing in both the index column list and include column list.

Before: 
image

After:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
@@ -442,12 +442,14 @@ namespace Microsoft.SqlTools.ServiceLayer.ExecutionPlan
 
         private static List<ExecutionPlanRecommendation> ParseRecommendations(ShowPlanGraph g, string fileName)
         {
-            return g.Description.MissingIndices.Select(mi => new ExecutionPlanRecommendation
+            var recommendations =  g.Description.MissingIndices.Select(mi => new ExecutionPlanRecommendation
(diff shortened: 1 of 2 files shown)


erinstellato-ms (approved):

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
```diff
@@ -469,6 +471,8 @@ private static string ParseMissingIndexQueryText(string fileName, string impact,
 GO
 */
 ";
+
+            return missingIndexQueryText;
```
kisantia, line 475: are the changes in this file also needed for the fix or are they code cleanup? Also, what is the commented out code for right above this?

kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/ExecutionPlan/ExecutionPlanGraphUtils.cs
lewis-sanchez: That's part of a verbatim string and it helps to generate the SQL to create the missing suggested index. It's commented out to prevent a user from accidentally creating indexes they don't need.
image

lewis-sanchez: The changes here were minor code cleanup to make it easier for me to debug without having to switch between files to see logic before the return.

PR #2223, opened 2023-09-09
Get the backup compression and checksum values directly from server

barbaravaldez:
Fix https://github.com/microsoft/azuredatastudio/issues/24321

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerPrototype.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerPrototype.cs
@@ -2326,8 +2326,8 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
 
             private void LoadDBSettingsProperties()
             {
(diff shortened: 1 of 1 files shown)


corivera (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerPrototype.cs
```diff
@@ -2326,8 +2326,8 @@ private void LoadSecurityProperties()
 
             private void LoadDBSettingsProperties()
             {
-                this.checkBackupChecksum = this.configService.GetServerSmoConfig(server, this.configService.BackupChecksumDefaultPropertyNumber).ConfigValue == 1;
-                this.checkCompressBackup = this.configService.GetServerSmoConfig(server, this.configService.BackupCompressionDefaultPropertyNumber).ConfigValue == 1;
+                this.checkBackupChecksum = server.Configuration.DefaultBackupChecksum.ConfigValue == 1;
```
corivera, line 2329: Doesn't need to happen in this PR, but are there other places we're getting the config from that can use SMO Configuration properties like this instead?

barbaravaldez: yup, I think we use it for Min and max memories

PR #2238, opened 2023-09-12
Add ability to only show folders in File Browser dialogs

corivera:
Currently we can only select files in this dialog, but some Basic Admin input fields need to be able to select folders, like the default database file location, for example.


src/Microsoft.SqlTools.ServiceLayer/FileBrowser/FileBrowserOperation.cs
```diff
@@ -238,7 +242,7 @@ public List<FileTreeNode> GetChildren(string filePath)
 
                 // if the node is a directory, or if we are browsing for files and the file name is allowed,
```
Charles-Gagnon, line 243: nit: update comment too

Charles-Gagnon (approved):

PR #2242, opened 2023-09-15
Updated TDE NuGet and contracts

smartura:
This pull request updates the version of NuGet used for TDE migrations.


Packages.props
```diff
@@ -30,7 +30,7 @@
 		<PackageReference Update="Microsoft.SqlServer.Migration.Assessment" Version="1.0.20230301.46" />
 		<PackageReference Update="Microsoft.SqlServer.Migration.Logins" Version="1.0.20230407.56" />
 		<PackageReference Update="Microsoft.SqlServer.Management.SqlParser" Version="170.9.0" />
-		<PackageReference Update="Microsoft.SqlServer.Migration.Tde" Version="1.0.0-preview.1.0.20230720.98" />
+		<PackageReference Update="Microsoft.SqlServer.Migration.Tde" Version="1.0.0-preview.1.0.20230914.107" />
```
siyao-Siyang, line 33: need add nuget package in bin folder too?

siyao-Siyang (approved):

Packages.props
smartura: I realized I never responded to this comment.  Sorry about that!

If you look at https://github.com/microsoft/sqltoolsservice/blob/main/nuget.config
you'll see it has `nuget.org` and `./bin/nuget` listed as packageSources.

The reason why I don't need the nuget package in the bin folder is because the nuget is public on nuget.org.

PR #2245, opened 2023-09-20
Remove redundant usage of object URNs in database handler

corivera:
We pass both the object URN and database name in several places in the database handler code. Since we know the object type is always a database the URN is redundant here.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -646,25 +646,28 @@ public void PurgeQueryStoreData(PurgeQueryStoreDataRequestParams purgeParams)
...
         {
             ConnectionInfo connectionInfo = this.GetConnectionInfo(connectionUri);
             var originalDatabaseName = connectionInfo.ConnectionDetails.DatabaseName;
             try
             {
+                string objectURN;
                 if (!isNewDatabase && !string.IsNullOrEmpty(databaseName))
                 {
                     connectionInfo.ConnectionDetails.DatabaseName = databaseName;
+                    objectURN = string.Format(System.Globalization.CultureInfo.InvariantCulture, "Server/Database[Name='{0}']", databaseName);
```
Charles-Gagnon, line 659: Make sure to quote escape the name

Charles-Gagnon (approved):

PR #2247, opened 2023-09-21
Added new test framework (test)

smartguest:
Replaces the xunit framework with nunit (to fix a pipeline failure)


Charles-Gagnon (approved):
You should be able to remove the xunit package from the project as well, right?

https://github.com/microsoft/sqltoolsservice/blob/main/test/Microsoft.SqlTools.ServiceLayer.PerfTests/Microsoft.SqlTools.ServiceLayer.PerfTests.csproj#L24

PR #2248, opened 2023-09-22
Adding missing package in perf csproj

aasimkhan30:
@smartguest , I am adding the missing package in perf project. This is causing some issues when I try to dotnet build.


smartguest:
@aasimkhan30 I'm reverting the change as we speak, apparently the test is too ingrained with xunit so that removing it causes more issues

PR #2249, opened 2023-09-22
Revert change to nunit and restore xunit

smartguest:
Addresses recent issues with build, also attempts to fix the underlying "fail" issue using the way that was suggested in the original failure


aasimkhan30 (approved):

smartguest:
awaiting for tests to complete to ensure the fail status works or not

Charles-Gagnon:
We should be preferring nunit over xunit (better asserting and aligns with what the rest of our tests use). Was it not possible to fix the build issues instead of going back to xunit?

PR #2252, opened 2023-09-26
Fixing connections for stateless object explorer to work with fabric tokens

aasimkhan30:
Currently the smo server connection do not 'Connect' if the underlying sql connection object is not 'Open'. This just makes sure that I am always opening the underlying sql connection before performing any actions that require.

The error I was seeing was.

```
Microsoft.SqlServer.Management.Common.ConnectionFailureException
  HResult=0x80131501
  Message=Failed to connect to server tcp:renzo-srv-e875f3f6-7cd7-44ed-bd53-1c2f22c11cc9-c86ae94119db.sqltest-eg1.mscds.com,1433.
  Source=Microsoft.SqlServer.ConnectionInfo
  StackTrace:
   at Microsoft.SqlServer.Management.Common.ConnectionManager.Connect()
   at Microsoft.SqlTools.SqlCore.ObjectExplorer.SmoModel.SmoWrapper.OpenConnection(SmoObjectBase smoObj) in /_/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/SmoModel/SmoWrapper.cs:line 52

  This exception was originally thrown at this call stack:
    [External Code]

Inner Exception 1:
SqlException: Login failed for user ''.
```

To summarize:

code that errored out was this.
```
using (SqlConnection conn = new SqlConnection(connectionString))
            {
                ServerConnection connection;
                if (accessToken != null)
                {
                    connection = new ServerConnection(conn, accessToken as IRenewableToken);
                }
                 connection.Connect();
            }
```

The new code that works is 

```
using (SqlConnection conn = new SqlConnection(connectionString))
            {
                conn.AccessToken = accessToken?.Token; // setting the access token to sql connection.
                conn.Open(); // openning the connection before creating the smo server connection object. 
                ServerConnection connection = new ServerConnection(conn);
                connection.Connect();
            }
```

### DIFF
--- a/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
+++ b/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
@@ -41,7 +41,6 @@ namespace Microsoft.SqlTools.SqlCore.ObjectExplorer
                 conn.AccessToken = accessToken?.Token;
                 conn.Open();
                 ServerConnection connection = new ServerConnection(conn);
-                connection.Connect();
                 connection.AccessToken = accessToken as IRenewableToken;
                 return await Expand(connection, accessToken, nodePath, serverInfo, options, filters);
             }
@@ -130,7 +129,6 @@ namespace Microsoft.SqlTools.SqlCore.ObjectExplorer
                                     var underlyingSqlConnection = nodeContext.Server.ConnectionContext.SqlConnectionObject;
                                     underlyingSqlConnection.AccessToken = securityToken.Token;
                                     await underlyingSqlConnection.OpenAsync();
-                                    nodeContext.Server.ConnectionContext.Connect();
                                 }
 
                                 return node.Expand(taskCancellationTokenSource.Token, securityToken?.Token, filters);
--- a/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
+++ b/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
```diff
@@ -37,78 +38,49 @@ public static async Task<TreeNode[]> Expand(string connectionString, SecurityTok
...
-                {
-                    if (connection.IsOpen)
-                    {
-                        connection.Disconnect();
-                    }
-                }
+                conn.AccessToken = accessToken?.Token;
+                conn.Open();
+                ServerConnection connection = new ServerConnection(conn);
+                connection.Connect();
```
shueybubbles, line 44: >   connection.Connect(); [](http://example.com/codeflow?start=13&length=24)

There are 2 changes here, really. 
1. You are setting the token explicitly on the SqlConnection instead of using IRenewableToken
2. You call SqlConnection.Connect explicitly instead of relying on ServerConnection to call it.

What happens if you just make change 1? Did this bug manifest only in AAD scenarios and not in SQL auth?
The behavior of ServerConnection re: the connection state is supposed to be auth-type-independent.

90% of SMO application code never calls `SqlConnection.Open` explicitly and relies on ServerConnection to call it as needed.

I suspect either your IRenewableToken implementation has a bug, or ServerConnection has a regression in the IRenewableToken code path. SSMS stopped using IRenewableToken a while back so the latter scenario is certainly a possibility.

shueybubbles: Also - have you considered implementing a custom interactive authentication provider to provide tokens directly to SqlClient instead of relying on IRenewableToken?

shueybubbles: whatever the issue is in ServerConnection - please go to the internal ADO repo for it and write a test for the expected behavior and see if it's a regression that should be fixed.

cheenamalhotra: Have you tried enabling 'EnableSqlAuthenticationProvider' on target database?

aasimkhan30: @cheenamalhotra , I am just using sql core package for fabric which does not contain Connection service yet. Therefore I can not use 'EnableSqlAuthenticationProvider'

kburtram (approved):
Please follow up with David regarding the ServerConnection issue we're hitting.  But this looks fine to workaround the issue in the meantime.

kburtram:
FYI, the test failure looks related to this change.  Please take a look prior to merging.

src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
cheenamalhotra: You could still register auth provider on SqlConnection (required only once globally) and consume it from Authentication project - but I'll leave it up-to you to decide.

PR #2253, opened 2023-09-28
Updating SQL Projects nuget dependency

Benjin:
Updating to `0.2.2-preview`, and changing references to DSP to use the new enum-based API

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.Data.Tools.Schema.SchemaModel;
 using Microsoft.SqlServer.Dac.Projects;
 using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.Hosting;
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
```diff
@@ -117,7 +118,14 @@ internal async Task HandleCreateSqlProjectRequest(Contracts.CreateSqlProjectPara
         {
             await RunWithErrorHandling(async () =>
             {
-                await SqlProject.CreateProjectAsync(requestParams.ProjectUri, new SqlServer.Dac.Projects.CreateSqlProjectParams() { ProjectType = requestParams.SqlProjectType, DspVersion = requestParams.DatabaseSchemaProvider });
+                SqlServer.Dac.Projects.CreateSqlProjectParams createParams = new()
+                {
+                    ProjectType = requestParams.SqlProjectType,
+                    TargetPlatform = requestParams.DatabaseSchemaProvider == null ? null : DspToSqlPlatform(requestParams.DatabaseSchemaProvider),
+                    BuildSdkVersion = requestParams.BuildSdkVersion
```
Benjin, line 125: Found that BuildSdkVersion wasn't getting carried over.  Whoops.

kisantia: oh wow good catch

kisantia (approved):

Benjin:
One more change in-flight for SqlProjects that'll let me remove the DSP-SqlPlatforms conversion code from here; gonna hold this for `0.2.2`

PR #2258, opened 2023-10-01
Enable Premium SSD V2 for SQL VM SKU recommendation

Ramudaykumar:
The SKU Recommendations include premium v2 disks for SQL VM.

### DIFF
--- a/test/Microsoft.SqlTools.Migration.IntegrationTests/Migration/MigrationServiceTests.cs
+++ b/test/Microsoft.SqlTools.Migration.IntegrationTests/Migration/MigrationServiceTests.cs
@@ -58,7 +58,8 @@ namespace Microsoft.SqlTools.Migration.IntegrationTests.Migration
                 EndTime = DateTime.UtcNow.ToString("yyyy-MM-dd HH:mm:ss"),
                 PerfQueryIntervalInSec = 30,
                 ScalingFactor = 1,
-                DatabaseAllowList = new List<string> { "test", "test1" }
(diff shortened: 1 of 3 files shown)


test/Microsoft.SqlTools.Migration.IntegrationTests/Migration/MigrationServiceTests.cs
```diff
@@ -58,7 +58,8 @@ public async Task TestHandleMigrationGetSkuRecommendationsRequest()
                 EndTime = DateTime.UtcNow.ToString("yyyy-MM-dd HH:mm:ss"),
                 PerfQueryIntervalInSec = 30,
                 ScalingFactor = 1,
-                DatabaseAllowList = new List<string> { "test", "test1" }
+                DatabaseAllowList = new List<string> { "test", "test1" },
+                IsPremiumSSDV2Enabled = true,
```
aasimkhan30, line 62: You might want to look into adding tests for when it is false.

aasimkhan30 (approved):

test/Microsoft.SqlTools.Migration.IntegrationTests/Migration/MigrationServiceTests.cs
aasimkhan30: Never mind. The test is ignored.

Ramudaykumar: False is just to make it backward compatible, which is any way ignored. :)

PR #2262, opened 2023-10-03
Only opening connections when access token is provided

aasimkhan30:
Doing this when access token (like sql auth) is not provided leads to connection errors.

### DIFF
--- a/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
+++ b/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
@@ -133,6 +133,7 @@ namespace Microsoft.SqlTools.SqlCore.ObjectExplorer
                                 if (!nodeContext.Server.ConnectionContext.IsOpen && securityToken != null)
                                 {
                                     var underlyingSqlConnection = nodeContext.Server.ConnectionContext.SqlConnectionObject;
+                                    underlyingSqlConnection.AccessToken = securityToken.Token;
                                     await underlyingSqlConnection.OpenAsync();
                                 }
 
--- a/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
+++ b/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
@@ -40,7 +40,7 @@ namespace Microsoft.SqlTools.SqlCore.ObjectExplorer
             {
                 // In case of MWC access token based connections, we need to open the sql connection first before creating  the server connection
                 // Not doing so will result in an exception when trying to call 'Connect' on the server connection
-                if(accessToken != null)
+                if (accessToken != null)
                 {
                     conn.AccessToken = accessToken.Token;
                     await conn.OpenAsync();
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
```diff
@@ -38,10 +38,16 @@ public static async Task<TreeNode[]> Expand(string connectionString, SecurityTok
         {
             using (SqlConnection conn = new SqlConnection(connectionString))
             {
-                conn.AccessToken = accessToken?.Token;
-                conn.Open();
+                if(accessToken != null)
+                {
+                    conn.AccessToken = accessToken.Token;
+                    await conn.OpenAsync();
```
Charles-Gagnon, line 44: Shouldn't you still be trying to open even if there isn't an access token? Ideally this code should be generic enough that it can be used for other scenarios - such as ones where there's a user id/password.

aasimkhan30: Ideally yes, but unfortunately opening the sql auth based connections leads to connection exception in this line.
https://github.com/microsoft/sqltoolsservice/blob/73c2a75fbaa903ea50f135f6f99fbf24e07a3523/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/SmoModel/DatabaseTreeNode.cs#L23
 I haven't investigated smo to see why that's happening but I am making this fix to unblock people on mwc

Charles-Gagnon: The number of SMO connection issues you're hitting is definitely concerning...

But regardless - so these failures would only be something that people hit in development? Since we won't have sql login connections once this goes live. Just want to make sure, although I guess this isn't really blocking. 

But please at least add a comment here explaining why it's only being opened when there's a token. (and ideally add context on why you're even opening it here in the first place too)

src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
```diff
@@ -127,7 +133,6 @@ public static async Task<TreeNode[]> ExpandTreeNode(TreeNode node, ObjectExplore
                                 if (!nodeContext.Server.ConnectionContext.IsOpen && securityToken != null)
                                 {
                                     var underlyingSqlConnection = nodeContext.Server.ConnectionContext.SqlConnectionObject;
-                                    underlyingSqlConnection.AccessToken = securityToken.Token;
```
Charles-Gagnon, line 130: Why is this being removed?

src/Microsoft.SqlTools.SqlCore/ObjectExplorer/StatelessObjectExplorer.cs
```diff
@@ -38,10 +38,18 @@ public static async Task<TreeNode[]> Expand(string connectionString, SecurityTok
         {
             using (SqlConnection conn = new SqlConnection(connectionString))
             {
-                conn.AccessToken = accessToken?.Token;
-                conn.Open();
+                // In case of MWC access token based connections, we need to open the sql connection first before creating  the server connection
```
Charles-Gagnon, line 41: Is this actually related to MWC? I'd just remove that to avoid confusion for people who don't know what MWC is.

Charles-Gagnon (approved):

PR #2263, opened 2023-10-04
Use original file paths when getting associated DB files

corivera:
We were doing some path Joins that were replacing the target machine's file path with local path separators, causing Linux paths to have windows file separators.


src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/CommonUtilities.cs
```diff
@@ -376,10 +375,7 @@ public static string[] GetAssociatedFilePaths(ServerConnection connection, strin
 
             foreach (DataRow currentRow in dataTable.Rows)
             {
-                var primaryFolder = Path.GetDirectoryName(primaryFilePath);
```
corivera, line 379: This previous code was a bit of SSMS logic that didn't quite translate without all the extra SSMS classes and checks.

Charles-Gagnon (approved):

PR #2264, opened 2023-10-06
Improve how server OS and Platform are displayed

corivera:
I imported some of the SSMS logic for how to display the OS and Platform so that it includes release info, and also handles the case where a down-level instance won't have the HostPlatform and HostDistribution properties through SMO. Only servers that are version 14 and above will have those, but we should still show OS information without them. So I moved the OS and Platform properties in ServerPrototype140 back into the base prototype.

Also fixed an issue where we were still querying isPolybaseEnabled on down-level instances that don't support it, as well as improving how to detect when to show isXTPSupported.


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Security/NetStandardUtils.cs
```diff
@@ -111,18 +111,11 @@ public static bool IsSql13SP1OrLater(Version version)
...
             }
-            bool isXTPSupported = false;
-
-            if (server.ConnectionContext.ExecuteScalar("SELECT SERVERPROPERTY('IsXTPSupported')") != DBNull.Value)
-            {
-                isXTPSupported = server.IsXTPSupported;
-            }
-
-            return isXTPSupported;
+            return server.IsSupportedProperty(nameof(server.IsXTPSupported)) && server.IsXTPSupported;
```
corivera, line 118: This part was directly ported over from the current SSMS code.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Server/ServerData.cs
```diff
@@ -1212,19 +1211,25 @@ private void LoadGeneralProperties()
...
+                this.platform = server.Platform;
+                this.operatingSystem = "Microsoft Windows NT " +
+                        Convert.ToString(server.OSVersion, System.Globalization.CultureInfo.InvariantCulture);
             }
-            if (server.VersionMajor >= 14)
+            else
             {
-                this.operatingSystem = server.HostDistribution;
                 this.platform = server.HostPlatform;
+                this.operatingSystem = string.Format("{0} ({1})", server.HostDistribution, server.HostRelease);
```
Charles-Gagnon, line 1223: Interpolated strings are great 😄 

```suggestion
                this.operatingSystem =$"{server.HostDistribution} ({server.HostRelease})";
```

Charles-Gagnon (approved):

barbaravaldez (approved):

PR #2267, opened 2023-10-12
adding server level filestream access property

ssreerama:
This PR adds the property of filestream access level to the database view info which helps to enable disable the filestream filegroup table functionality

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
@@ -306,11 +306,11 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                                         // ServerFilestreamAccessLevel uses xp_instance_regread so will throw if that isn't available. In that case we just default to Disabled to ensure the UI reflects that
                                         databaseViewInfo.ServerFilestreamAccessLevel = dataContainer.Server.FilestreamLevel;
                                     }
-                                    catch (PropertyCannotBeRetrievedException ex) {
+                                    catch (PropertyCannotBeRetrievedException ex)
+                                    {
                                         // Cannot retrieve FilestreamLevel, defaulting to disabled level
-                                        // Error: {"Property FilestreamLevel is not available for Server '[SampleServer]'. This property may not exist for this object, or may not be retrievable due to insufficient access rights."}
                                         databaseViewInfo.ServerFilestreamAccessLevel = FileStreamEffectiveLevel.Disabled;
-                                        Logger.Error(ex.Message);
+                                        Logger.Error($"Cannot retrieve FilestreamLevel, defaulting to disabled level. Error: '{ex.Message}'");
                                     }
                                 }
                             }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
@@ -301,6 +301,12 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                                     // Files tab is only supported in SQL Server, but files exists for all servers and used in detach database, cannot depend on files property to check the supportability
                                     ((DatabaseInfo)databaseViewInfo.ObjectInfo).IsFilesTabSupported = true;
                                     ((DatabaseInfo)databaseViewInfo.ObjectInfo).Filegroups = GetFileGroups(smoDatabase, databaseViewInfo);
+                                    try
+                                    {
+                                        // cannot retrieve ServerFilestreamAccessLevel property for localdb instance
+                                        databaseViewInfo.ServerFilestreamAccessLevel = (int)dataContainer.Server.FilestreamLevel;
+                                    }
+                                    catch (PropertyCannotBeRetrievedException) { }
                                 }
 
                                 if (prototype is DatabasePrototype160)
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
@@ -308,7 +308,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                                     }
                                     catch (PropertyCannotBeRetrievedException ex) {
                                         // Cannot retrieve FilestreamLevel, defaulting to disabled level
-                                        // Error: {"Property FilestreamLevel is not available for Server '[SqmpleServer]'. This property may not exist for this object, or may not be retrievable due to insufficient access rights."}
+                                        // Error: {"Property FilestreamLevel is not available for Server '[SampleServer]'. This property may not exist for this object, or may not be retrievable due to insufficient access rights."}
                                         databaseViewInfo.ServerFilestreamAccessLevel = FileStreamEffectiveLevel.Disabled;
                                         Logger.Error(ex.Message);
                                     }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
@@ -307,7 +307,8 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                                         databaseViewInfo.ServerFilestreamAccessLevel = dataContainer.Server.FilestreamLevel;
                                     }
                                     catch (PropertyCannotBeRetrievedException ex) {
-                                        // Setting filestream access level to disabled
+                                        // Cannot retrieve FilestreamLevel, defaulting to disabled level
+                                        // Error: {"Property FilestreamLevel is not available for Server '[SqmpleServer]'. This property may not exist for this object, or may not be retrievable due to insufficient access rights."}
                                         databaseViewInfo.ServerFilestreamAccessLevel = FileStreamEffectiveLevel.Disabled;
                                         Logger.Error(ex.Message);
                                     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -301,6 +301,12 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
                                     // Files tab is only supported in SQL Server, but files exists for all servers and used in detach database, cannot depend on files property to check the supportability
                                     ((DatabaseInfo)databaseViewInfo.ObjectInfo).IsFilesTabSupported = true;
                                     ((DatabaseInfo)databaseViewInfo.ObjectInfo).Filegroups = GetFileGroups(smoDatabase, databaseViewInfo);
+                                    try
+                                    {
+                                        // cannot retrieve ServerFilestreamAccessLevel property for localdb instance
```
Charles-Gagnon, line 306: Do you have a link to docs that can help explain this? And did you chat with @shueybubbles about whether this is possibly something that could be fixed in SMO instead?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -301,6 +301,12 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
+                                        databaseViewInfo.ServerFilestreamAccessLevel = (int)dataContainer.Server.FilestreamLevel;
+                                    }
+                                    catch (PropertyCannotBeRetrievedException) { }
```
Charles-Gagnon, line 309: Would be good to log the exception regardless for debug purposes

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
shueybubbles: filestreamlevel comes from the registry

[https://github.com/microsoft/sqlmanagementobjects/blob/d6f3e68fe6fae09d22355a160baa525d8032fd64/src/Microsoft/SqlServer/Management/SqlEnum/xml/inc_server.xml#L54C2-L54C2](https://github.com/microsoft/sqlmanagementobjects/blob/d6f3e68fe6fae09d22355a160baa525d8032fd64/src/Microsoft/SqlServer/Management/SqlEnum/xml/inc_server.xml#L54C2-L54C2)

is the comment referring specifically to the [localdb](https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/sql-server-express-localdb?view=sql-server-ver16) product? If so, it's probably a bad sign that the docs for localdb expressly call it out as "can't be managed remotely via SQL Server Management Studio."
I suspect it is missing lots of functionality expected by SMO.
Are you trying to support localdb with ADS?

Charles-Gagnon: @erinstellato-ms I've not heard of us explicitly targeting localdb with ADS - is that something we want to work to try and support?

erinstellato-ms: We don't need to support localdb unless @shueybubbles knows something I don't!

ssreerama: OH! I thought we support localDB too, when i tested LocalDB for database properties, noticed this property cannot be retrieved, so added try/catch block to bypass the issue.

ssreerama: I'm still confused to take an action on this one, please help me if I'm missing anything.

Before the change, I can open db properties on localhost localdb instance. with this change it is failing to load the properties and I thought it would be great to just bypass this parameter and not to affect the previous behavior. If we cannot use the logic for this scenario, then we will lose the working behavior.

So need your suggestions if I can leave the changes or remove the try/catch block and let ADS take care of failing the dialog to open. :)

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
ssreerama: Added the exception message to the logger

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
erinstellato-ms: @ssreerama I'm not sure if this answers your question, but localdb is a version of SQLExpress with a wrapper that has a Windows service (so it's dependent on a Windows services to activate).  Visual Studio integrates with it in a couple ways, and it can be used with SSDT offline development.  ADS is cross platform.  If we are interested in supporting a lightweight local database option in ADS, completing our SQLLite data connector (used for testing) would be valuable to users, but I suspect (?) that's out of scope here.  However (yes, I'm going to complicate things), I've seen requests for supporting SQLLite in the NPS verbatims.  Do we want to go down that road?

@Charles-Gagnon  and @shueybubbles I'm going to tag you here for engineering help because I'm about out of my depth on how to advise Sai.

Also: I don't know if the work has been done by the appropriate team (which is not us) to ship localdb 2022.

Charles-Gagnon: Hm, for a change this I guess we might as well fix it since it seems straightforward enough. But in general we should probably be careful not to spend too much time specifically trying to fix localdb issues until we decide to officially support it.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -301,6 +301,14 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
                                     // Files tab is only supported in SQL Server, but files exists for all servers and used in detach database, cannot depend on files property to check the supportability
                                     ((DatabaseInfo)databaseViewInfo.ObjectInfo).IsFilesTabSupported = true;
                                     ((DatabaseInfo)databaseViewInfo.ObjectInfo).Filegroups = GetFileGroups(smoDatabase, databaseViewInfo);
+                                    try
+                                    {
+                                        // Cannot retrieve ServerFilestreamAccessLevel property for localdb instance as filestream level comes from the registry
```
Charles-Gagnon, line 306: So if I'm understanding this correctly - what's happening here is that you're updating the dialog to specifically check if filestream is supported for enabling/disabling the table. But SMO uses the registry to populate the ServerFilestreamAccessLevel property - which means it will fail for anything that doesn't have access to that (which would include other platforms, including docker, I would expect).

So the try/catch here is just making it so we don't fail to open the dialog in those scenarios.

If the above is accurate then I have two suggestions: 

1. Update the comment to clearly indicate that - `ServerFilestreamAccessLevel uses the registry so will throw if that isn't available. In that case we just default to Disabled to ensure the UI reflects that`
2. Specifically set the property to disabled on a catch (but still log the error, although I would log it as an info and include a comment that this may be expected). Leaving it as undefined isn't the worst either - but I see no reason not to specifically just disable it in this scenario since we know it isn't available

@erinstellato-ms @shueybubbles This seem reasonable to you?

erinstellato-ms: @Charles-Gagnon making sure I understand...  there are scenarios where we will not be able to definitively confirm if the filestream option should be enabled or not, so in the case where we cannot, we will disable it, but include an info message.  Correct?  If so, then yes, seems reasonable.

shueybubbles: I didn't say "SMO uses the registry". SMO uses TSQL to invoke xp_reg_read on the server. Apparently `localdb` doesn't like xp_reg_read.
The question is whether ADS is trying to support localdb or not.

ssreerama: @Charles-Gagnon , you are completely right. I can update comment and also do the #2. Thanks.

waiting for @shueybubbles and @erinstellato-ms comment on this one!!

Charles-Gagnon: @shueybubbles Good point - we can update the text to be `ServerFilestreamAccessLevel uses xp_instance_regread so will throw if that isn't available. In that case we just default to Disabled to ensure the UI reflects that`

@erinstellato-ms Yup. Although I haven't investigated why localdb would throw in this case, since it should be able to access the registry...might just be an intended limitation of localdb itself.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -301,6 +301,16 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
                                     ((DatabaseInfo)databaseViewInfo.ObjectInfo).Filegroups = GetFileGroups(smoDatabase, databaseViewInfo);
+                                    try
+                                    {
+                                        // ServerFilestreamAccessLevel uses xp_instance_regread so will throw if that isn't available. In that case we just default to Disabled to ensure the UI reflects that
+                                        databaseViewInfo.ServerFilestreamAccessLevel = dataContainer.Server.FilestreamLevel;
+                                    }
+                                    catch (PropertyCannotBeRetrievedException ex) {
+                                        // Setting filestream access level to disabled
+                                        databaseViewInfo.ServerFilestreamAccessLevel = FileStreamEffectiveLevel.Disabled;
+                                        Logger.Error(ex.Message);
```
Charles-Gagnon, line 312: Usually it's nice to leave context about the error - for example "Cannot retrieve Server FilestreamLevel, defaulting to Disabled. Error : {ex.Message}"

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
Charles-Gagnon: @ssreerama Sorry - I meant in the actual log message itself. The log contains a LOT of information (especially with sqlclient logs added) - and so adding context is super important there so it isn't just some random error message in the middle of a bunch of other stuff.

PR #2268, opened 2023-10-13
Fix for properties dialog not working when database name has special characters.

ssreerama:
This change adds fixes to the database properties that are failing to open the properties dialog when the database name has special characters.
ADS issue : https://github.com/microsoft/azuredatastudio/issues/24569 
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
@@ -12,6 +12,7 @@ using System.Globalization;
 using System.IO;
 using System.Security;
 using System.Xml;
+using System.Xml.Linq;
 using Microsoft.SqlServer.Management.Common;
 using Microsoft.SqlServer.Management.Sdk.Sfc;
 using Microsoft.SqlServer.Management.Smo;
@@ -1204,6 +1205,8 @@ namespace Microsoft.SqlTools.ServiceLayer.Management
             }
             else
             {
+                var databaseNode = new XElement("database");
+                databaseNode.Value = connInfo.ConnectionDetails.DatabaseName;
                 xml =
                 string.Format(@"<?xml version=""1.0""?>
                 <formdescription><params>
@@ -1211,11 +1214,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Management
                 <connectionmoniker>{0} (SQLServer, user = {1})</connectionmoniker>
                 <servertype>sql</servertype>
                 <urn>Server[@Name='{0}']</urn>
-                <database>{2}</database>                
+                {2}                
                 </params></formdescription> ",
                 connInfo.ConnectionDetails.ServerName.ToUpper(CultureInfo.InvariantCulture),
                 connInfo.ConnectionDetails.UserName,
-                connInfo.ConnectionDetails.DatabaseName);
+                databaseNode);
             }
             var xmlDoc = new XmlDocument();
             xmlDoc.LoadXml(xml);
--- a/src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
@@ -1199,11 +1199,11 @@ namespace Microsoft.SqlTools.ServiceLayer.Management
                 {0}
                 <connectionmoniker>{1} (SQLServer, user = {2})</connectionmoniker>
                 <servertype>sql</servertype>
-                <urn>Server[@Name=&apos;{1}&apos;]</urn>
+                <urn>Server[@Name='{1}']</urn>
                 <itemtype>Database</itemtype>                
                 </params></formdescription> ",
                 serverNameNode,
-                serverNameNode.Value,
+                Urn.EscapeString(serverNameNode.Value),
                 connInfo.ConnectionDetails.UserName);   
             }
             else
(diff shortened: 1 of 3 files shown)


Charles-Gagnon:
How do the changes in this PR relate to the issue described?

ssreerama:
Oh No! I didn't notice, used a wrong branch...

corivera (approved):
Can you also update the PR title to reflect these changes?

src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
```diff
@@ -1204,18 +1205,20 @@ private static XmlDocument CreateDataContainerDocument(ConnectionInfo connInfo,
...
                 <connectionmoniker>{0} (SQLServer, user = {1})</connectionmoniker>
                 <servertype>sql</servertype>
                 <urn>Server[@Name='{0}']</urn>
-                <database>{2}</database>                
+                {2}                
                 </params></formdescription> ",
                 connInfo.ConnectionDetails.ServerName.ToUpper(CultureInfo.InvariantCulture),
                 connInfo.ConnectionDetails.UserName,
-                connInfo.ConnectionDetails.DatabaseName);
+                databaseNode);
```
Charles-Gagnon, line 1221: Shouldn't we be doing the same for the other nodes that have custom values? While I believe that currently we don't allow special characters like > in them that isn't something we can guarantee forever so seems reasonable to be defensive here.

We should also be escaping single quotes in the Server name passed to the urn node (again, just to be safe)

src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
```diff
@@ -1187,35 +1188,41 @@ internal static System.Security.SecureString BuildSecureStringFromPassword(strin
...
                 xml =
                 string.Format(@"<?xml version=""1.0""?>
                 <formdescription><params>
-                <servername>{0}</servername>
-                <connectionmoniker>{0} (SQLServer, user = {1})</connectionmoniker>
+                {0}
+                <connectionmoniker>{1} (SQLServer, user = {2})</connectionmoniker>
                 <servertype>sql</servertype>
-                <urn>Server[@Name='{0}']</urn>
+                <urn>Server[@Name=&apos;{1}&apos;]</urn>
```
Charles-Gagnon, line 1202: I meant that you need to escape single quotes in the server name field (using `Urn.EscapeString`) - the single quotes here are fine.

corivera (approved):

src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
```diff
@@ -1187,35 +1188,41 @@ internal static System.Security.SecureString BuildSecureStringFromPassword(strin
...
 
             if (!databaseExists)
             {
                 xml =
                 string.Format(@"<?xml version=""1.0""?>
                 <formdescription><params>
-                <servername>{0}</servername>
-                <connectionmoniker>{0} (SQLServer, user = {1})</connectionmoniker>
+                {0}
+                <connectionmoniker>{1} (SQLServer, user = {2})</connectionmoniker>
```
Charles-Gagnon, line 1200: You shouldn't use the escaped version of the name here - we want the full version here (since it's just a string)

src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
```diff
@@ -1187,35 +1188,41 @@ internal static System.Security.SecureString BuildSecureStringFromPassword(strin
...
                 <servertype>sql</servertype>
-                <urn>Server[@Name='{0}']</urn>
+                <urn>Server[@Name='{1}']</urn>
```
Charles-Gagnon, line 1202: Were you going to create an XElement for the urn & connectionmoniker nodes? I think that would be safer

src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
ssreerama: I don't see the place where we pass server name to the urn for ObjectUrn like database name, except here, revered the changes and added individual XElement nodes to the other xml elements. Thanks

src/Microsoft.SqlTools.ServiceLayer/Management/Common/DataContainer.cs
```diff
@@ -1187,35 +1188,47 @@ internal static System.Security.SecureString BuildSecureStringFromPassword(strin
...
         {
             string xml = string.Empty;
+            var serverNameNode = new XElement("servername");
+            serverNameNode.Value = connInfo.ConnectionDetails.ServerName.ToUpper(CultureInfo.InvariantCulture);
+
+            var connectionMonikerNode = new XElement("connectionmoniker");
+            connectionMonikerNode.Value = string.Format("{0} (SQLServer, user = {1})", serverNameNode.Value, connInfo.ConnectionDetails.UserName);
+
+            var urnNode = new XElement("urn");
+            urnNode.Value = string.Format("Server[@Name = '{0}']", serverNameNode.Value);
```
Charles-Gagnon, line 1198: You still need to URN escape the string here

ssreerama: is this what you are expecting here?

Urn.EscapeString(serverNameNode.Value)

PR #2269, opened 2023-10-13
SPID branch for STS

smartguest:
Adds STS to connection complete response. 

Part of fix to ADS SPID display


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -578,6 +578,8 @@ private ConnectionCompleteParams GetConnectionCompleteParams(string connectionTy
                     ? reliableConnection.GetUnderlyingConnection()
                     : connection;
 
+                response.SPID = (underlyingConnection as SqlConnection).ServerProcessId;
```
Charles-Gagnon, line 581: @cheenamalhotra Does the SPID still change when silent reconnections happen (which if I remember correctly was something that wasn't uncommon for Azure DBs)? Or is this guaranteed to stay the same until the user disconnects?

@Matteo-T @shueybubbles It looks like [SSMS refreshes the SPID on every query execution](https://msdata.visualstudio.com/SQLToolsAndLibraries/_git/SSMS?path=/Sql/ssms/core/SqlEditors/src/SqlScriptEditorControl.cs&version=GBmain&line=2909&lineEnd=2910&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) - is the above why or is there anything else that you've run in to with ensuring this is kept up to date?

cheenamalhotra: That is true, SPID is subject to change if driver reconnects internally on query execution and STS may not send any connection complete events.

So ideally, we should be sending/reading SPID from `query/complete` instead of `connection/complete` as SPID is reflective of active connection used by the query editor, which can change internally at driver level.

cheenamalhotra: Just tested, SSMS doesn't update it's title everytime query executes.. which means it can lead to mis-information:



If we decide to show this on titles, we need to be certain that query editor title will be refreshed on every query execution. But is it possible to do so without incurring UI/editor flickering on every query execution?

shueybubbles: we should probably remove the SPID from the title. The updated one is on the status bar. People would thank us for shortening the title text length!

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -489,6 +494,15 @@ private async Task ExecuteInternal()
                 // Call the query execution callback
                 if (QueryCompleted != null)
                 {
+                    if (sqlConn != null)
+                    {
+                        // Update SPID here as it may change upon restart and only updates upon query execution.
+                        SPID = (sqlConn.GetUnderlyingConnection() as SqlConnection).ServerProcessId;
+                        if (SPID != 0)
+                        {
+                            Console.WriteLine(SPID);
```
Charles-Gagnon, line 503: Console.WriteLine is going to break STS (stdout is used for the JSONRPC channel so this would be inserting random characters in that stream)

cheenamalhotra (approved):

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -489,6 +494,14 @@ private async Task ExecuteInternal()
                 // Call the query execution callback
                 if (QueryCompleted != null)
                 {
+                    if (sqlConn != null)
+                    {
+                        // Update PID here as it may change upon restart and only updates upon query execution.
```
Charles-Gagnon, line 499: restart? Did you mean reconnect?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/QueryCompleteEvent.cs
```diff
@@ -23,6 +23,11 @@ public class QueryCompleteParams
         /// Summaries of the result sets that were returned with the query
         /// </summary>
         public BatchSummary[] BatchSummaries { get; set; }
+
+        /// <summary>
+        /// Process ID assigned to the query editor.
+        /// </summary>
+        public string PID { get; set; }
```
Charles-Gagnon, line 30: This should be nullable too - we don't always set it (and other providers definitely won't necessarily be setting it so we need to be able to handle that)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -516,6 +529,10 @@ private async Task ExecuteInternal()
                 {
                     // Subscribe to database informational messages
                     sqlConn.GetUnderlyingConnection().InfoMessage -= OnInfoMessage;
+                    var intPID = (sqlConn.GetUnderlyingConnection() as SqlConnection).ServerProcessId;
```
Charles-Gagnon, line 532: Why are you duplicating the logic here if we're already doing it up above? It seems like this one should be removed - since I believe we need to set the PID before the QueryCompleted callback is done

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -489,6 +494,14 @@ private async Task ExecuteInternal()
                 // Call the query execution callback
                 if (QueryCompleted != null)
                 {
+                    if (sqlConn != null)
+                    {
+                        // Update PID here as it may change upon restart and only updates upon query execution.
+                        var intPID = (sqlConn.GetUnderlyingConnection() as SqlConnection).ServerProcessId;
```
Charles-Gagnon, line 500: We'll still possibly have a PID on a failed query, did you check to make sure the SPID is being updated in that case too?

alanrenmsft:
please add screenshots to demonstrate the feature for SQL Server and Postgres.

src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/ConnectionCompleteNotification.cs
```diff
@@ -64,6 +64,11 @@ public class ConnectionCompleteParams
         /// Additional optional message with details about why the version isn't supported.
         /// </summary>
         public string? UnsupportedVersionMessage { get; set; }
+
+        /// <summary>
+        /// Process ID for the connection, if the connection was successful.
+        /// </summary>
+        public string? ServerConnectionId { get; set; }
```
alanrenmsft, line 71: why do we have 3 different names for the same thing in STS? ServerProcessId (defined in MDS), ServerConnectionId and PID. 

If we defined it as PID in the protocol, I'd use it and get rid of the ServerConnectionId.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -577,6 +577,11 @@ private ConnectionCompleteParams GetConnectionCompleteParams(string connectionTy
                 DbConnection underlyingConnection = reliableConnection != null
                     ? reliableConnection.GetUnderlyingConnection()
                     : connection;
+                
+                var pid = (underlyingConnection as SqlConnection).ServerProcessId;
+                if (pid != 0) {
```
Charles-Gagnon, line 582: nit: A comment here explaining that 0 means the connection is inactive would be helpful for those that aren't aware

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/ExecuteRequests/QueryCompleteEvent.cs
```diff
@@ -23,6 +23,11 @@ public class QueryCompleteParams
         /// Summaries of the result sets that were returned with the query
         /// </summary>
         public BatchSummary[] BatchSummaries { get; set; }
+
+        /// <summary>
+        /// Process ID assigned to the query editor.
```
Charles-Gagnon, line 28: nit: 
```suggestion
        /// Process ID of the query that was executed
```

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -288,6 +288,11 @@ internal set
         /// </summary>
         public bool HasErrored { get; private set; }
 
+        /// <summary>
+        /// The Process ID of the query when connected.
+        /// </summary>
+        public string PID { get; private set; }
```
Charles-Gagnon, line 294: nit: Agree with Alan that this should be updated to be either ServerProcessId (to match MDS) or ServerConnectionId (to match protocol)

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Query.cs
```diff
@@ -489,6 +494,14 @@ private async Task ExecuteInternal()
                 // Call the query execution callback
                 if (QueryCompleted != null)
                 {
+                    if (sqlConn != null)
+                    {
+                        // Update PID here as it may change upon reconnect and only updates upon query execution.
+                        var intPID = (sqlConn.GetUnderlyingConnection() as SqlConnection).ServerProcessId;
+                        if (intPID != 0) {
```
Charles-Gagnon, line 501: nit: Same comment here about including comment about why 0 is skipped

Charles-Gagnon (approved):

Charles-Gagnon (approved):

PR #2270, opened 2023-10-18
Remove Hyphen from suggestion triggers

cheenamalhotra:
Addresses https://github.com/microsoft/azuredatastudio/issues/24716

Before:

before_hyphensuggestions

After:

after_hyphensuggestions


kisantia (approved):

Charles-Gagnon:
@kburtram @llali I see your names a lot with this early language service stuff, anything you can think of for why we had this be a trigger?

Charles-Gagnon (approved):
Looks good, but this list seems weird to me to begin with. Why are `:` `\` or `"` in there too?

cheenamalhotra:
I'll leave the others as-is for now unless they cause concerns :P 
But good to learn where they came from!

PR #2271, opened 2023-10-20
Scopes down context generation to a single DB in a connection.

lewis-sanchez:
Scopes down context generation to a single database instead of to an entire connection. This speeds up context generation for connections that have a high number of databases.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
@@ -17,7 +17,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Metadata
 {
     internal static class SmoScripterHelpers
     {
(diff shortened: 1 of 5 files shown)


kisantia (approved):

src/Microsoft.SqlTools.ServiceLayer/Metadata/SmoScripterHelpers.cs
```diff
@@ -172,12 +172,17 @@ private static IEnumerable<string> GenerateTableScripts(Server server)
...
         }
 
-        private static UrnCollection GetAllServerTableAndViewUrns(Server server)
+        private static UrnCollection GetDatabaseTableAndViewUrns(Server server, string databaseName)
         {
             UrnCollection urnCollection = new UrnCollection();
 
             foreach (Database db in server.Databases)
             {
+                if (db.Name != databaseName)
```
Charles-Gagnon, line 181: Instead of looping through all databases you can just access the DB directly through the indexer on Databases

PR #2273, opened 2023-10-23
Add CodeQL to the build pipeline

SakshiS-harma:


azure-pipelines/config/tsaoptions.json
```diff
@@ -0,0 +1,24 @@
+{
```
kisantia, line 1: nit: uppercase the "O" in "tsaOptions.json" for the file name

kisantia (approved):

azure-pipelines/config/tsaoptions.json
kisantia: oh but ignore this comment if that's the way it's supposed to be for the config to get picked up

SakshiS-harma: Yeah this is the configuration used in ADS too. I will keep it as-is for now, and if it works as expected, will update the name later :-).

cheenamalhotra (approved):

PR #2275, opened 2023-10-25
update DacFx version to 162.1.167

kisantia:
Updating DacFx version to the latest release. I did a quick test of the dacpac, schema compare, and sql projects extensions in ADS to confirm they are still working with this version.

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -28,7 +28,7 @@
 		<!-- TODO: Upgrade package and use Token Credential design -->
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
 		<PackageReference Update="Microsoft.SqlServer.Management.SmoMetadataProvider" Version="170.18.0" />
-		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="162.1.153-preview" />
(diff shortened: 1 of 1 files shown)


Packages.props
```diff
@@ -28,7 +28,7 @@
 		<!-- TODO: Upgrade package and use Token Credential design -->
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
 		<PackageReference Update="Microsoft.SqlServer.Management.SmoMetadataProvider" Version="170.18.0" />
-		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="162.1.153-preview" />
+		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="162.1.167" />
```
SakshiS-harma, line 31: Just curious why was the DacFx version in preview last time and not now?

SakshiS-harma (approved):

Packages.props
kisantia: The latest Dacfx nuget is the GA version of DacFx that was released last week, so that's why it isn't preview. Having this match the GA version will also help for investigating the ADS schema compare issues if we need to ask the users to compare the performance slowdown for the latest and past versions of sqlpackage, since there are only sqlpackage releases for the GA dacfx versions.

SakshiS-harma: Got it. Thanks!

PR #2276, opened 2023-10-27
Populate encryptors list for basic admin database objects

corivera:
These encryptors are the list of keys and certificates that are selected when setting the encryption option in the backup database dialog. I also made some changes to the backup service call to return the correct operation completion status and to improve error reporting.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
@@ -346,6 +346,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery
             catch (Exception e)
             {
                 await requestContext.SendError(e);
+                Logger.Error(e);
                 if (sqlConn != null)
                 {
                     sqlConn.Dispose();
--- a/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
@@ -309,6 +309,7 @@ namespace Microsoft.SqlTools.ServiceLayer.DisasterRecovery
         {
             BackupResponse response = new BackupResponse();
             ConnectionInfo connInfo;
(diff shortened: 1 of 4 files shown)


barbaravaldez (approved):

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -333,21 +334,23 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+                    response.Result = false;
                 }
 
+                await requestContext.SendResult(response);
             }
-            else
+            catch (Exception e)
             {
-                response.Result = false;
+                await requestContext.SendError(e);
```
Charles-Gagnon, line 348: Consider logging error here too

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
```diff
@@ -333,21 +334,23 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+                if (sqlConn != null)
+                {
+                    sqlConn.Dispose();
```
Charles-Gagnon, line 351: What about if the task starts up successfully but then an error occurs during execution? Are we guaranteed to have StatusChanged with IsCompleted = true called in that case?

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/DisasterRecovery/DisasterRecoveryService.cs
corivera: Looking at the SqlTask class, IsCompleted only gets set to false if the status is InProgress or NotStarted. On other results like Failed or Canceled it gets set to true.

corivera: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/TaskServices/SqlTask.cs#L355

PR #2281, opened 2023-11-02
Fix Webproxy to work with credentials

cheenamalhotra:
Addresses https://github.com/microsoft/azuredatastudio/issues/24854


cheenamalhotra:
Merging to prepare build for client testing

PR #2282, opened 2023-11-02
Fix for create database issue of managed instance.

ssreerama:
This PR adds fix for create database issue for Managed Instance.

ADS Issue: https://github.com/microsoft/azuredatastudio/issues/24559

CreateDB_MI

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
@@ -1767,15 +1767,10 @@ WHERE do.database_id = @DbID
             }
             else
             {
-                if (context.Server.DatabaseEngineType == DatabaseEngineType.SqlAzureDatabase)
+                if (context.Server.DatabaseEngineType == DatabaseEngineType.SqlAzureDatabase || context.Server.DatabaseEngineEdition == DatabaseEngineEdition.SqlManagedInstance)
                 {
-                    // Azure instances don't have a model database we can query, so just use the defaults
+                    // Azure instances and SQL Managed Instances don't have a model database we can query, so just use the defaults
                     this.originalState = new DatabaseData(context);
-                }
-                else if (context.Server.DatabaseEngineEdition == DatabaseEngineEdition.SqlManagedInstance)
-                {
-                    // SQl Managed Instance don't have a model database we can query, so just use the defaults to master
-                    this.originalState = new DatabaseData(context, "master");
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
```diff
@@ -1772,6 +1772,11 @@ public DatabasePrototype(CDataContainer context)
                     // Azure instances don't have a model database we can query, so just use the defaults
                     this.originalState = new DatabaseData(context);
                 }
+                else if (context.Server.DatabaseEngineEdition == DatabaseEngineEdition.SqlManagedInstance)
+                {
+                    // SQl Managed Instance don't have a model database we can query, so just use the defaults to master
```
Charles-Gagnon, line 1777:
```suggestion
                    // SQL Managed Instances don't have a model database we can query, so just default to master
```

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
```diff
@@ -1772,6 +1772,11 @@ public DatabasePrototype(CDataContainer context)
...
+                    this.originalState = new DatabaseData(context, "master");
```
Charles-Gagnon, line 1778: Is this what SSMS does?

@erinstellato-ms Does this seem like a reasonable approach? I can't remember if MI users will always be guaranteed to have access to master - do we want to consider either using the plain defaults like Azure does?

ssreerama: I'm not sure how they handled in SSMS, but only approach that works in our case is by using the master database. I can check if we can use the default like azure.

erinstellato-ms: @Charles-Gagnon @ssreerama  I'm checking, give me a few

corivera: Also do we need to specify master, or can we just do the defaults like the Azure if-case directly above this?

corivera (approved):

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
Charles-Gagnon: It makes sense to try and use "reasonable" defaults when we can, but my question is whether master is actually going to represent reasonable defaults. If it isn't then yeah, I think your suggestion Cory is what I'd lean towards.

Charles-Gagnon (approved):
Using the defaults seems fine for now, but we should still follow up on what the "best" approach is here. Is there some set of reasonable defaults we can use for MI?

src/Microsoft.SqlTools.ServiceLayer/Admin/Database/DatabasePrototype.cs
ssreerama: It worked with or without specifying master as default. I'll check with Erin to confirm if we have any MI specific defaults or this approach is the correct one. Merging this one for now. Thank you all

PR #2284, opened 2023-11-03
Moving Performance query code to Core project

Benjin:


Packages.props
```diff
@@ -35,7 +35,7 @@
 		<PackageReference Update="Microsoft.SqlServer.Assessment" Version="[1.1.17]" />
 		<PackageReference Update="Microsoft.SqlServer.Migration.Assessment" Version="1.0.20230907.105" />
 		<PackageReference Update="Microsoft.SqlServer.Migration.Logins" Version="1.0.20230407.56" />
-		<PackageReference Update="Microsoft.SqlServer.Management.QueryStoreModel" Version="163.26.1" />
+		<PackageReference Update="Microsoft.SqlServer.Management.QueryStoreModel" Version="163.55.1" />
```
Charles-Gagnon, line 38: Checked in nuget packages 😭

Benjin: Is an unlisted nuget package preferred?  I _suppose_ it could be listed publicly, but that implies that it's meant to be consumed by external devs.

Charles-Gagnon: *shrug* We push internal stuff to nuget, so it seems better for everyone to just bite the bullet and make it "official" (with a note about it being intended for internal use to avoid people complaining it doesn't work like they expect). If someone wants to try and use this then go for it I say.

(I understand this stuff comes with engineering overhead to set up the publishing pipelines and whatnot - but TBH that's well worth the cost)

build.json
```diff
@@ -46,7 +46,7 @@
     {
       "Name": "Microsoft.SqlTools.SqlCore",
       "Frameworks": [
-        "netstandard2.0"
+        "net6.0"
```
Charles-Gagnon, line 49: Are you updating the sqldb workload to net6 then? 

@aasimkhan30 FYI since you were asking about this as well

aasimkhan30: Unfortunately, there are blockers on the provisioning code that is preventing the migration to netcore

Benjin: I was just about to tag you about this; net6 can be supported over there since the DacFx workload is built on it, yeah?  Seems like the path we'd want anyway since the overall .NET advice is that anything that can be .NET rather than Standard or Framework should be.

aasimkhan30: I see. So, you plan on writing APIs in the dacfx workload for this feature, right? If that's the case using netcore is fine. @Charles-Gagnon FYI.

Charles-Gagnon (approved):

build.json
Charles-Gagnon: Resolved offline

Packages.props
Benjin: Discussed offline-but-technically-it-was-online-since-we-were-communicating-remotely-and-digitally-over-the-internet.

Not a blocking issue, and something I can pursue separately.

PR #2285, opened 2023-11-03
Hide permissions if a login doesn't support it

corivera:
Needed for https://github.com/microsoft/azuredatastudio/issues/24830

Hides the securables & permissions tables if it's not supported or we get an error

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Login/LoginHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Login/LoginHandler.cs
@@ -95,7 +95,7 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
             };
 
             // hide permissions for system logins or if the login doesn't support querying permissions
-            if (prototype.HidePermissions || (!dataContainer.IsNewObject && login.IsSystemObject))
+            if (prototype.HidePermissions || (!parameters.IsNewObject && login.IsSystemObject))
             {
                 loginInfo.SecurablePermissions = null;
             }
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Login/LoginHandler.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Login/LoginHandler.cs
@@ -82,9 +82,18 @@ namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement
                 ConnectPermission = prototype.WindowsGrantAccess,
                 IsEnabled = !prototype.IsDisabled,
                 IsLockedOut = prototype.IsLockedOut,
-                UserMapping = new ServerLoginDatabaseUserMapping[0],
-                SecurablePermissions = prototype.SecurablePermissions
+                UserMapping = new ServerLoginDatabaseUserMapping[0]
             };
+
+            // show permissions for new logins and existing non-system logins
+            if (!prototype.HidePermissions && (dataContainer.IsNewObject || !((dataContainer.Server.GetSmoObject(parameters.ObjectUrn) as Login).IsSystemObject)))
+            {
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Login/LoginHandler.cs
```diff
@@ -82,10 +82,19 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
                 ConnectPermission = prototype.WindowsGrantAccess,
                 IsEnabled = !prototype.IsDisabled,
                 IsLockedOut = prototype.IsLockedOut,
-                UserMapping = new ServerLoginDatabaseUserMapping[0],
-                SecurablePermissions = prototype.SecurablePermissions
+                UserMapping = new ServerLoginDatabaseUserMapping[0]
             };
 
+            // show permissions for new logins and existing non-system logins
+            if (!prototype.HidePermissions && (dataContainer.IsNewObject || !((dataContainer.Server.GetSmoObject(parameters.ObjectUrn) as Login).IsSystemObject)))
```
kburtram, line 89: it looks like a Login is being created in a similar condition on line 59 above, and cached in the state object in the prototype.  It may be cheaper to either cache the instance created above in this method, or to add an accessor to the one stored in the state in the prototype.  I'm assuming that GetSmoObject is running SQL queries in order to create the Login, if that's not the case than I don't think it matters as much.

kburtram (approved):
Looks fine, but you may be creating an extra SMO object that requires executing additional SQL statements during dialog initialization.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Login/LoginHandler.cs
corivera: Re-used the login object

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Login/LoginHandler.cs
```diff
@@ -82,10 +91,20 @@ public override Task<InitializeViewResult> InitializeObjectView(InitializeViewRe
...
                 IsLockedOut = prototype.IsLockedOut,
-                UserMapping = new ServerLoginDatabaseUserMapping[0],
-                SecurablePermissions = prototype.SecurablePermissions
+                UserMapping = new ServerLoginDatabaseUserMapping[0]
             };
 
+            // hide permissions for system logins or if the login doesn't support querying permissions
+            if (prototype.HidePermissions || (!dataContainer.IsNewObject && login.IsSystemObject))
+            {
+                loginInfo.HidePermissions = true;
```
Charles-Gagnon, line 100: Could you just have SecurablePermissions be nullable instead and then only show if it's defined on the extension side? I thought that's what we do for other similar situations.

Charles-Gagnon (approved):

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Security/LoginData.cs
```diff
@@ -2098,7 +2111,16 @@ public LoginPrototype(CDataContainer context, Login login)
...
-            this.securablePermissions = SecurableUtils.GetSecurablePermissions(this.exists, PrincipalType.Login, login, context);
+            try
+            {
+                this.securablePermissions = SecurableUtils.GetSecurablePermissions(this.exists, PrincipalType.Login, login, context);
+            }
+            catch (Exception ex)
+            {
+                Logger.Error($"Exception thrown while trying to get securables for login '{this.LoginName}'. Error message: '{ex.Message}'");
+                this.securablePermissions = new SecurablePermissions[0];
+                this.hidePermissions = true;
```
Charles-Gagnon, line 2122: Could you do the same here with the nullable? If this is just temporary code feel free to ignore though

PR #2294, opened 2023-11-15
Multi-targeting SqlCore

Benjin:
Adding back netstandard2.0 until that dependency can be dropped, and adding conditionals for QueryStore package reference and consumption

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -245,7 +245,7 @@ steps:
   displayName: 'ESRP SqlCore CodeSigning - sha256 only'
   inputs:
     ConnectedServiceName: 'Code Signing'
-    FolderPath: '$(Build.SourcesDirectory)/src/Microsoft.SqlTools.SqlCore/bin/$(buildConfiguration)/net6.0'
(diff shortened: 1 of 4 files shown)


Charles-Gagnon:
Please:

1. Make an issue (preferably with the other workload items) to track removing this once we've moved over to the SqlToolsCore workload
2. Add comment(s) linking to that work item - at least one wherever you IFDEF stuff out

aasimkhan30 (approved):

Benjin:
@Charles-Gagnon

> Please:
> 
> 1. Make an issue (preferably with the other workload items) to track removing this once we've moved over to the SqlToolsCore workload
> 2. Add comment(s) linking to that work item - at least one wherever you IFDEF stuff out

Done, tracking removal in SqlToolsCore as Task #1254787

Charles-Gagnon (approved):

Charles-Gagnon:
Can you make sure the test failures aren't related to this change?

Benjin:
@Charles-Gagnon 

> Can you make sure the test failures aren't related to this change?

Everything's passing locally except for the test that requires a storage account to be set up, which is failing because I don't have those reg keys populated, so it looks good.

PR #2298, opened 2023-11-27
Avoid event counters from being captured + update MDS

cheenamalhotra:
Updates MDS to latest version + fixes event source traces

When working with SqlClient traces, it's possible to capture System.Net.Http Event counters wherever Http clients are involved. This started happening recently, and is likely coming from .NET 7 as driver version hasn't updated. This causes lots of traces of EventPayload object type (internal to Event Counters implementation), leading to pollution of logs.

image

This PR skips capturing them, as they're not intended for logging.

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -65,7 +65,7 @@
 		packages\Microsoft.SqlTools.ManagedBatchParser\Microsoft.SqlTools.ManagedBatchParser.nuspec-->
 	<ItemGroup>
 		<PackageReference Update="Azure.Identity" Version="1.10.2" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.1" />
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.2" />
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="170.18.0" />
 		<PackageReference Update="Newtonsoft.Json" Version="13.0.3" />
 		<PackageReference Update="System.Configuration.ConfigurationManager" Version="7.0.0" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.Hosting/Utility/SqlClientEventListener.cs
```diff
@@ -36,22 +36,25 @@ protected override void OnEventSourceCreated(EventSource eventSource)
         /// <param name="eventData">The data for the event</param>
         protected override void OnEventWritten(EventWrittenEventArgs eventData)
         {
-            if (eventData.Payload == null)
+            // Skip EventCounters as they can come from any event source and will pollute traces captured.
+            if (!eventData.EventName.Equals(nameof(EventCounter)))
```
kisantia, line 40: nit: update the if condition checking for EventCounters to return early, like the `eventData.Payload == null` check below, for easier readability

kisantia (approved):

src/Microsoft.SqlTools.Hosting/Utility/SqlClientEventListener.cs
```diff
@@ -36,10 +36,12 @@ protected override void OnEventSourceCreated(EventSource eventSource)
...
         protected override void OnEventWritten(EventWrittenEventArgs eventData)
         {
-            if (eventData.Payload == null)
+            // Skip EventCounters as they can come from any event source and will pollute traces captured.
+            if (eventData.Payload == null || eventData.EventName.Equals(nameof(EventCounter)))
             {
                 return;
             }
+
             foreach (object payload in eventData.Payload)
```
Charles-Gagnon, line 45: Is the payload a string in the cases we want to log? If so could we just check that here and only log it if it's a string? That would save us from having to do this again if a different object type is added as a payload (although I suppose that maybe we want to leave this since the objects added might have a ToString implementation and actually be useful...)

Charles-Gagnon (approved):

src/Microsoft.SqlTools.Hosting/Utility/SqlClientEventListener.cs
cheenamalhotra: It's commonly a string for events like 'Trace', 'Error' and depends on what the driver writes using `EventSource.WriteEvent`. I believe we could always expect/request a `ToString()` implementation from driver if that happens in future if we see something like that again, so that should be fine. These counters came from .NET 7, and that too polluting logs largely without we enabling them.. seemed a bit strange.

PR #2301, opened 2023-12-05
Updated sql Nuget version and updated picking v2 perf queries

krritik:
Changes include -
1) Upgarding the version of sql migration assessment nuget (includes artihmetic overflow exception fix in perf queries)
2) Change in DataPointsCollector call. So that it picks the V2 (memory utilization perf query.

Tested - Tested locally the new nuget and changes with ADS Extension. Perf Data collection happening correctly with the changes.


src/Microsoft.SqlTools.Migration/SqlDataQueryController.cs
```diff
@@ -71,7 +71,10 @@ public class SqlDataQueryController : IDisposable
             this.messages = new List<KeyValuePair<string, DateTime>>();
             this.errors = new List<KeyValuePair<string, DateTime>>();
             perfDataCache = new SqlPerfDataPointsCache(this.outputFolder, _logger);
-            dataCollector = new DataPointsCollector(new string[] { connectionString }, _logger);
+            dataCollector = new DataPointsCollector(
```
Ramudaykumar, line 74: SqlDataQueryController is covered in test, and you are changing inside the constructor as hardcoded binary parameter. For now, it looks ok.

Ramudaykumar (approved):

PR #2302, opened 2023-12-07
Adding simple query-based object explorer 

aasimkhan30:
Adding simple query based object explorer for perf reasons in online scenarios.
The OE model is defined in src/Microsoft.SqlTools.SqlCore/ObjectExplorer2/ObjectExplorerModel.xml
The T4 parser is defined in ObjectExplorerModel.tt

TODO: Separate out tests and add more tests to make sure objects are constructed properly


Charles-Gagnon (commented):
Would be great to add a README or comments explaining why this exists and what differences it has from the "main" one. 

Also consider a better name too - like `SimpleObjectExplorerModel` and such to make it clearer how it's different (2 to me implies it's a "better" version, which doesn't really seem to be the case here)

kburtram (approved):

PR #2304, opened 2023-12-20
Fix some background task exceptions being swallowed

Charles-Gagnon:
Discovered a couple issues with the use of ContinueWith while doing some sqlparser stuff so fixing them up. @cheenamalhotra I took a quick look through other uses of ContinueWith and didn't see any other issues along these lines, but I'd still recommend going through yourself and making sure I didn't miss anything.

GetSignatureHelp wasn't checking if the task was faulted, so when it awaited on line 552 it threw the error, which got silently swallowed and no response was ever sent. I changed it to send an error response back so that the client knows there was an issue - which doesn't change behavior in ADS as far as I can tell but is a good practice so we can see it in the message logs too.

GetSemanticMarkers had two issues : 

1. Same as above with not handling faulted tasks, in this case though we just log it for debugging purposes since there isn't a result being waited for
image

3. Because this method was called within another background task, if any exception occurred while setting up it would silently be swallowed which could make debugging difficult. In addition, we'd then ignore all other files that were being parsed. So improved it to catch exceptions, log them and then move on so we get debug information and can allow other scripts to still be parsed if just some of them have issues.
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -543,6 +543,12 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                     GetSignatureHelp(textDocumentPosition, scriptFile)
                         .ContinueWith(async task =>
                     {
+                        if (task.IsFaulted)
+                        {
+                            Logger.Error($"Error getting signature help for script file {scriptFile}: {task.Exception}");
+                            await requestContext.SendError(task.Exception);
+                            return;
+                        }
                         var result = await task;
                         if (result != null)
                         {
@@ -1723,7 +1729,6 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
             ConnectionServiceInstance.TryFindConnection(
                 scriptFile.ClientUri,
                 out connInfo);
-
             var parseResult = await ParseAndBind(scriptFile, connInfo);
 
             // build a list of SQL script file markers from the errors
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -543,6 +543,12 @@ private static TelemetryProperties CreatePeekTelemetryProps(bool succeeded, bool
                     GetSignatureHelp(textDocumentPosition, scriptFile)
                         .ContinueWith(async task =>
                     {
+                        if (task.IsFaulted)
```
cheenamalhotra, line 546: Are you able to debug into this?
I think faulted tasks continue to "ContinueWithOnFaulted" continuation.

Charles-Gagnon: Yes, and I verified it works now (see the logs in the description).

It will go to ContinueWithOnFaulted if there is one - but since this doesn't then it will default to going into ContinueWith. So yes, we could add a separate ContinueWithOnFaulted - but I didn't see much benefit in doing that vs just manually checking IsFaulted

cheenamalhotra: That's strange.. but thanks for capturing it.

cheenamalhotra (approved):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
Charles-Gagnon: More context - ContinueWithOnFaulted is our code wrote - and it's just using ContinueWith and IsFaulted anyways

https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Utility/TaskExtensions.cs#L65

There's some additional stuff it doesn't, but I have a couple issues with it : 

1. I'm not sure why Exception would ever be non-null is IsFaulted is false...that seems redundant to me (and no examples I've seen do the same thing)
2. ContinueWithOnFaulted uses Wait() when calling the handler

One thing to keep in mind - ContinueWith isn't even called if an exception is thrown while starting the task. e.g. the situation I ran in to was that I had manually binplaced a file so there was a runtime exception trying to load a method that didn't exist. So when it tried to call GetSemanticMarkers it thew the runtime exception before the task actually started and thus ContinueWith never got called at all. That's why I had to add the try/catch. This is very much an edge case though so I'm not too worried about it - and generally most of our handler logic will capture any exceptions like that and send back an error immediately.

PR #2307, opened 2024-01-04
[Port] Update Azure.Identity (#2272)

kisantia:
Porting back Azure.Identity update


kisantia:
The CodeQL github workflow was removed from main last year in https://github.com/microsoft/sqltoolsservice/pull/2073 because it isn't used, so I don't think the failing check here is concerning. I can port over the commit to remove the check too if that's preferred.

Charles-Gagnon (approved):

PR #2318, opened 2024-01-12
Allow IsLedger value to be set when creating a database

corivera:
For https://github.com/microsoft/azuredatastudio/issues/24544


kisantia (approved):

kisantia:
can a small test be added to verify this is supported now?

corivera:
It's only supported for 2022 and above, so I'll have to add some versioning logic to the tests

PR #2319, opened 2024-01-12
add requests to create and retrieve s3 credentials

barbaravaldez:
s3-restore

image


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -1531,5 +1533,53 @@ private static string GetDscValue(int id, string value)
...
+                {
+                    using (var cmd = new SqlCommand { Connection = sqlConn })
+                    {
+                        cmd.CommandText = "SELECT [NAME] FROM sys.credentials";
+                        cmd.ExecuteNonQuery();
+                        using (IDataReader reader = cmd.ExecuteReader())
+                        {
+                            while (reader.Read())
+                            {
+                                if(reader.GetString(0).StartsWith("s3"))
```
Charles-Gagnon, line 1573: 1. The casing doesn't match the credential you create above, so is that going to work correctly to find it?
2. According to https://learn.microsoft.com/en-us/sql/relational-databases/backup-restore/sql-server-backup-to-url-s3-compatible-object-storage?view=sql-server-ver16#create-credential the identity must always exactly be `S3 Access Key`, so what about searching for that exactly?

Charles-Gagnon:
Are you going to support renewing/re-entering the access key through the dialog as well? Or will they have to go and manually drop the credential before starting the restore in order to create a new one?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/S3CredentialRequestParams.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// Target database
+        /// </summary>
+        public string ConnectionUri { get; set; }
+    }
+
+    public class NewS3CredentialRequestResponse { }
+
+    public class NewS3CredentialRequest
+    {
+        public static readonly RequestType<NewS3CredentialRequestParams, NewS3CredentialRequestResponse> Type = RequestType<NewS3CredentialRequestParams, NewS3CredentialRequestResponse>.Create("objectManagement/newS3CredentialRequest");
```
Charles-Gagnon, line 33: Instead of making this S3 specific what about just updating the base object management credential stuff as needed and using that? From what I can tell there's nothing here that's "special" - it's just a normal credential with a specific name (which can easily be specified on the UI side instead)

Charles-Gagnon: And then for the GetS3Credentials - I'd suggest making a generic message that can get a credential with a specific name and using that instead (if such a thing doesn't already exist). That way it's usable for other things too if needed in the future.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
barbaravaldez: yup, I added this in the ADS, but using the identity part makes more sense. I'll change it to that instead

barbaravaldez:
> Are you going to support renewing/re-entering the access key through the dialog as well? Or will they have to go and manually drop the credential before starting the restore in order to create a new one?

Yes, I would like to add that option so users can manage their credentials. I'm just not sure how to add the experience in the UI. Not sure if I should create a new dialog or use the create credential dialog to show the current credentials.

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/GetS3Credentials.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement.Contracts
+{
+    public class GetS3CredentialsRequestParams : GeneralRequestDetails
+    {
+        /// <summary>
+        /// Target database
+        /// </summary>
+        public string connectionUri { get; set; }
```
corivera, line 18:
```suggestion
        public string ConnectionUri { get; set; }
```

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/Contracts/S3CredentialRequestParams.cs
```diff
@@ -0,0 +1,35 @@
...
+#nullable disable
+using Microsoft.SqlTools.Hosting.Protocol.Contracts;
+using Microsoft.SqlTools.Utility;
+
+namespace Microsoft.SqlTools.ServiceLayer.ObjectManagement.Contracts
+{
+    public class NewS3CredentialRequestParams : GeneralRequestDetails
+    {
+        /// <summary>
+        /// URI of the underlying connection for this request
```
corivera, line 15: This comment might need updating. Looks like this was copied from ConnectionUri

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -1531,5 +1533,53 @@ private static string GetDscValue(int id, string value)
...
             }
         }
+
+        public void CreateNewCredential(NewS3CredentialRequestParams obj)
+        {
+            using (var dataContainer = CreateDatabaseDataContainer(obj.ConnectionUri, true, String.Empty))
+            {
+                SecureString secret = $"{obj.Secret}".StringToSecureString();
+                secret.MakeReadOnly();
+                string identity = "S3 Access Key";
```
corivera, line 1543: Since this might get reused below, could you make it a constant?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Database/DatabaseHandler.cs
```diff
@@ -1531,5 +1533,53 @@ private static string GetDscValue(int id, string value)
                     return value;
             }
         }
+
+        public void CreateNewCredential(NewS3CredentialRequestParams obj)
```
corivera, line 1537: Could you add a doc comment for this method as well?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Credential/CredentialHandler.cs
```diff
@@ -60,6 +63,36 @@ public override Task<string> Script(CredentialViewContext context, CredentialInf
...
+        {
+            List<string> credentials = new List<string>();
+            ConnectionInfo connectionInfo = this.GetConnectionInfo(parameters.ConnectionUri);
+            using (SqlConnection sqlConn = ConnectionService.OpenSqlConnection(connectionInfo))
+            {
+                if (sqlConn != null)
+                {
+                    using (var cmd = new SqlCommand { Connection = sqlConn })
+                    {
+                        cmd.CommandText = "SELECT [NAME] FROM sys.credentials";
```
Charles-Gagnon, line 81: Since this is just getting the name, maybe name this `getCredentialNames`? Otherwise it sounds like it should be returning the actual secret too

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypeHandler.cs
```diff
@@ -103,6 +104,27 @@ public virtual Task Drop(string connectionUri, string objectUrn, bool throwIfNot
             return Task.CompletedTask;
         }
 
+        public virtual Task CreateCredential(string connectionUri, string objectUrn, bool throwIfNotExist)
```
corivera, line 107: It's unclear how a credential is being created here. Could you add a comment providing more details?

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypes/Credential/CredentialInfo.cs
```diff
@@ -19,5 +19,7 @@ public class CredentialInfo : SqlObject
         public DateTime DateLastModified { get; set; }
         public DateTime CreateDate { get; set; }
         public string ProviderName { get; set; }
+        public string? Secret { get; set; }
+        public string? Url { get; set; }
```
Charles-Gagnon, line 23: From my comment in the other PR, isn't the URL just the name for the S3 secret scenario? (which seems to be backed up by this not seemingly being used anywhere here)

barbaravaldez: that makes sense

src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectTypeHandler.cs
barbaravaldez: Oh sorry, I forgot to remove this method when cleaning the code

PR #2324, opened 2024-02-14
Updating DacFx/Projects version to latest available one

ssreerama:
image
image

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -28,8 +28,8 @@
 		<!-- TODO: Upgrade package and use Token Credential design -->
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
 		<PackageReference Update="Microsoft.SqlServer.Management.SmoMetadataProvider" Version="170.18.0" />
-		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="162.2.82-preview" />
-		<PackageReference Update="Microsoft.SqlServer.DacFx.Projects" Version="0.2.4-preview" />
+		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="162.2.91-preview" />
+		<PackageReference Update="Microsoft.SqlServer.DacFx.Projects" Version="0.2.3-preview" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Data" Version="9.0.4" />
 		<PackageReference Update="Microsoft.Azure.Kusto.Language" Version="9.0.4" />
 		<PackageReference Update="Microsoft.SqlServer.Assessment" Version="[1.1.17]" />
@@ -68,7 +68,7 @@
 		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.4" />
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="170.18.0" />
 		<PackageReference Update="Newtonsoft.Json" Version="13.0.3" />
-		<PackageReference Update="System.Configuration.ConfigurationManager" Version="8.0.0" />
+		<PackageReference Update="System.Configuration.ConfigurationManager" Version="7.0.0" />
 	</ItemGroup>
 
 	<ItemGroup>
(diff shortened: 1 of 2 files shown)


ssreerama:
Project version update of 0.2.4-preview requires DacFx version to update to 162.2.27-preview which we wont use preview version in STS for the release( as per my knowledge). Converting it to draft for now.

Packages.props
```diff
@@ -29,7 +29,7 @@
 		<PackageReference Update="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version="1.1.1" />
 		<PackageReference Update="Microsoft.SqlServer.Management.SmoMetadataProvider" Version="170.18.0" />
 		<PackageReference Update="Microsoft.SqlServer.DacFx" Version="162.1.167" />
```
llali, line 31: dacfx should b updated to 162.2.82-preview

llali (approved):

Packages.props
```diff
@@ -64,11 +64,11 @@
...
 		packages\Microsoft.SqlTools.ManagedBatchParser\Microsoft.SqlTools.ManagedBatchParser.nuspec-->
 	<ItemGroup>
-		<PackageReference Update="Azure.Identity" Version="1.10.2" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.3" />
+		<PackageReference Update="Azure.Identity" Version="1.10.3" />
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.4" />
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="170.18.0" />
 		<PackageReference Update="Newtonsoft.Json" Version="13.0.3" />
-		<PackageReference Update="System.Configuration.ConfigurationManager" Version="7.0.0" />
+		<PackageReference Update="System.Configuration.ConfigurationManager" Version="8.0.0" />
```
cheenamalhotra, line 71: Was it a required change to resolve some conflict? If yes, which package references 8.0.0?

ssreerama: updating DacFx version to latest preview raised errors on build for the version downgrade of above three packages. I know we upgraded AzureIdentity and MDS in dacfx. Checking which package caused configurationManager version reference.

ssreerama: Actually all are for DacFx only:

error NU1605:  MicrosoftSqlToolsServiceLayer -> Microsoft.SqlServer.DacFx 162.2.82-preview -> System.Runtime.Caching 8.0.0 -> Sys
tem.Configuration.ConfigurationManager (>= 8.0.0)

cheenamalhotra: Ok, we should've 'ideally' been on 7.0.0 till we used .NET 7.0, but I hope they're backwards compatible and nothing much has changed. For future ref, do not update this package in DacFx to be beyond .NET major version as sometimes they may bring in breaking changes.

We will soon plan on updating .NET to 8.0 to avoid any issues with this package.

cheenamalhotra (approved):

Packages.props
ssreerama: Got it, the reason for updating it is now DacFx supports .NET8.0 too.

Charles-Gagnon: ...or at the very least, you should always be bumping the major version whenever a dependency with a new major version is brought in so it adheres to semver rules. Otherwise you could end up breaking customers if they have a dependency on a specific major version of a package.

llali: those packages still have target for .net 6. updating it to 8.0.0 should not break anything

llali: if ADS is still on .net6/7 we can make the version of those packages for target .net6 6.0.0. for .net8 we can still use 8.0.0. I think that would make more sense

Charles-Gagnon: STS is still net7, yeah. And that idea sounds great! Hopefully it isn't too much of a pain.

Packages.props
```diff
@@ -64,8 +64,8 @@
 	<!-- When updating version of Dependencies in the below section, please also update the version in the following files:
 		packages\Microsoft.SqlTools.ManagedBatchParser\Microsoft.SqlTools.ManagedBatchParser.nuspec-->
 	<ItemGroup>
-		<PackageReference Update="Azure.Identity" Version="1.10.2" />
-		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.3" />
+		<PackageReference Update="Azure.Identity" Version="1.10.3" />
+		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.4" />
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="170.18.0" />
 		<PackageReference Update="Newtonsoft.Json" Version="13.0.3" />
 		<PackageReference Update="System.Configuration.ConfigurationManager" Version="7.0.0" />
```
ssreerama, line 71: reverted to 7.0.0

llali (approved):

zijchen (approved):

PR #2328, opened 2024-02-21
Moving table designer to sqlcore

aasimkhan30:
Separating out the table designer into sqlcore. I have added a new class called TableDesignerManager that handles all the td models and caching. Now the tabledesigner service is just a barebones request handler and doesn't do any of logical work required for table designer.


kburtram (approved):

src/Microsoft.SqlTools.SqlCore/TableDesigner/Contracts/TableDesignerView.cs
```diff
@@ -3,48 +3,73 @@
...
+        [DataMember(Name = "showRemoveRowConfirmation")]
         public bool ShowRemoveRowConfirmation { get; set; } = false;
     }
 
+    [DataContract]
     public class DesignerTabView
     {
+        [DataMember(Name = "title")]
         public string Title { get; set; }
+        [DataMember(Name = "properties")]
```
caohai, line 72: This should be `components`

PR #2329, opened 2024-02-22
Packages version upgrade required for DacFx NET8.0 support, failing all dacfx extensions

ssreerama:
In this PR, we are updating the versions of below packages to support the ADS dacfx extensions to work. As they are failing to load the 8.0.0 version of packages.

1. System,IO.Packaging from 7.0.0 => 8.0.0
2. Adding System.Runtime.Caching 8.0.0
3. System.Configuration.ConfigurationManager from 7.0.0 => 8.0.0

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -8,7 +8,7 @@
 		<PackageReference Update="System.Security.Permissions" Version="7.0.0" />
 		<PackageReference Update="System.Data.SqlClient" Version="4.8.6" />
 		<PackageReference Update="System.Text.Encoding.CodePages" Version="7.0.0" />
-		<PackageReference Update="System.Text.Encodings.Web" Version="7.0.0" />	
+		<PackageReference Update="System.Text.Encodings.Web" Version="7.0.0" />
 		<PackageReference Update="System.Reactive.Core" Version="6.0.0" />
 
 		<!-- Deprecated: TODO Migrate to Azure.Core -->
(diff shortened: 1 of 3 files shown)


ssreerama:
ADS Issue: https://github.com/microsoft/azuredatastudio/issues/25375

Charles-Gagnon:
Why are we trying to load the 8.* versions of these libraries? The NET6 version of DacFx should be referencing the 6.* version instead : 

https://www.nuget.org/packages/Microsoft.SqlServer.DacFx/162.2.91-preview#dependencies-body-tab

image

Is there some reason we're using the NET8 version of DacFx instead?

Charles-Gagnon:
Ok, so it looks like the fix that @llali made (referenced here : https://github.com/microsoft/sqltoolsservice/pull/2324#discussion_r1492947378) didn't quite resolve the package versioning issues. I'm seeing that we reference the NET6 version of the Dac library, but it still has a reference to the NET8 version of some libraries : 

image

In the interest of trying to keep changes to a minimum (in case there's breaking changes in those assemblies) it might be best to go back and try to fix up the DacFx references to ensure they're not referencing any of the newer major versions of dependent assemblies.

ssreerama:
Hmm, yeah I see some how the Net6 is getting the 8.0.0 packaging dll only, all other dlls looks fine. checking for the fix on DacFx side.

ssreerama:
Not going in this route, since it is fixed in DacFx

PR #2336, opened 2024-02-28
Fix schema matching issue in ObjectExplorerModel

aasimkhan30:
Making sure child objects have the same schema as parents objects. Added some tests for this behavior too.


Charles-Gagnon (approved):
Looks like there's an extra using that needs to be removed

PR #2342, opened 2024-03-11
Add missing password field to ServerConnection string

smartguest:
Currently the Microsoft.SqlServer.Management.Common.ServerConnection class does not properly take in valid SQL connection strings that have empty password, it instead trims the password property entirely from the string and results in the connection having a connection string that does not connect properly when Connect is called on the connection. 

This will need to be addressed by the developers in charge of the SqlServer.Management.Common namespace (of the Dotnet API). For now, to handle this limitation, we add back in the empty password property to the connection string of the ServerConnection if we detect that the password is empty and that the Server Connection uses SQL Authentication.

Addresses: https://github.com/microsoft/azuredatastudio/issues/25048

StoredProcedureScript

### DIFF
--- a/src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
+++ b/src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
@@ -53,6 +53,12 @@ namespace Microsoft.SqlTools.SqlCore.Scripting
             }
 
             ServerConnection = new ServerConnection(sqlConnection);
+            if (string.IsNullOrEmpty(ServerConnection.Password)) {
+                // Manually add password as server connection class does not add it to string, required for successful connection.
+                if(ServerConnection.ConnectionString.IndexOf("Authentication=SqlPassword") != -1 && ServerConnection.ConnectionString.IndexOf("Password=") == -1) {
+                    ServerConnection.ConnectionString += ";Password=;";
+                }
+            }
             if (!string.IsNullOrEmpty(azureAccountToken))
             {
                 ServerConnection.AccessToken = new AzureAccessToken(azureAccountToken);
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -53,6 +53,12 @@ public ScriptAsScriptingOperation(ScriptingParams parameters, string azureAccoun
             }
 
             ServerConnection = new ServerConnection(sqlConnection);
+            if (string.IsNullOrEmpty(ServerConnection.Password)) {
+                // Manually add password as server connection class does not add it to string, required for successful connection.
+                if(ServerConnection.ConnectionString.IndexOf("Authentication=SqlPassword") != -1 && ServerConnection.ConnectionString.IndexOf("Password=") == -1) {
```
cheenamalhotra, line 58: Have you tested the case where the user provides empty password in connection string? i.e. 'Password=;'

cheenamalhotra, line 58: Instead of comparing strings, you can instantiate Sql Connection string builder with this connection string and then compare options.

cheenamalhotra (changes requested):

src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
cheenamalhotra, line 58: Please verify connection string where SqlPassword authentication mode is not specified, as it's auto determined by the driver based on username specified.

Correct logic should be to only check if user is specified, and password is not, you can set password to be empty - it will work or fail based on server authentication requirements.

Use SqlConnectionStringBuilder to compare, as mentioned before.

smartguest: It works fine, it seems based on stepping through my code that ServerConnection removes the Password from the connection string if it's empty no matter what.

cheenamalhotra (approved):

PR #2343, opened 2024-03-15
Update service connection to fix pipelines 

kisantia:
The old service connection was deleted yesterday, which caused the pipelines to start failing when trying to access AKV. I created a new one called "STS Pipelines" that should fix this.

### DIFF
--- a/azure-pipelines/integrationtests.yml
+++ b/azure-pipelines/integrationtests.yml
@@ -66,7 +66,7 @@ steps:
   - task: AzureKeyVault@1
     displayName: 'Azure Key Vault: SqlToolsSecretStore'
     inputs:
-      azureSubscription: 'ClientToolsInfra_670062 (88d5392f-a34f-4769-b405-f597fc533613)'
+      azureSubscription: 'STS Pipelines'
       KeyVaultName: SqlToolsSecretStore
       SecretsFilter: 'sqltools-backup-url-tests-blobcontaineruri,sqltools-backup-url-tests-storageaccountname,sqltools-backup-url-tests-storageaccountkey'
 
--- a/azure-pipelines/integrationtests.yml
+++ b/azure-pipelines/integrationtests.yml
@@ -66,7 +66,7 @@ steps:
   - task: AzureKeyVault@1
     displayName: 'Azure Key Vault: SqlToolsSecretStore'
     inputs:
-      azureSubscription: 'STS Pipelines'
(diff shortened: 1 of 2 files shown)


SakshiS-harma (approved):
Just to understand, why was the service connection removed?

kisantia:
> Just to understand, why was the service connection removed?

There was some cleanup of resources that happened over the last couple days, and from the names, it wasn't obvious what the app in Azure corresponding to the previous service connection was for, so the team decided to delete it and see what broke. And this was one of the things that broke :)

azure-pipelines/integrationtests.yml
```diff
@@ -66,7 +66,7 @@ steps:
   - task: AzureKeyVault@1
     displayName: 'Azure Key Vault: SqlToolsSecretStore'
     inputs:
-      azureSubscription: 'ClientToolsInfra_670062 (88d5392f-a34f-4769-b405-f597fc533613)'
+      azureSubscription: 'STS Pipelines'
```
Matteo-T, line 69: Can we add a link to either the pipeline or the wiki page? Even a couple of guids... any for or  breadcrumb would help the next time around (and GUID are better than strings since they are harder do mutate).

azure-pipelines/release.yml
```diff
@@ -2,7 +2,7 @@ steps:
 - task: AzureKeyVault@1
   displayName: 'Azure Key Vault: ado-secrets'
   inputs:
-    azureSubscription: 'ClientToolsInfra_670062 (88d5392f-a34f-4769-b405-f597fc533613)'
+    azureSubscription: 'STS Pipelines'
     KeyVaultName: 'ado-secrets'
     SecretsFilter: 'github-distro-mixin-password,ado-crossplatbuildscripts-password'
```
kisantia, line 7: I don't think the `github-distro-mixin-password` is needed, since I couldn't find any references here and in the sqltoolsserviceRelease.ps1 script in CrossPlatBuildScripts that gets called. The mixin step is only done in ADS, so it's also weird if a secret with 'mixin' in the name would be needed here. 

After I make sure the pipelines start passing again with this current change, I'll try removing it and see if anything breaks.

azure-pipelines/integrationtests.yml
kisantia: good idea! I added a comment on how to find this service connection. Since this is open source, I avoided putting a link to the actual pipeline

PR #2346, opened 2024-03-21
Upgrade Kusto and SQL Tools services to net8.0

corivera:
Official support for net7.0 ends in May, so I'm upgrading these projects since we publish tools for them to use in Polyglot Notebooks (formerly .NET Interactive). Any of the test projects that reference those also had to be updated to net8.0. A lot of the changes deal with fixing code analysis errors during compilation, since new functionality was added in net8.0. However, there were 900+ errors across the main Kusto and SQL Tools projects, so I just added suppression rules for those to save time.


build.cake
```diff
@@ -126,14 +126,6 @@ Task("PopulateRuntimes")
                 "win-x64",
                 "win-x86",
                 "win-arm64",
-                "ubuntu.14.04-x64",
```
corivera, line 129: The list of runtime IDs was simplified in net8.0, which is why these were removed. More info here: https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#known-rids

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
```diff
@@ -38,8 +38,8 @@ public class ExecutionEngine : IDisposable
         private ScriptExecutionResult result = ScriptExecutionResult.Failure;
         private bool isLocalParse;
         private ExecutionEngineConditions conditions = null;
-        private IList<Batch> preConditionBatches = new List<Batch>();
-        private IList<Batch> postConditionBatches = new List<Batch>();
+        private List<Batch> preConditionBatches = new List<Batch>();
```
Charles-Gagnon, line 41: Why did this change?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/ExecutionEngine.cs
```diff
@@ -191,7 +191,6 @@ private void ConfigureBatchParser()
 
             commandParser = new Parser(batchParser, batchParser, new StringReader(script), "[script]");
             commandParser.SetRecognizeSqlCmdSyntax(sqlCmdMode);
-            commandParser.SetBatchDelimiter(BatchSeparator);
```
Charles-Gagnon, line 194: intended?

src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs
```diff
@@ -541,7 +540,7 @@ private void ParseSetvar(Token setvarToken)
         }
 
 
-        private void ParseConnect(Token connectToken)
```
Charles-Gagnon, line 544: In the future I would suggest suppressing new issues as long as they aren't causing actual compilation failures or something, and then doing follow up PRs to unsuppress and fix them. Makes it much easier to review and prevents issues from sneaking in.

src/Microsoft.SqlTools.ManagedBatchParser/ReliableConnection/RetryPolicy.SqlAzureTemporaryAndIgnorableErrorDetectionStrategy.cs
```diff
@@ -20,10 +20,12 @@ public abstract partial class RetryPolicy
         /// </summary>
         public class SqlAzureTemporaryAndIgnorableErrorDetectionStrategy : ErrorDetectionStrategyBase
         {
+#pragma warning disable CA1859 // Use concrete types when possible for improved performance
```
Charles-Gagnon, line 23: Would suggest just disabling this in editorconfig - and then undoing all the other type changes you did. We can address those separately if needed.

src/Microsoft.SqlTools.ServiceLayer/Agent/Common/LogAggregator.cs
```diff
@@ -679,6 +679,7 @@ private ILogSource CreateUninitializedAggregation(string outputLogSourceName, IL
 #endregion
 
 #region Report Progress
+#pragma warning disable IDE0060 // Remove unused parameter
```
Charles-Gagnon, line 682: Suppress in editorconfig instead

Directory.Build.props
```diff
@@ -27,7 +27,7 @@
     <FileVersion>$(Version)</FileVersion>
     <InformationalVersion>$(Version)</InformationalVersion>
 
-    <ToolsServiceTargetRuntimes>win-x64;win-x86;win-arm64;ubuntu.14.04-x64;ubuntu.16.04-x64;centos.7-x64;rhel.7.2-x64;debian.8-x64;fedora.23-x64;opensuse.13.2-x64;osx.10.11-x64;osx-x64;osx-arm64;linux-x64;linux-arm64</ToolsServiceTargetRuntimes>
```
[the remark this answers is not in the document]
Charles-Gagnon: Saw your response above, glad to see that since it was always a bit iffy to have such specific versions anyways IMO.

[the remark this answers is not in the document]
corivera: They simplified the list of .NET runtime IDs in net8.0. See the docs here: https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#known-rids

Compilation fails if they're still included

src/Microsoft.Kusto.ServiceLayer/Microsoft.Kusto.ServiceLayer.csproj
```diff
@@ -35,14 +35,14 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.OperationalInsights" />
+    <PackageReference Include="Microsoft.Data.SqlClient" VersionOverride="5.2.0" />
```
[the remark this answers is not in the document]
corivera: @cheenamalhotra recommended upgrading to 5.2.0 to resolve MDS issues with .NET 8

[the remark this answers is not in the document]
Charles-Gagnon: Are you trying to limit the scope of the update? I think it'd be better to just update everything to the new version instead unless there's some blocking issue that would make that not be possible.

@cheenamalhotra thoughts?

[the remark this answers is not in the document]
corivera: I'm fine with updating everything for consistency, especially since net7.0 won't be supported anymore. Some people just had concerns about unnecessary updates in other things like Batch Parser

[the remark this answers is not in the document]
Charles-Gagnon: Since it's just 5.1.4 -> 5.2.0 I think that should be fine.

@Matteo-T any concerns on your end about updating sqlclient for ManagedBatchParser?

src/Microsoft.SqlTools.ResourceProvider.DefaultImpl/Microsoft.SqlTools.ResourceProvider.DefaultImpl.csproj
```diff
@@ -9,10 +9,10 @@
 		<OutputType>Library</OutputType>
 		<StartupObject />
 		<Description>Provides the default  for SqlTools applications.</Description>
-		<Copyright>� Microsoft Corporation. All rights reserved.</Copyright>
+		<Copyright>  Microsoft Corporation. All rights reserved.</Copyright>
```
Charles-Gagnon, line 12: We have this defined in the root build.props (with the correct (C) symbol)  so it shouldn't be necessary here

```suggestion
```

src/Microsoft.SqlTools.ResourceProvider/Microsoft.SqlTools.ResourceProvider.csproj
```diff
@@ -6,13 +6,13 @@
     <Company>Microsoft</Company>
     <Product>Sql Tools Service for Resource Provider services</Product>
     <Description>Provides Resource Provider and control plane support.</Description>
-    <Copyright>� Microsoft Corporation. All rights reserved.</Copyright>
+    <Copyright>  Microsoft Corporation. All rights reserved.</Copyright>
```
Charles-Gagnon, line 9: Same as other, already defined in root build.props
```suggestion
```

src/Microsoft.SqlTools.ServiceLayer/Microsoft.SqlTools.ServiceLayer.csproj
```diff
@@ -53,7 +53,7 @@
 		<PackageReference Include="Microsoft.SqlServer.Management.SqlParser" />
 		<PackageReference Include="SkiaSharp" />
 		<PackageReference Include="SkiaSharp.NativeAssets.Linux.NoDependencies" Condition="$([MSBuild]::IsOsPlatform('Linux'))" />
-		<PackageReference Include="System.Configuration.ConfigurationManager" />
+		<PackageReference Include="System.Configuration.ConfigurationManager" VersionOverride="8.0.0" />
```
Charles-Gagnon, line 56: You don't need the version override anymore, right?

Charles-Gagnon:
Looks like the UseDotNet task has some issues - according to this you need to specify `8.0.100` as the SDK for it to work....

https://github.com/microsoft/azure-pipelines-tasks/issues/19292#issuecomment-1817241159

.editorconfig
```diff
@@ -142,6 +142,24 @@ dotnet_diagnostic.CA2215.severity = none
...
+dotnet_diagnostic.CA1862.severity = none
+dotnet_diagnostic.CA1863.severity = none
+dotnet_diagnostic.CA1864.severity = none
+dotnet_diagnostic.CA1865.severity = none
+dotnet_diagnostic.CA1866.severity = none
+dotnet_diagnostic.CA1868.severity = none
+dotnet_diagnostic.CA2250.severity = none
+dotnet_diagnostic.IDE0038.severity = none
+dotnet_diagnostic.CA1510.severity = none
+dotnet_diagnostic.CA1513.severity = none
```
cheenamalhotra, line 162: How about making them a 'suggestion' instead of 'none'?

Packages.props
```diff
@@ -70,7 +70,7 @@
 		<PackageReference Update="Microsoft.Data.SqlClient" Version="5.1.4" />
 		<PackageReference Update="Microsoft.SqlServer.SqlManagementObjects" Version="170.18.0" />
 		<PackageReference Update="Newtonsoft.Json" Version="13.0.3" />
-		<PackageReference Update="System.Configuration.ConfigurationManager" Version="7.0.0" />
+		<PackageReference Update="System.Configuration.ConfigurationManager" Version="8.0.0" />
```
cheenamalhotra, line 73: Please also consider upgrading System.IO.Packaging, System.Composition, System.Security.Permissions, System.Text.Encoding.CodePages, System.Text.Encodings.Web, Microsoft.Extensions.DependencyModel and Microsoft.Extensions.FileSystemGlobbing to v8.0.0. Maybe combine them together with a comment to update on every .NET major version bump.

corivera: Why do those packages also need to be updated?

Packages.props
```diff
@@ -70,7 +62,18 @@
...
 		<PackageReference Update="Newtonsoft.Json" Version="13.0.3" />
-		<PackageReference Update="System.Configuration.ConfigurationManager" Version="7.0.0" />
+  </ItemGroup>
+  
+  <!-- The following packages always need to be updated to the current .NET SDK version. -->
+  <ItemGroup>
+    <PackageReference Update="Microsoft.Extensions.DependencyModel" Version="8.0.0" />
+    <PackageReference Update="Microsoft.Extensions.FileSystemGlobbing" Version="8.0.0" />
+    <PackageReference Update="System.Composition" Version="8.0.0" />
+		<PackageReference Update="System.Configuration.ConfigurationManager" Version="8.0.0" />
```
Charles-Gagnon, line 72: nit: Fix indentation - here and line 77

Packages.props
Charles-Gagnon: "Need" is a strong word, but in general keeping them updated is a good thing. And major version bumps are a safe time to do that.

src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<!-- Targeting both net7.0 and net472 so that other things such as PS cmdlets can use this
+		<!-- Targeting both net8.0 and net472 so that other things such as PS cmdlets can use this
 		which need to support a wider range of machines -->
-		<TargetFrameworks>net7.0;net472;netstandard2.0</TargetFrameworks>
```
Charles-Gagnon, line 5: We should also be doing a major version bump of this (defined in build-and-release.yml)

Charles-Gagnon: Do the STS version as well while you're at it

azure-pipelines/build-and-release.yml
```diff
@@ -4,7 +4,7 @@ pr: none
 variables:
   # Major version number for the release
 - name: Major
-  value: '4'
+  value: '5'
   # Minor version number for the release (should be incremented post a stable release)
 - name: Minor
   value: '12'
```
Charles-Gagnon, line 10: Reset to 0

Charles-Gagnon (approved):

Packages.props
cheenamalhotra: Because they are built from dotnet/runtime and when it ships v8.0.0, also releases the libraries v8.0.0, so they're best compatible when versions match.

colombod:
When will the package land on nuget?

PR #2348, opened 2024-03-22
Changes needed for Onebranch pipeline to work

SakshiS-harma:
Specify projects' filename for the build process, instead of just folder for Onebranch pipeline to work.
And the changes are compatible with the current pipeline as well.


shueybubbles (approved):
:shipit:

PR #2349, opened 2024-04-01
Add RollForward property to SQL Tools and Kusto projects

corivera:
This property allows us to run the services on a machine that doesn't have the required dotnet SDK version installed. This will future proof the dotnet tool versions of these projects, since the packages don't include a copy of the runtime when built as a dotnet tool. The "Major" option means the services will only target a higher SDK version if the specified SDK version is not present on the machine, as opposed to "LatestMajor" which always uses the highest version available no matter what. Doing it that way seemed a little safer for compatibility.

Verified by downgrading the projects to net7 and successfully running them on a machine with only net8 installed.


corivera:
Formatted the csproj files to fix indentation differences, so use the Ignore Whitespace option to view the relevant changes.

Charles-Gagnon (approved):

PR #2353, opened 2024-04-09
Fix the error message in the assert prompt

jiangtao69039:
The prompt information here should use the `afterTableCreationCompletionItems` object instead of `initialCompletionItems`.


Charles-Gagnon (approved):
Thank you for the contribution!

PR #2354, opened 2024-04-11
handle multiple templates in case of SQL DB

sahuroshan:
In case the assesed instance has more than 50 dbs then instead of one template we will be getting list of templates from the nuget, This PR is to handle that change only.

Added test case for next gen gp recommendation and for  mutiple templates case as well.

ADO task link - https://msdata.visualstudio.com/Database%20Systems/_workitems/edit/3100628

### DIFF
--- a/src/Microsoft.SqlTools.Migration/Contracts/GetArmTemplateRequest.cs
+++ b/src/Microsoft.SqlTools.Migration/Contracts/GetArmTemplateRequest.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System.Collections.Generic;
 using Microsoft.SqlTools.Hosting.Protocol.Contracts;
 
 namespace Microsoft.SqlTools.Migration.Contracts
@@ -10,7 +11,7 @@ namespace Microsoft.SqlTools.Migration.Contracts
     public class GetArmTemplateRequest
     {
         public static readonly
-            RequestType<string, string> Type =
-                RequestType<string, string>.Create("migration/getarmtemplate");
+            RequestType<string, List<string>> Type =
+                RequestType<string, List<string>>.Create("migration/getarmtemplate");
     }
 }
--- a/src/Microsoft.SqlTools.Migration/MigrationService.cs
+++ b/src/Microsoft.SqlTools.Migration/MigrationService.cs
@@ -607,21 +607,24 @@ namespace Microsoft.SqlTools.Migration
         /// </summary>
         internal async Task HandleGetArmTemplateRequest(
     string skuRecommendationReportFilePath,
-    RequestContext<string> requestContext)
+    RequestContext<List<string>> requestContext)
         {
             try
             {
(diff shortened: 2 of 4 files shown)


src/Microsoft.SqlTools.Migration/Contracts/GetArmTemplateRequest.cs
```diff
@@ -3,14 +3,15 @@
...
 
 namespace Microsoft.SqlTools.Migration.Contracts
 {
     public class GetArmTemplateRequest
     {
         public static readonly
-            RequestType<string, string> Type =
-                RequestType<string, string>.Create("migration/getarmtemplate");
+            RequestType<string, List<string>> Type =
+                RequestType<string, List<string>>.Create("migration/getarmtemplate");
```
Ramudaykumar, line 15: From SQL NuGet output is List<SqlArmTemplate>, are we doing any handling to get in List <String>? 
or automatically it is getting handled?

Ramudaykumar (approved):
Looks good to me.

src/Microsoft.SqlTools.Migration/Contracts/GetArmTemplateRequest.cs
sahuroshan: this contract is between sqltoolsservice and ADS.

src/Microsoft.SqlTools.Migration/MigrationService.cs
```diff
@@ -607,21 +607,24 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
-                SqlArmTemplate template = provider.GenerateProvisioningScript(recommendations);
-
-                string jsonOutput = JsonConvert.SerializeObject(
-                template,
-                Formatting.Indented,
-                new JsonSerializerSettings { NullValueHandling = NullValueHandling.Ignore, Culture = CultureInfo.InvariantCulture }
-                );
-
-                await requestContext.SendResult(jsonOutput);
+                List<SqlArmTemplate> templateList = provider.GenerateProvisioningScript(recommendations);
```
sahuroshan, line 616: @Ramudaykumar  the list from nuget is getting handled here. instead of single template here now we are getting  list of templates from nuget.

src/Microsoft.SqlTools.Migration/MigrationService.cs
```diff
@@ -607,21 +607,24 @@ public void InitializeService(IProtocolEndpoint serviceHost)
...
+                List<string> armTemplates = new List<string>();
+                foreach (SqlArmTemplate template in templateList)
```
stuti149, line 618: Should we check if templaList is undefined or it is handled?

Ramudaykumar: please add a null check here, nice catch Stuti!

sahuroshan: I don't think it would be null or undefined , it can be empty though.

PR #2355, opened 2024-04-12
Remove dotnet tool stuff from projects that don't need it

Charles-Gagnon:
Just some cleanup that was brought up in another PR.

Also removing the Migration project from the packaging list. It only needs to be uploaded as a zip currently so there's no reason to generate a nuget for it.

### DIFF
--- a/build.json
+++ b/build.json
@@ -54,12 +54,12 @@
     }
   ],
   "PackageProjects": [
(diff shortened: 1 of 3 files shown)


kisantia (approved):

corivera (approved):

SakshiS-harma (approved):
Thanks for making this change, Charles!

PR #2358, opened 2024-05-06
Removed Azure Secrets from Release YML task

smartguest:
Done on recommendation from @shueybubbles So far testing shows no issues, just that the current commit is not in the main branch. If there's anything I'm doing wrong with the implementation of the release task, please let me know what I can do to fix it. 

I based the inputs on what was found here: https://github.com/microsoft/go-sqlcmd/blob/688eef29013f99edf0b19384e743fcc117aae25a/build/azure-pipelines/build-product.yml#L231 and what the stsRelease.ps1 script does,

### DIFF
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -6,7 +6,7 @@ steps:
     KeyVaultName: 'ado-secrets'
     SecretsFilter: 'github-distro-mixin-password,ado-crossplatbuildscripts-password'
 
-- checkout: git://mssqltools/_git/CrossPlatBuildScripts
+- checkout: git://dev.azure.com/mssqltools/_git/CrossPlatBuildScripts
   displayName: Clone CrossPlatBuildScripts
 
 - task: DownloadBuildArtifacts@0
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -6,8 +6,6 @@ steps:
     KeyVaultName: 'ado-secrets'
     SecretsFilter: 'github-distro-mixin-password,ado-crossplatbuildscripts-password'
 
-- checkout: git://CrossPlatBuildScripts/CrossPlatBuildScripts
-  displayName: Clone CrossPlatBuildScripts
 
 - task: DownloadBuildArtifacts@0
   displayName: 'Download build drop artifacts'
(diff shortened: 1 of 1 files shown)


smartguest:
@corivera Probably not, I will remove it in a bit.

shueybubbles:
>     TargetFolder: '$(Build.SourcesDirectory)/artifacts/package'

why not `$(Build.ArtifactStagingDirectory)`? That's the default folder githubrelease pushes from.


---
Refers to: azure-pipelines/release.yml:15 in b20ae47. [](commit_id = b20ae4750e430ebcd03da23fd37350d60d37e05c, deletion_comment = False)

shueybubbles (approved):
:shipit:

kisantia:
@SakshiS-harma fyi this change will also need to be made in the new STS pipeline yml you're working on

SakshiS-harma:
> @SakshiS-harma fyi this change will also need to be made in the new STS pipeline yml you're working on

Thanks for letting me know. Release will be tackled next.

azure-pipelines/release.yml
```diff
@@ -20,23 +9,24 @@ steps:
     downloadPath: '$(Agent.TempDirectory)'
 
 - task: CopyFiles@2
-  displayName: 'Copy build drop artifacts to: $(Build.SourcesDirectory)/artifacts/package/artifacts/package'
+  displayName: 'Copy build drop artifacts to: $(Build.ArtifactStagingDirectory)'
   inputs:
     SourceFolder: '$(Agent.TempDirectory)/drop'
-    TargetFolder: '$(Build.SourcesDirectory)/artifacts/package'
+    TargetFolder: '$(Build.ArtifactStagingDirectory)'
```
SakshiS-harma, line 15: As far as I understand, you still the whole path- '$(Build. ArtifactStagingDirectory)/artifacts/package' here to navigate to the packages folder.

azure-pipelines/release.yml
```diff
@@ -20,23 +9,24 @@ steps:
...
 
 - script: |
-    cd $(Build.SourcesDirectory)/artifacts/package
+    cd $(Build.ArtifactStagingDirectory)
```
SakshiS-harma, line 18: Same as above. You need the complete path- staging directory doesn't redirect you to the artifacts/package folder.

SakshiS-harma (changes requested):
The build drop path needs to be updated.

smartguest:
@SakshiS-harma Do we need to add anything to the "assets" tag for GithubRelease@1? As far as I know, by default it points to $(Build.ArtifactStagingDirectory)/* 

I'm worried it might pick up other things besides the package artifacts unless they are the only things in the directory.

SakshiS-harma:
> @SakshiS-harma Do we need to add anything to the "assets" tag for GithubRelease@1? As far as I know, by default it points to $(Build.ArtifactStagingDirectory)/*
> 
> I'm worried it might pick up other things besides the package artifacts unless they are the only things in the directory.

My bad, I didn't realize that it was the target folder and it is picking the source from the drop folder. Please revert the last change I asked for.
Assets is optional that you can provide if you want to. It would be on the following lines:
assets: |
      $(Build.ArtifactStagingDirectory)/*.zip
      $(Build.ArtifactStagingDirectory)/*.tar.gz
But I am assuming that only zips and tar files are present in the drop folder, so assets tag shouldn't be necessary.

azure-pipelines/release.yml
```diff
@@ -20,23 +9,24 @@ steps:
...
     rm Microsoft.SqlTools.ServiceLayer-osx-arm64-unsigned-net8.0.tar.gz
     rm Microsoft.SqlTools.Migration-osx-arm64-unsigned-net8.0.tar.gz
   displayName: 'Delete the unsigned arm64-osx packages'
 
-- task: PowerShell@2
-  displayName: 'Run Automated Release Script'
```
SakshiS-harma, line 35: Can you add back an appropriate display name.

SakshiS-harma (approved):

SakshiS-harma:
Also add the assets tag as well to the GithubRelease@1 tag, to be compatible with the sqltoolsserviceRelease.ps1

PR #2360, opened 2024-05-20
Changes required for OneBranch Pipeline Migration for signing stage

SakshiS-harma:
This PR moves DotnetPack stage out of initial "build" stage, so that dlls can first be signed and then packed into nuget packages. 
#It also adds repositoryPath and globalRepository path to a local folder, so that it can download the 3rd party nuget packages to the local directory and can be signed appropriately#-> this part has been moved to STS ADO repo since the old pipeline was failing with the new changes.
Additionally adding some more logging to get more insights when failures happen.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -261,7 +261,7 @@ Task("BuildFx")
             var projectFolder = System.IO.Path.Combine(sourceFolder, project.Name);
             var logPath = System.IO.Path.Combine(logFolder, $"{project.Name}-{framework}-build.log");
             using (var logWriter = new StreamWriter(logPath)) {
-                Run(dotnetcli, $"build --framework {framework} --configuration {configuration} \"{projectFolder}\"",
+                Run(dotnetcli, $"build --framework {framework} --configuration {configuration} -bl \"{projectFolder}\"",
                     new RunOptions
                     {
                         StandardOutputWriter = logWriter,
--- a/build.cake
+++ b/build.cake
@@ -456,7 +456,7 @@ void PublishProject(string packageName, string[] projects)
                     publishArguments = $"{publishArguments} --runtime {runtime} --self-contained";
                 }
                 publishArguments = $"{publishArguments} --framework {framework} --configuration {configuration}";
-                publishArguments = $"{publishArguments} --output \"{outputFolder}\" \"{projectFile}\"";
+                publishArguments = $"{publishArguments} --output \"{outputFolder}\" -p:GeneratePackageOnBuild=false \"{projectFile}\"";
                 var publishStart = DateTime.Now;
                 Information($"=~= [{(publishStart - overallStart)}] Publishing {project} / {framework} / {runtime}");
                 Run(dotnetcli, publishArguments)
--- a/build.cake
+++ b/build.cake
@@ -289,11 +289,11 @@ Task("DotnetPack")
 });
 
 /// <summary>
-///  Packages projects specified in FxBuildProjects using available Nupecs, these projects require that publishing be done first. Note that we
+///  Packages projects specified in FxBuildProjects using available Nuspecs, these projects require that publishing be done first. Note that we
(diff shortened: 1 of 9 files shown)


nuget.config
```diff
@@ -1,5 +1,14 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
+  <config>
+    <!--
+      Where packages should go.  Visual Studio will provide a default value of $(SolutionDir) but command-line based
+      package restores could end up in a different location.  Setting this value keeps Visual Studio and NuGet.exe
+      in sync for packages.config based restores.
+    -->
+    <add key="globalPackagesFolder" value=".packages" />
+    <add key="repositoryPath" value=".packages" />
```
kisantia, line 10: what's the difference between these two?

build.cake
```diff
@@ -277,17 +277,14 @@ Task("BuildFx")
 ///  Packages projects specified in PackageProjects
 /// </summary>
 Task("DotnetPack")
-    .IsDependentOn("Cleanup")
-    .IsDependentOn("Setup")
-    .IsDependentOn("Restore")
```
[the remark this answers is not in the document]
SakshiS-harma: So these were removed because DotnetPack was moved out of the build step. The build step has Cleanup, Setup, Restore as pre-steps to be completed before DotnetPack can run. But when DotnetPack was moved out to go with DotnetPackNuspec, these dependencies had to be removed, else it was cleaning up everything that was built in the prior step.

nuget.config
SakshiS-harma: The globalPackagesFolder is a global cache of NuGet packages. This is where NuGet downloads NuGet packages to as a machine cache. The repositoryPath is the packages directory a solution will use, when using a packages.config file. Generally they are kept same. Combination of both of these helped to get rid of the signing problem I was facing where the 3rd party nuget packages were not getting signed properly, because they were getting downloaded at the default "{username}/.nuget/packages" location.

build.cake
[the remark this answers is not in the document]
Charles-Gagnon: I would suggest also renaming this to DotnetPackNoBuild in that case to make it explicitly clear that this step doesn't do a normal dotnet pack (which would rebuild)

scripts/runhelpers.cake
```diff
@@ -220,7 +221,7 @@ public void DotnetPack(string outputFolder, string projectFolder, string project
 }
 
 public void DotnetPackNuspec(string outputFolder, string projectFolder, string project) {
```
Charles-Gagnon, line 223: While you're here - could you rename this to "NugetPackNuspec"? It's not doing dotnet pack at all which could be very confusing

build.cake
```diff
@@ -322,7 +320,7 @@ Task("DotnetPackServiceTools")
     {
         var outputFolder = System.IO.Path.Combine(nugetPackageFolder);
         var projectFolder = System.IO.Path.Combine(sourceFolder, project);
-        DotnetPack(outputFolder, projectFolder, project);
```
Charles-Gagnon, line 325: Would suggest updating task name to `DotnetPackServiceToolsNoBuild`

build.cake
```diff
@@ -297,6 +294,7 @@ Task("DotnetPack")
 ///  currently.
 /// </summary>
 Task("DotnetPackNuspec")
```
Charles-Gagnon, line 296: Suggest updating task name to `NugetPackNuspec` as well if you change the name of the called method too

Charles-Gagnon (approved):

build.cake
SakshiS-harma: I feel updating task name to add "NoBuild" isn't necessary. The other task names don't have this much information in their name either. Let me know if you still strongly feel that this should be updated.

Charles-Gagnon: Either way is fine, but generally I find having names be more explicit isn't a bad thing. Not a blocker though.

PR #2368, opened 2024-06-14
Add generate xlf file step for localization

lewis-sanchez:
This PR adds a step to generate the *.xlf files, so that the necessary *.lcl files can be generated for localization.


Benjin (approved):
Can you run the loc pipeline against your branch to confirm that the necessary files are getting picked up by the OneLocBuild task?

PR #2369, opened 2024-06-14
Fix loc task config in pipeline script

lewis-sanchez:
This PR fixes the configuration for the localization task in the localization pipeline.


Benjin:
LGTM

PR #2371, opened 2024-06-19
Create PullRequest.yml

cssuh:
Creates a Pull Request YAML build file, and splits build.yml into 3 separate steps:

include-archive.yml
include-build.yml
include-nugetPackageGeneration.yml

This change was done so we can consolidate the pipeline code to one repo, and we will reference these files for the Official and Buddy builds of STS as well.


Benjin:
Are these three includes exact copies of the files from ADO?

PR #2373, opened 2024-06-28
Add PR Pipeline YML

cssuh:
Adds PR Pipeline which calls the following:

- include-build.yml
- include-nugetPackageGeneration.yml
- include-archive.yml


build.cmd
kburtram, line 1: It looks like deleting this file is causing a build break in the CI validation.  Why do we want to delete it?

kburtram:
@cssuh are you still doing something with this PR or should we close it out as abandoned?

cssuh:
@kburtram I'll close it

PR #2376, opened 2024-07-12
Update Azure.Identity to 1.12.0

kburtram:
Updating Microsoft.Identity to 1.12.0 due to Nuget restore warnings.  And updating ManagedBatchParser nuspec per comment in .props file.


aasimkhan30 (approved):

Matteo-T:
Hi @kburtram - I may need to sync up with you on the mechanics of icking up a new version of the Managed Parser package. It's been a while since I refreshed it and I forgot :). IIRC, @cheenamalhotra helped me out last time... Not urgent... I cannot easily move MSAL beyond 4.60.3 yet without risking to break the SQLServer Powershell module, but eventually I suppose the Az.* modules are going to move forward as well, so I'll be ready.

PR #2379, opened 2024-07-16
Apply query execution options once per connection session

kburtram:
Fix for https://github.com/microsoft/azuredatastudio/issues/25525.  This splits the execution options into two sets, one that runs on the connection and the other that runs for each query execution.


caohai (approved):

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/QueryExecution/QueryExecutionServiceTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            { // rerun the query without setting execution option and confirm previous option persists
+                var queryResult = await ExecuteAndVerifyQuery(
+                    "SELECT transaction_isolation_level FROM sys.dm_exec_sessions WHERE session_id=@@SPID;",
+                    connectionInfo.OwnerUri);
+
+                Assert.AreEqual(queryResult.RowCount, 1);
+                Int16 transactionLevel = (Int16)queryResult.Rows[0][0].RawObject;
+                Assert.AreEqual(4, transactionLevel);
+            }
```
caohai, line 65: Can we also have a test case covering doc renaming scenario if it can fit into existing tests easily?

caohai (approved):

PR #2380, opened 2024-07-18
Fix handling of NOEXEC option

kburtram:
The handling of the NOEXEC option isn't correct.  The desired behavior is that when the option is enabled, at the start of the settings code a NOEXEC OFF is emitted so the other settings can be executed, then NOEXEC ON is emitted so that the user query will not be executed.  This updates the logic to match SSMS.  This is a follow-up bug found while testing the fix for https://github.com/microsoft/azuredatastudio/issues/25525.

**ADS profile of simple execution**
image

**SSMS profile of simple execution**
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
@@ -196,30 +196,24 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution
             }
         }
 
-        public string SetNoExecString
+        public string GetSetNoExecString(bool on)
         {
-            get
-            {
-                return string.Format(System.Globalization.CultureInfo.InvariantCulture, s_SetNoExec, (this.settings.NoExec ? s_On : s_Off));
-            }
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QuerySettingsHelper.cs
```diff
@@ -196,30 +196,24 @@ public string SetQuotedIdentifierString
...
-            {
-                return string.Format(System.Globalization.CultureInfo.InvariantCulture, s_SetNoExec, (this.settings.NoExec ? s_On : s_Off));
-            }
+            return string.Format(System.Globalization.CultureInfo.InvariantCulture, s_SetNoExec, (on ? s_On : s_Off));
         }
 
-        public string GetSetStatisticsTimeString(bool? on)
+        public string GetSetStatisticsTimeString(bool on)
         {
-            on = on ?? this.settings.StatisticsTime;
```
kburtram, line 209: the logic to set `on` to the value of the setting is incorrect since the setting is to determine if a SET option should be emitted, not the value of the SET option.  Fortunately, this fallback logic wasn't being used anywhere and can safely be removed.

aasimkhan30 (approved):

cssuh (approved):

PR #2381, opened 2024-07-29
Remove NetStandard and NetFramework as targets for SqlCore

Benjin:
With in-deployment changes for SqlDbNative, SqlCore no longer needs to target anything other than .NET.

### DIFF
--- a/build.json
+++ b/build.json
@@ -42,15 +42,6 @@
         "net472",
         "net8.0"
       ]
-    },
-    {
-      "Name": "Microsoft.SqlTools.SqlCore",
-      "Frameworks": [
-        "net6.0",
-        "netstandard2.0",
-        "net472"
-      ],
-      "SkipPack": "true"
     }
   ],
   "PackageProjects": [
--- a/src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
+++ b/src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
@@ -12,7 +12,7 @@
 		<PreserveCompilationContext>true</PreserveCompilationContext>
 		<AssemblyTitle>SqlTools SqlCore Library</AssemblyTitle>
 		<Description>Provides core sql functionality for SQL server editors like Object explorer, Query Execution, and Scripting</Description>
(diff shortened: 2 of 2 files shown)


src/Microsoft.SqlTools.SqlCore/Microsoft.SqlTools.SqlCore.csproj
```diff
@@ -12,7 +12,7 @@
 		<PreserveCompilationContext>true</PreserveCompilationContext>
 		<AssemblyTitle>SqlTools SqlCore Library</AssemblyTitle>
 		<Description>Provides core sql functionality for SQL server editors like Object explorer, Query Execution, and Scripting</Description>
-		<TargetFrameworks>net6.0;netstandard2.0;net472</TargetFrameworks>		<!-- TODO: remove netstandard; tracked in SqlToolsCore as Task #1254787 -->
```
Benjin, line 15: Close Task 1254787 after this is checked in

build.json
```diff
@@ -42,15 +42,6 @@
         "net472",
         "net8.0"
       ]
-    },
-    {
-      "Name": "Microsoft.SqlTools.SqlCore",
-      "Frameworks": [
-        "net6.0",
```
Charles-Gagnon, line 49: We still need the net6 one for now - will it do that automatically since it's the only target framework?

If you have a test build you did for comparison that'd be great.

Charles-Gagnon (approved):

build.json
Benjin: Yup, it still builds .NET 6; I only removed it from the `FxBuildProjects` section.

The nuget produced via the [PR build](https://msdata.visualstudio.com/SQLToolsAndLibraries/_build/results?buildId=135257542&view=artifacts&pathAsName=false&type=publishedArtifacts) still has .NET 6 binaries in it, but I've kicked off a [Buddy build](https://msdata.visualstudio.com/SQLToolsAndLibraries/_build/results?buildId=135351899&view=results) for even more confirmation.

Charles-Gagnon: Ah, thanks! Looks good!

Benjin: Buddy build also looks good.

PR #2385, opened 2024-08-22
Undo changes made in PR #2218 and optimize copying query results

lewis-sanchez:
This PR is a part of https://github.com/microsoft/azuredatastudio/issues/25632

This PR undoes #2218, which removed functionality to copy query results from SQL Tools Service (STS). This PR brings back functionality to copy results into the clipboard from STS. This PR also removes a bottleneck that was making it take a long time to copy query results because the `ToString`  method was being repeatedly invoked on the string builder that is assembling the query results string that is added to the clipboard.

After optimizing the copy endpoint to stop repeatedly invoking `ToString()` on the `StringBuilder` the time needed to copy query results became significantly faster:

Before removing the repeated `ToString()` calls:
Copying 25,000 rows took 7,982 ms
Copying 50,000 rows took 34,863 ms
Copying 75,000 rows took 76,430 ms
Copying 100,000 rows took 163,462 ms

After removing repeated `ToString()` calls:
Copying 25,000 rows took 1.0278617 ms
Copying 50,000 rows took 1.9388735 ms
Copying 75,000 rows took 2.8113253 ms
Copying 100,000 rows took 3.8120204 ms

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -54,6 +54,7 @@
     <PackageReference Update="Microsoft.SqlServer.XEvent.XELite" Version="2023.1.30.3" />
     <PackageReference Update="SkiaSharp" Version="2.88.6" />
     <PackageReference Update="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="2.88.6" Condition="$([MSBuild]::IsOsPlatform('Linux'))" />
+    <PackageReference Update="TextCopy" Version="6.2.1" />
   </ItemGroup>
 
   <!-- When updating version of Dependencies in the below section, please also update the version in the following files:
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
@@ -31,6 +31,11 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
         /// The selections.
         /// </summary>
         public TableSelectionRange[] Selections { get; set; }
+
+        /// <summary>
+        /// Whether to copy the results directly to the clipboard from STS.
+        /// </summary>
+        public bool CopyDirectlyToClipboard { get; set; }
     }
 
     /// <summary>
@@ -48,11 +53,5 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
     {
         public static readonly RequestType<CopyResultsRequestParams, CopyResultsRequestResult> Type =
             RequestType<CopyResultsRequestParams, CopyResultsRequestResult>.Create("query/copy");
-    }
-
-    public class CopyResultsToClipboardEvent
-    {
-        public static readonly EventType<CopyResultsRequestParams> Type =
-            EventType<CopyResultsRequestParams>.Create("query/copyToClipboard");
     }
 }
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
@@ -31,11 +31,6 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
         /// The selections.
         /// </summary>
         public TableSelectionRange[] Selections { get; set; }
-
-        /// <summary>
-        /// Whether to copy results from the UI process.
-        /// </summary>
-        public bool CopyFromUIProcess { get; set; }
     }
 
     /// <summary>
@@ -43,7 +38,6 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
     /// </summary>
     public class CopyResultsRequestResult
     {
-        public string Results { get; set; }
     }
 
     /// <summary>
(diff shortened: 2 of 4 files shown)


Packages.props
```diff
@@ -54,6 +54,7 @@
     <PackageReference Update="Microsoft.SqlServer.XEvent.XELite" Version="2023.1.30.3" />
     <PackageReference Update="SkiaSharp" Version="2.88.6" />
     <PackageReference Update="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="2.88.6" Condition="$([MSBuild]::IsOsPlatform('Linux'))" />
+    <PackageReference Update="TextCopy" Version="6.2.1" />
```
caohai, line 57: I did a quick check and found that it's tricky to access clipboard on Linux, I believe this package doesn't work with Ubuntu 22.04 without additional package. See https://github.com/CopyText/TextCopy/issues/733

Can you give this package a test on all platforms and see
1) does it work?
2) is it worth it to add this package? or do we want to write something of our own (win should be easy, for mac see https://stackoverflow.com/questions/51909120/how-to-get-the-clipboard-text-on-osx-using-dllimport-with-c)

If we believe it makes sense to use this package, then you might also want to add the nuget page / repo page / license info of this package to the PR description.

lewis-sanchez: @caohai, you're right, this isn't going to work. TextCopy version 6.2.1 was used in [STS PR #2096](https://github.com/microsoft/sqltoolsservice/pull/2096/files#diff-202f452c0bbb139bd6a395d212e8b6cfb7909dc37a19a6b600ef1adf98c7662dR50) and it broke copying on Linux. A follow up PR was then made to copy from ADS instead of copying from STS here: https://github.com/microsoft/azuredatastudio/pull/24341.

I'll continue to look into how to fix this for Linux.

kburtram: @lewis-sanchez do you know why with @alanrenmsft PR (https://github.com/microsoft/sqltoolsservice/pull/2096/) we're still seeing slow copy experience?  That PR seems to be doing what I thought this PR was going to do, and the recent bug report isn't for Linux (https://github.com/microsoft/azuredatastudio/issues/25632).

kburtram:
@lewis-sanchez is it possible to do some basic benchmarking for this change?  For example, with some very large resultssets how long does copy take with & without this change on various environments (specifically Linux and Windows).  With all the previous PRs in this area linked above, I want to make sure we're address the issue reported in the most recent bug.

lewis-sanchez:
> @lewis-sanchez is it possible to do some basic benchmarking for this change? For example, with some very large resultssets how long does copy take with & without this change on various environments (specifically Linux and Windows). With all the previous PRs in this area linked above, I want to make sure we're address the issue reported in the most recent bug.

@kburtram, yes, it should be possible to do some benchmarking for this change.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResultsRequest.cs
```diff
@@ -38,7 +38,6 @@ public class CopyResultsRequestParams : SubsetParams
     /// </summary>
     public class CopyResultsRequestResult
     {
-        public string Results { get; set; }
```
kburtram, line 41: so if the copy operation is being done in the UI process (no Linux or remote extension host), how does the UI get the string to put on the clipboard now?

lewis-sanchez: Copying from the UI is done when this setting is unchecked in Azure Data Studio's settings
image

When the "Prefer Providers Copy Handler" setting is unchecked, the STS endpoint that copies results will not be invoked and a different code flow will handle copying results from the UI. The logic for that is found here in the Azure Data Studio repo: [Code in screenshot](https://github.com/microsoft/azuredatastudio/blob/570c27b39edb2e536a89e7f7df20f886fc219296/src/sql/workbench/services/query/common/queryRunner.ts#L608)

image

`handleCopyRequestByProvider` is the method that allows STS to copy results, and `copySelectionToClipboard` copies query results to the clipboard from the UI.

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -863,19 +864,17 @@ internal async Task HandleCopyResultsRequest(CopyResultsRequestParams requestPar
                             }
                         }
                         // Add line break if this is not the last row in all selections.
-                        if (rowIndex + pageStartRowIndex != lastRowIndex && (!builder.ToString().EndsWith(Environment.NewLine) || (!Settings?.QueryEditorSettings?.Results?.SkipNewLineAfterTrailingLineBreak ?? true)))
+                        var lastChar = builder[builder.Length - 1];
```
caohai, line 867: This implementation will behave differently from the old implementation on windows when the string ends with \n but not \r\n. (Old implementation would return false for the `EndsWith` check but new one would return true).

Instead of doing manual char level comparison, you can probably use [ToString(int,int)](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.tostring?view=net-8.0&redirectedfrom=MSDN#system-text-stringbuilder-tostring(system-int32-system-int32)) overload, which should still be able to avoid the perf overhead of full string generation of `ToString()`.

```
public static bool EndsWith(StringBuilder sb, string test)
{
    if (sb.Length < test.Length)
        return false;
    return sb.ToString(sb.Length - test.Length, test.Length).EndsWith(test);
}
```

PR #2388, opened 2024-09-10
Handle cross database view definition

asrichesson:
Fixes https://github.com/microsoft/vscode-mssql/issues/1132

Fixes issue where peeking/viewing the definition of an object from a different database would result in showing only the schema. 

Example:
```sql
USE MyDatabase
GO
SELECT * FROM [ExternalDatabase].dbo.SomeTable
```

Pressing f12 on the table will now properly show the table definition 
```sql
CREATE TABLE [dbo].[SomeTable](
	[column_1] [int] NULL
) ON [PRIMARY]
GO
```

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/SqlObjectIdentifier.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/SqlObjectIdentifier.cs
@@ -1,0 +1,19 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.SqlTools.ServiceLayer.Scripting
+{
+    internal class Sql4PartIdentifier : Sql3PartIdentifier
+    {
+        public string? ServerName { get; set; }
+    }
+
+    internal class Sql3PartIdentifier
+    {
+        public required string ObjectName { get; set; }
+        public string? SchemaName { get; set; }
+        public string? DatabaseName { get; set; }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/Scripting/SqlObjectIdentifier.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
kburtram, line 1: It looks like this header is expected follow a specific format and I'm getting the below build error.

```
.\sqltoolsservice\src\Microsoft.SqlTools.ServiceLayer\Scripting\SqlObjectIdentifier.cs(1,1): er
ror IDE0073: A source file contains a header that does not match the required text (https://learn.microsoft.com/dotnet/
fundamentals/code-analysis/style-rules/ide0073) [.\sqltoolsservice\src\Microsoft.SqlTools.Servi
ceLayer\Microsoft.SqlTools.ServiceLayer.csproj]
```

this is the expected format.

```
//
//
```

kburtram (approved):
LGTM, but needs the copywrite header updated to fix build error.

src/Microsoft.SqlTools.ServiceLayer/Scripting/SqlObjectIdentifier.cs
asrichesson: fixed

caohai (approved):

caohai:
@asrichesson Thanks for fixing this issue! Feel free to go ahead and merge the PR. Once it's merged, I'll queue a new build and bring the fix to vscode-mssql.

asrichesson:
@caohai I don't think I have permissions to complete the PR

PR #2397, opened 2024-09-20
Add XE 'Actions' into Profiler Event

### LINKED ISSUE #2396: 'ProfilerEvent' class does not include XEvent 'Action'(s)
`XeStreamObservable` method `OnEventRead(IXEvent xEvent)` will wrap the XE **Values** into `ProfilerEvent.Values` which is further dispatched to current observers. However, the dictionary does not include the information from XE related to the '**Actions**', only the 'Fields'.

https://github.com/microsoft/sqltoolsservice/blob/18ded668d887bea9bb0052f97bbabb62365bb3bd/src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs#L139

https://github.com/microsoft/sqltoolsservice/blob/18ded668d887bea9bb0052f97bbabb62365bb3bd/src/Microsoft.SqlTools.ServiceLayer/Profiler/Contracts/ProfilerEvent.cs#L40

image

emchavez:
fixes #2396 issue 

`ProfilerEvent` dictionary will not contain the information from the XEvent related to the '**Actions**'. 

If we attempt to open XEL files on an app consumer of SqlToolService  (ie.`Azure Data Studio`) the information will miss these columns. 

XEvent **Actions** are used commonly, for example to automate generic metadata such as 'app name' 'code package version', 'timestamp', etc. Potentially any piece/kind of telemetry, hence they should not be omitted. 

Example of ADS displaying the info (with local STS)
image

Example of ADS displaying the info (default)
image

at last, we chose to add a '(action)' postfix in case there is a key dup. SSMS is following this convention when displaying the info:
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
@@ -143,6 +143,17 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
             {
                 profileEvent.Values.Add(kvp.Key, kvp.Value.ToString());
             }
+            // Add the XE 'actions'.
+            foreach (var kvp in xEvent.Actions)
+            {
+                string key = kvp.Key;
+                if (profileEvent.Values.ContainsKey(key))
+                {
+                    // Append a postfix to avoid duplicate keys while keeping the data.
+                    key += " (action)";
+                }
+                profileEvent.Values.Add(key, kvp.Value.ToString());
+            }
             CurrentObservers.ForEach(o => o.OnNext(profileEvent));
             return Task.FromResult(0);
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
```diff
@@ -143,6 +143,17 @@ private Task OnEventRead(IXEvent xEvent)
             {
                 profileEvent.Values.Add(kvp.Key, kvp.Value.ToString());
             }
+            // Add the XE 'actions'.
+            foreach (var kvp in xEvent.Actions)
```
kburtram, line 147: I assume `Actions` can never be null here?

kburtram (approved):

kburtram:
@emchavez this change looks good to me.  Please merge when you're ready, or if you don't have permissions, please let me know and I'll merge for you.

src/Microsoft.SqlTools.ServiceLayer/Profiler/ObservableXEventSession.cs
emchavez: I will add a explicit check for this, otherwise this would generate exception

PR #2407, opened 2024-10-15
Fix Programmability not being available for Fabric DW and Synapse SQL Serverless Pools

cgettys-microsoft:
Fabric DW and Synapse SQL Serverless Pools both support a subset of the Programmability features of SQL Server. And they both share an engine edition (not addressing that at this time).

As such, both need most of the Programmability folder exposed.
This is already the behavior of SSMS:
image

But unfortunately, this is not the behavior of VS Code and Azure Data Studio.
It looks like SSMS and SQL Tools Service diverged at some point, probably by mistake.
So the OE in VS Code looks like this:
image

The fix is very simple - a few lines of configuration.
Tested manually in VS Code for both products.
With the change, a Synapse SQL Serverless Pool OE in VS Code looks like this:
image
And similar for Fabric DW.

This matches SSMS behavior.
When this reaches ADS and VS Code, it should close bugs like:
* [Cannot find Stored Proc and Function folders for Fabric SQL Data warehouse in Azure Data Studio · Issue #25942 · microsoft/azuredatastudio (github.com)](https://github.com/microsoft/azuredatastudio/issues/25942)
* [Cannot browse programmability (stored procedures) under ondemand pool · Issue #24557 · microsoft/azuredatastudio (github.com)](https://github.com/microsoft/azuredatastudio/issues/24557) 
   * Duplicate: [Synapse Serverless Pool Missing Programmability folder and objects · Issue #25034 · microsoft/azuredatastudio (github.com)](https://github.com/microsoft/azuredatastudio/issues/25034)

### DIFF
--- a/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/SmoModel/SmoTreeNodes.cs
+++ b/src/Microsoft.SqlTools.SqlCore/ObjectExplorer/SmoModel/SmoTreeNodes.cs
@@ -910,7 +910,7 @@ namespace Microsoft.SqlTools.SqlCore.ObjectExplorer.SmoModel
                 NodeValue = SR.SchemaHierarchy_Programmability,
                 NodeTypeId = NodeTypes.Programmability,
                 IsSystemObject = false,
-                ValidFor = ValidForFlag.NotSqlDemand,
+                ValidFor = ValidForFlag.All,
                 SortPriority = SmoTreeNode.NextSortPriority,
             });
             currentChildren.Add(new FolderNode {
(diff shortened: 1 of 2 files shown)


brooksie1821:
Do we have a tentative date for when this is going to be pushed?

aasimkhan30 (approved):

cgettys-microsoft:
> Do we have a tentative date for when this is going to be pushed?

My understanding is that should be in releases next month (November). But I'm on the Fabric DW side of things, so don't quote me - it's @kburtram and team's release process and I don't want to put words in their mouth.

brooksie1821:
awesome!  Glad to hear it.  So far this is the only reason I still have to use SSMS, be glad to make the switch.

cgettys-microsoft:
@brooksie1821 - you'll be happy to hear that Azure Data Studio 1.50 (as well as the latest mssql VS Code extension release) that came out today should have this fix. Hope that helps :).

PR #2408, opened 2024-10-15
Removing incorrect "attachedDBFileName" option

Benjin:
The correct option already existed - "attachDbFileName" (no "ed") - so this wasn't causing any breakage, just confusion


src/Microsoft.SqlTools.ServiceLayer/Localization/sr.resx
```diff
@@ -5506,7 +5506,7 @@ The Query Processor estimates that implementing the following index could improv
     <comment></comment>
   </data>
   <data name="ConnectionConfigOptions_connectTimeout_displayName" xml:space="preserve">
-    <value>Connect timeout</value>
+    <value>Connection timeout (seconds)</value>
```
Benjin, line 5509: some loc changes that missed the last PR

laurennat (approved):

PR #2416, opened 2024-10-18
Editing attach options to fix debugging on Mac.

aasimkhan30:
Pushing the fix done by @kburtram that fixes debugging on Mac. Tested on windows and debugging works fine there too.


kburtram (approved):
Thanks for fixing!

PR #2421, opened 2024-10-23
Fix template generation issue in Mac

sahuroshan:
**Issue -** 
The  **HandleGetArmTemplateRequest** was working perfectly in case of windows and we were receiving ArmTemplate in response , but this call was failing in case of MacOS.

**Reason -**
 On debugging we found the **ExtractSkuRecommendationsFromReport** call was adding some random Console message (**mesage - SKU recommendation report found at {filepath}**) in the response .

**Fix -**
We found that there was one Console.log line in **ExtractSkuRecommendationsFromReport**  function and we removed that  line . PR link - https://msdata.visualstudio.com/Database%20Systems/_git/sql-migrate-asmt/pullrequest/1491680

But even after removing that Console.log line , we encountered same issue on Mac , hence before calling ExtractSkuRecommendationsFromReport method we are setting **Console.Out to null** so that it does not log any console messages.

**Also removed the logs that were added just for debugging this issue.**

### DIFF
--- a/src/Microsoft.SqlTools.Migration/MigrationService.cs
+++ b/src/Microsoft.SqlTools.Migration/MigrationService.cs
@@ -611,15 +611,33 @@ namespace Microsoft.SqlTools.Migration
         {
             try
             {
-                Logger.Verbose("request received in toolsservice");
                 ProvisioningScriptServiceProvider provider = new ProvisioningScriptServiceProvider();
                 string searchPattern = $"*{targetType}-Baseline*.json";
                 string skuRecommendationReportFilePath = Directory.GetFiles(SqlAssessmentConfiguration.ReportsAndLogsRootFolderPath, searchPattern).FirstOrDefault();
-                Logger.Verbose($"Logging report file path -- {skuRecommendationReportFilePath}");
-                List<SkuRecommendationResult> recommendations = ExtractSkuRecommendationReportAction.ExtractSkuRecommendationsFromReport(skuRecommendationReportFilePath);
-                Logger.Verbose($"recommendations generated-- {recommendations.Count}");
+
+                // Save the current Console.Out
+                var originalOut = Console.Out;
+                List<SkuRecommendationResult> recommendations;
+                try
+                {
+                    // Redirect Console.Out to a null stream (no output)
+                    using (var nullWriter = new StreamWriter(Stream.Null))
+                    {
+                        Console.SetOut(nullWriter);
+
+                        /* This call was adding some console messages in the response hence the API call was failing in MacOS
+                         * setting Console.Out to a null stream so that this call does not add any random messages to response.*/
+                        recommendations = ExtractSkuRecommendationReportAction.ExtractSkuRecommendationsFromReport(skuRecommendationReportFilePath);
+                    }
+                }
+                finally
+                {
+                    // Restore the original Console.Out
+                    Console.SetOut(originalOut);
+                }
+
+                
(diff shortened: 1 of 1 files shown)


Bhavikshah123 (approved):

src/Microsoft.SqlTools.Migration/MigrationService.cs
```diff
@@ -611,15 +611,33 @@ internal async Task HandleGetArmTemplateRequest(
...
+                try
+                {
+                    // Redirect Console.Out to a null stream (no output)
+                    using (var nullWriter = new StreamWriter(Stream.Null))
+                    {
+                        Console.SetOut(nullWriter);
+
+                        /* This call was adding some console messages in the response hence the API call was failing in MacOS
+                         * setting Console.Out to a null stream so that this call does not add any random messages to response.*/
+                        recommendations = ExtractSkuRecommendationReportAction.ExtractSkuRecommendationsFromReport(skuRecommendationReportFilePath);
```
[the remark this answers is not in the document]
sahuroshan: I have mentioned in PR description we already tried that and still got the same error . 
 Explaining the issue here again -

We found that there was one Console.log line in ExtractSkuRecommendationsFromReport function and we removed that line . PR link - https://msdata.visualstudio.com/Database%20Systems/_git/sql-migrate-asmt/pullrequest/1491680

But even after removing that Console.log line , we encountered same issue on Mac , hence before calling ExtractSkuRecommendationsFromReport method we are setting Console.Out to null so that it does not log any console messages.

src/Microsoft.SqlTools.Migration/MigrationService.cs
```diff
@@ -611,15 +611,33 @@ internal async Task HandleGetArmTemplateRequest(
         {
             try
             {
-                Logger.Verbose("request received in toolsservice");
```
aasimkhan30, line 614: I would suggest to keep log statements as they help in debugging customer issues

src/Microsoft.SqlTools.Migration/MigrationService.cs
```diff
@@ -611,15 +611,33 @@ internal async Task HandleGetArmTemplateRequest(
...
-                List<SkuRecommendationResult> recommendations = ExtractSkuRecommendationReportAction.ExtractSkuRecommendationsFromReport(skuRecommendationReportFilePath);
-                Logger.Verbose($"recommendations generated-- {recommendations.Count}");
+
+                // Save the current Console.Out
+                var originalOut = Console.Out;
+                List<SkuRecommendationResult> recommendations;
+                try
+                {
+                    // Redirect Console.Out to a null stream (no output)
+                    using (var nullWriter = new StreamWriter(Stream.Null))
```
aasimkhan30, line 624: Nit: maybe create an issue for this so that you can track work for properly fixing the actual bug.

aasimkhan30 (approved):
Please add back the log statements

src/Microsoft.SqlTools.Migration/MigrationService.cs
sahuroshan: Makes sense will create

src/Microsoft.SqlTools.Migration/MigrationService.cs
sahuroshan: Will create a separate task fo logging. None of our APIs are adding logs as of now. May be we can have one PR which adds logs for all of APIs

PR #2428, opened 2024-12-04
Separating group name from displayName

Benjin:
Splitting the display name from the group name so that we have a known string that we can process on in the MSSQL layer that's separate from the translated display name


aasimkhan30 (approved):

aasimkhan30:
Probably we should only keep groupname with the fields and send a mapping of group names and translated names.

Benjin:
Yeah, that's the right way to do this... I was thinking about keeping the contract changes as simple as possible, but I think this isn't currently being used outside of MSSQL, so the risk is low.

PR #2436, opened 2025-01-14
Removing unused filtering options

aasimkhan30:
Fixes: https://github.com/microsoft/vscode-mssql/issues/18288


Benjin (approved):

Benjin:
These are also not used in ADS, right?

aasimkhan30:
@Benjin, these options aren't used in ADS.

PR #2438, opened 2025-01-15
Automatically disable language service when non T-SQL sql language is detected

laurennat:
Notify client when non tsql language is detected.

MSSQL PR: 

tooManySyntaxErrosCase
languageServiceCase1


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/NonTSqlNotification.cs
```diff
@@ -0,0 +1,741 @@
...
+        /// A list of keywords that exist in MySQL and PostgreSQL documentation
+        /// but are not present in T-SQL documentation.
+        /// Sources:
+        /// - MySQL: https://dev.mysql.com/doc/refman/8.4/en/keywords.html
+        /// - PostgreSQL: https://www.postgresql.org/docs/current/sql-keywords-appendix.html
+        /// - T-SQL: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql
+        /// </summary>
+        public static readonly HashSet<string> Keywords = new HashSet<string>
+        {
+            "COPARTITION",
```
aasimkhan30, line 57: Can you make this list in alphabetical order? vscode should have an extension that will let you sort lines alphabetically.

If possible, can you separate this list into Postgres/MYSQL, etc?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1726,6 +1726,40 @@ private async Task<CompletionItem[]> ApplyCompletionExtensions(ConnectionInfo co
...
         #region Diagnostic Provider methods
 
+        /// <summary>
+        /// Checks for non T-SQL syntax within the Parse Result, and 
+        /// sends notification if non T-SQL syntax is detected
+        /// </summary>
+        /// <param name="scriptFile"></param>
+        private async Task CheckForNonTSqlLanguage(string uri, ParseResult parseResult)
+        {
+            if (parseResult.Errors.Count() > TSqlDetectionConstants.SqlFileErrorLimit)
```
aasimkhan30, line 1736: Can you please write unit tests for this feature?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/NonTSqlNotification.cs
```diff
@@ -0,0 +1,741 @@
...
+            "SOURCE_COMPRESSION_ALGORITHMS",
+            "OVERRIDING",
+            "KEY_TYPE",
+            "RLIKE",
+            "ISNULL",
+            "MAX_SIZE",
+            "DUPLICATE",
+            "FIXED",
+            "TRIGGER_SCHEMA",
+            "COST",
```
aasimkhan30, line 171: We should think about removing such common 'object names' from the list.

test/Microsoft.SqlTools.ServiceLayer.UnitTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -210,5 +214,45 @@ public void GetDefaultCompletionListWithMatchesTest()
...
         }
+
+        [Test]
+        public async Task CheckForNonTSqlLanguageTest()
+        {
+            // get the test service 
+            LanguageService service = TestObjects.GetTestLanguageService();
+            var connInfo = new ConnectionInfo(null, null, null);
+
+            const string tsqlText = "SELECT * FROM sys.objects\n--returning\nselect * s\n";
```
aasimkhan30, line 225: use multiline string to make this more readable.
```csharp
string tsqlText = @"
   SELECT * from sys.objects
   --returning
   select * s
";
```

test/Microsoft.SqlTools.ServiceLayer.UnitTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -210,5 +214,45 @@ public void GetDefaultCompletionListWithMatchesTest()
...
+            const string nonTSqlText = "returning";
+            string errorLimitText = string.Concat(Enumerable.Repeat("select * s\n", 51));
+
+            var scriptFileTSql = new ScriptFile();
```
aasimkhan30, line 229: Ideally, we should separate out every test case instead of combining them. Help in debugging when tests are failing.
Something like 4 separate test cases will be nice.

1. TestValidSqlSyntax.
2. TestInvalidSqlSyntax.
3. Test error threshold
4. Test Non Tsql keyword as object name in valid sql

test/Microsoft.SqlTools.ServiceLayer.UnitTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -210,5 +214,45 @@ public void GetDefaultCompletionListWithMatchesTest()
...
+            ParseResult errorParseResult = service.ParseAndBind(errorScriptFile, connInfo).GetAwaiter().GetResult();
+            
+            // This should not notify, so it won't error out
+            await service.CheckForNonTSqlLanguage(scriptFileTSql.ClientUri, tSqlParseResult);
+
+            // Check that send event is called; this will error out because the Language Service's 
+            // service host is not set up; it will only error out if a Non T Sql Notification is trying to be sent out
+            var exceptionErrorFile = Assert.ThrowsAsync<System.InvalidOperationException>(async () =>
+               await service.CheckForNonTSqlLanguage(errorScriptFile.ClientUri, errorParseResult));
+            Assert.AreEqual("SendEvent called when ProtocolChannel was not yet connected", exceptionErrorFile.Message);
```
aasimkhan30, line 249: What are we testing here exactly? Shouldn't we just assert true/false? You can try isolating your core logic to bare minimum so that it is easier to test.

aasimkhan30: We shouldn't be testing indirect events that might occur because of certain things in our core logic. 

It will be better if we are just testing true/false for these tests.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/NonTSqlNotification.cs
```diff
@@ -0,0 +1,687 @@
...
+    /// </summary>
+    public class NonTSqlParams
+    {
+        /// <summary>
+        /// URI identifying the file detected as non-TSQL.
+        /// </summary>
+        public string OwnerUri { get; set; }
+
+        /// <summary>
+        /// Indicates whether the file was flagged due to containing
```
Benjin, line 24: Update comment to explicitly include what true and false mean:

> If true, file contains non-TSQL keywords.  If false, file hit the error limit.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1738,6 +1778,8 @@ internal async Task<ScriptFileMarker[]> GetSemanticMarkers(ScriptFile scriptFile
                 out connInfo);
             var parseResult = await ParseAndBind(scriptFile, connInfo);
 
+            await CheckForNonTSqlLanguage(scriptFile.ClientUri, parseResult);
```
Benjin, line 1781: given that nothing depends on the return value, we probably don't need to `await` this.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1726,6 +1726,46 @@ private async Task<CompletionItem[]> ApplyCompletionExtensions(ConnectionInfo co
...
+                                       OwnerUri = uri,
+                                       ContainsNonTSqlKeywords = false
+                                   });
+
+                return true;
+            }
+            var significantTokenTexts = parseResult.Script.Tokens
+             .Where(token => token.IsSignificant)
+             .Select(token => token.Text.ToUpperInvariant())
+             .ToList();
```
Benjin, line 1753: I think this is an important path in which to pay attention to performance because we want this to be repsonsive.  Linq is a wonderful tool, but ends up creating new lists each time you chain a call.  Since you only care about finding a single keyword, you can tighten the performance here a couple ways:

* Initialize the HashSet with a case-insensitive comparer to avoid needing to call `ToUpperInvariant()` on every token
* convert the chained Linq calls to a for-loop that checks `IsSignificant` and then immediately if the HashSet contains that keyword.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -336,6 +337,75 @@ await langService.DoHandleRebuildIntellisenseNotification(
...
+        [TestCase(
+            @"
+            SELECT * from [returning]
+            --returning
+            ", 
+            false, 
+            "Should not detect non-T-SQL in a valid T-SQL script, with a Non-T-SQL keyword as an object name"
+        )]
+        [TestCase(
+            "returning",
```
Benjin, line 382: Can you expand this case so that it's an actual MySQL script instead of just a singular keyword?

Benjin (approved):

Benjin:
I'm hitting a false positive with a sample DacFx deployment script (DMed to you).

aasimkhan30 (approved):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1742,14 +1742,21 @@ await ServiceHostInstance.SendEvent(
...
+            IList<SqlIdentifier> scriptIdentifiers = parseResult.Script.RetrieveAllIdentifiers();
+            foreach (var identifier in scriptIdentifiers)
+            {
+                identifiers.Add(identifier.ToString());
+            }
+
             foreach (var token in parseResult.Script.Tokens)
             {
-                if (token.IsSignificant && TSqlDetectionConstants.Keywords.Contains(token.Text))
+                if (token.IsSignificant && TSqlDetectionConstants.Keywords.Contains(token.Text) && !identifiers.Contains(token.Text))
```
Benjin, line 1759: Can you add a test case that covers the identifiers case?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1742,14 +1742,21 @@ await ServiceHostInstance.SendEvent(
...
-                                       NonTSqlKeyword = null
+                                    NonTSqlKeyword = null,
                                    });
-
                 return true;
             }
+
+            HashSet<string> identifiers = new HashSet<string>();
+            IList<SqlIdentifier> scriptIdentifiers = parseResult.Script.RetrieveAllIdentifiers();
+            foreach (var identifier in scriptIdentifiers)
```
Benjin, line 1752: what type is `identifiers`?  We should generally avoid `var` for that reason :)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1742,14 +1742,21 @@ await ServiceHostInstance.SendEvent(
                                    new NonTSqlParams
                                    {
                                        OwnerUri = uri,
-                                       NonTSqlKeyword = null
+                                    NonTSqlKeyword = null,
```
Benjin, line 1745: fix indentation

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
Benjin, line 1752: Generally nice for readability to avoid `var` unless it's apparently from the same line what type it is.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1742,14 +1742,21 @@ await ServiceHostInstance.SendEvent(
...
                                    });
-
                 return true;
             }
+
+            HashSet<string> identifiers = new HashSet<string>();
```
Benjin, line 1750: This should probably be another case-insensitive hashset

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -404,6 +404,24 @@ public async Task CheckForNonTSqlLanguageTest(string scriptText, bool expectedRe
...
+
+        [Test]
+        [TestCase(
+            "./TestScript.txt", 
+            false, 
+            "Long Sql Server script; should not detect Non-T-SQL syntax"
+        )]
+        public async Task CheckForNonTSqlLanguageFromFilePathTest(string scriptText, bool expectedResult, string message)
+        {            
+            if (File.Exists(scriptText))
```
Benjin: Test should fail if the file doesn't exist.  If someone moves files around so that it's no longer in the right spot, then this test will pass.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -404,6 +404,24 @@ public async Task CheckForNonTSqlLanguageTest(string scriptText, bool expectedRe
...
+            return result;
+        }
+
+        [Test]
+        [TestCase(
+            "./TestScript.txt", 
+            false, 
+            "Long Sql Server script; should not detect Non-T-SQL syntax"
+        )]
+        public async Task CheckForNonTSqlLanguageFromFilePathTest(string scriptText, bool expectedResult, string message)
```
Benjin: fix name of `scriptText` input, since it's now a filepath

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -404,6 +404,24 @@ public async Task CheckForNonTSqlLanguageTest(string scriptText, bool expectedRe
...
 
             Assert.AreEqual(expectedResult, result, message);
+            return result;
+        }
+
+        [Test]
+        [TestCase(
+            "./TestScript.txt", 
+            false, 
+            "Long Sql Server script; should not detect Non-T-SQL syntax"
```
Benjin, line 414: May as well state that this "long SQL Server script" is "AdventureWorks DacFx deployment script" and name the .txt file appropriately.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -404,6 +404,24 @@ public async Task CheckForNonTSqlLanguageTest(string scriptText, bool expectedRe
             var result = await langService.CheckForNonTSqlLanguage(scriptFile.ClientUri, parseResult);
 
             Assert.AreEqual(expectedResult, result, message);
+            return result;
+        }
+
+        [Test]
+        [TestCase(
+            "./TestScript.txt",
```
Benjin, line 412: Why not make the file a .sql since that's what it is?

Benjin (approved):

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -336,6 +337,101 @@ await langService.DoHandleRebuildIntellisenseNotification(
...
+                Assert.Fail($"File not found: {filePath}");
+            }
+        }
+
+        public async Task<bool>CheckForNonTSqlHelper(string scriptText)
+        {
+            var connInfo = new ConnectionInfo(null, null, null);
+
+            // Dynamically generate the script if it's null (for error limit test)
+            scriptText ??= string.Concat(Enumerable.Repeat("select * s\n", 51));
```
Benjin, line 423: This should be `TSQLDetectionContants.SqlFileErrorLimit + 1` so the test doesn't fail if someone changes a limit

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -336,6 +337,101 @@ await langService.DoHandleRebuildIntellisenseNotification(
...
+
+            var scriptFile = new ScriptFile();
+            scriptFile.SetFileContents(scriptText);
+
+            var langService = CreateLanguageService(scriptFile);
+            ParseResult parseResult = await langService.ParseAndBind(scriptFile, connInfo);
+
+            var result = await langService.CheckForNonTSqlLanguage(scriptFile.ClientUri, parseResult);
```
Benjin, line 431: I have to scroll up to the function header to see what type the result is; `bool result`

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/LanguageServer/LanguageServiceTests.cs
```diff
@@ -24,6 +24,7 @@
 using Microsoft.SqlTools.ServiceLayer.Connection.Contracts;
 using Moq;
 using NUnit.Framework;
+using Microsoft.SqlServer.Management.SqlParser.Parser;
```
Benjin, line 27: This should be in order with the rest (general rule is System.* first, then everything else alphabetically)

Benjin (approved):

PR #2442, opened 2025-02-03
Made detection for Non-TSQL files less aggressive

laurennat:
Trimmed down Non-TSQL word list, and only
languageserviceupdate
 flag files with more than one non tsql keyword

Fixes https://github.com/microsoft/vscode-mssql/issues/18610


kburtram (approved):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1754,18 +1754,25 @@ await ServiceHostInstance.SendEvent(
                 identifiers.Add(identifier.ToString());
             }
 
+            int count = 0;
+            string[] nonTSqlKeywords = new string[2];
```
Benjin, line 1758: Use a shared const for the threshold to make it easier in case we change it again.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/NonTSqlNotification.cs
```diff
@@ -58,15 +58,12 @@ public static class TSqlDetectionConstants
             // MYSQL words
             "ACCESSIBLE",
             "AGAINST",
-            "ALGORITHM",
```
Benjin, line 61: How'd you choose these keywords to remove?  If you used a new source for TSQL keywords, please update the comment with the link

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Contracts/NonTSqlNotification.cs
laurennat: from testing on https://github.com/microsoft/sql-server-samples/tree/master

PR #2444, opened 2025-02-14
Bump SqlParser to 172.14.1

caohai:
https://github.com/microsoft/vscode-mssql/issues/18508
This PR bumps SqlParser to `172.14.1` to bring in new features like Vector support.
https://www.nuget.org/packages/Microsoft.SqlServer.Management.SqlParser/172.14.1

I had to make two code changes with this vbump as `ParamStartLocation` and `ParamEndLocation` fileds are of type `Location` instead of `Location?` now.


Charles-Gagnon (approved):

Charles-Gagnon:
I would also suggest adding dependabot support for nuget package updates as well to avoid things like this getting super out of date again.

See https://github.com/Azure/azure-functions-sql-extension/blob/main/.github/dependabot.yml for an example

Benjin (approved):

PR #2447, opened 2025-02-20
Adding missing SecureEnclaves option to connection provider options

Benjin:
Option wasn't showing up in MSSQL connection dialog

First half of the fix for https://github.com/microsoft/vscode-mssql/issues/18641


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
```diff
@@ -155,6 +155,19 @@ internal static ConnectionProviderOptions BuildConnectionProviderOptions()
...
                     new ConnectionOption
+                    {
+                        Name = "secureEnclaves",
+                        DisplayName = SR.ConnectionConfigOptions_secureEnclaves_displayName,
+                        Description = SR.ConnectionConfigOptions_secureEnclaves_description,
+                        ValueType = ConnectionOption.ValueTypeCategory,
+                        GroupName = GroupNames.Security,
+                        CategoryValues =
+                        [
+                            new CategoryValue { Name = "Disabled", DisplayName = SR.ConnectionConfigOptions_columnEncryptionSetting_category_Disabled },
```
cssuh, line 166: should this be `ConnectionConfigOptions_secureEnclaves_category_Disabled`?

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
```diff
@@ -155,6 +155,19 @@ internal static ConnectionProviderOptions BuildConnectionProviderOptions()
...
+                            new CategoryValue { Name = "Enabled", DisplayName = SR.ConnectionConfigOptions_columnEncryptionSetting_category_Enabled }
```
cssuh, line 167: same for here `ConnectionConfigOptions_secureEnclaves_category_Enabled`

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionProviderOptionsHelper.cs
Benjin: They could be, but it's the same set of values.  I'll fix the names to make them more clear.

cssuh (approved):

PR #2455, opened 2025-03-19
Fixing issue where GetScriptAs operations are being set to an exception state too early

afayyad:
We received GetScript errors for Fabric SQL DBs in some cases with the message of "An attempt was made to transition a task to a final state when it had already completed".

This is masking the actual error, and it looks like it's because the scriptAsOperation ProgressNotification is setting the task to a final state with SetException, and then in the CompleteNotification event handler, it tries to SetException or SetResult again. However, at that point, the task is already in a final state, so it errors out with the above message.

These changes are to capture the error in the ProgressNotification handler and then only transition the task to a final state once.

### DIFF
--- a/src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
+++ b/src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
@@ -36,12 +36,11 @@ namespace Microsoft.SqlTools.SqlCore.Scripting
             StringBuilder scriptAsTaskProgressError = new StringBuilder();
             scriptAsOperation.CompleteNotification += (sender, args) =>
             {
-                if (args.HasError || !string.IsNullOrEmpty(scriptAsTaskProgressError.ToString()))
+                if (args.HasError) {
+                    scriptAsTaskProgressError.AppendLine(args.ErrorMessage);
+                }
+                if (scriptAsTaskProgressError.Length != 0)
                 {
-                    if (args.HasError)
-                    {
-                        scriptAsTaskProgressError.AppendLine(args.ErrorMessage);
-                    }
                     scriptAsTask.SetException(new Exception(scriptAsTaskProgressError.ToString()));
                 }
                 else
--- a/src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
+++ b/src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
@@ -33,12 +33,16 @@ namespace Microsoft.SqlTools.SqlCore.Scripting
         private static async Task<string> ExecuteScriptAs(ScriptAsScriptingOperation scriptAsOperation)
         {
             TaskCompletionSource<string> scriptAsTask = new TaskCompletionSource<string>();
-            string scriptAsTaskProgressError = string.Empty;
+            StringBuilder scriptAsTaskProgressError = new StringBuilder();
             scriptAsOperation.CompleteNotification += (sender, args) =>
             {
-                if (args.HasError || scriptAsTaskProgressError != string.Empty)
+                if (args.HasError || !string.IsNullOrEmpty(scriptAsTaskProgressError.ToString()))
                 {
-                    scriptAsTask.SetException(new Exception(args.HasError ? args.ErrorMessage : scriptAsTaskProgressError));
+                    if (args.HasError)
+                    {
+                        scriptAsTaskProgressError.AppendLine(args.ErrorMessage);
+                    }
+                    scriptAsTask.SetException(new Exception(scriptAsTaskProgressError.ToString()));
                 }
                 else
                 {
@@ -51,7 +55,7 @@ namespace Microsoft.SqlTools.SqlCore.Scripting
             {
                 if (args.ErrorMessage != null)
                 {
-                    scriptAsTaskProgressError = args.ErrorMessage;
+                    scriptAsTaskProgressError.AppendLine(args.ErrorMessage);
                 }
             };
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
```diff
@@ -50,7 +51,7 @@ private static async Task<string> ExecuteScriptAs(ScriptAsScriptingOperation scr
             {
                 if (args.ErrorMessage != null)
                 {
-                    scriptAsTask.SetException(new Exception(args.ErrorMessage));
+                    scriptAsTaskProgressError = args.ErrorMessage;
```
aasimkhan30, line 54: I don't know if it is possible but can we have multiple progress notifications with errors? Should we make this a list

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
```diff
@@ -33,11 +33,12 @@ public static async Task<string> GetScriptAsScript(ScriptingParams parameters, S
...
         {
             TaskCompletionSource<string> scriptAsTask = new TaskCompletionSource<string>();
+            string scriptAsTaskProgressError = string.Empty;
             scriptAsOperation.CompleteNotification += (sender, args) =>
             {
-                if (args.HasError)
+                if (args.HasError || scriptAsTaskProgressError != string.Empty)
                 {
-                    scriptAsTask.SetException(new Exception(args.ErrorMessage));
+                    scriptAsTask.SetException(new Exception(args.HasError ? args.ErrorMessage : scriptAsTaskProgressError));
```
aasimkhan30, line 41: What if there are both `args.ErrorMessage` and `scriptAsTaskProgressError`. Maybe we should expose both of them in that situation

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
afayyad: I don't think so, based on the code flow I've seen. Looks like we only send up to 1 progress notification, in Create, Alter, or Delete use cases where we go into GenerateScriptAs method, in ScriptAsScriptingOperation

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
afayyad: Good point, I'll do this, thanks

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
```diff
@@ -33,11 +33,12 @@ public static async Task<string> GetScriptAsScript(ScriptingParams parameters, S
         private static async Task<string> ExecuteScriptAs(ScriptAsScriptingOperation scriptAsOperation)
         {
             TaskCompletionSource<string> scriptAsTask = new TaskCompletionSource<string>();
+            string scriptAsTaskProgressError = string.Empty;
             scriptAsOperation.CompleteNotification += (sender, args) =>
             {
-                if (args.HasError)
+                if (args.HasError || scriptAsTaskProgressError != string.Empty)
```
Charles-Gagnon, line 39:
```suggestion
                if (args.HasError || !string.IsNullOrEmpty(scriptAsTaskProgressError))
```

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
Benjin: Agreed; we should either unroll to the proper error or have the full stack

afayyad: Done, put them in a stringbuilder

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
afayyad: Done, added it to the stringbuilder

afayyad:
image

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
```diff
@@ -33,11 +33,16 @@ public static async Task<string> GetScriptAsScript(ScriptingParams parameters, S
         private static async Task<string> ExecuteScriptAs(ScriptAsScriptingOperation scriptAsOperation)
         {
             TaskCompletionSource<string> scriptAsTask = new TaskCompletionSource<string>();
+            StringBuilder scriptAsTaskProgressError = new StringBuilder();
             scriptAsOperation.CompleteNotification += (sender, args) =>
             {
-                if (args.HasError)
```
Charles-Gagnon, line 38: I'd suggest creating a local var with the ToString result - otherwise you're building the string twice.

Charles-Gagnon (approved):

src/Microsoft.SqlTools.SqlCore/Scripting/AsyncScriptAsScriptingOperation.cs
afayyad: But we might to add another error to the stringbuilder, so we might need to build it again, right? Maybe I can just check the length of the stringbuilder instead of checking isNullOrEmpty on the sb.toString()?

Charles-Gagnon: Or move the additional append out and then just have this check the error string length.

aasimkhan30 (approved):

PR #2466, opened 2025-04-07
Add Schema Compare Service endpoint to include/exclude all comparison diffs

lewis-sanchez:
This PR adds a new request endpoint to the Schema Compare Service that permits including/excluding all schema compare differences. This PR is a necessary part to add include/exclude all functionality in the Schema Compare UI that is being added to the MSSQL extension.

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeAllNodesOperation.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeAllNodesOperation.cs
@@ -1,0 +1,120 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
+using Microsoft.SqlServer.Dac.Compare;
+using Microsoft.SqlTools.ServiceLayer.SchemaCompare.Contracts;
+using Microsoft.SqlTools.ServiceLayer.TaskServices;
+using Microsoft.SqlTools.Utility;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.SqlTools.ServiceLayer.SchemaCompare
+{
+    /// <summary>
+    /// Class to represent an in-progress schema compare include/exclude all Node operation
+    /// </summary>
+    class SchemaCompareIncludeExcludeAllNodesOperation : ITaskOperation
+    {
+        private CancellationTokenSource cancellation = new CancellationTokenSource();
+        private bool disposed = false;
+
+        /// <summary>
+        /// Gets the unique id associated with this instance.
+        /// </summary>
+        public string OperationId { get; private set; }
+
+        public SchemaCompareIncludeExcludeAllNodesParams Parameters { get; }
+
+        protected CancellationToken CancellationToken { get { return this.cancellation.Token; } }
+
+        public string ErrorMessage { get; set; }
+
+        public SqlTask SqlTask { get; set; }
+
+        public SchemaComparisonResult ComparisonResult { get; set; }
+
+        public bool Success { get; set; }
+
+        public List<DiffEntry> AllIncludedOrExcludedDifferences;
+
+
+        public SchemaCompareIncludeExcludeAllNodesOperation(SchemaCompareIncludeExcludeAllNodesParams parameters, SchemaComparisonResult comparisonResult)
+        {
+            Validate.IsNotNull("parameters", parameters);
+            this.Parameters = parameters;
+            Validate.IsNotNull("comparisonResult", comparisonResult);
+            this.ComparisonResult = comparisonResult;
+        }
+
+        /// <summary>
+        /// Execute will include/exclude all differences in the schema compare result.
+        /// </summary>
+        /// <param name="mode"></param>
+        public void Execute(TaskExecutionMode mode)
+        {
+            this.CancellationToken.ThrowIfCancellationRequested();
+
+            try
+            {
+                var schemaDifferences = new List<SchemaDifference>(this.ComparisonResult.Differences);
+                this.IncludeExcludeAllDifferences(schemaDifferences);
+            }
+            catch (Exception e)
+            {
+                ErrorMessage = e.Message;
+                Logger.Error(string.Format("Schema compare includ/exclude all operation {0} failed with exception {1}", this.OperationId, e.Message));
+                throw;
+            }
+
+            this.AllIncludedOrExcludedDifferences = new List<DiffEntry>();
+            if (this.ComparisonResult.Differences != null)
+            {
+                foreach (SchemaDifference difference in this.ComparisonResult.Differences)
+                {
+                    DiffEntry diffEntry = SchemaCompareUtils.CreateDiffEntry(difference, null, this.ComparisonResult);
+                    this.AllIncludedOrExcludedDifferences.Add(diffEntry);
+                }
+            }
+        }
+
+        private void IncludeExcludeAllDifferences(List<SchemaDifference> schemaDifferences)
+        {
+            var problematicDifferences = new List<SchemaDifference>();
+            foreach (var difference in schemaDifferences)
+            {
+                this.Success = this.Parameters.IncludeRequest ? this.ComparisonResult.Include(difference) : this.ComparisonResult.Exclude(difference);
+
+                if (!this.Success)
+                {
+                    problematicDifferences.Add(difference);
+                }
+            }
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeAllNodesOperation.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            try
+            {
+                var schemaDifferences = new List<SchemaDifference>(this.ComparisonResult.Differences);
+                this.IncludeExcludeAllDifferences(schemaDifferences);
+            }
+            catch (Exception e)
+            {
+                ErrorMessage = e.Message;
+                Logger.Error(string.Format("Schema compare includ/exclude all operation {0} failed with exception {1}", this.OperationId, e.Message));
```
Benjin, line 69: typo: includ[e]

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeAllNodesOperation.cs
```diff
@@ -0,0 +1,120 @@
...
+                    DiffEntry diffEntry = SchemaCompareUtils.CreateDiffEntry(difference, null, this.ComparisonResult);
+                    this.AllIncludedOrExcludedDifferences.Add(diffEntry);
+                }
+            }
+        }
+
+        private void IncludeExcludeAllDifferences(List<SchemaDifference> schemaDifferences)
+        {
+            var problematicDifferences = new List<SchemaDifference>();
+            foreach (var difference in schemaDifferences)
```
Benjin, line 87: use `SchemaDifference` instead of `var` if the type isn't already on the same line so readers don't have to hunt for the type.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeAllNodesOperation.cs
```diff
@@ -0,0 +1,120 @@
...
+                }
+            }
+
+            if (problematicDifferences.Count != 0)
+            {
+                IncludeExcludeAllDifferences(problematicDifferences);
+            }
+        }
+
+        // The schema compare public api doesn't currently take a cancellation token so the operation can't be cancelled
```
Benjin, line 103: Make this into a docstring

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1371,6 +1371,16 @@ public async Task VerifySchemaCompareServiceCalls()
 
                 await SchemaCompareService.Instance.HandleSchemaCompareIncludeExcludeNodeRequest(excludeParams, publishRequestContext.Object);
 
+                // Include/Exclude all service call
```
Benjin, line 1374: Given that the include/excludeAll logic lives at this layer, there should be tests that validate the actual input/output.  Should be easy to set up with some contrived dacpacs or sqlprojects.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/SchemaCompareIncludeExcludeAllNodesOperation.cs
lewis-sanchez: Sounds good, but the type is present on line 84 where the List is passed in to the method.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
lewis-sanchez: The requested test was added in the this commit: https://github.com/microsoft/sqltoolsservice/pull/2466/commits/b94508958df7f77ebfb242c144bc2b4eda95cefb

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1680,6 +1690,65 @@ public async Task VerifySchemaCompareWarningsBeingExcluded()
...
         }
 
+        /// <summary>
+        /// Verify that include/exclude all excludes all comparison differences.
+        /// </summary>
+        [Test]
+        public async Task IncludeExcludeAllWithDacpacToDacpacComparison()
+        {
+            // create dacpacs from databases
+            SqlTestDb sourceDb = await SqlTestDb.CreateNewAsync(TestServerType.OnPrem, false, null, SourceScript, "SchemaCompareSource");
```
Benjin, line 1700: Glancing at `SourceScript` and `TargetScript`, I'm not sure that they would actually trigger the graph traversal logic, which is the most interesting thing to be tests.  Can you verify that `problematicDifferences` is actually getting employed in this test, and update the test cases if it isn't?  You might need a slightly more contrived test case to ensure it gets triggered.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1680,6 +1690,65 @@ public async Task VerifySchemaCompareWarningsBeingExcluded()
...
+                {
+                    SourceEndpointInfo = sourceInfo,
+                    TargetEndpointInfo = targetInfo
+                };
+
+                var schemaCompareOperation = new SchemaCompareOperation(schemaCompareParams, null, null);
+                schemaCompareOperation.Execute(TaskExecutionMode.Execute);
+                Assert.True(schemaCompareOperation.ComparisonResult.IsValid);
+                Assert.False(schemaCompareOperation.ComparisonResult.IsEqual);
+                Assert.NotNull(schemaCompareOperation.ComparisonResult.Differences);
```
Benjin, line 1725: Add an assert with the specific expected number of differences rather than just "not null" so that we get more confident coverage.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1680,6 +1690,65 @@ public async Task VerifySchemaCompareWarningsBeingExcluded()
...
+                    OperationId = schemaCompareOperation.OperationId,
+                    IncludeRequest = false,
+                    TaskExecutionMode = TaskExecutionMode.Execute
+                };
+
+                var includeExcludeAllNodesOperation = new SchemaCompareIncludeExcludeAllNodesOperation(includeExcludeAllNodesParams, schemaCompareOperation.ComparisonResult);
+                includeExcludeAllNodesOperation.Execute(TaskExecutionMode.Execute);
+
+                Assert.True(includeExcludeAllNodesOperation.Success, "Include/Exclude all operation should succeed");
+                Assert.True(includeExcludeAllNodesOperation.AllIncludedOrExcludedDifferences.All(x => x.Included == false), "All differences should be excluded");
```
Benjin, line 1739: expected number of differences can be used again down here as a second point of verification.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceTests.cs
```diff
@@ -1680,6 +1690,65 @@ public async Task VerifySchemaCompareWarningsBeingExcluded()
...
+                schemaCompareOperation.Execute(TaskExecutionMode.Execute);
+                Assert.True(schemaCompareOperation.ComparisonResult.IsValid);
+                Assert.False(schemaCompareOperation.ComparisonResult.IsEqual);
+                Assert.NotNull(schemaCompareOperation.ComparisonResult.Differences);
+                Assert.IsNull(schemaCompareOperation.ErrorMessage);
+
+                var includeExcludeAllNodesParams = new SchemaCompareIncludeExcludeAllNodesParams()
+                {
+                    OperationId = schemaCompareOperation.OperationId,
+                    IncludeRequest = false,
```
Benjin, line 1731: It'd be good for the test case to also try the IncludeAll case to confirm that both directions work

Benjin (approved):

PR #2470, opened 2025-04-15
Adding missing null check

Benjin:
Previously, a connection string specifying Azure MFA auth but lacking a user ID would cause this to throw.


aasimkhan30 (approved):

aasimkhan30 (approved):

Benjin:
Looks like Cheena had some expected behavior explicitly about this situation; I'm not sure why, so I'll work around this in MSSQL instead.

PR #2473, opened 2025-04-28
Fixed Table Designer Schema Bug

laurenastrid1:
Fixes https://github.com/microsoft/vscode-mssql/issues/18510?reload=1?reload=1

Table Designer wasn't picking up new/dropped schemas

### DIFF
--- a/src/Microsoft.SqlTools.SqlCore/TableDesigner/TableDesignerManager.cs
+++ b/src/Microsoft.SqlTools.SqlCore/TableDesigner/TableDesignerManager.cs
@@ -21,6 +21,7 @@ namespace Microsoft.SqlTools.SqlCore.TableDesigner
         public const string TableDesignerApplicationNameSuffix = "TableDesigner";
         private Dictionary<string, Dac.TableDesigner> idTableMap = new Dictionary<string, Dac.TableDesigner>();
         private const string CheckCreateTablePermissionInDbQuery = "SELECT HAS_PERMS_BY_NAME(QUOTENAME(@dbname), 'DATABASE', 'CREATE TABLE')";
+        private const string GetSchemaNamesQuery = "SELECT NAME FROM SYS.SCHEMAS";
         private const string CheckAlterTablePermissionQuery = "SELECT HAS_PERMS_BY_NAME(QUOTENAME(@schema) + '.' + QUOTENAME(@table), 'OBJECT', 'ALTER')";
         public bool AllowDisableAndReenableDdlTriggers { get; set; } = true;
         public TableDesignerInfo InitializeTableDesigner(TableInfo tableInfo)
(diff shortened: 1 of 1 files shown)


aasimkhan30:
While this would display all schemas, the DacFX model wouldn’t recognize the newly added schemas until it’s reinitialized. As a result, validations like detecting duplicate table names, index or broken foreign keys wouldn’t work, leading users into error states that the model can’t properly handle. I think the existing behavior(while not ideal) is better.  By not showing schemas that DacFX isn’t aware of, we avoid putting users into those problematic situations.

src/Microsoft.SqlTools.SqlCore/TableDesigner/TableDesignerManager.cs
```diff
@@ -1822,6 +1825,32 @@ private void CheckCreateAlterTablePermission(TableInfo tableInfo, string connect
...
+        {
+            var connectionStringBuilder = new SqlConnectionStringBuilder(tableInfo.ConnectionString);
+            connectionStringBuilder.InitialCatalog = tableInfo.Database;
+            connectionStringBuilder.ApplicationName = TableDesignerManager.TableDesignerApplicationNameSuffix;
+            return connectionStringBuilder;
+        }
+
+        private List<string> GetSchemaNames(TableInfo tableInfo, string connectionString)
+        {
+            var schemaNames = new List<string>();
```
kburtram, line 1839: looks fine, and I guess we don't have a SMO object model here to use that for getting Schemas, so this looks quicker.

src/Microsoft.SqlTools.SqlCore/TableDesigner/TableDesignerManager.cs
```diff
@@ -1822,6 +1825,32 @@ private void CheckCreateAlterTablePermission(TableInfo tableInfo, string connect
                 });
             }
         }
+
+        private SqlConnectionStringBuilder GetConnectionStringBuilderForTable(TableInfo tableInfo)
+        {
+            var connectionStringBuilder = new SqlConnectionStringBuilder(tableInfo.ConnectionString);
+            connectionStringBuilder.InitialCatalog = tableInfo.Database;
+            connectionStringBuilder.ApplicationName = TableDesignerManager.TableDesignerApplicationNameSuffix;
```
kburtram, line 1833: `TableDesignerApplicationNameSuffix` seems to suggest a suffix, do we need to stick something after this value, or it's good as-is

kburtram (approved):
LGTM, but have Aasim take a look at this.

src/Microsoft.SqlTools.SqlCore/TableDesigner/TableDesignerManager.cs
laurenastrid1: This code already existed before, I just moved it into a function. so it's probably ok this way.

PR #2474, opened 2025-04-28
Add Chat Service to handle GH Copilot chat participant

kburtram:
Add Chat Service to handle GH Copilot chat participant API.  This includes both the Chat Service, and well as the SemanticKernel "VS Code" connector.  The Connector code will be move to different repo later, but for near future I'm planning to build from here.


src/Microsoft.SqlTools.ServiceLayer/Copilot/VSCodeChatCompletionStream.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+
+            // Add explanation message if last message was system
+            if (result.Count > 0 && result[^1].Role == MessageRole.System)
+            {
+                result.Add(new LanguageModelRequestMessage
+                {
+                    Text = "Above messages contain the results of calling a function. " +
+                          "The user cannot see this result, so you should explain it to the user if referencing it in your answer.",
+                    Role = MessageRole.User
```
caohai, line 115: Should this role be `MessageRole.System`? Otherwise this message is showing up in the user messages section.

kburtram: I think there was an issue where the final message in the chat needs to be a user message.  So this was a workaround for that.  This isn't showing up in the users output.

caohai: Ah I see, I found this line when I was debugging messages. I tested changing this role to `System` and it still worked the same.

caohai (approved):

PR #2489, opened 2025-05-08
Update system prompt to be more specific on allowed actions

kburtram:


src/Microsoft.SqlTools.ServiceLayer/Copilot/CopilotConversationManager.cs
```diff
@@ -207,9 +207,118 @@ private async Task InitConversation(CopilotConversation conversation)
...
+                        - For large tables or complex queries, favor efficient query structures and indexing strategies.
+                        - Warn users about potentially expensive or non-SARGable queries when applicable.
+                        - Suggest using batching or pagination for large data operations.
+                        - Default SELECT examples should include a TOP clause (e.g., TOP 100) unless otherwise specified.
+
+                        VERSION AWARENESS:
+                        - When using T-SQL features, prefer syntax compatible with SQL Server 2016 or newer unless a specific version is known.
+                        - Identify Azure SQL-specific syntax where applicable and clarify when the behavior may differ.";
+
+                                        // Initialize chat history
```
caohai, line 319: indentation seems to be off here?

caohai (approved):

lewis-sanchez (approved):

src/Microsoft.SqlTools.ServiceLayer/Copilot/CopilotConversationManager.cs
kburtram: I'm indenting align with where the code nested would put the string, vs. left aligned.  The model shouldn't care about the white space.  Not sure what's more readable for embedded text blocks like this.  I'll do left aligned if we need to update for other reasons, otherwise, probably leave like this.

caohai (approved):

PR #2496, opened 2025-05-22
Fix hover response

### LINKED ISSUE #2495: Hover sends back two responses for the same request
In [`HandleHoverRequest`](https://github.com/microsoft/sqltoolsservice/blob/48f446723cfa04ae3f0e3734cf61488fcf178819/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs#L567), if the hover request succeeds then two responses are sent. The first is the hover info, the second is a `null`. In the language server protocol, only one response should be sent for one request. The second null response is causing [errors in editors such as neovim](https://github.com/Kurren123/mssql.nvim/issues/39). I think the logic in this function needs to be reworked to only send the null response if the hover info was not sent.

Kurren123:

## Description

Fixes #2495. I didn't add any unit tests as this seems like it needs to be covered by an integration test. Please confirm. 

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.Kusto.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.Kusto.ServiceLayer/LanguageServices/LanguageService.cs
@@ -439,23 +439,19 @@ namespace Microsoft.Kusto.ServiceLayer.LanguageServices
         {
             try
             {
+                Hover hover = null;
                 // check if Quick Info hover tooltips are enabled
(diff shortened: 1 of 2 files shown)


Kurren123:
@microsoft-github-policy-service agree

kburtram (approved):
Thanks for fixing!  I guess the original logic assumed a return after the SendResult; but this single function exit logic seems even better.

PR #2497, opened 2025-05-28
Adding parallel message handling to schema designer apis. 

aasimkhan30:

## Description

Previously, sd apis were running on main thread and were blocking. So if there is a long SD initialization, the entire extension becomes unresponsive. This PR fixes that behavior.  Copying the pattern from table designer for it. 

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SchemaDesigner/SchemaDesignerService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SchemaDesigner/SchemaDesignerService.cs
@@ -30,104 +30,140 @@ namespace Microsoft.SqlTools.ServiceLayer.SchemaDesigner
         {
             Logger.Verbose("Initializing Schema Designer Service");
             this.serviceHost = serviceHost;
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/SchemaDesigner/SchemaDesignerService.cs
```diff
@@ -54,80 +54,85 @@ await requestContext.SendResult(new CreateSessionResponse()
...
                 await requestContext.SendResult(report);
-            }
-            catch (Exception e)
+            });
+        }
+
+        private Task HandleRequest<T>(RequestContext<T> requestContext, Func<Task> action)
+        {
+            // The request handling will take some time to return, we need to use a separate task to run the request handler so that it won't block the main thread.
+            // For any specific table designer instance, ADS UI can make sure there are at most one request being processed at any given time, so we don't have to worry about race conditions.
```
Benjin, line 122: Update comment (ADS, Table Designer).  Also, can this instead be moved someplace else so that it's shared with Table Designer instead of copied?

Benjin (approved):

Benjin (approved):

PR #2506, opened 2025-07-01
Fixing completion items not getting latest token and adding more sql keywords

aasimkhan30:

### Summary
This PR fixes an issue where keyword completion was off by one character. The problem was caused by updating the parser context after retrieving the token text, which resulted in stale context and incorrect completion suggestions.

In addition to the fix, this PR also adds support for several commonly used SQL keywords, including newer ones and multi-word constructs like GROUP BY, to improve completion relevance.

Fixes https://github.com/microsoft/vscode-mssql/issues/1100

## Code Changes Checklist


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1664,7 +1664,6 @@ public async Task<CompletionItem[]> GetCompletionItems(
             // cache the current script parse info object to resolve completions later
             this.currentCompletionParseInfo = scriptParseInfo;
             resultCompletionItems = result.CompletionItems;
-
```
Benjin, line 1667: nit: this probably shouldn't have been removed.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
```diff
@@ -31,8 +31,11 @@ public static class AutoCompleteHelper
 
         public static readonly string[] DefaultCompletionText = new string[]
         {
-            "abs",
+            "@@error",
```
Benjin, line 34: Where'd you source these missing keywords from?  If there's a stable source, maybe that can be added as a comment.

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
aasimkhan30: mostly AI generated and request from the customer in https://github.com/microsoft/vscode-mssql/issues/1100

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
```diff
@@ -420,22 +480,26 @@ public static class AutoCompleteHelper
             "state",
             "statement",
             "static",
-            "statistics",
             "statistics_norecompute",
+            "statistics",
```
Benjin, line 484: Hmm, whatever did the alphabetization chose to put `word_something` ahead of `word` each time.  That's kinda weird...

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/AutoCompleteHelper.cs
aasimkhan30: vscode does smart matching of completion items with actual text in editor. So the ordering in the list doesn't matter much.

Benjin: Yup, I know; it just caught me off-guard.

Benjin (approved):

PR #2509, opened 2025-07-14
Fixing special file URI like GIST for query execution. 

aasimkhan30:

## Description

Issue: https://github.com/microsoft/vscode-mssql/issues/18544

Previously, all connections were tied to files, which meant that even non-file-based connections (like Object Explorer or URI-based ones) had to go through the File.Open command. To handle this, a check was added to ignore unknown file URIs.

However, that's no longer the case. With the current model, this check is now incorrectly filtering out valid non-file URIs—like those used by the GIST extension. Due to which the file related to such special URIs is always empty as it is actively ignored by STS.  This leads to no result set when executing queries in such files. 

I'm removing the check for now, and I've verified that everything works as expected in both ADS and VS Code MSSQL.

FYI: For more info why this check was added. 
https://github.com/microsoft/sqltoolsservice/pull/361

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Workspace/Workspace.cs
@@ -28,13 +28,6 @@ namespace Microsoft.SqlTools.ServiceLayer.Workspace
         #region Private Fields
 
         private const string UntitledScheme = "untitled";
-        private static readonly HashSet<string> fileUriSchemes = new HashSet<string>(StringComparer.OrdinalIgnoreCase) 
-        {
-            "file",
-            UntitledScheme,
-            "tsqloutput",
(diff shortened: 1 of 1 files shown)


caohai:
> no result set when executing queries in such files

Is there any exception thrown (in dev console) when this happens?

caohai (approved):

lewis-sanchez (approved):

Benjin:
@aasimkhan30 
>Previously, all connections were tied to files, which meant that even non-file-based connections had to go through the File.Open command. To handle this, a check was added to ignore unknown file URIs.  However, that's no longer the case.

When you say "that's no longer the case" in the PR description, it's no longer the case because VS Code has progressed beyond that (remote files, gists, etc.) model and we didn't update our side, or because we missed something when removing that restriction since Kevin added that code in STS?

PR #2520, opened 2025-10-07
moving grid summary and copy cells to sts

aasimkhan30:

## Description

1. Moving grid selection summary logic to backend to improve perf. 

Frontend summary:

https://github.com/user-attachments/assets/9dbeefd0-6645-4fce-8a1f-6386b1f8a8a1

Backend Summary:

https://github.com/user-attachments/assets/45de76c1-142b-4d0c-95a9-4e7dc5ff67fb

2. This PR also moves copy cells request with richer type support (text, json, csv) to backend to improve performance and avoid reading all data to frontend and then to clipboard

This api is named copy2 to avoid changing existing copy that is used in ads

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResults2Request.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/CopyResults2Request.cs
@@ -67,7 +67,8 @@ namespace Microsoft.SqlTools.ServiceLayer.QueryExecution.Contracts
     }
 
     /// <summary>
-    /// Copy Results Request
+    /// Copy Results Request. This API is named "copy2" to avoid conflict with the existing
+    /// "copy" API which is currently in used by Azure Data Studio.
     /// </summary>
     public class CopyResults2Request
     {


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -1134,60 +1152,43 @@ private bool TryGetNumericValue(object rawValue, out double numericValue)
...
-                if (rawValue is decimal m)
-                {
-                    numericValue = (double)m;
-                    return true;
+                    try
+                    {
+                        numericValue = convertible.ToDouble(CultureInfo.InvariantCulture);
+                        return true;
+                    }
+                    catch (Exception ex) when (ex is InvalidCastException || ex is FormatException || ex is OverflowException)
```
Benjin, line 1247: what happened to the previous concern about exception handling being slow?

aasimkhan30: Yeah, this will most likely not lead to an exception as we are type checking with the hash map. Doing an overall parse to double will be extremely slow as most string values will lead to exception, slowing the entire process down.

PR #2530, opened 2025-10-23
Edit Data Service Updates for Edit Data in the MSSQL for VS Code extension

lewis-sanchez:

## Description

This PR adds functionality to be able to generate the DML statements for pending updates that users have made with Edit Data but haven't saved to the database table. The PR also adds additional properties to result objects to get column information about column table names, and which table columns can be edited, and which cannot. Other changes include returning the original row when changes in a row or cell are reverted.

*Provide a clear, concise summary of the changes in this PR. What problem does it solve? Why is it needed? Link any related issues using [issue closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).*

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
@@ -527,6 +527,8 @@ namespace Microsoft.SqlTools.ServiceLayer.EditData
         {
             if (rowId >= NextRowId || rowId < 0)
             {
+                string errorMessage = $"Invalid row ID {rowId} provided. Valid range is 0 to {NextRowId - 1} (NextRowId: {NextRowId})";
+                Logger.Warning(errorMessage);
                 throw new ArgumentOutOfRangeException(nameof(rowId), SR.EditDataRowOutOfRange);
             }
         }


src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditColumnInfo.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class EditColumnInfo
+    {
+        /// <summary>
+        /// The name of the column
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Whether or not the column can be edited
```
Benjin: Can you update the comment to give an example or two of why a column might not be editable?  (e.g. computed column, idk what else)

src/Microsoft.SqlTools.ServiceLayer/EditData/Contracts/EditRevertRowRequest.cs
```diff
@@ -21,6 +21,10 @@ public class EditRevertRowParams : RowOperationParams
     /// </summary>
     public class EditRevertRowResult
     {
+        /// <summary>
+        /// The row after the revert was applied
```
Benjin: Is this the same as the original, unedited row?  If so, that might be good clarification.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -86,6 +86,7 @@ public void InitializeService(ServiceHost serviceHost)
             serviceHost.SetRequestHandler(EditSubsetRequest.Type, HandleSubsetRequest, true);
             serviceHost.SetRequestHandler(EditUpdateCellRequest.Type, HandleUpdateCellRequest, true);
             serviceHost.SetRequestHandler(EditCommitRequest.Type, HandleCommitRequest, true);
+            serviceHost.SetRequestHandler(EditScriptRequest.Type, HandleEditScriptRequest, true);
```
Benjin: While you're here, can you update all these SetRequestHandler() calls to name the magic constant?

e.g. `isParallelProcessingSupported: true`

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -100,10 +101,23 @@ internal async Task HandleSessionRequest<TResult>(SessionOperationParams session
...
+        internal async Task HandleCreateRowRequest(EditCreateRowParams createParams,
             RequestContext<EditCreateRowResult> requestContext)
         {
-            return HandleSessionRequest(createParams, requestContext, s => s.CreateRow());
+            try
+            {
+                EditSession editSession = GetActiveSessionOrThrow(createParams.OwnerUri);
+                EditCreateRowResult result = editSession.CreateRow();
+
+                EditRow[] rows = await editSession.GetRows(result.NewRowId, 1);
```
Benjin: name the magic constant

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
```diff
@@ -172,14 +186,23 @@ internal Task HandleRevertCellRequest(EditRevertCellParams revertParams,
...
-            });
+                EditSession editSession = GetActiveSessionOrThrow(revertParams.OwnerUri);
+                EditRow revertedRow = await editSession.RevertRow(revertParams.RowId);
+
+                await requestContext.SendResult(new EditRevertRowResult
+                {
+                    Row = revertedRow
+                });
+            }
+            catch (Exception e)
```
Benjin: Would it be worth upgrading HandleSessionRequest() to do the error handling as well?  Unsure if that would break other users of it, but then you could clean up some of the boilerplate in this PR.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -350,6 +366,12 @@ public EditRevertCellResult RevertCell(long rowId, int columnId)
         {
             ThrowIfNotInitialized();
 
+            // Sanity check the row ID
```
Benjin: Do you know how we might end up with this error?  Is there more logging or error handling that might need to be done?

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -366,16 +388,23 @@ public EditRevertCellResult RevertCell(long rowId, int columnId)
...
         /// </exception>
         /// <param name="rowId">The internal ID of the row to reset</param>
-        public void RevertRow(long rowId)
+        /// <returns>The row after the revert was applied</returns>
+        public async Task<EditRow> RevertRow(long rowId)
         {
             ThrowIfNotInitialized();
 
+            // Sanity check the row ID
+            if (rowId >= NextRowId || rowId < 0)
```
Benjin: duplicate sanity check logic; move to helper function.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -388,13 +417,38 @@ public void RevertRow(long rowId)
...
+        public string[] ScriptEdits()
+        {
+            ThrowIfNotInitialized();
+
+            var scripts = new List<string>();
+
+            List<RowEditBase> editOperations = EditCache.Values.ToList();
+            editOperations.Sort();
+
+            foreach (var rowEdit in editOperations)
```
Benjin: use a real type here instead of `var` since it isn't apparent from the same row what the type is.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
```diff
@@ -465,6 +519,20 @@ public EditUpdateCellResult UpdateCell(long rowId, int columnId, string newValue
...
+        /// Gets the column information including name and editability for the associated result set
+        /// </summary>
+        /// <returns>Array of column information</returns>
+        public EditColumnInfo[] GetColumnInfo()
+        {
+            ThrowIfNotInitialized();
+            return associatedResultSet?.Columns?.Select(c => new EditColumnInfo
+            {
+                Name = c.ColumnName,
+                IsEditable = c.IsUpdatable
```
Benjin, line 513: Out of curiosity, if the column already has an `isUpdatable` property, why rename to row's prop to `isEditable`?

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/ServiceIntegrationTests.cs
```diff
@@ -189,7 +189,12 @@ public async Task RevertRowSucceeds()
 
             // If: I ask to revert a row that has a pending edit
             var efv = new EventFlowValidator<EditRevertRowResult>()
-                .AddResultValidation(Assert.NotNull)
+                .AddResultValidation(result =>
+                {
+                    Assert.NotNull(result);
```
Benjin: Nit: use fluent-style assertions to make it easier to read and make it obvious which value is expected vs. actual:

e.g.: `Assert.That(something, Is.Not.True);` and `Assert.That(something, Is.EqualTo(exepectedVal))`

test/Microsoft.SqlTools.ServiceLayer.UnitTests/EditData/SessionTests.cs
```diff
@@ -1020,6 +1034,83 @@ public async Task ScriptProvidedOutputPath()
             }
         }
 
+        [Test]
+        public void ScriptEditsNoParamNotInitialized()
```
Benjin, line 1035: Do a quick pass of all these comments to make sure that we're only including the useful/descriptive ones.  We don't need to keep all the generated comments that just repeat the code (especially when it doubles the number of lines, lol)

src/Microsoft.SqlTools.ServiceLayer/EditData/EditDataService.cs
allancascante: can this be a lint rule?

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
lewis-sanchez: I honestly don't know how we would end up with this error in normal operation. Before I started making changes in [EditSession.cs](https://github.com/microsoft/sqltoolsservice/blob/357755cae2ede63c6630e7d3f852164b5324568d/src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs#L273), I noticed the sanity check being made in `DeleteRow`. I added it to `RevertCell`, `RevertRow`, and `UpdateCell` as a defensive safeguard because we get a better error message (EditDataRowOutOfRange) than the generic array index exception.

src/Microsoft.SqlTools.ServiceLayer/EditData/EditSession.cs
lewis-sanchez: `EditColumnInfo` is part of our client API contract. I renamed it to `IsEditable` because the client-side API is focused on edit operations (`EditSession`, `EditRow`, `EditCell`, etc.), so `IsEditable` maintains consistency. I also found `IsEditable` to be more intuitive while doing development in the extension because it clearly answers, "can I edit this?"

Benjin (approved):

PR #2534, opened 2025-11-04
Adding support for async script operation.

aasimkhan30:
## Description

This PR adds ReturnScriptAsEvent param to scripting operation, which makes it immediately returns an operation ID. This ID can be then used to listen for progress and completion events and also to cancel the scripting operation from the frontend. 

Making minimal changes in order to avoid breaking ads compatibility. 

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
@@ -189,19 +189,15 @@ namespace Microsoft.SqlTools.ServiceLayer.Scripting
         private async void SendScriptingCompleteEvent<TParams>(RequestContext<ScriptingResult> requestContext, EventType<TParams> eventType, TParams parameters,
                                                                SmoScriptingOperation operation, ScriptingParams scriptingParams)
         {
-            // If ReturnScriptAsEvent is enabled, include script in the complete event
+            // If ReturnScriptAsynchronously is enabled, include script in the complete event
             if (scriptingParams.ReturnScriptAsynchronously && parameters is ScriptingCompleteParams completeParams)
             {
                 completeParams.Script = operation.ScriptText;
+                await requestContext.SendEvent(eventType, parameters);
+                return;
             }
 
             await requestContext.SendEvent(eventType, parameters);
-
-            // If ReturnScriptAsEvent is enabled, we already sent the result with operation ID earlier
-            if (scriptingParams.ReturnScriptAsynchronously)
-            {
-                return;
-            }
 
             switch (scriptingParams.ScriptDestination)
             {


src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -137,9 +137,16 @@ public Task HandleScriptExecuteRequest(ScriptingParams parameters, RequestContex
 
             operation.PlanNotification += (sender, e) => requestContext.SendEvent(ScriptingPlanNotificationEvent.Type, e).Wait();
             operation.ProgressNotification += (sender, e) => requestContext.SendEvent(ScriptingProgressNotificationEvent.Type, e).Wait();
-            operation.CompleteNotification += (sender, e) => this.SendScriptingCompleteEvent(requestContext, ScriptingCompleteEvent.Type, e, operation, parameters.ScriptDestination);
+            operation.CompleteNotification += (sender, e) => this.SendScriptingCompleteEvent(requestContext, ScriptingCompleteEvent.Type, e, operation, parameters);
 
             RunTask(requestContext, operation);
+
+            // If ReturnScriptAsEvent is enabled, return operation ID immediately
+            if (parameters.ReturnScriptAsEvent)
```
Benjin: Is this an option just to preserve behavior in ADS?  Or do we have scenarios in MSSQL that require both sync and async scripting calls?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
```diff
@@ -180,10 +187,23 @@ public async Task HandleScriptCancelRequest(ScriptingCancelParams parameters, Re
...
+            }
+
             await requestContext.SendEvent(eventType, parameters);
-            switch (scriptDestination)
+
+            // If ReturnScriptAsEvent is enabled, we already sent the result with operation ID earlier
+            if (scriptingParams.ReturnScriptAsEvent)
+            {
+                return;
+            }
```
Benjin: May as well merge these blocks.

src/Microsoft.SqlTools.SqlCore/Scripting/Contracts/ScriptingParams.cs
```diff
@@ -77,5 +77,13 @@ public class ScriptingParams
         /// </summary>
         public ScriptingOperationType Operation { get; set; } = ScriptingOperationType.Create;
 
+        /// <summary>
+        /// Gets or sets whether to return the operation ID immediately and deliver the script via the complete event.
+        /// When true, the scripting/script request returns immediately with an operation ID,
+        /// and the script content is delivered later via the scripting/scriptComplete event.
+        /// When false (default), the request waits for completion and returns both operation ID and script together.
+        /// </summary>
+        public bool ReturnScriptAsEvent { get; set; }
```
Benjin: The name here feels more literal than descriptive.  Maybe `ReturnScriptAsynchronously`?

src/Microsoft.SqlTools.ServiceLayer/Scripting/ScriptingService.cs
aasimkhan30: Just to preserve existing behavior of this API in ADS.

Benjin (approved):

PR #2549, opened 2025-12-03
Don't default to AzureMFA

allancascante:
## Description

Auth should not default to Azure MFA, in particular when parsing a connection string using things like Authentication=Active Directory Service Principal 

## Code Changes Checklist


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1749,16 +1749,31 @@ await Task.Factory.StartNew(() => Task.Run(async () => {
         public ConnectionDetails ParseConnectionString(string connectionString)
         {
             SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connectionString);
+            bool isAzureMFA = builder.Authentication == SqlAuthenticationMethod.ActiveDirectoryInteractive; // Azure Multi-Factor Authentication
+            bool isSqlLogin = builder.Authentication == SqlAuthenticationMethod.SqlPassword;// SQL username/password
+            bool isIntegrated = builder.IntegratedSecurity; // Windows Integrated Security
+            string Authenticationtype = null;
```
Benjin: nit: casing (should be `authenticationType`)

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1749,16 +1749,31 @@ await Task.Factory.StartNew(() => Task.Run(async () => {
         public ConnectionDetails ParseConnectionString(string connectionString)
         {
             SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connectionString);
+            bool isAzureMFA = builder.Authentication == SqlAuthenticationMethod.ActiveDirectoryInteractive; // Azure Multi-Factor Authentication
+            bool isSqlLogin = builder.Authentication == SqlAuthenticationMethod.SqlPassword;// SQL username/password
+            bool isIntegrated = builder.IntegratedSecurity; // Windows Integrated Security
```
Benjin: may as well get rid of these variables and just put the conditions directly in the if-conditions below

PR #2561, opened 2026-01-07
Change on how system versioned tables are scripted as create

allancascante:
## Description

System Versioned tables must have a key before it could be created. With these changes system versioned (tables) will include the key as part of the create script being created. 

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
+++ b/src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
@@ -474,6 +474,22 @@ namespace Microsoft.SqlTools.SqlCore.Scripting
                     }
                 }
 
+                // For system-versioned (temporal) tables, use Table.Script() directly to ensure
+                // constraints are scripted inline with CREATE TABLE. This is required because temporal
+                // tables need the primary key to exist before SYSTEM_VERSIONING can be enabled.
+                // The Scripter.Script(urns) pipeline scripts indexes/constraints as separate ALTER
+                // statements which fails for temporal tables. (see GitHub issue azuredatastudio#20315)
+                if (urns.Count == 1 && this.Parameters.Operation == ScriptingOperationType.Create)
+                {
+                    SqlSmoObject smoObject = server.GetSmoObject(urns[0]);
+                    if (smoObject is Table table && table.IsSystemVersioned)
+                    {
+                        var resultVersioned = table.Script(options);
+                        resultScript = GetScript(options, resultVersioned);
+                        return resultScript;
+                    }
+                }
+
                 scripter.Options = options;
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
```diff
@@ -474,6 +474,22 @@ private string GenerateScriptAs(Server server, UrnCollection urns, ScriptingOpti
                     }
                 }
 
+                // For system-versioned (temporal) tables, use Table.Script() directly to ensure
+                // constraints are scripted inline with CREATE TABLE. This is required because temporal
+                // tables need the primary key to exist before SYSTEM_VERSIONING can be enabled.
+                // The Scripter.Script(urns) pipeline scripts indexes/constraints as separate ALTER
+                // statements which fails for temporal tables. (see GitHub issue azuredatastudio#20315)
+                if (urns.Count == 1 && this.Parameters.Operation == ScriptingOperationType.Create)
```
ssreerama, line 482: Are there cases where temporal tables might come with multiple URNs? here we are checking for count-1, what happens if we are scripting more than one temporal tables?

allancascante: For vs code extension, there is no way to script as create more than one table at the same time, at least from the existing UI, currently we could only select one table from a database a script as create that one table.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/Scripting/TemporalTableScriptingTests.cs
```diff
@@ -0,0 +1,217 @@
...
+                        ScriptOptions = new ScriptOptions
+                        {
+                            ScriptCreateDrop = "ScriptCreate",
+                        },
+                        ScriptingObjects = new List<ScriptingObject> { scriptingObject }
+                    };
+
+                    ScriptingService service = new ScriptingService();
+                    await service.HandleScriptExecuteRequest(scriptingParams, requestContext.Object);
+                    Thread.Sleep(2000);
```
ssreerama, line 183: Do we need this? seems it is added for debugging purpose.

src/Microsoft.SqlTools.SqlCore/Scripting/ScriptAsScriptingOperation.cs
ssreerama: What about the history table? When scripting a temporal table from the UI, does the selection automatically include the associated history table from the temporal folder structure, or does it only script the main table?

allancascante: It only script the main table

PR #2562, opened 2026-01-07
XElite integration

allancascante:
## Description
This PR integrates XELite's XELiveEventStreamer to provide push-based live event streaming for the SQL Server Profiler, replacing SMO ring-buffer polling with direct event delivery for improved performance and reduced latency.

**New Files**
LiveStreamXEventSession.cs - New XEvent session implementation that uses XELite's XELiveEventStreamer for push-based event delivery

- Implements IObservableXEventSession interface for reactive event streaming
- Includes LiveStreamObservable class with automatic reconnection on transient failures (configurable max attempts and exponential backoff with jitter)
- Handles event conversion from IXEvent to ProfilerEvent

**Modified Files**
IXEventSession.cs

- Added IObservableXEventSession interface extending IXEventSession with ObservableSessionEvents property
- Existing IXEventSession interface remains unchanged for backward compatibility

ProfilerService.cs

- Implemented OpenLiveStreamSession() method that:
 - Opens a SQL connection and validates the XEvent session exists
 - Creates a LiveStreamXEventSession with connection string and session name
 - Configures reconnection settings
- Added IsPaused property to PauseProfilingResult response

ProfilerSession.cs

- Added SessionObserver class implementing IObserver<ProfilerEvent> for push-based streaming
- Events buffered using lock-free atomic buffer swap pattern
- Added callback mechanism (onSessionActivity) that triggers immediate processing when events arrive
- Added Completed and Error properties for session lifecycle tracking

ProfilerSessionMonitor.cs

- Removed background polling thread (ProcessSessions loop)
- Replaced with callback-driven ProcessSession() that runs on-demand when events arrive
- Added session completion/error notification to listeners

IProfilerSessionMonitor.cs

- Updated PauseViewer method signature to return pause state via out bool isPaused parameter

Test Files

LiveStreamXEventSessionTests.cs - Added unit tests covering:

- Push-based event delivery
- Reconnection with exponential backoff
- Multiple observer support
- Session completion/error handling
- Edge cases and error conditions

ProfilerTestObjects.cs - Mock objects and test helpers for profiler testing
TestLiveEventFetcher - Simulates XELite streaming with configurable delays and failures
TestLiveStreamSessionFactory - Creates test sessions for integration tests
ProfilerServiceTests.cs - Updated with additional test coverage for pause state

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
@@ -15,19 +15,33 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
 {
     /// <summary>
     /// XEvent session that uses XELite's XELiveEventStreamer for push-based live event streaming.
+    /// 
+    /// This class combines two components:
+    /// - Session: The XEvent Session object used for session lifecycle management (start/stop the session on SQL Server)
+    /// - observableSession: The LiveStreamObservable that wraps XELite for push-based event streaming
+    /// 
+    /// The Session object is needed because XELite only handles event streaming from an existing session;
+    /// it does not manage session lifecycle on the server. The Session object provides the ability to
+    /// start and stop the XEvent session on SQL Server.
     /// </summary>
     public class LiveStreamXEventSession : IObservableXEventSession
     {
+        /// <summary>
+        /// Observable wrapper that uses XELite for push-based event delivery.
+        /// This handles the actual event streaming from SQL Server.
+        /// </summary>
         private readonly LiveStreamObservable observableSession;
         private readonly SessionId sessionId;
 
         /// <summary>
-        /// Gets or sets the underlying SMO XEvent Session object (optional, for session management).
+        /// Gets or sets the underlying XEvent Session object used for session lifecycle management.
+        /// This is used to start/stop the XEvent session on SQL Server.
+        /// XELite only streams events; it cannot manage session state.
         /// </summary>
         public Session Session { get; set; }
 
         /// <summary>
-        /// Gets or sets the SQL connection used for SMO operations.
+        /// Gets or sets the SQL connection used for XEvent session management.
         /// This connection is disposed when the session is stopped.
         /// </summary>
         public Microsoft.Data.SqlClient.SqlConnection SqlConnection { get; set; }
@@ -81,7 +95,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
         /// </summary>
         public void Start()
         {
-            // Ensure the underlying SMO session is running before starting the stream
+            // Ensure the underlying XEvent session is running before starting the stream
             if (Session != null && !Session.IsRunning)
             {
                 Session.Start();
@@ -103,6 +117,12 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
     /// <summary>
     /// Observable wrapper for XELiveEventStreamer that provides push-based event delivery
     /// with automatic reconnection on transient failures.
+    /// 
+    /// Note: This class implements its own reconnection logic rather than relying on ReliableConnection
+    /// because XELite manages its own internal SQL connection for event streaming. The reconnection
+    /// here handles XELite stream interruptions (network blips, server restarts, session timeouts)
+    /// which are distinct from the SQL connection used for session management. ReliableConnection
+    /// cannot intercept or retry XELite's internal connection failures.
     /// </summary>
     internal class LiveStreamObservable : IObservable<ProfilerEvent>
     {
@@ -182,6 +202,12 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
 
         private void StartStreamInternal()
         {
+            // Fire-and-forget: start the async streaming task without blocking
+            _ = StartStreamAsync();
+        }
+
+        private async Task StartStreamAsync()
+        {
             IXEventFetcher fetcher;
             try
             {
@@ -197,14 +223,20 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
             try
             {
                 var token = cancellationTokenSource.Token;
-                var streamTask = fetcher.ReadEventStream(OnEventReceived, token);
-                streamTask.ContinueWith(t => OnStreamEnded(t, fetcher), TaskScheduler.Default);
+                await fetcher.ReadEventStream(OnEventReceived, token);
+                // Stream completed normally
+                OnStreamEnded(completed: true, exception: null, fetcher: fetcher);
+            }
+            catch (OperationCanceledException)
+            {
+                // Stream was cancelled (expected during Stop)
+                (fetcher as IDisposable)?.Dispose();
+                OnStreamEnded(completed: true, exception: null, fetcher: null);
             }
             catch (Exception ex)
             {
-                // Handle synchronous exceptions from ReadEventStream
-                (fetcher as IDisposable)?.Dispose();
-                HandleFactoryException(ex);
+                // Stream failed with an error
+                OnStreamEnded(completed: false, exception: ex, fetcher: fetcher);
             }
         }
 
@@ -290,7 +322,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
             return profilerEvent;
         }
 
-        private void OnStreamEnded(Task fetcherTask, IXEventFetcher fetcher)
+        private void OnStreamEnded(bool completed, Exception exception, IXEventFetcher fetcher)
         {
             // Dispose the fetcher if it implements IDisposable
             (fetcher as IDisposable)?.Dispose();
@@ -309,17 +341,16 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
             }
 
             // Handle stream failure with reconnection
-            if (fetcherTask.IsFaulted)
+            if (exception != null)
             {
                 if (currentReconnectAttempt < maxReconnectAttempts)
                 {
                     ScheduleReconnect();
                     return;
                 }
-                
+
                 // Max retries exceeded - notify error and complete
-                NotifyError(fetcherTask.Exception?.GetBaseException() ?? 
-                    new Exception("XEvent stream failed after maximum reconnection attempts"));
+                NotifyError(exception);
             }
 
             NotifyCompleted();
@@ -335,14 +366,22 @@ namespace Microsoft.SqlTools.ServiceLayer.Profiler
                 (0.5 + Random.Shared.NextDouble() * 0.5));
 
             var token = cancellationTokenSource?.Token ?? CancellationToken.None;
-            
-            Task.Delay(delay, token).ContinueWith(t =>
+
+            // Fire-and-forget: schedule reconnection after delay
+            _ = ReconnectAfterDelayAsync(delay, token);
+        }
+
+        private async Task ReconnectAfterDelayAsync(TimeSpan delay, CancellationToken token)
+        {
+            try
             {
-                if (!t.IsCanceled)
-                {
-                    StartStreamInternal();
-                }
-            }, TaskScheduler.Default);
+                await Task.Delay(delay, token);
+                await StartStreamAsync();
+            }
+            catch (OperationCanceledException)
+            {
+                // Reconnection was cancelled - this is expected during Stop()
+            }
         }
 
         private List<IObserver<ProfilerEvent>> GetCurrentObservers()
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/Profiler/IXEventSessionFactory.cs
```diff
@@ -31,5 +31,14 @@ public interface IXEventSessionFactory
...
         IXEventSession OpenLocalFileSession(string filePath);
+
+        /// <summary>
+        /// Opens a live streaming session using XELite's XELiveEventStreamer for push-based
+        /// event delivery. This replaces SMO ring-buffer polling with direct event streaming.
+        /// </summary>
+        /// <param name="sessionName">Name of the XEvent session to stream from</param>
+        /// <param name="connInfo">Connection information for the SQL Server</param>
+        /// <returns>A live streaming XEvent session</returns>
+        IXEventSession OpenLiveStreamSession(string sessionName, ConnectionInfo connInfo);
```
kburtram: What’s the rationale for introducing a new method for “live stream” sessions on the session factory interface instead of using the existing CreateXEventSession method?

It’s not clear to me when a consumer of this interface would prefer one over the other. If the concern is supporting SMO and XELite sessions concurrently, would separate factories for each session type be a cleaner design?

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,366 @@
...
+using Microsoft.SqlServer.XEvent.XELite;
+using Microsoft.SqlTools.ServiceLayer.Profiler.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Profiler
+{
+    /// <summary>
+    /// XEvent session that uses XELite's XELiveEventStreamer for push-based live event streaming.
+    /// This replaces SMO ring-buffer polling with direct event delivery via XELite's event_stream target.
+    /// </summary>
+    public class LiveStreamXEventSession : XEventSession, IObservableXEventSession
```
kburtram: Should we just replace the SMO-based session implementation with the XELite one?  Any reason not to remove the old implementation when adding the new one?

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,366 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.SqlServer.XEvent.XELite;
+using Microsoft.SqlTools.ServiceLayer.Profiler.Contracts;
+
+namespace Microsoft.SqlTools.ServiceLayer.Profiler
+{
+    /// <summary>
+    /// XEvent session that uses XELite's XELiveEventStreamer for push-based live event streaming.
+    /// This replaces SMO ring-buffer polling with direct event delivery via XELite's event_stream target.
```
kburtram: nit: this comment is referencing SMO code that is likely being removed and will be out-of-date once that old code no longer exists.

ssreerama:
can you update the PR name, it is currently showing your branch name

src/Microsoft.SqlTools.ServiceLayer/Profiler/Contracts/ProfilerEventsAvailableNotification.cs
```diff
@@ -15,8 +15,6 @@ public class ProfilerEventsAvailableParams
         public string OwnerUri { get; set; }
 
         public List<ProfilerEvent> Events { get; set; }
-
-        public bool EventsLost { get; set; }
```
ssreerama, line 19: I believe this Boolean returns if the events have been lost or not, with this removal, how are we going to know if any event is lost or not. Does XElite events already have that info?

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,376 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#nullable disable
```
ssreerama: Just checking whether we plan to keep nullable disabled here or clean it up later? we may miss potential null‑reference issues in this file, if that is the intension here.

ssreerama (approved):

src/Microsoft.SqlTools.ServiceLayer/Profiler/Contracts/ProfilerEventsAvailableNotification.cs
allancascante: The original implementation used a ring buffer of 1,000 events. A flag was set to indicate when events were being dropped from the buffer. This was necessary because the system followed a pull‑based model: clients polled for new events every second. On very busy servers, it was possible to generate more than 1,000 events within a single second, causing excess events to be dropped and the flag to be set.
With the introduction of the XElite streamer, this model changes fundamentally. We no longer rely on a local buffer or polling. Instead, clients are notified using a push‑based approach, receiving events from a stream as they are produced.
Under this new model, even for very busy servers with a high event rate, events are delivered to clients immediately as they arrive. Rather than storing events in a local buffer and waiting for clients to pull them, the streamer continuously listens to the event stream and forwards events downstream in real time. As a result, the previous buffer‑overflow scenario and the associated flag are no longer applicable.

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
allancascante: I removed it and fix the code so this is not longer needed

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,411 @@
...
+    /// <summary>
+    /// XEvent session that uses XELite's XELiveEventStreamer for push-based live event streaming.
+    /// </summary>
+    public class LiveStreamXEventSession : IObservableXEventSession
+    {
+        private readonly LiveStreamObservable observableSession;
+        private readonly SessionId sessionId;
+
+        /// <summary>
+        /// Gets or sets the underlying SMO XEvent Session object (optional, for session management).
```
kburtram: Should we remove all the references to SMO, if we're using XElite?  Or is there still a SMO dependency here?

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,411 @@
...
+        /// </summary>
+        public Session Session { get; set; }
+
+        /// <summary>
+        /// Gets or sets the SQL connection used for SMO operations.
```
kburtram: I won't flag every reference to SMO, but please clean up comments globally if they aren't still valid.

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,411 @@
...
+                reconnectDelay);
+        }
+
+        /// <summary>
+        /// Starts the live event stream. The underlying XEvent session should already be running.
+        /// </summary>
+        public void Start()
+        {
+            // Ensure the underlying SMO session is running before starting the stream
+            if (Session != null && !Session.IsRunning)
```
kburtram, line 99: I'm unclear the difference between the Session and the observable session?  Is there a reason we want to have two session objects?

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,411 @@
...
+            {
+                // Handle synchronous exceptions from factory - no fetcher to dispose
+                HandleFactoryException(ex);
+                return;
+            }
+
+            try
+            {
+                var token = cancellationTokenSource.Token;
+                var streamTask = fetcher.ReadEventStream(OnEventReceived, token);
```
kburtram: Did you consider using await here instead of ContinueWith?

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
```diff
@@ -0,0 +1,411 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Creates a new LiveStreamXEventSession with a custom streamer factory (for testing).
+        /// </summary>
+        public LiveStreamXEventSession(
+            Func<IXEventFetcher> streamerFactory,
+            SessionId sessionId,
+            int maxReconnectAttempts = ProfilerConstants.DefaultMaxReconnectAttempts,
```
kburtram, line 83: I notice there’s quite a bit of reconnect logic in the profiling service. Is there a reason this isn’t handled by the connection components (e.g., ReliableConnection), so we don’t duplicate retry/backoff logic across services?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
```diff
@@ -176,97 +196,78 @@ private bool RemoveSession(SessionId sessionId, out ProfilerSession session)
...
-                        }                        
+                            else
+                            {
+                                emptyReadsInARow++;
+
+                                if (session.Completed)
+                                {
+                                    // Session completed - check one more time for any late events
+                                    // before signaling completion (to handle race conditions)
+                                    if (emptyReadsInARow >= 2)
```
kburtram, line 251: Is this how the API is documented?  That two empty reads in a row means that a XEvent session on the server has been stopped?

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerService.cs
```diff
@@ -326,30 +338,39 @@ private static BaseXEStore CreateXEventStore(ConnectionInfo connInfo, SqlStoreCo
...
+            // Create the live streaming session using XELite
+            var liveSession = new LiveStreamXEventSession(
+                connectionString,
+                sessionName,
+                new SessionId(session.ID.ToString()),
+                maxReconnectAttempts: ProfilerConstants.DefaultMaxReconnectAttempts,
+                reconnectDelay: ProfilerConstants.DefaultReconnectDelay);
+
+            // Set the SMO session for target XML retrieval
+            liveSession.Session = session;
```
kburtram, line 370: I thought we were planning to use XELite sessions rather than SMO sessions here, but I may be missing the design intent. Could you clarify?

allancascante: XELite's XELiveEventStreamer connects to an existing running session and streams events - it cannot create, start, or stop sessions. It only receives events via ReadEventStream().

src/Microsoft.SqlTools.ServiceLayer/Profiler/ProfilerSessionMonitor.cs
allancascante: I added a comment clarifying, this is just to make sure that no events are left in the buffer before cleaning all resources.

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
allancascante: I refactored the code to use await.

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
allancascante: Session: The XEvent Session object for session lifecycle management (start/stop/create)
observableSession: The LiveStreamObservable that wraps XELite for push-based event streaming

I added comments to point this out

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
allancascante: Added comments for this on the code, the reconnection here is for XELite stream interruptions (network blips, server restarts, session timeouts) which are distinct from the SQL connection used for session management. AFAIK ReliableConnection cannot intercept or retry XELite's internal connection failures.

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
allancascante: updated the comments

src/Microsoft.SqlTools.ServiceLayer/Profiler/LiveStreamXEventSession.cs
allancascante: update the comments

PR #2571, opened 2026-01-21
Adding flag to diff SC and publish deploy options, SC default behaior…

ssreerama:
… is to modify few options to non defaults but publish doesn'shouldn't have those changes

## Description
- STS prepares deployment options for schema compare by normalizing a small set of options (up to 7) and adjusting their defaults to support schema compare behavior. 
- Since the Publish dialog consumes the same underlying options list, those normalized values appear as non-defaults in the Publish UI.
- By introducing this flag, we now clearly differentiate option preparation for schema compare versus publish scenarios.
- With this change, we no longer need to denormalize options during SqlPackage command generation. That logic has been removed, as STS now supplies pure DacFx publish options, which are passed back to DacFx directly along with any user modifications.
- Note: we are not changing the existing behavior with this change, only publish call needs to send the true to get the clean options.

## Code Changes Checklist

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
@@ -4,7 +4,6 @@
 //
 
 #nullable disable
-using Microsoft.SqlTools.Hosting.Protocol;
 using Microsoft.SqlTools.ServiceLayer.DacFx;
 using Microsoft.SqlTools.ServiceLayer.DacFx.Contracts;
 using Microsoft.SqlTools.ServiceLayer.SchemaCompare;
@@ -13,7 +12,6 @@ using Microsoft.SqlTools.ServiceLayer.TaskServices;
 using Microsoft.SqlTools.ServiceLayer.Test.Common;
 using Microsoft.SqlTools.ServiceLayer.Test.Common.RequestContextMocking;
 using Microsoft.SqlServer.Dac;
-using Moq;
 using System;
 using System.IO;
 using System.Threading.Tasks;


src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -99,25 +99,38 @@ public class DeploymentOptions
...
-            options.DropPermissionsNotInSource = true;
-            options.DropRoleMembersNotInSource = true;
-            options.IgnoreKeywordCasing = false;
-            options.IgnoreSemicolonBetweenStatements = false;
+            // Only apply STS overrides when requested (for Schema Compare operations)
+            // For Publish operations, we want pure DacFx native defaults
+            if (applySTSOverrides)
+            {
+                // Adding these defaults to ensure behavior similarity with other tools. Dacfx and SSMS import/export wizards use these defaults.
+                // Tracking the full fix : https://github.com/microsoft/azuredatastudio/issues/5599
```
Benjin: Is this still relevant?  Maybe we should transfer the bug to the MSSQL repo.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
```diff
@@ -99,25 +99,38 @@ public class DeploymentOptions
         public Dictionary<string, string> ObjectTypesDictionary = new Dictionary<string, string>(StringComparer.InvariantCultureIgnoreCase);
         #endregion
 
-        public DeploymentOptions()
+        /// <summary>
+        /// Creates DeploymentOptions with either STS defaults or DacFx native defaults.
+        /// </summary>
+        /// <param name="applySTSOverrides">When true, applies 7 STS-specific overrides for Schema Compare. When false, uses pure DacFx defaults for Publish operations.</param>
+        internal DeploymentOptions(bool applySTSOverrides)
```
Benjin: I think we can/should improve the naming and docstrings here to make it more clear what's happening and why:

Publish and Schema Compare should use different defaults:
* schema compare should use modified defaults
* publish should use dacfx's original defaults

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptionsRequest.cs
```diff
@@ -29,6 +34,8 @@ public class SchemaCompareOptionsResult : ResultStatus
 
     /// <summary>
     /// Defines the Schema Compare request type
+    /// TODO: Consider renaming to a more generic name (e.g., "dacfx/getDefaultDeploymentOptions")
+    /// since this endpoint is now used by both Schema Compare and Publish operations.
```
Benjin: Since we're holding this change off for February, we can clean this up with the rename now.  I think we should use an enum named `scenario` instead of a boolean, with the values being deployment (default) and schemaCompare.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Seems this is fixed, the issue was closed with a comment, will clean up the comment

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: As we discussed, in the teams, will update the namings and make them generic method for both SC and publish.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/DeploymentOptions.cs
ssreerama: Seems the short-term description (https://github.com/microsoft/azuredatastudio/issues/5599) is misleading here, DacFx doesn't have different defaults for the schema compare, and not sure how the SSMS have them. Do we still need to depend on these non-defaults.

Tagging @kisantia @dzsquared if have any idea about these. To give some context 7 deployment options are modified to non-defaults in STS for schema compare usage in ADS/Vscode

kisantia: This was a while ago so I don't remember, but you can take a look at the default values SSMS sets in DeployModel.cs in the SSMS repo and compare those to the defaults of DacDeployOptions

ssreerama: Thank you @kisantia , I see that these property defaults differ in SSMS.

@llali , based on the issue discussion, the proposed long-term fix is to align these defaults within DacFx. I’d like to get your thoughts on this approach. If DacFx updates the defaults, we’ll have consistent behavior across all surfaces. Otherwise, SSMS and VS Code schema compare would align, while DacFx schema compare would remain different.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareOptionsRequest.cs
ssreerama: done, made the request generic with enum as deployment to default. 

moved the generic request to dacfx service and by providing the scenario enum as param.

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GetDeploymentOptionsRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Parameters for getting deployment options based on scenario
+    /// </summary>
+    public class GetDeploymentOptionsParams
+    {
+        /// <summary>
+        /// Specifies the scenario for which to retrieve default deployment options.
+        /// Deployment (default): Returns DacFx native defaults (for Publish operations).
+        /// SchemaCompare: Returns modified defaults that match SSMS behavior.
```
Benjin: I think we should keep the mentions of SSMS to a minimum in STS. `DeploymentOptions.cs` is fine, but I think repeating it risks confusing others because that's a bit of a implementation nuance; the important thing is whether we're getting defaults for SchemaCompare or Deployment

src/Microsoft.SqlTools.ServiceLayer/DacFx/DacFxService.cs
```diff
@@ -281,6 +282,34 @@ await requestContext.SendResult(new DacFxOptionsResult()
...
         }
 
+        /// <summary>
+        /// Gets deployment options based on the specified scenario
+        /// </summary>
+        public async Task HandleGetDeploymentOptionsRequest(GetDeploymentOptionsParams parameters, RequestContext<GetDeploymentOptionsResult> requestContext)
+        {
+            try
+            {
+                // this does not need to be an async operation since this only creates and returns the default object
```
Benjin: This comment doesn't really add anything useful; let's remove it.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -799,6 +799,47 @@ public async Task ValidateGetDefaultPublishOptionsCallFromService()
...
+                .Returns(Task.FromResult(new object()));
+
+            await service.HandleGetDeploymentOptionsRequest(deploymentParams, deploymentRequestContext.Object);
+
+            Assert.That(deploymentResult, Is.Not.Null, "Deployment result should not be null");
+            Assert.That(deploymentResult.Success, Is.True, "Deployment request should succeed");
+            Assert.That(deploymentResult.DefaultDeploymentOptions.BooleanOptionsDictionary[nameof(DacDeployOptions.AllowDropBlockingAssemblies)].Value, 
+                Is.False, "AllowDropBlockingAssemblies should be false for Deployment (DacFx native default)");
+
+            // Test Schema Compare scenario - should return SSMS-matching modified defaults
```
Benjin: Mentioning SSMS here is likely confusing.  Just Schema Compare is the important part.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
```diff
@@ -799,6 +799,47 @@ public async Task ValidateGetDefaultPublishOptionsCallFromService()
             await service.HandleGetDefaultPublishOptionsRequest(p, dacfxRequestContext.Object);
         }
 
+        /// <summary>
+        /// Verify the deployment options endpoint with different scenarios
+        /// </summary>
+        [Test]
+        public async Task ValidateGetDeploymentOptionsWithScenarios()
```
Benjin, line 806: Since GetDeploymentOptionsResult extends ResultStatus, can you use `MockResult<GetDeploymentOptionsRequest>` and `myMock.AssertSuccess()` to make this cleaner?

See `SqlProjectsServiceTests.cs` for examples

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
```diff
@@ -400,64 +401,66 @@ public void ValidateSchemaCompareOptionsDefaultAgainstDacFx()
         public async Task ValidateSchemaCompareGetDefaultOptionsCallFromService()
         {
             DeploymentOptions deployOptions = new DeploymentOptions();
-            var schemaCompareRequestContext = new Mock<RequestContext<SchemaCompareOptionsResult>>();
-            schemaCompareRequestContext.Setup(x => x.SendResult(It.IsAny<SchemaCompareOptionsResult>())).Returns(Task.FromResult(new object()));
-            schemaCompareRequestContext.Setup((RequestContext<SchemaCompareOptionsResult> x) => x.SendResult(It.Is<SchemaCompareOptionsResult>((options) => SchemaCompareTestUtils.ValidateOptionsEqualsDefault(options) == true))).Returns(Task.FromResult(new object()));
+            var schemaCompareRequestContext = new Mock<RequestContext<GetDeploymentOptionsResult>>();
```
Benjin: Same here about MockRequest

src/Microsoft.SqlTools.ServiceLayer/DacFx/Contracts/GetDeploymentOptionsRequest.cs
ssreerama: Agreed! minimized the usage of ssms word mentioning :)

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/DacFx/DacFxServiceTests.cs
ssreerama: cleaner now after using the MockRequest and assertsuccess.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SchemaCompare/SchemaCompareServiceOptionsTests.cs
ssreerama: done!

Benjin (approved):

PR #2588, opened 2026-02-19
Remove URI escaping from workspace service. 

aasimkhan30:
## Description

Follow up to  PR in vscode-mssql, remove escaping from file URI. https://github.com/microsoft/vscode-mssql/pull/21229

Unfortunately, we were still doing that over here which broke query execution as the execution service was not able to find a URI match to get the SQL text. 

## Code Changes Checklist

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -56,7 +56,7 @@
     <PackageReference Update="SkiaSharp" Version="2.88.6" />
     <PackageReference Update="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="2.88.6" Condition="$([MSBuild]::IsOsPlatform('Linux'))" />
     <PackageReference Update="TextCopy" Version="6.2.1" />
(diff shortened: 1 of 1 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,9 @@
+{
```
caohai: unrelated change?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -1582,20 +1582,16 @@ private async Task SaveResultsHelper(SaveResultsRequestParams saveParams,
         // Internal for testing purposes
         internal string GetSqlText(ExecuteRequestParamsBase request)
         {
-            // This URI doesn't come in escaped - so if it's a file path with reserved characters (such as %)
```
caohai, line 1585: Does file path with % work with the new change?

caohai (approved):

PR #2595, opened 2026-02-28
Enable parallel result subset fetches

aasimkhan30:
## Description

Adds parallel fetches with query results subsets with serialized based on owner uri to make sure we do not send responses out of sequence. 

Might help: https://github.com/microsoft/vscode-mssql/issues/21361

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Localization/sr.cs
@@ -223,6 +223,14 @@ namespace Microsoft.SqlTools.ServiceLayer
             }
         }
 
+        public static string QueryServiceSubsetRowsTimeout
+        {
+            get
+            {
+                return Keys.GetString(Keys.QueryServiceSubsetRowsTimeout);
+            }
+        }
+
         public static string QueryServiceDataReaderByteCountInvalid
         {
             get
@@ -11100,6 +11108,9 @@ namespace Microsoft.SqlTools.ServiceLayer
 
 
             public const string QueryServiceSubsetResultSetOutOfRange = "QueryServiceSubsetResultSetOutOfRange";
+
+
+            public const string QueryServiceSubsetRowsTimeout = "QueryServiceSubsetRowsTimeout";
 
 
             public const string QueryServiceDataReaderByteCountInvalid = "QueryServiceDataReaderByteCountInvalid";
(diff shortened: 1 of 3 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
```diff
@@ -487,13 +494,36 @@ internal Task HandleConnectionUriChangedNotification(ConnectionUriChangedParams
...
+            using (await subsetLock.LockAsync())
             {
-                ResultSubset = subset
-            };
-            await requestContext.SendResult(result);
-            Logger.Stop($"Done Handler for Subset request with for Query:'{subsetParams.OwnerUri}', Batch:'{subsetParams.BatchIndex}', ResultSetIndex:'{subsetParams.ResultSetIndex}', RowsStartIndex'{subsetParams.RowsStartIndex}', Requested RowsCount:'{subsetParams.RowsCount}'\r\n\t\t with subset response of:[ RowCount:'{subset.RowCount}', Rows array of length:'{subset.Rows.Length}']");
+                using (CancellationTokenSource timeoutCancellationSource = new CancellationTokenSource())
+                using (CancellationTokenSource delayCancellationSource = new CancellationTokenSource())
+                {
+                    Task<ResultSetSubset> subsetTask = InterServiceResultSubset(subsetParams, timeoutCancellationSource.Token);
```
kburtram, line 503: if the task returns faulted, do we properly detect and log the error?  I guess it throws somewhere and goes the higher level exception handler in the dispatcher queue?

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/ResultSet.cs
```diff
@@ -294,17 +307,28 @@ public Task<ResultSetSubset> GetSubset(long startRow, int rowCount)
                         IEnumerable<long> rowOffsets = fileOffsets.LongSkip(startRow).Take(rowCount);
 
                         // Iterate over the rows we need and process them into output
-                        // ReSharper disable once AccessToDisposedClosure   The lambda is used immediately in .ToArray call
-                        rows = rowOffsets.Select((offset, id) => fileStreamReader.ReadRow(offset, id, Columns).ToArray()).ToArray();
+                        List<DbCellValue[]> subsetRows = new List<DbCellValue[]>();
+                        int rowId = 0;
```
kburtram, line 311: this shouldn't be `startRow`?

aasimkhan30: This is index from offset. So I think we are correctly starting from 0

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs
aasimkhan30: yes, it should surface through the dispatcher queue.

PR #2607, opened 2026-03-09
Adding Set project properties handler for .sqlproj updates

ssreerama:
## Description

This pull request:
- Introduces a new request that allows setting multiple generic properties in a `.sqlproj` file through a dedicated request handler. 
- The implementation includes special handling for properties that are init-only in DacFx, such as `ProjectGuid`, by directly editing the project XML. 
- Comprehensive integration tests have been added to verify the correct behavior for both standard and special-case properties.

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
@@ -26,6 +26,7 @@ namespace Microsoft.SqlTools.ServiceLayer.SqlProjects
         private static readonly Lazy<SqlProjectsService> instance = new Lazy<SqlProjectsService>(() => new SqlProjectsService());
         private const string RunSqlCodeAnalysisPropertyName = "RunSqlCodeAnalysis";
         private const string SqlCodeAnalysisRulesPropertyName = "SqlCodeAnalysisRules";
+        private const string ProjectGuidPropertyName = "ProjectGuid";
 
         /// <summary>
         /// Gets the singleton instance object
@@ -194,19 +195,25 @@ namespace Microsoft.SqlTools.ServiceLayer.SqlProjects
             {
                 SqlProject project = GetProject(requestParams.ProjectUri, onlyLoadProperties: true);
 
+                // First pass: apply all DacFx-managed properties so they are fully flushed
+                // to disk before any raw XML edits are made against the same file.
                 foreach (KeyValuePair<string, string> entry in requestParams.Properties)
                 {
-                    if (string.Equals(entry.Key, "ProjectGuid", StringComparison.OrdinalIgnoreCase))
-                    {
-                        // DacFx exposes ProjectGuid as an init-only field with no writable API,
-                        // so fall back to editing the .sqlproj XML directly. Evict the cached
-                        // project afterward so the updated value is picked up on the next load.
-                        SetReadOnlyPropertyInXml(requestParams.ProjectUri, entry.Key, entry.Value);
-                        Projects.TryRemove(requestParams.ProjectUri, out _);
-                    }
-                    else
+                    if (!string.Equals(entry.Key, ProjectGuidPropertyName, StringComparison.OrdinalIgnoreCase))
                     {
                         project.Properties.SetProperty(entry.Key, entry.Value);
+                    }
+                }
+
+                // Second pass: apply XML-only properties (e.g. ProjectGuid) now that DacFx
+                // has finished writing, then evict the cached project so the next load picks
+                // up the updated file.
+                foreach (KeyValuePair<string, string> entry in requestParams.Properties)
+                {
+                    if (string.Equals(entry.Key, ProjectGuidPropertyName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        SetReadOnlyPropertyInXml(requestParams.ProjectUri, entry.Key, entry.Value);
+                        Projects.TryRemove(requestParams.ProjectUri, out _);
                     }
                 }
             }, requestContext);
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
```diff
@@ -186,6 +188,30 @@ internal async Task HandleSetDatabaseSchemaProviderRequest(SetDatabaseSchemaProv
...
+
+                foreach (KeyValuePair<string, string> entry in requestParams.Properties)
+                {
+                    if (string.Equals(entry.Key, "ProjectGuid", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // DacFx exposes ProjectGuid as an init-only field with no writable API,
+                        // so fall back to editing the .sqlproj XML directly. Evict the cached
+                        // project afterward so the updated value is picked up on the next load.
+                        SetReadOnlyPropertyInXml(requestParams.ProjectUri, entry.Key, entry.Value);
+                        Projects.TryRemove(requestParams.ProjectUri, out _);
```
lewis-sanchez: What happens if a projectUri is removed first and then in the next loop iteration the project properties are updated?

Maybe we should update all properties first before any `TryRemove` method calls are made to be on the safe side?

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
```diff
@@ -1270,6 +1271,93 @@ public async Task TestLoadOnlyProperties()
...
+            Assert.AreEqual("TestSource", project.Properties.DatabaseSource, "DatabaseSource should be unaffected by updating Description");
+        }
+
+        /// <summary>
+        /// Verifies the ProjectGuid special-case in HandleSetProjectPropertiesRequest:
+        /// because DacFx exposes ProjectGuid as an init-only field, the handler writes it
+        /// directly to the .sqlproj XML via SetReadOnlyPropertyInXml. This test confirms
+        /// the value round-trips correctly through a set → evict → reload sequence.
+        /// </summary>
+        [TestCase(ProjectType.LegacyStyle)]
```
lewis-sanchez, line 1353: You're missing the test attribute: `[Test]`

Above this line.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
```diff
@@ -1270,6 +1271,93 @@ public async Task TestLoadOnlyProperties()
             Assert.IsFalse(service.Projects[projectUri].OnlyPropertiesLoaded, "Project should be fully-loaded when initially opened for a list of files");
         }
 
+        [Test]
+        [TestCase(ProjectType.LegacyStyle)]
+        [TestCase(ProjectType.SdkStyle)]
+        public async Task TestSetProjectProperties(ProjectType projectType)
```
lewis-sanchez: Break this test down into individual unit tests. If any random case in this test fails, figuring out where the test failed is going to be difficult.

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
```diff
@@ -186,6 +188,30 @@ internal async Task HandleSetDatabaseSchemaProviderRequest(SetDatabaseSchemaProv
...
+                    }
+                    else
+                    {
+                        project.Properties.SetProperty(entry.Key, entry.Value);
```
lewis-sanchez: Can `Properties` be null by any chance? This might throw a `NullReferenceException` when `SetProperty` is called and `Properties` is null.

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
```diff
@@ -186,6 +188,30 @@ internal async Task HandleSetDatabaseSchemaProviderRequest(SetDatabaseSchemaProv
...
 
+        internal async Task HandleSetProjectPropertiesRequest(SetProjectPropertiesParams requestParams, RequestContext<ResultStatus> requestContext)
+        {
+            await RunWithErrorHandling(() =>
+            {
+                SqlProject project = GetProject(requestParams.ProjectUri, onlyLoadProperties: true);
+
+                foreach (KeyValuePair<string, string> entry in requestParams.Properties)
+                {
+                    if (string.Equals(entry.Key, "ProjectGuid", StringComparison.OrdinalIgnoreCase))
```
lewis-sanchez: Make `the "ProjectGuid"` literal a const.

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
ssreerama: No, SqlProject.Properties cannot be null in practice here, As GetProject() calls SqlProject.OpenProject, which throws if the project can't be loaded, it never returns a partially-constructed object with a null Properties, every other calls access proeprties without a guard and none of them have been an issue.

test/Microsoft.SqlTools.ServiceLayer.IntegrationTests/SqlProjects/SqlProjectsServiceTests.cs
ssreerama: done

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
ssreerama: there could be chance of it if DacFx doesn't update the sqlproj directly and flushes out the old properties data. I've modified the pattern to update the properties first and then doing edit xml at last. Thanks for pointing this out.

PR #2623, opened 2026-03-17
fix: strip XML-illegal control characters from Excel cell values (#17853)

kburtram:
Fixes https://github.com/microsoft/vscode-mssql/issues/17853: Control characters forbidden by XML 1.0 (U+0000–U+0008, U+000B–U+000C, U+000E–U+001F, U+FFFE–U+FFFF) in SQL result data caused the XmlWriter to enter an unrecoverable error state, failing the entire export. Add SanitizeXmlString() to both SqlTools Excel stream writer helpers and call it on every string cell value before writing.

### DIFF
--- a/test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelperTests.cs
+++ b/test/Microsoft.SqlTools.ServiceLayer.UnitTests/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelperTests.cs
@@ -495,7 +495,7 @@ namespace Microsoft.SqlTools.ServiceLayer.UnitTests.QueryExecution.DataStorage
 
         [TestCase("hello\x00world", "helloworld")]
         [TestCase("\x01\x02\x03", "")]
-        [TestCase("ok\x00\x01\x14end", "okend")]
+        [TestCase("ok\u0001\u0014end", "okend")]
         [TestCase("no control chars", "no control chars")]
         [TestCase("tab\there", "tab\there")]    // tab (\x09) is legal and must be preserved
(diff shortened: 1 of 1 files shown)


aasimkhan30 (dismissed):

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
```diff
@@ -448,6 +449,30 @@ private void EndRowIfNeeded()
...
+                    char c = input[i];
+                    bool legal = c == '\t' || c == '\n' || c == '\r' ||
+                                 (c >= '\x20' && c <= '\xD7FF') ||
+                                 (c >= '\xE000' && c <= '\xFFFD');
+                    if (!legal && sb == null)
+                    {
+                        sb = new StringBuilder(input.Length);
+                        sb.Append(input, 0, i);
+                    }
+                    if (legal) sb?.Append(c);
```
lewis-sanchez, line 472: Nit: Could make it immediately obvious that every char is either appended or triggers lazy init by restructuring this `if` and the one above this way:

```
if (legal)
{
    sb?.Append(c);
}
else if (sb == null)
{
    sb = new StringBuilder(input.Length);
    sb.Append(input, 0, i);
}
```

lewis-sanchez (dismissed):

src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsExcelFileStreamWriterHelper.cs
kburtram: This looks basically the same to me, so I'll leave as-is to avoid retesting.

lewis-sanchez (approved):

PR #2671, opened 2026-04-23
Restructuring db management tasks for better frontend handling.

aasimkhan30:
## Description

Adding task id in all db operations so that we can link them with sql tasks view items. 

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
@@ -15,7 +15,6 @@ using Microsoft.SqlTools.ServiceLayer.Management;
 using System.Linq;
 using Microsoft.SqlTools.ServiceLayer.ObjectManagement.PermissionsData;
 using Microsoft.SqlTools.ServiceLayer.TaskServices;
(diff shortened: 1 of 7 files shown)


src/Microsoft.SqlTools.ServiceLayer/ObjectManagement/ObjectManagementService.cs
```diff
@@ -115,11 +114,13 @@ await requestContext.SendResult(new RenameDatabaseResponse
...
             };
-            SqlTask sqlTask = SqlTaskManager.Instance.CreateAndRun<SqlTask>(metadata);
+            SqlTask sqlTask = SqlTaskManager.Instance.CreateTask<SqlTask>(metadata);
+            TaskExecutionResult taskResult = await RunTaskAsync(sqlTask);
 
             await requestContext.SendResult(new RenameDatabaseResponse
             {
-                TaskId = sqlTask.TaskId.ToString(),
+                TaskId = taskResult.TaskId,
+                ErrorMessage = taskResult.ErrorMessage,
```
Benjin, line 123: What's the general rule for when we should call `SendResult()` with an `ErrorMessage` property versus calling `SendError()`?

Instead, should SendError() be updated to include a TaskID?

Benjin (approved):

PR #2672, opened 2026-04-24
PR2 - Add Go-to-Definition/Peek/Hover/Completion support for SQL Project files

ssreerama:
## Description

This PR Introduces features like:
- IntelliSense (completions, hover, signature help) for SQL Project files (offline, no server connection required)
- Go-to-Definition and Peek Definition for objects defined in SQL Project files
- Schema-qualified name resolution (e.g. dbo.Orders) for all of the above

Our Approach in this PR:
- Added AddProjectContext() to ConnectedBindingQueue — registers an offline binding context using the LazySchemaModelMetadataProvider (built in a prior PR) instead of a live server connection
- Added InitializeProjectFileContexts() to LanguageService — called on project open; wires each project file's ScriptParseInfo to its project binding context
- Added QueueProjectTask() to LanguageService — handles F12/Peek for project files by matching resolved declarations against _sourceLocations index and returning file-path locations
- Extended SqlProjectsService to call InitializeProjectFileContexts() on project load, providing the TSqlModel-backed binder and metadata provider
- Fixed LazyCollection<T> name indexer to return null on miss instead of throwing — required for schema-qualified resolution to work end-to-end
- Fixed BuildSchemas() to infer implicit schemas (e.g. dbo) from object name parts — DacFx does not enumerate schemas without an explicit CREATE SCHEMA statement
- Added ScriptParseInfo.ProjectDatabaseName — used as the database name when binding project files (no connection = no database name from connection details)

## Code Changes Checklist

## Reviewers: [Please read our reviewer guidelines](https://github.com/microsoft/sqltoolsservice/blob/main/.github/REVIEW_GUIDELINES.md)

Go-to-Definition/Peek/Hover/Completion:
Gotodef_hover_pee_completions


.github/intellisense-projects.instructions.md
```diff
@@ -0,0 +1,585 @@
+# SQL Tools Service — Project IntelliSense Integration Guide
```
aasimkhan30, line 1: Why do we need this? The agent should be capable enough to understand the code structure on its own. It also becomes a maintenance burden to keep this document updated. A lot of the content here is hardcoded and will likely become outdated very quickly.

ssreerama: I'll not include this in the final checkout(will remove it from the PR), I added it to make sure copilot is not deviating from the previous investigation and remember the complete process. I had many issues with copilot where is jumps into assumptions and changing the code unnecessarily. It helps me to work on a focused path.

src/Microsoft.SqlTools.SqlCore/IntelliSense/Lazy/TSqlModelServer.cs
```diff
@@ -113,6 +113,19 @@ private IMetadataCollection<ISchema> BuildSchemas()
                     schemas.Add(new LazyModelSchema(this, _model, schemaName, DacQueryScopes.BuiltIn));
             }
```
aasimkhan30, line 133: Can we name the file TSqlModelServer or ProjectModelServer? 

In a similar way can we name other objects like

TSqlModelDatabase
TSqlModelSchema
TSqlModelTable.....

I prefer TsqlModel* because we are wrapping up the TsqlModel and it can be used outside of projects as well.

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
```diff
@@ -153,18 +156,50 @@ private async Task BuildProjectIntelliSenseAsync(string projectUri)
             try
             {
                 SqlProject project = GetProject(projectUri);
+
                 string databaseName = Path.GetFileNameWithoutExtension(projectUri);
+                string contextKey = $"project_{projectUri}";
+                string projectDir = Path.GetDirectoryName(projectUri) ?? string.Empty;
```
aasimkhan30: Potential NIT. Should we throw when we don't get ProjectDir?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
```diff
@@ -46,7 +46,7 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
         /// Map from context keys to binding context instances
         /// Internal for testing purposes only
         /// </summary>
-        internal ConcurrentDictionary<string, IBindingContext> BindingContextMap { get; set; }
+        public ConcurrentDictionary<string, IBindingContext> BindingContextMap { get; set; }
```
aasimkhan30, line 50: are you sure this change is needed? We already have InternalsVisibleTo for tests

src/Microsoft.SqlTools.SqlCore/IntelliSense/Lazy/TSqlModelServer.cs
aasimkhan30, line 1: While looking at the code, there seems to be hardcoded defaults like "dbo". is this intended or will be fixed in follow up PRs?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
```diff
@@ -175,8 +175,8 @@ private AutoCompletionResult CreateCompletionsFromSqlParser(
             }
 
             result.CompleteResult(completionList);
-            //The bucket for number of milliseconds will take to send back auto complete list
-            connInfo.IntellisenseMetrics.UpdateMetrics(result.Duration, 1, (k2, v2) => v2 + 1);
+            // connInfo is null for project files (offline IntelliSense — no server connection).
+            connInfo?.IntellisenseMetrics?.UpdateMetrics(result.Duration, 1, (k2, v2) => v2 + 1);
```
aasimkhan30: Are we overloading the ScriptParseInfo.IsConnected to also mean that it has a TsqlModel now? 

We should improve the overall architecture to make it obvious

otherwise, it doesn't make sense to do add null checks

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
```diff
@@ -153,18 +156,50 @@ private async Task BuildProjectIntelliSenseAsync(string projectUri)
...
+                
+                var fileUriList = allScripts
+                    .Select(path => new Uri(Path.IsPathRooted(path)
+                        ? path
+                        : Path.Combine(projectDir, path)).AbsoluteUri)
+                    .ToList();
+
+                // Stamp files before the model finishes loading so that F12 requests arriving
+                // during the load are queued on the binding context rather than rejected.
+                LanguageService.Instance.InitializeProjectFileContexts(fileUriList, contextKey, databaseName);
```
aasimkhan30: Should we clean this up on project close? It seems we only dispose the model.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
```diff
@@ -171,6 +179,10 @@ public ParseOptions ParseOptions
         {
             get
             {
+                if (this.OverrideParseOptions != null)
```
aasimkhan30: Can we make parse options part of context construction instead of an override? A project binding context should simply have its own options; it should not need to override connection-derived behavior.

aasimkhan30: Could we add a separate concept, e.g. `BindingContextKey`, `HasBindingContext`, or `BindingContextKind`, and reserve `IsConnected` for real server connections? Otherwise every later caller has to remember that `IsConnected == true` might still mean `connInfo == null` and `ServerConnection == null`.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1280,6 +1360,14 @@ private bool ShouldSkipIntellisense(string uri)
                 || ShouldSkipNonMssqlFile(uri);
         }
 
+        /// <summary>
+        /// Returns true when the ConnectionKey identifies a SQL project context (offline, TSqlModel-backed).
+        /// Use this for routing decisions instead of checking connInfo == null.
+        /// </summary>
+        private static bool IsProjectContext(string connectionKey)
+            => !string.IsNullOrEmpty(connectionKey) &&
+               connectionKey.StartsWith("project_", StringComparison.Ordinal);
```
aasimkhan30, line 1401: Maybe we can add better props to ScriptParseInfo instead of overloading existing props and using a prefix to route the flow? 

Something like 
```
internal enum BindingContextKind
{
    None,
    LiveConnection,
    Project
}

public BindingContextKind BindingContextKind { get; set; }
```

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
ssreerama: Correct, we should throw exception when we get a malformed project uri

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/BindingQueue.cs
ssreerama: public was unnecessary. Changing to internal.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/Completion/CompletionService.cs
ssreerama: Good catch. Fixed, IsConnected now strictly means "live SMO connection". Added IsProjectContext (set only for offline project files) with a clear doc comment.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
ssreerama: Done! I agree with you, using the override parse options could cause confusion. Now, any caller(today or in the future) can trust that IsConnected means exactly "there is a live SMO server connection". Project contexts are unambiguously identified by IsProjectContext and will never accidentally trigger SMO queries.

src/Microsoft.SqlTools.ServiceLayer/SqlProjects/SqlProjectsService.cs
ssreerama: Since the metadataprovide is not Idisposable we didn't do it, but now im cleaning up all (bindingcontext, map, metadaprovider, model) on project close.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/ConnectedBindingContext.cs
ssreerama: Done

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
ssreerama: Good one, I'll do this in a followup PR as it touched the existing prop and needs little more file updates.

aasimkhan30 (approved):

PR #2681, opened 2026-05-06
Allowing nuspecs to be packaged when targeting preview nugets

Benjin:
## Description

Fixes an issue where nuspecs would fail to pack if any dependencies had a preview flag.  STS regularly consumes preview versions of DacFx and SqlProject nuget packages, so this should have set to not error originally, but wasn't surfaced until recently.

## Code Changes Checklist


aasimkhan30 (approved):

azure-pipelines/build.yml
```diff
@@ -76,30 +76,13 @@ steps:
...
-    if ($hasPreview) {
-      $version = $baseVersion + '-preview'
-      Write-Host "##[warning]SqlCore has preview dependencies - package version: $version"
-    } else {
-      $version = $baseVersion
-      Write-Host "SqlCore has stable dependencies - package version: $version"
-    }
-    
-    $props.Project.PropertyGroup.Version = $version
+    $version = ($(Major)).ToString() + '.' + ($(Minor)).ToString() + '.' + $(Patch) + $(Build.BuildNumber);
```
Benjin, line 82: Reverting to previous version of this step.  We don't need to set SqlCore as a preview just because a dependency is in preview.

PR #2697, opened 2026-05-22
Fix go-to-definition on REFERENCES targets and eliminate false FK "no primary/candidate keys" errors in SQL projects

### LINKED ISSUE #2694: SQlL Project editor shows red line for foreign key
Using AdventureWorks 2025 SQL Project which builds successfully, redlines show up for some objects like foreign keys but there's nothing wrong with the key in the project, so the error is wrong:

Image

ssreerama:
## Description

This PR fixes two IntelliSense issues in SQL project files:

1. Go-to-definition didn't work on tables referenced in FOREIGN KEY ... REFERENCES clauses. The previous implementation relied solely on Resolver.FindCompletions, which returns no results in DDL contexts (e.g. inside REFERENCES dbo.Orders). Cursor navigation on the table name produced "No definition found."

2. False "There are no primary or candidate keys in the referenced table" errors appeared as squiggles on every FK constraint. The SqlParser binder validates FK references by iterating ITable.Indexes — not ITable.Constraints. Our TSqlModelMetadataProvider returned an empty Indexes collection, so the binder could never find the PK/UK and always fired the error.

## Code Changes Checklist

## Reviewers: [Please read our reviewer guidelines](https://github.com/microsoft/sqltoolsservice/blob/main/.github/REVIEW_GUIDELINES.md)

Before:
IntelliSense_No_PK_UK_Error

After:
IntelliSense_No_PK_UK_Errors_After_Fix

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -1747,16 +1747,18 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
 
                         Microsoft.SqlServer.Dac.SourceInformation? sourceInfo = null;
 
-                        // Step 2: Walk backward from tokenIndex to reconstruct schema.object
-                        // and look it up in the source-location index. Works in DDL REFERENCES
-                        // clauses where FindCompletions returns nothing.
-                        string schemaName = GetPrecedingSchemaName(scriptParseInfo.ParseResult.Script.TokenManager, tokenIndex);
-                        if (schemaName != null)
-                        {
-                            string qualifiedName = $"{schemaName}.{tokenText}";
-                            if (!lazyProvider.TryGetSourceInformation(qualifiedName, out sourceInfo) || sourceInfo?.SourceName == null)
-                                lazyProvider.TryGetSourceInformation(StripDatabasePrefix(qualifiedName), out sourceInfo);
-                        }
+                        // Step 2: Walk backward from tokenIndex to reconstruct the full
+                        // schema-qualified name, including dotted schema names such as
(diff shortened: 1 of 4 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1813,13 +1812,59 @@ private DefinitionResult QueueProjectTask(
...
+            TSqlModelMetadataProvider provider,
+            string candidate,
+            out Microsoft.SqlServer.Dac.SourceInformation? sourceInfo)
+        {
+            string current = candidate;
+            while (!string.IsNullOrEmpty(current))
+            {
+                if (provider.TryGetSourceInformation(current, out sourceInfo) && sourceInfo?.SourceName != null)
+                    return true;
+                int dot = current.IndexOf('.');
```
Benjin, line 1832: Will this work if a name has a dot in it?  e.g. `[Schema.With.Dot].[Table.Having.A.Dot]`

Benjin (commented):
This has a lot more code in a very critical area than I'd want to take Friday before Ask mode for a conference.  What do you think on holding this review until after Build?

ssreerama:
Yes we can take this in after the Build. Meanwhile I can fix all bugbash bugs in a single stretch and can do a single sts bump for them. These changes are pretty straighforward for the go-to definition and touching only proejcts intellisesne code and will not cause any regression at all.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
ssreerama: It should work as we are collecting the schema and objects from the tokens, as I have a test for dotted schema [schemaModel.Models].[TableName]. But I'll make sure to test this case with dotted table name, if is a valid case, and add a test for it.

Benjin (approved):

PR #2705, opened 2026-06-02
feat: Implement textDocument/references (Find All References) for SQL project files

ssreerama:
## Summary

Implements `textDocument/references` (Find All References) for SQL project files. Returns all token locations across project files where the symbol under the cursor is referenced.

**Note:**  For connected/online SQL files the handler returns an empty result immediately — the feature is project-only by design.

## Changes

- Adds Find All References support for SQL project files in VS Code.
-  Right-click a table/view/procedure name in a .sql  project file → Find All References → VS Code opens the References panel showing every file and line in the project that uses that object
- Works for both qualified names (dbo.Customers) and unqualified names (Customers)
- Returns empty for connected/online SQL files — project files only
- Unopened project files are parsed on demand, so references are never missed

**Why this is needed for Rename/Refactor**
Rename (textDocument/rename) requires knowing every location where a symbol is used before it can rewrite them. This PR implements that location-discovery foundation — the rename handler will call the same reference-finding logic to collect all occurrences across the project and then apply the text edits.

- **`ProjectLanguageServiceTests.cs`** — Add `FindReferencesTests` with 6 tests:
  1. IntelliSense-disabled guard
  2. Unknown-file guard
  3. Non-project (online) file guard
  4. Cross-file references — both qualified (`dbo.Customers`) and unqualified (`Customers`) name paths
  5. On-demand parse of files not yet opened by the user
  6. Isolation regression — unrelated tables must not appear in each other's results
  
  With the feature implementation:
FindAllReferences

  
  Same files returns when we do Ctrl+Shift+F (Find all - search)
image

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
@@ -1699,14 +1699,10 @@ namespace Microsoft.SqlTools.ServiceLayer.LanguageServices
                 return Array.Empty<Location>();
             }
 
-            var resolvedData = Tuple.Create(tokenText, candidateFiles.ToList());
-            if (resolvedData == null)
-                return Array.Empty<Location>();
-
             var results = new List<Location>();
-            foreach (string filePath in resolvedData.Item2)
+            foreach (string filePath in candidateFiles)
             {
-                try { results.AddRange(FindTokenLocationsInFile(filePath, resolvedData.Item1)); }
+                try { results.AddRange(FindTokenLocationsInFile(filePath, tokenText)); }
                 catch (Exception ex) { Logger.Verbose($"FindProjectSymbolLocations: error scanning '{filePath}': {ex.Message}"); }
             }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -258,9 +258,12 @@ public void InitializeService(ServiceHost serviceHost, SqlToolsContext context)
             // Register the requests that this service will handle
 
             // turn off until needed (10/28/2016)
-            // serviceHost.SetRequestHandler(ReferencesRequest.Type, HandleReferencesRequest);
             // serviceHost.SetRequestHandler(DocumentHighlightRequest.Type, HandleDocumentHighlightRequest);
 
+            // Returns all locations where the symbol under the cursor is referenced (currently supported for SQL project files only; returns empty for connected files).
+            // Parallel safe: the cursor file is re-parsed under a per-ScriptParseInfo lock when needed; token scanning across candidate files is read-only.
+            serviceHost.SetRequestHandler(ReferencesRequest.Type, HandleReferencesRequest, isParallelProcessingSupported: true);
```
aasimkhan30, line 265: is this a standard vscode api?

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -1599,6 +1602,196 @@ private DefinitionResult QueueTask(TextDocumentPosition textDocumentPosition, Sc
...
+                    // model by the last name part. This works for project contexts where
+                    // MetadataDisplayInfoProvider is not available (FindCompletions would return nothing).
+                    qualifiedName ??= provider.FindQualifiedNameByLastPart(tokenText) ?? tokenText;
+
+                    // Step 3: collect candidate files via DacFx dependency graph
+                    var candidateFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                    if (provider.TryGetSourceInformation(qualifiedName, out var defInfo) &&
+                        defInfo?.SourceName != null)
+                        candidateFiles.Add(defInfo.SourceName);
+                    foreach (string path in provider.GetReferencingFilePaths(qualifiedName))
```
aasimkhan30: There are multiple iterations on project objects. Maybe they can reduced in a follow up PR.

aasimkhan30:
The changes look good to me. We can take the comments as a follow up task

aasimkhan30 (dismissed):

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
ssreerama: The HandleReferencesRequest handler already existed in sqltoolsservice but was disabled (commented out) since there was no implementation behind it. As part of this work, we enabled it and added the actual implementation that performs project-level symbol lookup using the TSqlModel metadata provider, on front end yes it wiring to the vscode standard api (vscode.languages.registerReferenceProvider)

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
ssreerama: looked into it, and removed unnecessary conversions and using directly the object coming form dacfx to parse.

aasimkhan30 (approved):

aasimkhan30:
Please make sure that find references work fine in non-project sql files.

PR #2714, opened 2026-06-09
Fixing token acquisition support for Dataverse connections

Benjin:
## Description

Dataverse/dynamics connections have a different resource/scope for acquiring connection authentication tokens from Azure.  This PR adds support to STS for the database to specify the resource necessary for obtaining an MFA auth token.

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
@@ -788,7 +788,7 @@ namespace Microsoft.SqlTools.ServiceLayer.Connection
         /// advertises via FedAuthInfo (e.g. Dataverse TDS endpoints under <c>*.crm.dynamics.com</c>).
         ///
         /// When <paramref name="connInfo"/> is supplied, the resource is also captured into
-        /// <see cref="ConnectionInfo.LastAzureResourceRequested"/> so SMO's resource-less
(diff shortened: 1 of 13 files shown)


src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -2204,7 +2264,8 @@ internal static void ConfigureSqlConnectionAuth(SqlConnection sqlConn, Connectio
         {
             if (connInfo.AzureTokenFetcher != null && connInfo.ConnectionDetails.AuthenticationType == AzureMFA)
             {
-                sqlConn.AccessTokenCallback = ToSqlAccessTokenCallback(connInfo.AzureTokenFetcher);
+                var (token, _) = connInfo.AzureTokenFetcher(connInfo.AzureResourceUri).GetAwaiter().GetResult();
+                sqlConn.AccessToken = token;
```
caohai: For non-SMO `OpenSqlConnection` callers, would this static token assignment take away the auto-refresh capability from AccessTokenCallback?

Benjin: I think you're right, but it was still working before.  In any case, I've fixed that to be more intentional how it works.

caohai (approved):

PR #2738, opened 2026-06-25
Move base ConnectionInfo/ConnectionDetails into LanguageService

Charles-Gagnon:
## Description

Language service depends on a set of functionality from ConnectionInfo, but not the entire thing. Since moving it all would be a much more massive (and unnecessary) change, I'm instead opting to split it into a base abstract class in LanguageService and then have the hosts (ServiceLayer in STSs case) provide the "full" implementation with whatever else is needed on top of it.

There were also a few other smaller changes done - mostly just moving helper methods into LanguageService (and often times in places that made more sense than their original place)

## Code Changes Checklist


ssreerama (dismissed):

src/Microsoft.SqlTools.LanguageService/LanguageServices/Completion/CompletionService.cs
```diff
@@ -10,18 +10,15 @@
...
 using Microsoft.SqlTools.Utility;
 
-namespace Microsoft.SqlTools.ServiceLayer.LanguageServices.Completion
+namespace Microsoft.SqlTools.LanguageService.LanguageServices.Completion
 {
     /// <summary>
     /// A service to create auto complete list for given script document 
     /// </summary>
-    internal class CompletionService
+    internal sealed class CompletionService
```
Benjin, line 21: Out of curiosity, why seal this?

Benjin (dismissed):

src/Microsoft.SqlTools.LanguageService/LanguageServices/Completion/CompletionService.cs
Charles-Gagnon: I'll admit this was Copilot doing this, and I didn't see the harm in keeping it given how they're used. Happy to revert these though if you'd rather keep the changes minimal.

Benjin: It's okay, I guess.  Odd that copilot would decide to change the accessibility of a class, though.

aasimkhan30 (approved):

PR #2740, opened 2026-06-26
Move TSqlFormatterService into LanguageService

Charles-Gagnon:
## Description

Not too bad, but there's a tangled dependency with the ShouldSkipNonMssqlFile functionality. In order to keep these PRs small I refactored it out into a separate interface to make this initial move cleaner - and then will be following up with removing that once the rest of the core LanguageService stuff is moved over.

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
@@ -46,6 +46,8 @@ using Microsoft.SqlTools.ServiceLayer.SqlContext;
 using Microsoft.SqlTools.ServiceLayer.SqlProjects;
 using Microsoft.SqlTools.ServiceLayer.TableDesigner;
 using Microsoft.SqlTools.ServiceLayer.Utility;
+using Microsoft.SqlTools.LanguageService.Formatter;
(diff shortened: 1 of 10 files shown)


aasimkhan30 (approved):

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
```diff
@@ -183,6 +188,19 @@ private static void InitializeRequestHandlersAndServices(ServiceHost serviceHost
             InitializeHostedServices(serviceProvider, serviceHost);
             serviceHost.ServiceProvider = serviceProvider;
 
+            // Wire the formatter's host-specific seams here (inverted control) since the formatter lives in the
```
Benjin, line 191: What does this comment mean by "seams"?

Benjin (approved):

src/Microsoft.SqlTools.ServiceLayer/HostLoader.cs
Charles-Gagnon: Copilot loves this term for describing the public extensibility points that are used to hook the language service library up to its host. Not a term I've really heard used before, but it makes sense to me and I don't really have a cleaner alternative so figured it was fine.

PR #2741, opened 2026-06-26
Create abstraction for language service only settings

Charles-Gagnon:
## Description

The Language Service layer doesn't need to know about all the settings for the entire host, so create an abstraction that hosts implement for getting those settings.

Note that the SqlToolsSettings is only deserialized coming in from the configuration update messages - so the read-only properties we're adding here don't affect that at all since they just read the values from those specific implementations.

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.LanguageService/LanguageServices/ILanguageServiceSettings.cs
+++ b/src/Microsoft.SqlTools.LanguageService/LanguageServices/ILanguageServiceSettings.cs
@@ -1,0 +1,38 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.SqlTools.LanguageService.Formatter;
+
+namespace Microsoft.SqlTools.LanguageService.LanguageServices
+{
+    /// <summary>
+    /// Exposes the subset of workspace settings that the language service needs in order to
+    /// decide whether to provide diagnostics, suggestions and quick info, and to react to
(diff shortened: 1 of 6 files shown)


ssreerama (dismissed):

src/Microsoft.SqlTools.LanguageService/LanguageServices/ILanguageServiceSettings.cs
```diff
@@ -0,0 +1,38 @@
...
+        bool IsSuggestionsEnabled { get; }
+
+        /// <summary>Gets a flag determining if quick info (hover) is enabled.</summary>
+        bool IsQuickInfoEnabled { get; }
+
+        /// <summary>Gets a flag determining if IntelliSense is enabled.</summary>
+        bool IsIntelliSenseEnabled { get; }
+
+        /// <summary>Gets a flag determining if error checking is enabled, or <c>null</c> if unset.</summary>
+        bool? IsErrorCheckingEnabled { get; }
```
Benjin: this is the only one that needs to consider an unset state?

Benjin (dismissed):

src/Microsoft.SqlTools.LanguageService/LanguageServices/ILanguageServiceSettings.cs
Charles-Gagnon: Good catch - this seemed to be because of how it's used (the other ones had implementations already).

I changed it match the others though to better align usage for all this stuff.

Benjin (approved):

PR #2742, opened 2026-06-26
Move Scripter over to LanguageService

Charles-Gagnon:
## Description

Main move here was Scripter - rest was just supporting stuff (such as the Connection Info helper methods).

I was hoping to avoid taking a dependency on SqlCore to keep things simpler - but that's not straightforward. So adding that dependency for now and I'll revisit later once I'm done with the move.

## Code Changes Checklist


src/Microsoft.SqlTools.LanguageService/Connection/ConnectionStringHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+    public static class ConnectionStringHelper
+    {
+        /// <summary>
+        /// Build a connection string from a connection details instance
+        /// </summary>
+        /// <param name="connectionDetails">Connection details</param>
+        /// <param name="enableSqlAuthenticationProvider">Whether the configured 'Sql Authentication Provider' for 'Active Directory Interactive' authentication mode should be used when the user chooses 'Azure MFA'.</param>
+        /// <param name="enableConnectionPooling">Whether connection pooling is enabled for all SQL connections.</param>
+        /// <param name="forceDisablePooling">Whether to disable connection pooling, defaults to true.</param>
+        public static string BuildConnectionString(ConnectionDetails connectionDetails, bool enableSqlAuthenticationProvider, bool enableConnectionPooling, bool forceDisablePooling = true)
```
Charles-Gagnon: This was moved over directly from ConnectionService - it just had the new parameters added but otherwise the same.

src/Microsoft.SqlTools.LanguageService/Connection/ConnectionStringHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+            if (!string.IsNullOrEmpty(connectionDetails.ConnectionString))
+            {
+                connectionBuilder = new SqlConnectionStringBuilder(connectionDetails.ConnectionString);
+            }
+            else
+            {
+                // add alternate port to data source property if provided
+                string dataSource = !connectionDetails.Port.HasValue
+                    ? connectionDetails.ServerName
+                    : string.Format("{0},{1}", connectionDetails.ServerName, connectionDetails.Port.Value);
```
Benjin: While we're here, can you have this use string interpolation instead?

src/Microsoft.SqlTools.LanguageService/Connection/ConnectionStringHelper.cs
Benjin: Can you remind me why there are separate flags for `enableConnectionPooling` and `forceDisablePooling`?  At first glance, those seem like they'd control the same behavior; if that's not the case, can you explain the difference in a comment (and maybe we'd need a better name)?

src/Microsoft.SqlTools.LanguageService/Connection/ConnectionStringHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+
+#nullable disable
+
+using System;
+using System.Globalization;
+using Microsoft.Data.SqlClient;
+using Microsoft.SqlTools.LanguageService.Connection.Contracts;
+using static Microsoft.SqlTools.Utility.SqlConstants;
+
+namespace Microsoft.SqlTools.LanguageService.Connection
```
Benjin, line 14: It's a bit odd to me that connection stuff is in the lang service project, but oh well, I guess. 🙃

Charles-Gagnon: Agreed. Once I have this all untangled I plan on taking a look to see if we can "rationalize" this stuff more. I've considered even making a separate Connection library that contains just the core connection stuff - but we'll see.

For now this is just helper methods so I wasn't too concerned about it being in LS for now.

src/Microsoft.SqlTools.LanguageService/Connection/ConnectionStringHelper.cs
Charles-Gagnon: Uh, yeah I think the naming here is wonky.

EnableConnectionPooling is a global parameter passed in at startup that ties to the "Enable Connection Pooling" setting. If set to true (default false) then it will leave the pooling value as default for sqlconnections - which is enabled.

forceDisablePooling only matters if EnableConnectionPooling is false currently - since all the places we use it  look like `!EnableConnectionPooling && forceDisablePooling`. So really it's more of a default that can be overridden.

Which means that EnableConnectionPooling should really be "ForceDefaultConnectionPoolingEnabled" and forceDisablePooling should be something like "defaultConnectionPoolingEnabled"

I can do that as a follow up PR so it can be examined more closely, but for now I've consolidated the parameters here into a single parameter and moved the logic to calculate that back out into the caller ConnectionService.

src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs
```diff
@@ -1548,316 +1548,7 @@ public static string BuildConnectionString(ConnectionDetails connectionDetails,
...
-                connectionBuilder.TypeSystemVersion = connectionDetails.TypeSystemVersion;
-            }
-            if (!EnableConnectionPooling && forceDisablePooling)
-            {
-                connectionBuilder.Pooling = false;
-            }
-
-            return connectionBuilder;
+            return Microsoft.SqlTools.LanguageService.Connection.ConnectionStringHelper.CreateConnectionStringBuilder(connectionDetails, Instance.EnableSqlAuthenticationProvider, EnableConnectionPooling, forceDisablePooling);
         }
```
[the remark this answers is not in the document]
Charles-Gagnon: It'll catch up in the next loc pass

src/Microsoft.SqlTools.LanguageService/LanguageServices/ConnectionInfoBase.cs
```diff
@@ -44,6 +45,19 @@ protected ConnectionInfoBase(string ownerUri, ConnectionDetails connectionDetail
...
+        /// </summary>
+        public abstract bool IsCloud { get; set; }
+
+        /// <summary>
+        /// Tries to get the open <see cref="DbConnection"/> associated with the given connection type.
+        /// </summary>
+        /// <param name="connectionType">The connection type (e.g. "Query", "Default").</param>
+        /// <param name="connection">The located connection, or <c>null</c> if none.</param>
+        /// <returns>True if a connection of the given type was found; false otherwise.</returns>
+        public abstract bool TryGetConnection(string connectionType, out DbConnection connection);
```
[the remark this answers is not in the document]
Charles-Gagnon: This isn't a concern - we don't ship this to be used externally currently

test/Microsoft.SqlTools.LanguageService.UnitTests/Utility/SrTests.cs
```diff
@@ -24,6 +24,20 @@ public void SrPropertiesTest()
             Assert.NotNull(LanguageServiceSr.WorkspaceServiceBufferPositionOutOfOrder(0, 0, 0, 0));
             Assert.NotNull(LanguageServiceSr.WorkspaceServicePositionLineOutOfRange);
             Assert.NotNull(LanguageServiceSr.WorkspaceServicePositionColumnOutOfRange(0));
+            Assert.NotNull(LanguageServiceSr.ConnectionServiceConnStringMissingAttestationUrlWithAttestationProtocol);
+            Assert.NotNull(LanguageServiceSr.ConnectionServiceConnStringInvalidAttestationProtocolNoneWithUrl);
+            Assert.NotNull(LanguageServiceSr.ConnectionServiceConnStringMissingAttestationProtocolWithSecureEnclaves);
+            Assert.NotNull(LanguageServiceSr.ConnectionServiceConnStringInvalidEnclaveAttestationProtocol(""));
+            Assert.NotNull(LanguageServiceSr.ConnectionServiceConnStringInvalidColumnEncryptionSetting(""));
```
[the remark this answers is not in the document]
Charles-Gagnon: Fixed and added other missing strings as well

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -2200,7 +2201,7 @@ private DefinitionResult GetDefinitionFromTokenList(TextDocumentPosition textDoc
                         return new DefinitionResult
                         {
                             IsErrorResult = true,
-                            Message = SR.PeekDefinitionError(ex.Message),
+                            Message = PeekDefSR.PeekDefinitionError(ex.Message),
                             Locations = null
                         };
```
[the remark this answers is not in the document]
Charles-Gagnon: Loc will catch up

aasimkhan30 (approved):

Benjin (approved):

PR #2745, opened 2026-06-29
Show specific error messages for unsupported rename targets

ssreerama:
## Description

Previously, all rejected rename operations returned `null` from the server, causing the client to display a single generic "not a project file" error regardless of the actual reason. This change returns a specific `errorMessage` string in the response so the client can show a contextual message.

- Verified on SQL project files with cursor on: keyword (`SELECT`), schema name (`dbo`), valid table column, non-identifier whitespace
- Each scenario now shows the appropriate contextual error in VS Code
- Adding support to SkipExistingModelValidation for the Tsqlmodel update methods.

## Code Changes Checklist


src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -95,6 +95,19 @@ public static LanguageService Instance
 
         internal const int CompletionExtTimeout = 200;
 
+        internal const string RenameNotSupportedLiveServer =
+            "Rename is not supported for files connected to a live server. Open the file as part of a SQL project first.";
+
+        internal const string RenameNotSupportedSchemaNames =
+            "Rename is not supported for schema names. Use 'Move to Schema' to move objects between schemas.";
+
+        internal const string RenameNotSupportedKeywords =
```
aasimkhan30: lets keep the message short 

Image

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -95,6 +95,19 @@ public static LanguageService Instance
 
         internal const int CompletionExtTimeout = 200;
 
+        internal const string RenameNotSupportedLiveServer =
+            "Rename is not supported for files connected to a live server. Open the file as part of a SQL project first.";
+
+        internal const string RenameNotSupportedSchemaNames =
+            "Rename is not supported for schema names. Use 'Move to Schema' to move objects between schemas.";
```
aasimkhan30: don't see the point of showing "move to schema" here. Renaming the schema is  a very different operation than moving an object between schemas.

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
```diff
@@ -95,6 +95,19 @@ public static LanguageService Instance
 
         internal const int CompletionExtTimeout = 200;
 
+        internal const string RenameNotSupportedLiveServer =
+            "Rename is not supported for files connected to a live server. Open the file as part of a SQL project first.";
```
aasimkhan30: what happens when a sqlproj file is connected to a live server. if we allow that  case, we should change this error message.

aasimkhan30:
it will be nice if we can get the error messages reviewed by @dzsquared

src/Microsoft.SqlTools.ServiceLayer/LanguageServices/LanguageService.cs
dzsquared: help me understand the situation of this error message - is it a .sql file that is not part of a .sqlproj?

or is it just any connected editor? even when an editor is connected, if the file is in a sqlproj we should enable the rename.

however, if there's no sqlproj - we'd want an error message that redirects the user to sp_rename etc

ssreerama: When we open a .sql file from database and try to rename, right now we show the info message like this.
image

If the file is opened form sqlproj and connect to server, we consider it as a project file only and allow rename operation.

aasimkhan30 (dismissed):

aasimkhan30 (approved):

PR #2763, opened 2026-07-18
Fix #2154: Strip carriage return characters in CSV export

### LINKED ISSUE #2154: "Save as CSV" does not remove CR from data
A CR (carriage return) character is not removed when exporting data to a CSV.
This is an unexpected behavior especially as SQL Mgmt Studio does remove it.

Please fix the "Save as CSV" function.

Many thanks in advance!

saishparadkar10:
﻿Fixes #2154

## Problem
The "Save as CSV" function does not remove CR (\r) characters from data.
This causes malformed CSV files where rows containing embedded \r are split
across multiple lines when opened in Excel or other CSV parsers.

SQL Server Management Studio already strips CR characters during CSV export,
so users expect the same behaviour from the VS Code / Azure Data Studio tooling.

## Fix
Strip all \r characters from field values in EncodeCsvField() before any
quoting/escaping decisions are made. This normalises Windows-style \r\n line
endings to Unix-style \n (which is safe inside a double-quoted CSV field)
and removes any standalone \r characters.

## Testing
- Updated existing unit tests to reflect the new CR-stripping behaviour
- Added 4 new parametrised test cases covering embedded, leading, trailing,
  and \r\n-normalised CR scenarios
- All 33 SaveAsCsvFileStreamWriterTests pass

### DIFF
--- a/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsCsvFileStreamWriter.cs
+++ b/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsCsvFileStreamWriter.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.ServiceLayer/QueryExecution/DataStorage/SaveAsCsvFileStreamWriter.cs
```diff
@@ -125,13 +125,20 @@ internal string EncodeCsvField(string field)
                 return "NULL";
             }
 
-            // Replace all quotes in the original field with double quotes
-            string ret = field.Replace(textIdentifierString, textIdentifierString + textIdentifierString);
+            // Strip carriage return characters (issue #2154). CR characters inside data cause malformed
+            // CSV rows because some parsers treat \r as a line ending even inside double-quoted fields.
+            // Removing \r normalises Windows-style \r\n line endings to Unix-style \n, which is safe
+            // inside a quoted CSV field, and removes any standalone \r characters.
+            string sanitized = field.Replace("\r", "");
```
aasimkhan30, line 132: We should not universally modify the text and probably add a configurable option for it.

aasimkhan30:
The current code handles CR and LF according to the CSV specification. We must not remove CR from all values. This change also does not provide SSMS parity. SSMS has an option to remove both CR and LF. 

I will close this PR. We can consider a user setting for this behavior. However, we already have many settings, so it needs some PM work.

PR #2755, opened 2026-07-08
Schema Compare: Fabric Warehouse platform + constraint handling via DacFx

roramsay:
## Summary
Adds Microsoft Fabric Warehouse (`SqlDwUnified`) support to Schema Compare. STS surfaces the platform each endpoint targets and renders the per-difference scripts DacFx returns, with no Fabric-specific script logic of its own.

## Changes

**Platform surfacing (reflection-free)**
- `SchemaCompareOperation` adds `SourcePlatform` / `TargetPlatform`, read from `ComparisonResult.SourceModel.Version` / `TargetModel.Version` (short `SqlServerVersion` name, e.g. `Sql160`, `SqlDwUnified`).
- Surfaced on the JSON-RPC contract: `SourcePlatform` / `TargetPlatform` added to `SchemaCompareResult` (SqlCore `SchemaCompareResults.cs` + ServiceLayer `SchemaCompareRequest.cs`), wired through `SchemaCompareService`. Lets the client show which dialect a comparison is using (e.g. a Fabric Warehouse badge).

**Constraint scripts**
- `SchemaCompareUtils.CreateDiffEntry` now assigns the raw `GetDiffEntrySourceScript` / `GetDiffEntryTargetScript` result directly (`if (script != null)`), replacing the previous `StartsWith("alter")` strip. DacFx returns the standalone `ALTER TABLE ... ADD CONSTRAINT` for a `SqlDwUnified` constraint and empty for a child scripted inline in its parent's `CREATE`, so STS needs no platform/constraint logic.

**Package dependencies (`Packages.props`)**
- `Microsoft.SqlServer.DacFx`: `170.5.38-preview` -> `170.5.60-preview` (shipped; includes the SqlDwUnified version mapping and the `GetDiffEntryScript` inline-suppression behavior STS relies on).
- `Microsoft.SqlServer.TransactSql.ScriptDom`: `180.37.3` -> `180.59.2` (transitively required by DacFx 170.5.60-preview; restore fails NU1605 otherwise).

**Tests** (`SchemaCompareFabricPlatformTests.cs`, offline via SqlDwUnified dacpacs/projects, no live warehouse)
- `FabricWarehouse_SurfacesPlatformAndPreservesConstraintAlterScript` - SourcePlatform/TargetPlatform == `SqlDwUnified`; constraint difference carries its standalone `ADD CONSTRAINT` script.
[...]

PR #2771, opened 2026-07-30
Schema Compare: Fabric Warehouse platform + constraint handling via DacFx

roramsay:
> Replaces #2755, which was accidentally closed and could not be reopened. Same branch (`roramsay/schema-compare-fabric-dacfx-approach-b`), synced to latest `main`.

## Summary
Adds Microsoft Fabric Warehouse (`SqlDwUnified`) support to Schema Compare. STS renders the per-difference scripts DacFx returns, with no Fabric-specific script logic of its own.

## Changes

**Constraint scripts**
- `SchemaCompareUtils.CreateDiffEntry` now assigns the raw `GetDiffEntrySourceScript` / `GetDiffEntryTargetScript` result directly (`if (script != null)`), replacing the previous `StartsWith("alter")` strip. DacFx returns the standalone `ALTER TABLE ... ADD CONSTRAINT` for a `SqlDwUnified` constraint and empty for a child scripted inline in its parent's `CREATE`, so STS needs no platform/constraint logic.

**Package dependencies (`Packages.props`)**
- `Microsoft.SqlServer.DacFx`: `170.5.38-preview` -> `170.5.60-preview` (shipped; includes the `GetDiffEntryScript` inline-suppression behavior STS relies on).
- `Microsoft.SqlServer.TransactSql.ScriptDom`: transitively required by DacFx 170.5.60-preview (restore fails NU1605 otherwise).

**Tests** (`SchemaCompareFabricPlatformTests.cs`, offline via SqlDwUnified dacpacs/projects, no live warehouse)
- `FabricWarehouse_PreservesConstraintAlterScript` - a Fabric constraint difference carries its standalone `ADD CONSTRAINT` script (not stripped).
- `FabricWarehouse_GeneratesStandaloneConstraintDeployScript_Offline` - generated deploy script emits constraints as standalone `ALTER TABLE ... ADD CONSTRAINT ... NONCLUSTERED ... NOT ENFORCED`.
- `FabricWarehouse_AppliesConstraintsToTargetProject_Offline` - publishing the differences into a target SqlDwUnified project converges (table + constraints round-trip).

## Testing
- SchemaCompare unit tests: 6/6 pass.
- Fabric offline integration tests: 3/3 pass.
- ServiceLayer + SqlCore + IntegrationTests + UnitTests build with 0 errors against DacFx 170.5.60-preview.

## Live coverage
Database-source Fabric behavior (the DB branch of `GetDiffEntryScript`) is not exercisable offline; it is covered by DacFx's own live Fabric Warehouse Schema Compare tests in the DacFx Fabric Nightly pipeline.


src/Microsoft.SqlTools.SqlCore/SchemaCompare/SchemaCompareOperation.cs
```diff
@@ -117,6 +131,11 @@ public void Execute()
                     }
                 }
 
+                // Surface the platform each model targets (read from TSqlModel.Version) so the UI
+                // can show the dialect (e.g. SqlDwUnified for Fabric Warehouse).
+                this.SourcePlatform = this.ComparisonResult.SourceModel?.Version.ToString();
+                this.TargetPlatform = this.ComparisonResult.TargetModel?.Version.ToString();
```
Benjin: I think it'd be better to keep these values as the original enums here, and call toString() on them only when we're constructing the response object before sending.  Keeps the value guaranteed to be valid as long as possible.

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareRequest.cs
```diff
@@ -49,6 +49,18 @@ public class SchemaCompareResult : ResultStatus
         public bool AreEqual { get; set; }
 
         public List<DiffEntry> Differences { get; set; }
+
+        /// <summary>
+        /// The SQL platform (T-SQL dialect) the source model targets, as the short
+        /// SqlServerVersion name (e.g. "Sql160", "SqlAzureV12", "SqlDwUnified"). Lets the client
+        /// surface which dialect Schema Compare is using (e.g. a Fabric Warehouse badge).
+        /// </summary>
+        public string SourcePlatform { get; set; }
```
Benjin: Can you add the path of the enum definition within DacFx's codebase here so it's easier to look up/reference?

e.g.

`Valid values found in DacFx's PlatformEnum enum at /path/to/some/PlatformEnums.cs`

Side note: I didn't see these added in the reciprocol contract in the VS Code side (where you'll want to add the same comment)

src/Microsoft.SqlTools.SqlCore/SchemaCompare/SchemaCompareOperation.cs
roramsay: Done in 3b4f3fc2 — `SourcePlatform`/`TargetPlatform` on `SchemaCompareOperation` are now `SqlServerVersion?` and are set directly from `ComparisonResult.SourceModel?.Version` (no `ToString`). The `.ToString()` now happens only in `SchemaCompareService` when constructing the response object. The JSON-RPC contract fields stay `string` (the VS Code client consumes the short name).

src/Microsoft.SqlTools.ServiceLayer/SchemaCompare/Contracts/SchemaCompareRequest.cs
roramsay: Done in 3b4f3fc2 — added the enum path to the `SourcePlatform` doc-comment in both contracts (ServiceLayer `SchemaCompareRequest.cs` and SqlCore `SchemaCompareResults.cs`): "Valid values are the names of DacFx's `Microsoft.SqlServer.Dac.Model.SqlServerVersion` enum (DacFx source: `Source/SchemaApi/Model/SqlServerVersion.cs`)." For the reciprocal VS Code contract: the same enum-path reference has been added to `SchemaCompareResult.sourcePlatform`/`targetPlatform` in `vscode-mssql.d.ts` on PR #22340.

src/Microsoft.SqlTools.SqlCore/SchemaCompare/SchemaCompareUtils.cs
```diff
@@ -51,25 +51,22 @@ internal static DiffEntry CreateDiffEntry(SchemaDifference difference, DiffEntry
 
             if (difference.DifferenceType == SchemaDifferenceType.Object)
             {
-                // set source and target scripts
+                // DacFx's GetDiffEntry{Source,Target}Script returns the standalone script for a
+                // difference that owns one (e.g. ALTER TABLE ... ADD CONSTRAINT for a SqlDwUnified
```
Benjin, line 55: Is there a reason DacFx can't return the constraints for DW as inlined within the parent's CREATE?  I'm concerned that removing these checks could result in some unexpected/buggy behavior for SQL DB.

Benjin: After discussing offline, it sounds like:

1. Inline constraints are not valid for DW, so they must be `ALTER` syntax
2. DacFx now filters out ALTER for invalid endpoints, so those checks are no longer necessary here

Benjin (approved):

src/Microsoft.SqlTools.SqlCore/SchemaCompare/SchemaCompareUtils.cs
roramsay: For Warehouses the constraints must be in ALTER statements.  The reason for moving to the latest DacFx release in this PR is because it contains changes that make the correct syntax get passed through regardless of platform, so the ALTER check in STS is no longer necessary.

PR #2778, opened 2026-08-06
Release IntelliSense refresh lock before metadata rebuild

aasimkhan30:
## Description

- Release the thread-affine `BuildingMetadataLock` before starting the asynchronous IntelliSense metadata rebuild.
- Keep binding-context replacement and parse-state removal protected by the short synchronous monitor scope.
- Preserve URI-specific asynchronous serialization so refresh requests for the same editor cannot overlap.
- Add deterministic regression coverage for monitor release and same-URI refresh serialization.

Related to microsoft/vscode-mssql#22236.

### Evidence and scope

Public reports such as microsoft/vscode-mssql#22236 describe `Updating IntelliSense` remaining stuck and requiring a window reload, but they do not include a stack trace identifying this lock defect. We independently reproduced the refresh failure on both a one-table local database and Fabric: the async continuation attempted to release `BuildingMetadataLock` from a different thread, threw `SynchronizationLockException`, and left refresh incomplete. This PR fixes that confirmed refresh defect but does not claim to resolve every symptom covered by #22236.

Validation:

- `dotnet test test/Microsoft.SqlTools.ServiceLayer.UnitTests/Microsoft.SqlTools.ServiceLayer.UnitTests.csproj --no-restore --filter "FullyQualifiedName~RebuildIntelliSenseTests"` — 2 passed.
- `dotnet test test/Microsoft.SqlTools.ServiceLayer.UnitTests/Microsoft.SqlTools.ServiceLayer.UnitTests.csproj --no-build --no-restore --filter "FullyQualifiedName~Microsoft.SqlTools.ServiceLayer.UnitTests.LanguageServer"` — 61 passed.

*Provide a clear, concise summary of the changes in this PR. What problem does it solve? Why is it needed? Link any related issues using [issue closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).*

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
+++ b/src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
@@ -999,17 +999,24 @@ namespace Microsoft.SqlTools.LanguageService.LanguageServices
                     {
                         try
                         {
-                            this.BindingQueue.AddConnectionContext(connInfo, featureName: Constants.LanguageServiceFeature, overwrite: true);
-                            RemoveScriptParseInfo(rebuildParams.OwnerUri);
-                        }
-                        finally
-                        {
-                            // A Monitor is owned by the thread that entered it, so it must be
-                            // released before the asynchronous metadata rebuild can change threads.
-                            Monitor.Exit(scriptInfo.BuildingMetadataLock);
-                        }
+                            try
+                            {
+                                this.BindingQueue.AddConnectionContext(connInfo, featureName: Constants.LanguageServiceFeature, overwrite: true);
+                                RemoveScriptParseInfo(rebuildParams.OwnerUri);
+                            }
+                            finally
+                            {
+                                // A Monitor is owned by the thread that entered it, so it must be
+                                // released before the asynchronous metadata rebuild can change threads.
+                                Monitor.Exit(scriptInfo.BuildingMetadataLock);
+                            }
 
-                        await UpdateLanguageServiceOnConnection(connInfo);
+                            await UpdateLanguageServiceOnConnection(connInfo);
+                        }
+                        catch (Exception ex)
+                        {
+                            Logger.Error("Unknown error " + ex.ToString());
+                        }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
```diff
@@ -999,19 +999,24 @@ public virtual async Task DoHandleRebuildIntellisenseNotification(RebuildIntelli
...
+                            try
+                            {
+                                this.BindingQueue.AddConnectionContext(connInfo, featureName: Constants.LanguageServiceFeature, overwrite: true);
+                                RemoveScriptParseInfo(rebuildParams.OwnerUri);
+                            }
+                            finally
+                            {
+                                // A Monitor is owned by the thread that entered it, so it must be
+                                // released before the asynchronous metadata rebuild can change threads.
+                                Monitor.Exit(scriptInfo.BuildingMetadataLock);
```
Charles-Gagnon, line 1011: I question this whole locking stuff to begin with...a couple of questions : 

1. Why is it called BuildingMetadataLock if that's no longer actually be used to lock the actual building of the metadata?
2. The actual metadata should be shared among all editors that share a connection context, but we seem to be locking based on URI. So that seems like another potential problem - different editors connected to the same server could interfere with each other

Anyways - this fix is fine in that it definitely addresses the synchronization exception. But I think this whole binding stuff is worth reviewing in its entirety at some point since I think there's a number of issues with how it's actually implemented.

aasimkhan30: Totally onboard with the idea of sharing intellisense cache between editors. It especially becomes important when trying to solve something like this https://github.com/microsoft/vscode-mssql/issues/21958 (Updating cache on schema changes).  

I think `BindingQueue` and `BuildingMetadataLock` definitely have a naming problem as they are used beyond just binding operations.    Makes the code harder to read.

caohai (approved):

PR #2782, opened 2026-08-07
STS2 foundation — contracts, abstractions, and deterministic core

kburtram:
# PR: STS2 foundation — contracts, abstractions, and deterministic core

## Stack position

- Branch: **dev/karlb/sts2_foundation**
- Commit: **f19c4c1a**
- PR base: **main**
- Review order: **1 of 5**
- Change size: **35 files, 3,684 insertions**

## Summary

This PR establishes the architectural foundation for STS2 without adding a transport, database provider, background runtime, or ServiceLayer activation path.

It introduces three deliberately separated assemblies:

- **Microsoft.SqlTools.Sts2.Contracts** defines stable method names, error identities, limits, and other protocol constants.
- **Microsoft.SqlTools.Sts2.Abstractions** defines the provider-neutral database driver port and streaming execution event model.
- **Microsoft.SqlTools.Sts2.Core** implements the deterministic connection/query state machine as a synchronous reducer over serializable data.

The central review question is whether these boundaries are strong enough to support later runtime, transport, and provider layers without letting nondeterminism or provider-specific behavior leak into the protocol state machine.

## Why this is a separate PR

The eventual STS2 feature spans protocol handling, journaling, replay, stdio multiplexing, SQL providers, and legacy ServiceLayer integration. Reviewing that as one change would make the most important boundary—the pure state machine—difficult to evaluate.

This first slice is intentionally executable but inert:

- It builds independently through **sqltoolsservice-sts2.slnf**.
- Its reducer and architecture rules have focused tests.
- Nothing starts at process launch.
- Nothing opens files, sockets, streams, or database connections.
- No existing ServiceLayer behavior changes.

Subsequent PRs can therefore be reviewed as consumers of a stable foundation rather than as simultaneous changes to the protocol model.

### Stable contracts

The Contracts assembly collects the vocabulary that clients and implementations must agree on:

- v2 method names
- stable Sts2.* error codes and their JSON-RPC mappings
- pinned defaults and limits
- wire-format constants

Contracts is BCL-only. It does not depend on Core, Runtime, drivers, StreamJsonRpc, or the legacy service. Its public API is tracked explicitly through PublicAPI.Shipped.txt and PublicAPI.Unshipped.txt.

### Provider-neutral driver port

The Abstractions assembly defines:

- **IDbDriver** for opening a provider session
- **IDbSession** for forward-only query execution and cancellation
- sanitized open/query request records
- provider-neutral server metadata
- a stream of execution events for result-set boundaries, row pages, server messages, completion, and cancellation
- stable driver failures that carry Sts2 error identities instead of provider exception types

Provider types do not cross this boundary. In particular, ADO.NET connections, readers, commands, provider exceptions, and SQL Server-specific CLR types cannot appear in the public signatures.

The event/value model also makes loss and truncation explicit. Large values carry retained prefixes plus total byte count and full-value digest. Vector and spatial values cross the boundary either as complete provider-neutral representations or as explicit “unavailable” records with stable reasons; they are never silently approximated.

### Pure deterministic reducer

The Core assembly models the service as:

    CoreState + CoreEnvelope -> CoreDecision

A decision contains the next immutable state and a list of declarative outputs:

- JSON-RPC result
- JSON-RPC error
- JSON-RPC notification
- effect request
- diagnostic
- capture-configuration change

Core never executes those outputs. Later runtime code journals and performs them.

The reducer covers initialization, diagnostics, connection open/cancel/close, query execute/ack/cancel/dispose, driver observations, capture configuration, lifecycle controls, and terminal/error behavior.

### Serializable state only

Core state contains logical facts, not resources:

- connection and query identifiers
- lifecycle phases
- correlation identifiers
- page-credit counters
- completion/disposal guards
- capture policy and configuration version
- registered driver descriptors

It never contains a database connection, cancellation token, task, stream, timer, secret, row cell, or provider object.

Connection and query maps are immutable sorted dictionaries. That gives stable iteration and state-dump ordering instead of depending on incidental dictionary insertion behavior.

### Architecture enforcement

This PR adds compile-time and test-time guardrails:

- banned API analyzers reject ambient time, randomness, async/task primitives, channels, locks, cancellation tokens, and I/O in deterministic assemblies
- the stricter Core rules also reject GUID creation and other ambient identity sources
- public API analyzers track the intended surface of Contracts, Abstractions, and Core
- dependency-matrix tests verify the allowed project/package edges
- driver-isolation tests ensure provider and legacy namespaces do not enter the foundational projects
- every product assembly provides a COMPONENT.md responsibility/dependency fragment

The architecture rules are part of the build rather than review conventions that can drift.

### 1. IDs derive from journal sequence, not ambient randomness

Core derives logical identifiers such as connection and query IDs from the input sequence. It does not call Guid.NewGuid or consult a random source.

This makes a replayed envelope stream produce the same identifiers, state transitions, and outputs as the live run.

### 2. Session configuration enters as data

Core always starts from one fixed **CoreState.Initial**. Service version, available drivers, limits, and capture policy arrive through a journalable session.start control envelope.

This avoids a subtle replay split where the live state machine is initialized with host-only constructor values that are absent from the journal.

### 3. Cancellation is a state-machine event

Core does not receive CancellationToken. Cancellation enters as a request/control envelope, changes explicit state, and produces an effect request.

That makes cancel/complete/dispose races observable and replayable instead of depending on task scheduling.

### 4. Effects are declarative

Core asks for driver.open, driver.close, driver.queryStart, driver.queryAdvance, driver.queryCancel, and driver.queryDispose effects. Effect results later re-enter Core as envelopes.

This is the boundary that keeps I/O out of the reducer while still allowing the reducer to own lifecycle semantics.

### 5. Exactly-once terminal behavior is represented in state

Query state carries phase and CompleteSent explicitly. Dispose, cancel, driver completion, and connection-close races all pass through the same reducer guards.

This is preferable to relying on transport-side suppression because the invariant remains visible in state dumps and replay.

### 6. Capture policy is part of deterministic state

Row/SQL capture mode, maximum allowed capture, and config version live in Core state. Configuration changes produce a trace-visible output and increment a monotonic version.

The host can later impose a privacy ceiling without introducing a second, invisible policy state machine.

## Most interesting areas for review

1. **Reducer transitions in Sts2CoreReducer.cs**
   - Are connection and query terminal states complete?
   - Are cancel, close, dispose, and completion races resolved consistently?
   - Does each accepted request receive exactly one terminal response?

2. **The Core/Runtime boundary**
   - Are all effects declarative and serializable?
   - Is any live resource or ambient observation represented in Core?
   - Are output correlations and causal IDs sufficient for later replay?

3. **Driver abstractions**
   - Is the event stream expressive enough for SQL Server and portable providers?
   - Are error and value types provider-neutral?
   - Are large/vector/spatial value failure modes honest rather than lossy?

4. **Determinism enforcement**
   - Do the banned-symbol lists cover the APIs most likely to create replay drift?
   - Are the analyzer opt-in properties applied to the correct projects?

5. **Public API shape**
   - These are new public surfaces. Naming and ownership are much cheaper to adjust in this PR than after the runtime and providers depend on them.

## Compatibility and risk

- This PR adds new assemblies and solution entries only.
- No existing ServiceLayer project references the new assemblies.
- No command-line option, environment variable, process startup behavior, or stdio behavior changes.
- The primary risk is architectural: an incomplete contract or impure state transition would constrain all later layers.

## Validation

The branch was validated with:

    dotnet build sqltoolsservice-sts2.slnf -c Debug --nologo

    dotnet test test/sts2/Microsoft.SqlTools.Sts2.UnitTests/Microsoft.SqlTools.Sts2.UnitTests.csproj -c Debug --nologo

Results:

- Build succeeded with **0 warnings and 0 errors**.
- **38 tests passed**.
- The worktree and diff checks were clean.

## Out of scope

The following intentionally arrive in later stacked PRs:

- coordinator pump and write-ahead journal
- replay tooling and executable scenarios
- stdio multiplexing
- SQL Client and SQLite implementations
- StreamJsonRpc hosting
- legacy ServiceLayer activation

## Reviewer checklist

- [ ] Contracts contain protocol vocabulary, not implementation behavior.
- [ ] Abstractions expose no provider-specific or live-resource types.
- [ ] Core is synchronous, deterministic, and serializable.
- [ ] State transitions cover cancellation/disposal/close races.
- [ ] Stable error identities are preserved independently of provider failures.
- [ ] Analyzer and architecture tests enforce the stated dependency rules.
- [ ] Public API files reflect an intentional long-term surface.

### DIFF
--- a/sqltoolsservice.sln
+++ b/sqltoolsservice.sln
@@ -1,4 +1,3 @@
-﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.5.33209.295
--- a/sqltoolsservice.sln
+++ b/sqltoolsservice.sln
@@ -1,3 +1,4 @@
+﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.5.33209.295
@@ -102,6 +103,16 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlTools.LanguageService", "src\Microsoft.SqlTools.LanguageService\Microsoft.SqlTools.LanguageService.csproj", "{158EE5EB-B512-4372-AE3B-82941CB6727F}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlTools.LanguageService.UnitTests", "test\Microsoft.SqlTools.LanguageService.UnitTests\Microsoft.SqlTools.LanguageService.UnitTests.csproj", "{6D0421FC-C398-4357-A80D-7B301616632B}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "STS2", "STS2", "{8AFACB56-5A36-36F4-E3C3-98E87E8C2DB5}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.SqlTools.Sts2.Contracts", "src\sts2\Microsoft.SqlTools.Sts2.Contracts\Microsoft.SqlTools.Sts2.Contracts.csproj", "{AA652A61-F7D5-434F-A40C-3CCBD51B2703}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.SqlTools.Sts2.Abstractions", "src\sts2\Microsoft.SqlTools.Sts2.Abstractions\Microsoft.SqlTools.Sts2.Abstractions.csproj", "{FCAE08E9-3B4D-4641-BF7C-0CB0A71252D8}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.SqlTools.Sts2.Core", "src\sts2\Microsoft.SqlTools.Sts2.Core\Microsoft.SqlTools.Sts2.Core.csproj", "{C58D5B7E-1C8C-4A60-86BC-3898793FDD0F}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.SqlTools.Sts2.UnitTests", "test\sts2\Microsoft.SqlTools.Sts2.UnitTests\Microsoft.SqlTools.Sts2.UnitTests.csproj", "{7042FEDD-EDB6-4B8A-AB19-5B9FBF54D2DE}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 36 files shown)


src/sts2/Microsoft.SqlTools.Sts2.Abstractions/DriverPort.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <summary>
+        /// Numeric engine edition when known (serverproperty('EngineEdition')):
+        /// 5 = Azure SQL Database, 8 = Managed Instance, … Clients use this for
+        /// exact platform gating (USE vs reconnect on database switch) — the
+        /// display name cannot distinguish SQL DB from MI (both "SQL Azure").
+        /// </summary>
+        public int? EngineEditionId { get; init; }
+
+        /// <summary>SQL dialect the session speaks.</summary>
+        public required string Dialect { get; init; }
```
Benjin, line 113: Meaning tsql vs. pgsql/mysql/others?  Is the thinking here to leave the door open for using this as a cross-engine core in the future?

kburtram: This code has support for SQL Lite and SQL Client Driver.  I don't think we'd plug pg-sql, etc. into this level of the stack, though I suppose anything is possible.

aasimkhan30 (dismissed):

sqltoolsservice.sln
```diff
@@ -103,6 +104,16 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlTools.Language
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlTools.LanguageService.UnitTests", "test\Microsoft.SqlTools.LanguageService.UnitTests\Microsoft.SqlTools.LanguageService.UnitTests.csproj", "{6D0421FC-C398-4357-A80D-7B301616632B}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "STS2", "STS2", "{8AFACB56-5A36-36F4-E3C3-98E87E8C2DB5}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.SqlTools.Sts2.Contracts", "src\sts2\Microsoft.SqlTools.Sts2.Contracts\Microsoft.SqlTools.Sts2.Contracts.csproj", "{AA652A61-F7D5-434F-A40C-3CCBD51B2703}"
```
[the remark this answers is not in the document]
kburtram: I checked the .NET Project System behavior: because each project declares <Project Sdk="Microsoft.NET.Sdk">, Visual Studio detects and loads it with CPS even when the solution uses the generic C# {FAE04EC0...} project type; {9A19103F...} only forces CPS. See https://github.com/dotnet/project-system/blob/main/docs/opening-with-new-project-system.md. The solution and full stack build cleanly, so I’m leaving these GUIDs unchanged.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1096 @@
...
+            string maxRowCapture = state.MaxRowCapture;
+            string maxSqlCapture = state.MaxSqlCapture;
+            if (envelope.Payload is { ValueKind: JsonValueKind.Object } capturePayload
+                && capturePayload.TryGetProperty("capture", out JsonElement capture)
+                && capture.ValueKind == JsonValueKind.Object)
+            {
+                rowCapture = GetString(capture, "row") ?? rowCapture;
+                sqlCapture = GetString(capture, "sql") ?? sqlCapture;
+                maxRowCapture = GetString(capture, "maxRow") ?? maxRowCapture;
+                maxSqlCapture = GetString(capture, "maxSql") ?? maxSqlCapture;
```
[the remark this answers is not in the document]
kburtram: Fixed in 471dbd82. session.start now validates all four capture modes and both policy relationships. Invalid or self-contradictory tuples emit a deterministic diagnostic and fail closed to digest-only capture and ceilings. Six regression cases cover invalid values, policy violations, and a subsequent no-op setCapture. The full rewritten stack passes 361 unit/performance-smoke tests and 5 E2E tests.

src/sts2/Microsoft.SqlTools.Sts2.Abstractions/DriverPort.cs
```diff
@@ -0,0 +1,190 @@
...
+    {
+        /// <summary><c>sqlLogin</c>, <c>accessToken</c>, or <c>integrated</c>.</summary>
+        public required string Kind { get; init; }
+
+        /// <summary>User name for sqlLogin, display user otherwise.</summary>
+        public string? User { get; init; }
+
+        /// <summary>Password or token; null for integrated auth. Never logged or journaled.</summary>
+        public string? Secret { get; init; }
+    }
```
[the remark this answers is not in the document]
kburtram: Fixed in d5a26086. SecretMaterial now overrides ToString() with the credential value redacted. Regression tests verify that neither direct SecretMaterial formatting nor nested ConnectionOpenRequest formatting contains the password/token. The rewritten full stack passes restore, both STS2 and legacy ServiceLayer builds with 0 warnings/errors, 363 unit tests, and 5 E2E tests.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1121 @@
...
+    public static class Sts2CoreReducer
+    {
+        /// <summary>Decides the next state and outputs for one journaled input envelope.</summary>
+        public static CoreDecision Decide(CoreState state, CoreEnvelope envelope)
+        {
+            ArgumentNullException.ThrowIfNull(state);
+            ArgumentNullException.ThrowIfNull(envelope);
+
+            CoreState advanced = state with { LastSeq = envelope.Seq };
+            return envelope.Kind switch
```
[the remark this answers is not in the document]
kburtram: Fixed in 1b2f3438. Decide now emits a deterministic core.unexpectedInput diagnostic for an envelope whose sequence is below LastSeq and returns the existing state, so LastSeq cannot regress. Added regression coverage.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1121 @@
...
+                }
+
+                case "rows":
+                {
+                    QueryInfo updated = query with
+                    {
+                        PagesSent = query.PagesSent + 1,
+                        CreditOutstanding = Math.Max(0, query.CreditOutstanding - 1),
+                    };
+                    CoreState next = state with { Queries = state.Queries.SetItem(queryId, updated) };
```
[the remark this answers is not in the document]
kburtram: Fixed in 1b2f3438. A rows event with no outstanding credit now emits core.unexpectedInput and is ignored, preserving page counters and the configured window invariant. Added regression coverage.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1131 @@
...
+            string effectId = string.Create(CultureInfo.InvariantCulture, $"drv-qdispose-{envelope.Seq}");
+            string args = string.Create(CultureInfo.InvariantCulture, $$"""{"queryId":{{JsonSerializer.Serialize(queryId)}}}""");
+
+            // The query already emitted its terminal (completed/error/canceled). Dispose just
+            // releases resources — mark Disposed, free the connection, no second complete.
+            if (query.CompleteSent)
+            {
+                CoreState done = state with
+                {
+                    Queries = state.Queries.SetItem(queryId, query with { Phase = QueryPhase.Disposed }),
```
kburtram: [P1] Disposed queries are never removed from Core state. Both this completed-query path and the active-dispose acknowledgement replace the entry with Phase = Disposed, but nothing ever removes it. vscode-mssql disposes every completed query, so a long-lived service accumulates one immutable-map entry per execution forever; diagnostics.state, state copies, and health work all grow with total historical queries. CoreState.Queries even documents retention only until disposal, and the journal already preserves history. Please remove the tombstone once disposal is complete (keeping a separate cumulative counter if totalQueries must remain cumulative) and add a many-query regression.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1131 @@
...
+            {
+                // High-water (0-based pageSeq). Clamp to [current, PagesSent] so a duplicate,
+                // out-of-order, or impossibly-large client ack can never push pagesAcked past
+                // what was actually sent and over-grant credit beyond the window (I9, R011).
+                int requested = throughPageSeq >= 0 ? throughPageSeq + 1 : pagesAcked;
+                pagesAcked = Math.Min(query.PagesSent, Math.Max(pagesAcked, requested));
+            }
+            else
+            {
+                pagesAcked = Math.Min(query.PagesSent, pagesAcked + 1); // per-page credit
```
kburtram: [P1] The per-page ack form ignores its pageSeq and blindly increments PagesAcked. Replaying the same pageSeq 0 while several pages are outstanding therefore acknowledges a different page each time and grants fresh driver credit for a page the client never accepted. The numeric window still looks bounded in Core, but the actual unacked pages can exceed I9. Treat the per-page ordinal as an idempotent per-query high-water (D-0015), or otherwise track its identity; malformed or missing ordinals should not grant credit. Please cover a duplicate per-page ack before all outstanding pages are acknowledged.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1131 @@
...
+        private static CoreDecision DecideRequest(CoreState state, CoreEnvelope envelope)
+        {
+            if (envelope.Corr is null)
+            {
+                return Unexpected(state, envelope, "request without corr");
+            }
+
+            // SPEC §7.1: unknown fields are ignored unless they demand understanding.
+            if (envelope.Payload is { ValueKind: JsonValueKind.Object } payload)
+            {
```
kburtram, line 56: [P2] mustUnderstand_ is checked only on the request root. A client can send an unknown field such as options.mustUnderstand_newPagingMode or profile.mustUnderstand_newAuthRule, and the service silently ignores it even though SPEC section 7.1 promises rejection for request fields carrying this prefix. That defeats the compatibility escape hatch precisely for most extensible request shapes. Please validate nested protocol objects recursively with a regression, or explicitly narrow and document the contract if root-only behavior is intended.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1131 @@
...
+    /// </summary>
+    public static class Sts2CoreReducer
+    {
+        /// <summary>Decides the next state and outputs for one journaled input envelope.</summary>
+        public static CoreDecision Decide(CoreState state, CoreEnvelope envelope)
+        {
+            ArgumentNullException.ThrowIfNull(state);
+            ArgumentNullException.ThrowIfNull(envelope);
+
+            if (envelope.Seq < state.LastSeq)
```
kburtram: [P2] This stale-sequence guard still accepts envelope.Seq equal to state.LastSeq. Since generated connection, query, and effect IDs derive from seq, processing a duplicate sequence can re-emit a terminal response or attempt to add the same logical ID twice. I5 says journal sequence is gapless and monotonic; at minimum, equal sequences should be diagnosed just like lower ones and covered by the sequencing regression.

src/sts2/BannedSymbols.Core.txt
```diff
@@ -0,0 +1,19 @@
...
+T:System.Threading.ThreadPool;Core is synchronous (SPEC §9.3.3); no thread pool
+T:System.Threading.Timer;Core must not schedule time (SPEC §9.3.3); timers are envelopes
+T:System.Threading.PeriodicTimer;Core must not schedule time (SPEC §9.3.3); timers are envelopes
+T:System.Threading.Channels.Channel;Core must not use channels (SPEC §9.3.3)
+T:System.Threading.Channels.Channel`1;Core must not use channels (SPEC §9.3.3)
+T:System.Threading.Monitor;Core is single-threaded synchronous (SPEC §9.3.3); no locks
+T:System.Threading.CancellationToken;Cancellation is a journaled message, not a token (SPEC §9.3.6)
+T:System.IO.File;Core performs no I/O (SPEC §9.2)
+T:System.IO.Directory;Core performs no I/O (SPEC §9.2)
+T:System.IO.Stream;Core performs no I/O (SPEC §9.2)
```
kburtram, line 21: [P2] The analyzer list does not enforce the I/O and network boundary described by this PR and SPEC sections 4 and 9. It bans File, Directory, and the abstract Stream, but Core can still compile FileInfo, DirectoryInfo, FileStream, PipeStream, HttpClient, sockets, and other concrete I/O APIs; several synchronization primitives are similarly outside the list. Because the PR presents this as a build-enforced boundary, please broaden the banned symbols and ideally add a source or metadata architecture regression so the stated guarantee cannot be bypassed accidentally.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
kburtram: Fixed in d2e9873c. The reducer rejects every non-increasing sequence, including equality, without advancing state or re-emitting an RPC result. The equal-sequence regression is included with the existing stale-sequence coverage.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
kburtram: Fixed in d2e9873c. Required-extension validation now recursively walks nested objects and arrays. Regressions cover options, nested profile/auth data, and objects inside extension arrays.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
kburtram: Fixed in d2e9873c. Both completed-query disposal and active-dispose acknowledgement now remove the query from Core state; unknown IDs retain the existing idempotent behavior and the journal remains the history. A 1,000-query regression also covers late cleanup acks and repeated dispose requests, and proves the query map remains empty.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
kburtram: Fixed in d2e9873c. QueryInfo now retains only the window-bounded outstanding-page ledger, keyed by cumulative per-query ordinal with resultSetId:pageSeq identities. Per-page acks remove only a matching outstanding page; duplicate, malformed, future, mixed high-water, and cross-result-set cases are covered without granting false credit.

src/sts2/BannedSymbols.Core.txt
kburtram: Fixed in d2e9873c. Core now bans the System.IO, System.Net, and System.Threading namespace families, with an architecture regression pinning the build configuration. I also ran a negative compile probe: FileInfo, HttpClient, and SemaphoreSlim each failed with RS0030. Release build is clean and all 55 foundation tests pass.

src/sts2/Microsoft.SqlTools.Sts2.Core/Sts2CoreReducer.cs
```diff
@@ -0,0 +1,1234 @@
...
+                return Unexpected(state, envelope, "driver.open result for unknown or non-opening connection");
+            }
+
+            string status = GetString(envelope.Payload, "status") ?? "error";
+            if (status == "ok")
+            {
+                string? handleId = GetString(envelope.Payload, "handleId");
+                string serverInfo = envelope.Payload!.Value.TryGetProperty("serverInfo", out JsonElement si)
+                    ? si.GetRawText()
+                    : "null";
```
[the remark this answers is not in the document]
kburtram: Fixed in 65981c0f. A successful driver.open now requires a nonblank handle before Core admits the connection; malformed responses remove both connection/open-id state and deterministically fail the original open request with Sts2.Internal, so no later effect can carry a null handle. Four missing/null/empty/whitespace regressions pass, as does the full 59-test foundation suite and a zero-warning Release build. Resolving.

sqltoolsservice.sln
```diff
@@ -1,3 +1,4 @@
+﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
kburtram: Fixed in 24ffea3f. The accidental BOM-only blank line is removed, so byte zero is now M from the required Microsoft Visual Studio Solution File header. dotnet sln sqltoolsservice.sln list parses the corrected file successfully and diff hygiene is clean. Resolving.

src/sts2/Microsoft.SqlTools.Sts2.Core/CoreState.cs
```diff
@@ -0,0 +1,194 @@
...
+        /// through construction — otherwise replay would start from a different state
+        /// than the live run did.
+        /// </summary>
+        public static CoreState Initial { get; } = new()
+        {
+            LastSeq = 0,
+            ShuttingDown = false,
+            Initialized = false,
+            ServiceVersion = "0.0.0.0",
+            MaxConnections = Contracts.Sts2Defaults.MaxConnections,
```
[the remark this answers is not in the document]
kburtram: No code change. C# simple-name namespace lookup checks enclosing namespaces, so inside Microsoft.SqlTools.Sts2.Core the qualifier Contracts resolves to the sibling Microsoft.SqlTools.Sts2.Contracts namespace. Core has the corresponding ProjectReference, and the exact Core/foundation Release build succeeds with zero warnings and zero errors. This is a compile-analysis false positive; resolving.

aasimkhan30 (approved):

PR #2783, opened 2026-08-07
STS2 runtime — write-ahead coordination, replay, and executable scenarios

kburtram:
# PR: STS2 runtime — write-ahead coordination, replay, and executable scenarios

## Stack position

- Branch: **dev/karlb/sts2_runtime**
- Commit: **15515a23**
- PR base: **dev/karlb/sts2_foundation**
- Review order: **2 of 5**
- Change size: **138 files, 13,239 insertions, 2 deletions**

## Summary

This PR turns the pure STS2 state machine from the foundation PR into an executable, observable, and replayable runtime.

The main additions are:

- a single ordered coordinator pump
- write-ahead envelope journaling
- deterministic canonical JSON and payload digests
- effect execution and resource ownership
- query streaming with explicit page credit
- secret tokenization and capture elision
- replay, diff, explain, and export validation tooling
- live-tail and metrics sinks
- a reusable testing assembly
- 54 executable YAML scenarios and invariant checking
- generated contract/state-machine/review documentation

There is still no stdio multiplexer, concrete SQL provider, StreamJsonRpc host, or ServiceLayer startup seam. The runtime is exercised through an in-memory RPC edge and FakeDriver, which keeps this PR focused on ordering, durability, privacy, replay, and lifecycle correctness.

### Ordered coordinator pump

**Coordinator** is the single-threaded heart of the runtime. Producers post RPC inputs, effect observations, and lifecycle controls to a bounded Channel. One reader processes them in order.

For each input the pump:

1. assigns run ID, sequence, timestamp, cause, configuration version, and digest
2. applies capture/redaction rules
3. appends the input envelope to the journal
4. invokes the pure Core reducer
5. journals every Core output
6. only then emits RPC output or schedules an effect

This ordering makes the journal the truth. An outbound result or driver action should never become externally visible before the decision that caused it is durable in the trace.

The bounded queue uses wait-on-full behavior rather than unbounded allocation or silent dropping. The coordinator exposes queue depth and fault counters for health reporting.

### Envelope and causal model

The runtime defines one closed envelope shape for inbound RPC, outbound RPC, effects, controls, configuration changes, metrics, diagnostics, and snapshots.

Every envelope includes:

- monotonically increasing sequence
- run ID
- timestamp
- kind and type
- correlation ID when applicable
- causal sequence link
- configuration version
- canonical payload digest
- sanitized payload

The cause field makes it possible to reconstruct why an output or observation exists instead of relying on timing proximity.

### Write-ahead journal

Journals are newline-delimited canonical JSON with:

- configurable segment rotation
- bounded deferred flushing
- active-segment reading
- per-segment byte counts and SHA-256 hashes
- a hash chain across segments
- a manifest with run/build/environment metadata

Lifecycle barriers are stronger than enqueue acknowledgements. A barrier completes only after the control envelope and all causally prior outputs have been journaled and flushed. This is important for shutdown paths where the process may terminate immediately after legacy handling continues.

### Deterministic JSON and replay

**CanonicalJson** produces stable UTF-8 representations and digests independent of property insertion order or culture.

**JournalReplayer** rebuilds Core state from the recorded envelope stream and checks that replayed outputs match the recorded outputs. Metric-only envelopes are intentionally ignored by Core replay because they describe runtime behavior rather than state-machine input.

The **sts2-replay** CLI provides:

- run — replay one journal
- verify — verify all journals under a directory
- until — reconstruct state at a sequence
- diff — locate the first divergence
- explain — render the causal tree around an envelope
- export-check — validate an export bundle’s hashes, privacy report, and replayability

Replay is treated as an executable correctness property, not merely a diagnostic reader.

### Effect runner and resource ownership

**DriverEffectRunner** owns all live driver sessions, open attempts, query pumps, and cancellation sources. Core only receives serializable handle IDs.

The runner:

- resolves driver.open/close/query effects
- posts every observation back through the coordinator inbox
- tracks open leases and in-flight opens
- runs forward-only query pumps
- enforces page-credit backpressure
- bounds cancellation and close behavior
- waits for a query pump to unwind before acknowledging disposal
- suppresses output once disposal has taken ownership of a query

This separates “what should happen” in Core from “perform the I/O and report what happened” in Runtime.

### Backpressure

Core grants page credit through driver.queryAdvance effects. The runner must acquire one credit before it can post one row page.

Acknowledgement can be per-page or high-water. The state machine tracks pages sent, pages acknowledged, and outstanding credit so window behavior remains replay-visible.

The design deliberately avoids “read everything and buffer later.” Provider implementations can stream forward-only and stop pulling when the client’s window is exhausted.

### Wire value encoding

**WireValueEncoder** converts provider-neutral values into stable JSON forms:

- lossless JSON-native primitives remain native
- decimal, date/time, GUID, binary, and non-finite floating-point values use explicit typed wrappers
- large strings and binary values use bounded prefixes plus honest total-byte and digest metadata
- UTF-8 truncation never splits a code point
- vector and spatial representations are complete or explicitly unavailable

The encoder is provider-neutral; concrete provider extraction arrives in the driver PR.

### Privacy boundaries

Secrets are removed before an RPC payload becomes an envelope.

**SecretSideTable** stores the actual material in memory under opaque, random, non-correlatable tokens. Only the effect edge can resolve those tokens. The table is never serialized, logged, exported, or exposed through state diagnostics.

**SecretRedactor** treats known credential keys and non-public fields under auth objects as secret material. It also tracks newly created tokens so rejected requests do not leave entries behind.

Capture elision applies the same “journal less than the wire sees” principle to SQL text and row cells. The journal retains a deterministic digest wrapper, while a short-lived side map substitutes the original fragment back at the effect or RPC edge. Captured fragments are released after the turn.

### Observability

The journal is the privileged first sink. Auxiliary observers run afterward and are fault-isolated.

Built-in observers include:

- **MetricsEnvelopeSink** for totals, kinds, and stable error-code histograms
- **BroadcastEnvelopeSink** for live-tail subscribers
- **MailboxEnvelopeSink** for bounded test/diagnostic capture
- **Sts2EventSource** for process-level counters

Live-tail subscribers have bounded drop-oldest buffers. A slow viewer cannot block the coordinator; it receives a drop count and can re-synchronize from the journal.

Runtime-only facts such as queue depth, open leases, pump count, and sink/emit/effect faults are overlaid on live health/state responses. They are deliberately not written into the deterministic Core result.

### Export bundles

**ExportBundleWriter** packages journal segments, manifest information, generated docs, and a privacy report. Validation checks hashes, canary results, and replayability.

The export path never makes unavailable source data appear: asking to include SQL cannot recover text that capture policy already reduced to a digest.

### Testing infrastructure

The new **Microsoft.SqlTools.Sts2.Testing** assembly includes:

- FakeDriver and scripted query behavior
- deterministic connection simulator
- YAML scenario parser and runner
- invariant checker
- secret canaries
- scenario catalog
- generated review-document builders

Each active scenario executes against the real coordinator, Core, journal, and FakeDriver; then its journal is invariant-checked and replayed.

The 54-scenario corpus covers normal operation and adversarial races, including:

- duplicate, late, unknown, and high-water acknowledgements
- cancellation versus completion
- close/dispose while streaming
- connection limits and duplicate open IDs
- auth/network/timeout/query failure classification
- slow consumers and exhausted windows
- multiple result sets and huge pages
- capture changes
- secret canaries
- replay digests
- shutdown/exit during work

### Generated review docs

The branch commits generated human review surfaces:

- CONTRACT.md
- TRACE-SCHEMA.md
- INVARIANTS.md
- SCENARIO-MATRIX.md
- STATE-MACHINE.md
- COMPONENTS.md

Tests compare committed files to the generators. Changing protocol constants, scenarios, state-machine text, or project references without regenerating the docs fails validation.

### 1. Journal before action

Both inputs and outputs are appended before external action. Auxiliary sinks also run only after the journal append.

This is the most important ordering guarantee in the PR. Reviewers should look for any RPC emission, effect dispatch, or observer callback that bypasses it.

### 2. One serialized decision lane

All state-machine input passes through one coordinator reader. Driver operations remain asynchronous, but their observations return to the same ordered inbox.

This replaces lock-based shared state with explicit message ordering and makes races reproducible as envelope order.

### 3. Runtime facts do not pollute deterministic state

Queue depth, task/pump counts, elapsed time, allocations, and observer faults are useful operational data but not Core inputs. They are composed at the runtime edge or emitted as replay-skipped metric envelopes.

### 4. Privacy is enforced before journaling

The journal never receives raw credentials. Product capture can also retain only SQL/row digests while still sending the original material to the provider/client during the live turn.

### 5. Backpressure controls provider pulling

Credit gates the query enumerator rather than merely throttling outbound writes. This bounds work and memory closer to the data source.

### 6. Scenario artifacts are first-class evidence

The scenario runner produces actual journals and immediately validates invariants and replay. The tests therefore exercise the same forensic artifacts intended for diagnosis.

## Most interesting areas for review

1. **Coordinator ordering**
   - Is every external action preceded by the correct journal append?
   - Are barrier and flush semantics strong enough?
   - Can auxiliary observers stall or fault the pump?

2. **Capture elision and secret lifetime**
   - Can raw values reach a journal, metric, state dump, or export?
   - Are tokens/fragments released on rejection, completion, cancellation, and failure?
   - Does substitution occur only at trusted edges?

3. **Query runner races**
   - Cancel versus complete
   - Dispose while streaming
   - Connection close with an active query
   - Pump teardown before resource acknowledgement

4. **Journal integrity**
   - Active segment behavior
   - rotation and hash chaining
   - manifest updates and crash recovery
   - replay treatment of runtime-only envelopes

5. **Wire value honesty**
   - exact byte limits
   - UTF-8 prefix safety
   - complete-or-unavailable vector/spatial handling
   - deterministic typed wrappers

6. **Scenario/invariant coverage**
   - The corpus is large by design. Review whether important state-machine branches are represented and whether invariant checks fail closed on unknown identifiers.

## Compatibility and risk

- No existing ServiceLayer startup path changes.
- No stdio ownership or routing changes.
- No real database provider is introduced.
- The principal risks are durability ordering, secret leakage, resource lifetime, and race behavior.
- The large change size is primarily the runtime implementation, focused tests, and 54 scenario files; it does not include unrelated refactoring.

## Validation

Commands used:

    dotnet build sqltoolsservice-sts2.slnf -c Debug --nologo

    dotnet test test/sts2/Microsoft.SqlTools.Sts2.UnitTests/Microsoft.SqlTools.Sts2.UnitTests.csproj -c Debug --no-build --nologo --filter "FullyQualifiedName!~Perf"

    dotnet test test/sts2/Microsoft.SqlTools.Sts2.UnitTests/Microsoft.SqlTools.Sts2.UnitTests.csproj -c Debug --no-build --nologo --filter "FullyQualifiedName~Perf"

    .\scripts\update-sts2-docs.ps1

Results:

- Build succeeded with **0 warnings and 0 errors**.
- **239 non-performance tests passed**.
- **1 performance smoke test passed**.
- Generated-document verification passed.
- Diff checks were clean.

## Out of scope

- routing STS2 and legacy traffic over one stdio pair
- concrete SQL Client or SQLite I/O
- StreamJsonRpc method exposure
- production process composition
- activation from the existing ServiceLayer executable

## Reviewer checklist

- [ ] Journal-before-action ordering holds for every input and output kind.
- [ ] Core remains the sole owner of logical state transitions.
- [ ] Runtime alone owns tasks, channels, handles, tokens, and streams.
- [ ] Secrets and elided values cannot enter persisted or diagnostic surfaces.
- [ ] Backpressure gates provider enumeration.
- [ ] Cancellation/disposal cleanly unwind query pumps and leases.
- [ ] Replay compares the same deterministic decisions made live.
- [ ] Generated docs and scenario/invariant coverage match the implementation.


src/sts2/Microsoft.SqlTools.Sts2.Testing/SecretCanaries.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Microsoft.SqlTools.Sts2.Testing
+{
+    /// <summary>
+    /// Known fake secrets planted by tests (SPEC §8.5). The canary scan greps every
+    /// produced artifact for these values; any hit fails the I6 secret-safety invariant.
+    /// </summary>
+    public static class SecretCanaries
+    {
+        /// <summary>A fake SQL login password.</summary>
+        public const string Password = "CANARY-pw-1f9b3c7d2e";
```
kburtram, line 19: This is will likely trigger a CodeQL warning.  I'll see if I can track down the suppression snippet.  But might do as a follow-up PR if it doesn't block the build pipelines.

tools/sts2-replay/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+                    ["run", string path] => Run(path),
+                    ["verify", string path] => Verify(path),
+                    ["until", string path, "--seq", string seq] => Until(path, long.Parse(seq, CultureInfo.InvariantCulture)),
+                    ["diff", string path] => Diff(path),
+                    ["explain", string path, "--seq", string seq] => Explain(path, long.Parse(seq, CultureInfo.InvariantCulture)),
+                    ["export-check", string bundle] => ExportCheck(bundle),
+                    _ => Usage(),
+                };
+            }
+            catch (Exception ex) when (ex is IOException or InvalidDataException or FormatException)
```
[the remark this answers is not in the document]
kburtram: Fixed in 462deb62. Sequence parsing now uses long.TryParse, so malformed and out-of-range values become a clean FormatException handled by the CLI boundary. UnauthorizedAccessException is also handled, and explain now detects duplicate sequence numbers explicitly and reports an InvalidDataException. These paths return exit code 2 instead of crashing.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Observability/MailboxEnvelopeSink.cs
```diff
@@ -0,0 +1,94 @@
...
+        public ValueTask OnEnvelopeAsync(Sts2Envelope envelope, bool flush)
+        {
+            // Non-blocking: the pump never awaits the inner sink. On overflow, drop the oldest
+            // to keep the freshest (a live observer wants recency) and count the drop.
+            if (!mailbox.Writer.TryWrite((envelope, flush)))
+            {
+                mailbox.Reader.TryRead(out _);
+                Interlocked.Increment(ref dropped);
+                mailbox.Writer.TryWrite((envelope, flush));
+            }
```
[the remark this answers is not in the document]
kburtram: Fixed in 462deb62. The mailbox now uses the channel's atomic DropOldest mode with an item-dropped callback. The callback increments Dropped only when the channel actually evicts an item, eliminating the producer/consumer TryRead race while keeping publication non-blocking.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Observability/EnvelopeSubscription.cs
```diff
@@ -0,0 +1,77 @@
...
+            ChannelWriter<Sts2Envelope> writer = channel.Writer;
+            if (writer.TryWrite(envelope))
+            {
+                return true;
+            }
+            // Full: evict the oldest then admit the newest (single-writer guarantees room).
+            channel.Reader.TryRead(out _);
+            Interlocked.Increment(ref dropped);
+            writer.TryWrite(envelope);
+            return false;
```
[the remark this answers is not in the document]
kburtram: Fixed in 462deb62. The subscription now uses atomic DropOldest with a callback that updates both the per-subscription and broadcast aggregate counters only for a real eviction. The producer-side TryRead and its race-prone boolean result were removed.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Observability/MetricsEnvelopeSink.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.SqlTools.Sts2.Runtime.Envelopes;
+
+namespace Microsoft.SqlTools.Sts2.Runtime.Observability
+{
+    /// <summary>
+    /// The metrics channel (SPEC §12.3): tallies the envelope stream by kind and counts
+    /// outbound errors by code, feeding both the process-wide <see cref="Sts2EventSource"/>
+    /// and the in-memory snapshot that health and <c>metric</c> envelopes report. Counting
+    /// is lock-free and never blocks the pump.
+    /// </summary>
```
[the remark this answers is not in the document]
kburtram: No code change. The wording is imprecise, but the implementation intentionally uses a short synchronous lock only around its two dictionaries; OnEnvelopeAsync performs no await or blocking I/O. This is documentation-only churn with no pump-correctness impact, so I am closing it under the stricter review bar.

test/sts2/Microsoft.SqlTools.Sts2.UnitTests/Scenarios/SimulatorSeedRegressionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                    Task<SimulatorResult> run = ConnectionSimulator.RunSeedAsync(7496, Path.Combine(root, "iter-" + i));
+                    // A wedged run stalls for its full 90s drain/settle budget before
+                    // reporting violations; a 60s guard fails fast and names the hang
+                    // instead of eating the whole gate.
+                    Task winner = await Task.WhenAny(run, Task.Delay(TimeSpan.FromSeconds(60)));
+                    if (winner != run)
+                    {
+                        failures.Add($"iter {i}: wedged (>60s; journals under {root})");
+                        continue;
+                    }
```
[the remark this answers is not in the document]
kburtram: Fixed in 435122d6. The regression sweep now stops after the first 60-second timeout instead of starting additional uncancelable simulator runs. The focused regression and full 251-test runtime suite pass.

test/sts2/Microsoft.SqlTools.Sts2.UnitTests/Perf/PerfSmokeTests.cs
```diff
@@ -0,0 +1,97 @@
...
+                        $$"""{"queryId":"{{queryId}}","throughPageSeq":{{(acked - 1).ToString(CultureInfo.InvariantCulture)}}}""");
+                }
+                else
+                {
+                    await Task.Delay(1);
+                }
+            }
+            stopwatch.Stop();
+
+            int pagesDelivered = session.Emitted.Count(m => m.Type == "v2/query.rows");
```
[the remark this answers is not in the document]
kburtram: No code change. The queue is bounded by emitted page messages (about 1,000 here), not the one million individual rows, and this performance-only harness passes its 50k rows/s gate in the full suite. Consolidating these small snapshot enumerations is not important enough for another review-churn cycle.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Envelopes/EnvelopeJsonCodec.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        private static string GetRequiredString(JsonElement root, string name) =>
+            root.TryGetProperty(name, out JsonElement value) && value.ValueKind == JsonValueKind.String
+                ? value.GetString()!
+                : throw new InvalidDataException($"Envelope line is missing required string field '{name}'.");
+
+        private static string? GetOptionalString(JsonElement root, string name) =>
+            root.TryGetProperty(name, out JsonElement value) && value.ValueKind == JsonValueKind.String
+                ? value.GetString()
+                : null;
```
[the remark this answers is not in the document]
kburtram: Fixed in 649bda42. Optional envelope identifiers now accept only an absent field, null, or a JSON string; any other JSON type produces InvalidDataException. Regression coverage includes both sessionId and corr.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Journaling/JournalReader.cs
```diff
@@ -0,0 +1,110 @@
...
+                .Distinct(StringComparer.Ordinal)
+                .ToArray();
+            if (runs.Length > 1)
+            {
+                throw new InvalidDataException(
+                    $"Journal directory '{directory}' contains {runs.Length} runs ({string.Join(", ", runs)}); " +
+                    "reads must be scoped to a single run. Use ReadRun(directory, runId) or one directory per run.");
+            }
+
+            return ReadSegments(segments);
```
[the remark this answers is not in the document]
kburtram: Fixed in 649bda42. ReadAll now rejects every glob-matching segment whose file name cannot be parsed as journal-{runId}-{NNNN}.jsonl before run detection or content reading, preventing malformed names from bypassing run isolation. Added regression coverage.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Journaling/JournalWriter.cs
```diff
@@ -0,0 +1,182 @@
...
+            await segmentStream!.WriteAsync(line).ConfigureAwait(false);
+            segmentHash!.AppendData(line);
+            segmentBytes += line.Length;
+
+            DateTimeOffset now = options.TimeProvider.GetUtcNow();
+            if (flush || (now - lastFlush).TotalMilliseconds >= options.FlushIntervalMs)
+            {
+                await segmentStream.FlushAsync().ConfigureAwait(false);
+                lastFlush = now;
+            }
```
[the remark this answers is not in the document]
kburtram: Fixed in 79fc428f. Explicit checkpoint appends, the public lifecycle/barrier flush, and segment close now call FileStream.Flush(flushToDisk: true), matching the promised stable-storage durability semantics. Periodic non-checkpoint flushing remains asynchronous. Focused durability tests and the full runtime suite pass.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Export/ExportBundleWriter.cs
```diff
@@ -0,0 +1,250 @@
...
+            {
+                File.Delete(bundlePath);
+            }
+
+            var canaryHits = new List<string>();
+            var segmentEntries = new JsonArray();
+
+            using (var zip = ZipFile.Open(bundlePath, ZipArchiveMode.Create))
+            {
+                foreach (string segment in Directory.EnumerateFiles(request.JournalDirectory).Order(StringComparer.Ordinal))
```
kburtram: [P1] Exclude the output bundle from the source enumeration. The production Sts2Session leaves ExportDirectory unset, so OutputDirectory defaults to JournalDirectory. Write opens sts2-export-<runId>.zip at line 78 and then EnumerateFiles on that same directory here, which includes the open output ZIP. File.ReadAllBytes will either fail on the writer share or copy a partial version of the bundle into itself; either way v2/diagnostics.exportLog cannot produce a valid production bundle. The current test passes a separate export directory and misses the product wiring. Please enumerate only this run's journal/diagnostic inputs (or stage the ZIP outside the source directory) and add a same-directory integration test.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/WireValueEncoder.cs
```diff
@@ -0,0 +1,380 @@
...
+                    return;
+                case Abstractions.DriverSpatialUnavailableValue spatialUnavailable:
+                    WriteSpatialUnavailable(writer, spatialUnavailable);
+                    return;
+
+                // Lossless JSON natives.
+                case bool b:
+                    writer.WriteBooleanValue(b);
+                    return;
+                case long l:
```
kburtram: [P1] Do not send arbitrary Int64 values as JSON numbers. SqlClient returns SQL bigint as System.Int64, so long.MaxValue is written as 9223372036854775807. vscode-mssql parses that token in Node as 9223372036854776000 before its number:approx hint can stringify it, permanently changing the grid/copy/export value. The dev/query TS-native backend deliberately carries bigint as an invariant string for this reason. Keep only JavaScript-safe integers native and encode out-of-range longs as invariant text/a typed wrapper. Add a cross-boundary test; the existing type-edge scenario injects prebuilt JsonNode wrappers and bypasses this encoder path.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/DriverEffectRunner.cs
```diff
@@ -0,0 +1,1389 @@
...
+        /// <summary>
+        /// Cancels any running query pumps and in-flight opens, disposes any sessions
+        /// still held, and returns how many sessions there were. Called at run end:
+        /// cancelling the pumps releases their credit-semaphore waits so no background
+        /// task is orphaned (a leak found by the 200-seed simulator). A non-zero return
+        /// is an I8 lease violation for scenarios that closed all their connections.
+        /// </summary>
+        public async ValueTask<int> DisposeLeakedSessionsAsync()
+        {
+            foreach (string queryId in queryPumps.Keys.ToArray())
```
kburtram: [P1] Teardown must await and dispose the work it owns before returning. This loop removes each query pump and cancels it, but DisposeLeakedSessionsAsync never awaits PumpTask and never disposes the pump CancellationTokenSource or SemaphoreSlim. Production Sts2Session closes the coordinator first, so a still-running enumerator then continues past session disposal and can only post into a closed inbox. Open tasks are likewise started without a tracked Task. Collect and cancel the live work, bounded-await its unwind, dispose the per-query resources, and make I8 assert that both open tasks and pump tasks are actually gone.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/DriverEffectRunner.cs
```diff
@@ -0,0 +1,1389 @@
...
+                            break;
+
+                        default:
+                            break;
+                    }
+                    lastEventDoneTicks = Stopwatch.GetTimestamp();
+                }
```
kburtram, line 645: [P1] A driver stream that ends normally without yielding ExecCompleted leaves the accepted query permanently unterminated. The finally block removes the runner pump, but Core remains Running, vscode-mssql waits forever for v2/query.complete, and the connection remains Busy. IDbSession.ExecuteAsync does not contractually guarantee that terminal event, so an empty iterator or provider regression reaches this path. Track whether a terminal event was observed and post a classified error on clean EOF; add a FakeDriver script with no terminal step.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Replay/JournalReplayer.cs
```diff
@@ -0,0 +1,223 @@
...
+                    case EnvelopeKinds.ConfigChanged:
+                    {
+                        if (pendingOutputs.Count == 0)
+                        {
+                            return Diverged(envelope.Seq, causeBySeq, outboundDigests, state, lastSeq,
+                                recorded: $"{envelope.Kind}/{envelope.Type} digest={envelope.Digest}",
+                                replayed: "no output at this causal position");
+                        }
+                        (string kind, string type, string? corr, string digest) = pendingOutputs.Dequeue();
+                        if (kind != envelope.Kind || type != envelope.Type || corr != envelope.Corr || digest != envelope.Digest)
```
kburtram: [P1] Replay verifies the stored digest string, not the recorded payload or causal identity. envelope.Payload is never rehashed, envelope.Cause is not compared with the input that produced the queued output, and the loop does not enforce one run with gapless sequences. Changing a recorded output payload while leaving Digest unchanged therefore still returns Verified; exporting that corrupted source simply computes a new segment hash and blesses it. Recompute every recorded payload digest and validate it against Digest, then enforce seq/run/cause invariants before reduction.

tools/sts2-replay/Program.cs
```diff
@@ -0,0 +1,189 @@
...
+        private static int Run(string path)
+        {
+            ReplayResult result = JournalReplayer.Replay(JournalReader.ReadAll(path));
+            foreach (string digest in result.OutboundDigests)
+            {
+                Console.WriteLine(digest);
+            }
+            if (!result.Identical)
+            {
+                Console.Error.WriteLine($"DIVERGED at seq {result.Divergence!.Seq}: recorded {result.Divergence.Recorded}; replayed {result.Divergence.Replayed}");
```
kburtram: [P1] Handle ReplayOutcome.Incomplete without dereferencing Divergence. Incomplete intentionally has Divergence == null, so a truncated journal reaches this line and Run crashes with NullReferenceException instead of explaining the recoverable tail condition. Verify and Diff make the same assumption. Branch on outcome and report PendingOutputCount/LastSeq for incomplete input, with CLI tests for run, verify, and diff over a truncated journal.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Observability/MailboxEnvelopeSink.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        /// <inheritdoc/>
+        public async ValueTask DisposeAsync()
+        {
+            mailbox.Writer.TryComplete();
+            try
+            {
+                await worker.ConfigureAwait(false);
```
kburtram: [P1] Disposal is unbounded even though this wrapper explicitly supports a hanging inner sink. Production Sts2Session awaits each MailboxEnvelopeSink.DisposeAsync during shutdown; if a third-party observer is blocked in OnEnvelopeAsync, worker never completes here and the whole STS2/bootstrap shutdown hangs forever. The existing hanging-sink test releases its gate before disposal, so it misses this path. Make shutdown best-effort with a bounded wait/cancellation strategy and test disposal while the inner sink remains wedged.

src/sts2/Microsoft.SqlTools.Sts2.Testing/InvariantChecker.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                switch (invariant)
+                {
+                    case "I1": // request terminality
+                        if (terminalsByCorr is null)
+                        {
+                            violations.Add("I1: runner did not supply terminal tracking");
+                            break;
+                        }
+                        foreach ((string corr, int count) in terminalsByCorr.Where(kv => kv.Value != 1))
```
kburtram: [P1] I1 cannot detect a request that receives zero terminal responses. terminalsByCorr is populated only when an output is emitted, so iterating its entries catches duplicates but a missing result/error is absent and passes. ConnectionSimulator.WaitForQuiescence explicitly times out and delegates this condition to I1, which means a stranded request can leave the seed green. Derive expected correlation IDs from rpc.in.request envelopes (or pass the full request set) and compare counts including zero; add a fire-without-terminal regression.

src/sts2/Microsoft.SqlTools.Sts2.Testing/FakeDriver.cs
```diff
@@ -0,0 +1,281 @@
...
+                        }
+                    }
+                    rows.Add(cells);
+                }
+                return rows;
+            }
+
+            public ValueTask CancelAsync(string queryId, CancellationToken cancellationToken)
+            {
+                queryCancel.Cancel();
```
kburtram: [P2] FakeSession uses one cancellation source for its entire lifetime and never resets it. After any query is canceled, every later ExecuteAsync links against an already-canceled token, so the fake cannot exercise the required connection reuse path after cancel/dispose and can hide product state-machine bugs there. Use a per-active-query cancellation source that is cleared/disposed after enumeration, then add cancel -> dispose -> second execute on the same connection.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Redaction/SecretRedactor.cs
```diff
@@ -0,0 +1,82 @@
...
+        private static JsonNode? RedactNode(JsonNode? node, SecretSideTable sideTable, bool underAuth, ICollection<string>? createdTokens)
+        {
+            switch (node)
+            {
+                case JsonObject obj:
+                    var redactedObject = new JsonObject();
+                    foreach ((string key, JsonNode? value) in obj)
+                    {
+                        bool valueIsAuthObject = string.Equals(key, "auth", StringComparison.OrdinalIgnoreCase);
+                        bool isSecret = value is JsonValue
```
kburtram: [P1] Redaction fails open for non-string credential values. The isSecret condition requires JsonValueKind.String, so malformed inputs such as {"password":123456} are journaled verbatim before the effect runner rejects them. Because redaction is the pre-validation privacy boundary, it should fail closed by credential key regardless of JSON type. Please replace any secret-key value with an opaque marker/token that still lets validation reject it, and add an artifact-scan regression for non-string credentials.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Journaling/JournalManifest.cs
```diff
@@ -0,0 +1,73 @@
...
+        public required string Os { get; init; }
+
+        /// <summary>Runtime framework description.</summary>
+        public required string RuntimeVersion { get; init; }
+
+        /// <summary>Command-line flags with secrets removed.</summary>
+        public IReadOnlyList<string> CommandLine { get; init; } = [];
+
+        /// <summary>Segments in write order with their hash chain.</summary>
+        public IReadOnlyList<JournalSegment> Segments { get; init; } = [];
```
kburtram, line 89: [P2] The manifest omits required replay context. SPEC section 8.3 requires effective configuration snapshots and driver package versions, and section 11.1 says every effective startup config is written to the manifest, but this schema contains neither. That makes an exported/replayed run unable to establish the exact runtime and provider configuration that produced it. Please add privacy-safe config snapshots and resolved driver package/version metadata to JournalRunInfo/JournalManifest and cover their serialization.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/DriverEffectRunner.cs
```diff
@@ -0,0 +1,1389 @@
...
+        private async Task CloseAsync(EffectWorkItem effect, ICoordinatorInbox inbox)
+        {
+            string connectionId = GetString(effect.Args, "connectionId") ?? "?";
+            string? handleId = GetString(effect.Args, "handleId");
+            if (handleId is not null && sessions.TryRemove(handleId, out IDbSession? session))
+            {
+                try
+                {
+                    // Bounded close (sts2.runtime.closeTimeoutMs): never wedge shutdown on a driver.
+                    await session.DisposeAsync().AsTask().WaitAsync(TimeSpan.FromMilliseconds(Sts2Defaults.CloseTimeoutMs)).ConfigureAwait(false);
```
kburtram, line 1199: [P1] An unclassified provider exception here strands the close operation. driver.close runs this method fire-and-forget, but the catch only handles TimeoutException, DbDriverException, and ObjectDisposedException. If DisposeAsync throws (for example) InvalidOperationException, the task faults without posting the driver.close effect response, leaving Core and the client in Closing indefinitely. Treat arbitrary provider exceptions as a failed-but-complete cleanup boundary, record a redacted diagnostic, and always post exactly one close response in a finally path.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Redaction/SecretRedactor.cs
kburtram: Fixed in ec3974ad. Secret-key and auth-context classification is now independent of JSON value type: valid strings become side-table tokens, while scalars/objects/arrays/null become an opaque redacted marker before validation. Malformed auth values are redacted wholesale. Tests scan both the real journal and exported ZIP and confirm the original value never appears.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/DriverEffectRunner.cs
kburtram: Fixed in ec3974ad. Close removes ownership first, catches every provider exception without recording its potentially sensitive message, records only the exception type, and posts exactly one successful cleanup response from finally so Core cannot remain Closing. A FakeDriver InvalidOperationException regression verifies the response count, redaction, and released handle.

src/sts2/Microsoft.SqlTools.Sts2.Testing/InvariantChecker.cs
kburtram: Fixed in ec3974ad. I1 now derives the expected correlation set from every rpc.in.request and unions it with observed terminal IDs, so both zero-response and stray/duplicate-response cases fail. A zero-terminal journal regression proves the former blind spot is closed.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Replay/JournalReplayer.cs
kburtram: Fixed in ec3974ad. Replay now recomputes every payload digest, enforces schema/kind validity, a single nonempty run ID, gapless sequences, prior-only required causes, and exact output-to-input causal identity. Tests cover retained-digest payload tampering, duplicate/gapped seq, mixed runs, missing/future/wrong causes, and unknown schema.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Journaling/JournalManifest.cs
kburtram: Fixed in ec3974ad. JournalRunInfo and JournalManifest now expose privacy-safe EffectiveConfiguration and DriverPackageVersions maps; JournalWriter makes deterministic ordinal-key copies and round-trip tests cover both. The downstream hosting composition will populate these fields from its effective session options and resolved driver assemblies when that stack layer is updated.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/DriverEffectRunner.cs
kburtram: Fixed in ec3974ad. A clean stream EOF without ExecCompleted now becomes one classified Sts2.QueryFailed.Transport terminal event, clearing Core's active query. ExecCompleted now also stops enumeration immediately so a faulty provider cannot emit post-terminal events. Both paths have regressions.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/DriverEffectRunner.cs
kburtram: Fixed in ec3974ad. The runner now tracks all owned background work, cancels and bounded-awaits opens/pumps during teardown, disposes every pump CTS/semaphore exactly once, bounded-disposes sessions, and observes task faults. Cleanup also covers a coordinator rejecting the pump's initial post. Lifecycle tests and ten repeated stress runs are green.

tools/sts2-replay/Program.cs
kburtram: Fixed in ec3974ad. run, verify, and diff now branch explicitly on ReplayOutcome.Incomplete and report LastSeq plus PendingOutputCount with exit code 1 instead of dereferencing a null Divergence. CLI-level tests exercise all three commands over a genuinely truncated journal.

src/sts2/Microsoft.SqlTools.Sts2.Testing/FakeDriver.cs
kburtram: Fixed in ec3974ad. FakeSession now owns a CTS per active execution, clears/disposes it in iterator cleanup, and only cancels the current query. The cancel-flow test now executes and successfully completes a second query on the same connection.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Export/ExportBundleWriter.cs
kburtram: Fixed in ec3974ad. Export now snapshots source files before opening the destination and excludes the exact output path using platform-appropriate path comparison. A same-directory regression exercises the production default, and the exported credential-artifact test also uses that path.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Observability/MailboxEnvelopeSink.cs
kburtram: Fixed in ec3974ad. Mailbox disposal now has a bounded grace period (the standard 5-second close bound) and safely observes an abandoned worker if an untrusted callback cannot be canceled. A short-bound internal regression disposes while the callback remains wedged and proves shutdown returns.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Effects/WireValueEncoder.cs
kburtram: Fixed in ec3974ad. Int64 values within JavaScript's safe range remain JSON numbers; values outside it use an invariant int64 typed wrapper. Tests cover both safe boundaries, +/-2^53, long.MinValue, and long.MaxValue. I also passed all four unsafe values through Node JSON.parse and vscode-mssql's actual QueryStudio cellDisplayText implementation without losing a digit.

src/sts2/Microsoft.SqlTools.Sts2.Runtime/Envelopes/EnvelopeJsonCodec.cs
```diff
@@ -0,0 +1,145 @@
...
+                Schema = schema,
+                RunId = GetRequiredString(root, "runId"),
+                Seq = root.GetProperty("seq").GetInt64(),
+                Ts = DateTimeOffset.Parse(GetRequiredString(root, "ts"), CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind),
+                Kind = kind,
+                SessionId = GetOptionalString(root, "sessionId"),
+                Corr = GetOptionalString(root, "corr"),
+                Cause = root.TryGetProperty("cause", out JsonElement causeElement) && causeElement.ValueKind == JsonValueKind.Number
+                    ? causeElement.GetInt64()
+                    : null,
```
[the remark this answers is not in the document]
kburtram: Fixed in 82a5c6e0. Envelope cause now accepts only missing/null or an Int64-valued JSON number; strings, objects, arrays, booleans, fractional values, and out-of-range numeric forms fail with InvalidDataException instead of being silently normalized to null. Five malformed-shape regressions pass, as does the full 291-test runtime suite and zero-warning Release build. Resolving.

test/sts2/Microsoft.SqlTools.Sts2.UnitTests/Scenarios/SimulatorSeedRegressionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                    Task<SimulatorResult> run = ConnectionSimulator.RunSeedAsync(7496, Path.Combine(root, "iter-" + i));
+                    // A wedged run stalls for its full 90s drain/settle budget before
+                    // reporting violations; a 60s guard fails fast and names the hang
+                    // instead of eating the whole gate.
+                    Task winner = await Task.WhenAny(run, Task.Delay(TimeSpan.FromSeconds(60)));
+                    if (winner != run)
+                    {
+                        failures.Add($"iter {i}: wedged (>60s; journals under {root})");
+                        break;
+                    }
```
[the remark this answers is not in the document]
kburtram: Fixed in 13e51425. When the 60-second fast-fail guard wins, the still-bounded simulator task now gets an ExecuteSynchronously fault-only continuation that reads Exception, so any eventual fault is observed without extending the failing test. The 20-iteration seed regression passes and the Release build is warning-free. Resolving.

aasimkhan30 (approved):

PR #2791, opened 2026-08-14
Add optional fast path for completions in large files

Charles-Gagnon:
## Description

In SSMS we were seeing a large delay when typing in large files. This was because we were parsing and binding on every keystroke, and currently SqlParser has some performance issues with large files where it allocates a lot of stuff on the heap (which triggers constant GCs) and eats a lot of CPU.

I will be looking at improving SqlParser performance too, but this is a quicker workaround and results in a much more reliable experience for large files no matter the size.

So why wasn't this a problem for the old SSMS language service? Because it was much more strict on when it did the full parsing - normal keystrokes just did the minimal amount of parsing (line scanning). Then, the VS background editor thread would trigger on idle and cause the full parse (semantic tokens + diagnostics) to occur. The newer language service model doesn't do this for you though - it expects the language servers themselves to handle any perf related stuff itself for each message. I considered making have a toggle to fully behave like the radlangsvc, but that isn't a great design overall (you're basically introducing a constant lag to any requests - even if the text is small). So instead I'd rather invest in trying to improve SqlParser first.

As a quicker and safer workaround in the meantime though, I've added an optional setting to allow doing a "fast path" for large files that will skip doing the parse/rebind while typing. During this the completions list will always just return the default list. And then once the user stops typing for >1sec the background parse/rebind will trigger - after which the user will be able to see their suggestions pop up.

I have this currently disabled for STS because with STS being a separate process from VS Code there isn't nearly as much of a direct impact (SSMS runs the language service in proc, so the GCs freeze the UI while running). This does still technically affect VS Code though - but given that I didn't really see a noticeable impact beyond the delay to load the completions list I figured it'd be better to leave this as is and then when I can make the sqlparser improvements that will help VS Code more directly.

## Code Changes Checklist

### DIFF
--- a/src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
+++ b/src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
@@ -80,6 +80,18 @@ namespace Microsoft.SqlTools.LanguageService.LanguageServices
 
         internal const int DiagnosticParseDelay = 750;
 
+        // Diagnostics (and the tree refresh completion relies on for large scripts) use a longer debounce so the
+        // expensive parse only runs after the user truly stops typing, not during a brief pause in a typing session
+        // (which would make resuming feel sluggish when it collides with an in-flight parse).
+        internal const int LargeScriptDiagnosticParseDelay = 1000;
+
+        // Scripts larger than this (in characters) are treated as "large" for completion: parsing them on every
+        // keystroke allocates enough to cause GC stalls that make typing lag, so completion does not parse inline.
+        // Instead it returns the default list and relies on the debounced diagnostics parse (which resets on every
+        // keystroke, so it only runs once typing stops) to refresh the cached tree; a request after typing stops is
+        // then accurate. Smaller scripts parse inline and stay fully accurate.
+        internal const int LargeScriptCompletionThresholdBytes = 250_000;
+
         internal const int HoverTimeout = 500;
 
         internal const int OnConnectionWaitTimeout = 300 * OneSecond;
@@ -2922,6 +2934,19 @@ namespace Microsoft.SqlTools.LanguageService.LanguageServices
             // reparse and bind the SQL statement if needed
             if (RequiresReparse(scriptParseInfo, scriptFile))
             {
+                // Parsing a very large script allocates enough that doing it on every keystroke causes GC stalls
+                // and laggy typing. For large scripts, don't parse inline: return the default list now and let a
+                // debounced background parse refresh the cached result, so a request after a brief pause is
+                // accurate. Never proceed with a stale tree here — cursor positions are computed against the
+                // current text, so a mismatched tree would locate the wrong token.
+                if (CurrentWorkspaceSettings.IsLargeScriptOptimizationEnabled
+                    && (scriptFile.Contents?.Length ?? 0) > LargeScriptCompletionThresholdBytes)
+                {
+                    ScriptDocumentInfo largeDocInfo = ScriptDocumentInfo.CreateDefaultDocumentInfo(textDocumentPosition, scriptFile);
+                    CompletionItem[] largeDefaultItems = AutoCompleteHelper.GetDefaultCompletionItems(largeDocInfo, useLowerCaseSuggestions);
+                    return await ApplyCompletionExtensions(connInfo, largeDefaultItems, largeDocInfo);
+                }
+
                 ParseResult parseResult = await ParseAndBind(scriptFile, connInfo);
                 if (parseResult == null || RequiresReparse(scriptParseInfo, scriptFile))
                 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
```diff
@@ -2922,6 +2934,19 @@ public virtual async Task<CompletionItem[]> GetCompletionItems(
...
+                // and laggy typing. For large scripts, don't parse inline: return the default list now and let a
+                // debounced background parse refresh the cached result, so a request after a brief pause is
+                // accurate. Never proceed with a stale tree here — cursor positions are computed against the
+                // current text, so a mismatched tree would locate the wrong token.
+                if (CurrentWorkspaceSettings.IsLargeScriptOptimizationEnabled
+                    && (scriptFile.Contents?.Length ?? 0) > LargeScriptCompletionThresholdBytes)
+                {
+                    ScriptDocumentInfo largeDocInfo = ScriptDocumentInfo.CreateDefaultDocumentInfo(textDocumentPosition, scriptFile);
+                    CompletionItem[] largeDefaultItems = AutoCompleteHelper.GetDefaultCompletionItems(largeDocInfo, useLowerCaseSuggestions);
+                    return await ApplyCompletionExtensions(connInfo, largeDefaultItems, largeDocInfo);
```
aasimkhan30, line 2951: Are there any extensions getting registered for ssms. Just curious

src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
```diff
@@ -2922,6 +2934,19 @@ public virtual async Task<CompletionItem[]> GetCompletionItems(
...
             {
+                // Parsing a very large script allocates enough that doing it on every keystroke causes GC stalls
+                // and laggy typing. For large scripts, don't parse inline: return the default list now and let a
+                // debounced background parse refresh the cached result, so a request after a brief pause is
+                // accurate. Never proceed with a stale tree here — cursor positions are computed against the
+                // current text, so a mismatched tree would locate the wrong token.
+                if (CurrentWorkspaceSettings.IsLargeScriptOptimizationEnabled
+                    && (scriptFile.Contents?.Length ?? 0) > LargeScriptCompletionThresholdBytes)
+                {
+                    ScriptDocumentInfo largeDocInfo = ScriptDocumentInfo.CreateDefaultDocumentInfo(textDocumentPosition, scriptFile);
```
aasimkhan30, line 2949: I wonder how useful are default items.

aasimkhan30 (dismissed):

aasimkhan30:
We should think about optimizing sqlparser. Like parsing only changed batches or something. I did notice that sqlparser incremental parser is slow

src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
```diff
@@ -80,6 +80,18 @@ public static TSqlLanguageService Instance
...
+        // expensive parse only runs after the user truly stops typing, not during a brief pause in a typing session
+        // (which would make resuming feel sluggish when it collides with an in-flight parse).
+        internal const int LargeScriptDiagnosticParseDelay = 1000;
+
+        // Scripts larger than this (in characters) are treated as "large" for completion: parsing them on every
+        // keystroke allocates enough to cause GC stalls that make typing lag, so completion does not parse inline.
+        // Instead it returns the default list and relies on the debounced diagnostics parse (which resets on every
+        // keystroke, so it only runs once typing stops) to refresh the cached tree; a request after typing stops is
+        // then accurate. Smaller scripts parse inline and stay fully accurate.
+        internal const int LargeScriptCompletionThresholdBytes = 100_000;
```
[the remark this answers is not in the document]
Charles-Gagnon: Updated name

src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
```diff
@@ -2922,6 +2934,19 @@ public virtual async Task<CompletionItem[]> GetCompletionItems(
...
             if (RequiresReparse(scriptParseInfo, scriptFile))
             {
+                // Parsing a very large script allocates enough that doing it on every keystroke causes GC stalls
+                // and laggy typing. For large scripts, don't parse inline: return the default list now and let a
+                // debounced background parse refresh the cached result, so a request after a brief pause is
+                // accurate. Never proceed with a stale tree here — cursor positions are computed against the
+                // current text, so a mismatched tree would locate the wrong token.
+                if (CurrentWorkspaceSettings.IsLargeScriptOptimizationEnabled
+                    && (scriptFile.Contents?.Length ?? 0) > LargeScriptCompletionThresholdBytes)
+                {
```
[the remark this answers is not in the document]
Charles-Gagnon: Added the additional check

src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
Charles-Gagnon: They can be useful for keywords. For example if I'm typing se for SELECT then that's included in the default completion items so will show up.

Less useful for object names and the like though.

I plan on spending a bit of time looking into improving stuff like this so the default items make sense in the context, but that's a much later task.

src/Microsoft.SqlTools.LanguageService/LanguageServices/TSqlLanguageService.cs
Charles-Gagnon: Not currently (and likely never? Hard to say. But VS provides its own ways of intercepting and modifying completion stuff so it hardly seems necessary)

aasimkhan30 (approved):

Charles-Gagnon:
I investigated some more and realized I was slightly mistaken on the true  cause of the lag - it wasn't completions being synchronous in SSMS at all (they aren't). It was the CPU usage and GCs caused by SqlParser that were then freezing the UI (since it runs in proc).

I've updated the description to match my findings.

I'm still going to get this in since it does make a noticeable improvement, but I'm also working on sqlparser updates on the side - with the eventual goal being to get rid of the need for this toggle completely at some point if we can get sqlparser in a better state.

@aasimkhan30 @Benjin FYI

